PR #11, opened 2020-01-31
feat: add bookmarking for units

abutterworth:
Adds bookmarking for units. The sequence nav will reflect bookmarks as they are added.

bookmarking 2020-01-31 16_54_25


abutterworth:
Now with unit titles above

image

src/courseware/sequence/Unit.jsx
```diff
@@ -1,10 +1,12 @@
...
 import PropTypes from 'prop-types';
 import { getConfig } from '@edx/frontend-platform';
+import BookmarkButton from './bookmark/BookmarkButton';
+import useBookmark from './bookmark/useBookmark';
 
-export default function Unit({ id, pageTitle }) {
+export default function Unit({ id, pageTitle, isBookmarked, onBookmarkChanged }) {
   const iframeRef = useRef(null);
-  const iframeUrl = `${getConfig().LMS_BASE_URL}/xblock/${id}`;
+  const iframeUrl = `${getConfig().LMS_BASE_URL}/xblock/${id}?is_microfrontend_embed=1`;
```
davidjoy: As per @ormsbee 's suggestion, were we going to rename `is_microfrontend_embed` to something like `show_title`?

src/courseware/sequence/Unit.jsx
```diff
@@ -17,21 +19,39 @@ export default function Unit({ id, pageTitle }) {
...
-      scrolling="no"
-      referrerPolicy="origin"
-    />
+    <div>
+      <div className="container-fluid mb-2">
+        <h2 className="mb-0">{pageTitle}</h2>
+        <BookmarkButton
+          onClick={toggleBookmark}
+          isBookmarked={isBookmarked}
+          isWorking={requestIsInFlight}
```
davidjoy: `processing={bookmarkProcessing}` or something? :)

src/courseware/sequence/bookmark/useBookmark.js
```diff
@@ -0,0 +1,36 @@
+import { useState } from 'react';
+import { getConfig } from '@edx/frontend-platform';
+import { getAuthenticatedHttpClient, getAuthenticatedUser } from '@edx/frontend-platform/auth';
+
+export default function useBookmark(usageId, isBookmarked, onBookmarkChanged) {
+  const [requestIsInFlight, setRequestInFlight] = useState(false);
```
davidjoy: `loading` or `processing` might be a bit less verbose.

src/courseware/sequence/bookmark/useBookmark.js
```diff
@@ -0,0 +1,36 @@
...
+  const httpClient = getAuthenticatedHttpClient();
+  const user = getAuthenticatedUser();
+
+  const toggle = () => {
```
davidjoy: Probably want `useCallback` here.

davidjoy (approved):
A few nits and questions

src/data/course-blocks/slice.js
```diff
@@ -0,0 +1,134 @@
...
+    fetchBlockMetadataSuccess: (draftState, action) => {
+      const { blockId, metadata, relatedBlocksMetadata } = action.payload;
+      if (!draftState.blocks[blockId]) {
+        draftState.blocks[blockId] = {};
+      }
+      draftState.blocks[blockId] = {
+        ...draftState.blocks[blockId],
+        ...metadata,
+        fetchState: 'loaded',
+        loaded: true,
```
davidjoy, line 53: What's the difference between the `fetchState` here and `loaded`?

src/data/course-blocks/slice.js
```diff
@@ -0,0 +1,134 @@
...
+      };
+      if (relatedBlocksMetadata) {
+        relatedBlocksMetadata.forEach((blockMetadata) => {
+          if (draftState.blocks[blockMetadata.id] === undefined) {
+            draftState.blocks[blockMetadata.id] = {};
+          }
+          draftState.blocks[blockMetadata.id] = {
+            ...draftState.blocks[blockMetadata.id],
+            ...blockMetadata,
```
davidjoy, line 62: Should we add a `fetchState` to this block too?  Seems like this code for related blocks and the above code for the primary block could be merged into a single function that updates metadata for any given block.

src/data/course-blocks/slice.js
```diff
@@ -0,0 +1,134 @@
...
+     * The update state of a property is added to the block in the store with
+     * a dynamic property name: ${propertyToUpdate}UpdateState.
+     * (e.g. bookmarkedUpdateState)
+     *
+     * Used in:
+     *  - saveSequencePosition
+     *  - addBookmark
+     *  - removeBookmark
+     */
+    updateBlockRequest: (draftState, action) => {
```
davidjoy, line 92: My "will this be hard to understand later?" sense is tingling on these updateBlock* routines. :)

davidjoy: That said, I also like it.

src/data/course-blocks/thunks.js
```diff
@@ -0,0 +1,119 @@
...
+import {
+  getCourseBlocks,
+  getSequenceMetadata,
+  getBlockCompletion,
+  updateSequencePosition,
+  createBookmark,
+  deleteBookmark,
+} from './api';
+
+export const fetchCourseBlocks = courseUsageKey => async (dispatch) => {
```
davidjoy: I think in the past we'd been talking about making things like this functions instead of consts.  It might make them more legible, perhaps?  Not sure.  I always have a hard time processing functions that return functions and such.

davidjoy (approved):

src/data/course-blocks/slice.js
abutterworth: True, we could dispatch actions for all these blocks in the originating thunk.

src/data/course-blocks/slice.js
abutterworth: agreed, I'm not totally sure this is a great idea, but I think we can change it in the future if we find it frustrating.

src/data/course-blocks/thunks.js
abutterworth: Will change 👍

src/data/course-blocks/slice.js
abutterworth: Will leave this for a later refactor

PR #13, opened 2020-02-27
If the user does not have access to the course, then redirect to the course outline.

davidjoy:
In a subsequent PR, if this API call is made on the course outline page in the MFE, we’ll need to be able to prevent the redirect.  But that view of the MFE doesn’t exist yet.

### DIFF
--- a/src/data/course-meta/thunks.js
+++ b/src/data/course-meta/thunks.js
@@ -1,4 +1,5 @@
 /* eslint-disable import/prefer-default-export */
+import { getConfig } from '@edx/frontend-platform';
 import {
   fetchCourseMetadataRequest,
   fetchCourseMetadataSuccess,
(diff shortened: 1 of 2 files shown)


src/data/course-meta/thunks.js
```diff
@@ -13,7 +14,12 @@ export function fetchCourseMetadata(courseUsageKey) {
     dispatch(fetchCourseMetadataRequest({ courseUsageKey }));
     try {
       const courseMetadata = await getCourseMetadata(courseUsageKey);
-      dispatch(fetchCourseMetadataSuccess(courseMetadata));
+
+      if (!courseMetadata.userHasAccess) {
+        global.location.assign(`${getConfig().LMS_BASE_URL}/courses/${courseUsageKey}/course/`);
```
abutterworth: Would this be better handled in the React side of things? We already handle some redirects there I think.

abutterworth (approved):

src/data/course-meta/thunks.js
abutterworth: I imagine we'll want the course metadata in the redux store even if we don't have access to the content once we get the course outline in the MFE.

davidjoy: I mean, in principle this is what the thunk is for... "Thunks are primarily used to delay a calculation until its result is needed, or to _insert operations at the beginning or end of the other subroutine._" (emphasis mine)

It feels a bit odd to me to let all this data get through redux, update the components, only to redirect then.

abutterworth:
My thinking was that we asked for course metadata and we got it
successfully. When we add course home to this MFE we'll need this metadata
in the redux store anyway. I'm not sure this thunk/routine should be
concerned with where to route you. Though adding redirects in useEffect
hooks doesn't feel great either.

davidjoy:
Ah, dang, you're right.  The course outline page doesn't do any redirects if you don't have access, it just shows you alternate content.  When we bring that in, we'll want to always load the metadata successfully and then make decisions.  It'll be easier later if this just works that way from the get go. 👍

PR #14, opened 2020-02-28
Added a verified upgrade component

davestgermain:
Requires API support in https://github.com/edx/edx-platform/pull/23238

Screen Shot 2020-03-03 at 11 09 28 AM


src/courseware/course/CourseSock.jsx
```diff
@@ -0,0 +1,94 @@
...
+          </button>
+        </div>
+        {this.state.showUpsell && (
+        <div className="d-flex justify-content-around">
+          <div className="mt-3">
+            <h2 className="font-weight-lighter">edX Verified Certificate</h2>
+            <h3>Why upgrade?</h3>
+            <ul>
+              <li>Official proof of completion</li>
+              <li>Easily shareable certificate</li>
```
abutterworth: This needs to be i18n friendly. See:
- https://edx.github.io/frontend-platform/module-Internationalization.html
- which re-exports react-intl components documented here: https://github.com/formatjs/react-intl/blob/master/docs/Components.md#formattedmessage
- example: https://github.com/edx/frontend-app-payment/blob/master/src/payment/checkout/Checkout.jsx#L107 - https://github.com/edx/frontend-app-payment/blob/master/src/payment/checkout/Checkout.jsx#L8

abutterworth (changes requested):

src/courseware/course/CourseSock.jsx
```diff
@@ -0,0 +1,177 @@
+import React, { Component } from 'react';
```
davidjoy: Code organization nit - 

The course sock feels like a pretty self contained module, so we may want to formalize it's organization a bit:

```
/course
    /course-sock
        /index.js - exports the public API of the module, which is only the CourseSock component
        /CourseSock.jsx
        /messages.jsx - messages used by the course sock module
        /assets
            /learner-quote.png
            /learner-quote2.png
            /verified-cert.png
```

Then it gets imported as:

`import CourseSock from './course-sock';`

See the courseware/sequence/content-lock module for a pretty simple example.

src/courseware/course/CourseSock.jsx
```diff
@@ -0,0 +1,177 @@
...
+export default class CourseSock extends Component {
+  constructor(props) {
+    super(props);
+    this.verifiedMode = props.verifiedMode;
+    this.state = { showUpsell: false, buttonClass: 'btn-success' };
+
+    this.handleClick = this.handleClick.bind(this);
+  }
+
+  handleClick() {
```
davidjoy: Fun fact -

You can avoid having to remember to `bind(this)` on handlers like this if they're declared as:

`handleClick = () => { ... }`

Because they're an arrow function, they don't need to be bound!  I've started doing that in class components just so I don't have to remember the `bind` line. :)

src/courseware/course/CourseSock.jsx
```diff
@@ -0,0 +1,177 @@
...
+    this.setState(state => ({
+      showUpsell: !state.showUpsell,
+      buttonClass: state.showUpsell ? 'btn-success' : 'btn-outline-success',
```
davidjoy: Given that `buttonClass` is 'derived state' directly from `showUpsell`, it'd be perfectly reasonable to pull it out of `state` and have a line at the top of `render` that reads:

```
render() {
  const buttonClass = state.showUpsell ? 'btn-success' : 'btn-outline-success'
  // rest of render(), using that const instead of `this.state.buttonClass`
}
```

Would simplify things a touch!

davidjoy (approved):
A few nits and mild suggestions that I'd probably want to clean up if I were looking at it later. :)

abutterworth (approved):

PR #15, opened 2020-03-03
Fire tracking log events in sequence navigation

stvstnfrd:
to mirror existing behavior in `edx-platform`.

This adds support for the following events:
- `edx.ui.lms.sequence.next_selected`
- `edx.ui.lms.sequence.previous_selected`
- `edx.ui.lms.sequence.tab_selected`

## Outstanding Events
- `edx.ui.lms.outline.*`
- `edx.ui.lms.link_clicked`

## References
- https://openedx.atlassian.net/browse/TNL-7066
- https://edx.readthedocs.io/projects/devdata/en/latest/internal_data_formats/tracking_logs/student_event_types.html#course-navigation-events

### DIFF
--- a/src/courseware/sequence/Sequence.jsx
+++ b/src/courseware/sequence/Sequence.jsx
@@ -3,6 +3,7 @@ import React, {
   useEffect, useContext, Suspense, useState,
 } from 'react';
 import PropTypes from 'prop-types';
+import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 import { injectIntl, intlShape, FormattedMessage } from '@edx/frontend-platform/i18n';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faChevronLeft, faChevronRight } from '@fortawesome/free-solid-svg-icons';
@@ -54,6 +55,23 @@ function Sequence({
     onNavigateUnit(unitId);
   };
 
+  const logEvent = (eventName, widgetPlacement, targetUnitId) => {
+    // Note: tabs are tracked with a 1-indexed position
+    // as opposed to a 0-index used throughout this MFE
+    const currentIndex = unitIds.indexOf(activeUnitId);
+    const payload = {
+      current_tab: currentIndex + 1,
+      id: activeUnitId,
+      tab_count: unitIds.length,
+      widget_placement: widgetPlacement,
+    };
+    if (targetUnitId) {
+      const targetIndex = unitIds.indexOf(targetUnitId);
+      payload.target_tab = targetIndex + 1;
+    }
+    sendTrackEvent(eventName, payload);
+  };
+
   const { add, remove } = useContext(UserMessagesContext);
   useEffect(() => {
     let id = null;
@@ -85,9 +103,18 @@ function Sequence({
     <>
       <SequenceNavigation
         className="mb-4"
-        onNext={handleNext}
-        onNavigate={handleNavigate}
-        onPrevious={handlePrevious}
+        onNext={() => {
+          logEvent('edx.ui.lms.sequence.next_selected', 'top');
+          handleNext();
+        }}
+        onNavigate={(unitId) => {
+          logEvent('edx.ui.lms.sequence.tab_selected', 'top', unitId);
+          handleNavigate(unitId);
+        }}
+        onPrevious={() => {
+          logEvent('edx.ui.lms.sequence.previous_selected', 'top');
+          handlePrevious();
+        }}
         unitIds={unitIds}
         activeUnitId={activeUnitId}
         isLocked={isGated}
@@ -122,7 +149,10 @@ function Sequence({
         <div className="unit-content-container below-unit-navigation">
           <Button
             className="btn-outline-secondary previous-button w-25 mr-2"
-            onClick={handlePrevious}
+            onClick={() => {
+              logEvent('edx.ui.lms.sequence.previous_selected', 'bottom');
+              handlePrevious();
+            }}
           >
             <FontAwesomeIcon icon={faChevronLeft} className="mr-2" size="sm" />
             <FormattedMessage
@@ -133,7 +163,10 @@ function Sequence({
           </Button>
           <Button
             className="btn-outline-primary next-button w-75"
-            onClick={handleNext}
+            onClick={() => {
+              logEvent('edx.ui.lms.sequence.next_selected', 'bottom');
+              handleNext();
+            }}
           >
             <FormattedMessage
               id="learn.sequence.navigation.after.unit.next"


src/courseware/sequence/Sequence.jsx
```diff
@@ -2,6 +2,7 @@
 import React, { useEffect, useContext, Suspense } from 'react';
 import PropTypes from 'prop-types';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import { sendTrackEvent } from '@edx/frontend-platform/analytics';
```
stvstnfrd, line 6: Is there a style guide recommendation on import order?

src/courseware/sequence/Sequence.jsx
```diff
@@ -26,7 +27,19 @@ function Sequence({
   bannerText,
   intl,
 }) {
+  const logEvent = (eventName) => {
+    const currentIndex = Math.max(0, unitIds.indexOf(activeUnitId));
```
stvstnfrd: I'm not sure if we care to have any more resilient error handling than this.
I recall running into this once during development, where the `activeUnitId` wasn't found so the index came back `-1`.
_Probably_ shouldn't happen, but, you know ;)

This just naively sets a floor of `0`, so we always have a valid (err, non-negative) index..

src/courseware/sequence/SequenceNavigation.jsx
```diff
@@ -19,6 +19,8 @@ export default function SequenceNavigation({
     <UnitButton
       key={unitId}
       unitId={unitId}
+      unitIds={unitIds}
+      activeUnitId={activeUnitId}
```
stvstnfrd: Disclosure: I'm cargo-culting and hacking away here; React is new for me :)

Is there a different way to pass this context along?

src/courseware/sequence/UnitButton.jsx
```diff
@@ -17,8 +18,24 @@ function UnitButton({
...
   showCompletion,
   unitId,
+  unitIds,
+  activeUnitId,
 }) {
+  const logEvent = (eventName) => {
+    const targetIndex = Math.max(0, unitIds.indexOf(unitId));
+    const currentIndex = Math.max(0, unitIds.indexOf(activeUnitId));
+    const payload = {
+      target_tab: targetIndex + 1,
```
stvstnfrd: This is the one difference between this logging helper and the one in `Sequence.jsx`

stvstnfrd (commented):
As mentioned, this is a new framework for me, so please pass along any feedback, particularly if there are more idiomatic ways of doing things.

Thanks!

src/courseware/sequence/Sequence.jsx
abutterworth: None right now afaik

src/courseware/sequence/Sequence.jsx
abutterworth: I think -1 is important information, maybe we should let it through. Otherwise we won't be able to distinguish between these out of bounds units vs the first one.

src/courseware/sequence/Sequence.jsx
```diff
@@ -26,7 +27,19 @@ function Sequence({
...
+      current_tab: currentIndex + 1,
+      id: activeUnitId,
+      tab_count: unitIds.length,
+      widget_placement: 'top',
+    };
+    sendTrackEvent(eventName, payload);
+  };
+
   const handleNext = () => {
+    logEvent('edx.ui.lms.sequence.next_selected');
```
abutterworth: Just to be sure – these next_selected and previous_select events are triggered when navigating between units and navigation between sequences?

src/courseware/sequence/SequenceNavigation.jsx
abutterworth: The less react's leaf components are responsible for/ aware of the more maintainable our code will be. It may be better to not send any of this in and send these track events in the `handleNavigate` in `Sequence.jsx` or do all the logging in SequenceNavigation before calling `onNext` `onPrevious` or `onNavigate`

src/courseware/sequence/Sequence.jsx
davidjoy: The only real guideline I have is that external dependencies (like this one) come before relative dependencies/other files in the project.

src/courseware/sequence/Sequence.jsx
stvstnfrd: +1, flooring removed in new commit.

src/courseware/sequence/Sequence.jsx
stvstnfrd: Cool, thanks!

src/courseware/sequence/Sequence.jsx
stvstnfrd: Yes. The code below where it checks the bounds:
- `if (previousIndex >= 0) {`
- `if (nextIndex < unitIds.length) {`

handles the two cases:
- within the sequence
- across sequences

I confirmed the events firing in each case.

davidjoy:
👋 I think Adam covered any comments I might have - happy to help with refactoring if it's helpful.  It feels like the UnitButton doesn't need to "own" sending the events itself, it can use handlers/callbacks to delegate that back to a "smarter" layer like the SequenceNavigation.  That way we don't need to pass quite so much context down.

src/courseware/sequence/SequenceNavigation.jsx
stvstnfrd: Yeah, it felt hacky as-was, so I'm glad I asked :)

I had started the integration "lower" because it looked like I may have needed some contextual switching at that level, which ended up being unnecessary. Taking a step back, it seems like we can refactor this to move the integration into `Sequence`.

Done.

src/courseware/sequence/UnitButton.jsx
stvstnfrd: I think I can combine these two helpers in the newly consolidated code.

src/courseware/sequence/Sequence.jsx
```diff
@@ -27,27 +28,64 @@ function Sequence({
...
     if (nextIndex < unitIds.length) {
       const newUnitId = unitIds[nextIndex];
-      handleNavigate(newUnitId);
+      onNavigateUnit(newUnitId);
     } else {
       onNext();
     }
   };
 
+  const handleNextBottom = () => {
```
stvstnfrd: Now with #16, we need to determine whether the event is being fired from the top or the bottom.

I considered other ways of doing this, like `className` introspection, but that seems like more of a hack.
I'm open to other ideas, but dedicated handlers (for top vs. bottom) doesn't seem like a crazy idea?
:man_shrugging: 

Thoughts?

src/courseware/sequence/Sequence.jsx
```diff
@@ -27,27 +28,64 @@ function Sequence({
   bannerText,
   intl,
 }) {
+  const currentIndex = unitIds.indexOf(activeUnitId);
+
+  const logEvent = (eventName, widgetPlacement, targetUnitId) => {
+    const payload = {
+      current_tab: currentIndex + 1,
```
abutterworth:
```suggestion
    const payload = {
      // tabs are tracked with a 1-indexed position as opposed to a 0-index used throughout this mfe
      current_tab: currentIndex + 1,
```
May be worth commenting why we're adding one. I was hoping to keep our 1-indexed "position" out of the mfe but this seems unavoidable

abutterworth (approved):
lgtm

src/courseware/sequence/Sequence.jsx
```diff
@@ -27,27 +28,64 @@ function Sequence({
...
+    logEvent('edx.ui.lms.sequence.next_selected', 'top');
+    handleNext();
+  };
+
   const handlePrevious = () => {
     const previousIndex = unitIds.indexOf(activeUnitId) - 1;
     if (previousIndex >= 0) {
       const newUnitId = unitIds[previousIndex];
-      handleNavigate(newUnitId);
+      onNavigateUnit(newUnitId);
```
stvstnfrd: Note below that `handleNavigate` was previously just a wrapper for `onNavigateUnit`, with no additional logic.

Since `handleNavigate` _was_ both the handler for the `UnitButton.click` _and_ called internally by `handlePrevious` and `handleNext`, we needed a way to differentiate between the calls.
We only want to log the event on `click`.

Now, `handlePrevious` and `handleNext` call `onNavigateUnit` directly, internally, and `handleNavigate` logs an event before calling `onNavigateUnit`.

src/courseware/sequence/Sequence.jsx
```diff
@@ -77,9 +115,9 @@ function Sequence({
         <AlertList topic="sequence" className="mt-3" />
         <SequenceNavigation
           className="mb-3"
-          onNext={handleNext}
+          onNext={handleNextTop}
```
stvstnfrd: Each of these updated to call the new (logging-aware) handlers.

src/courseware/sequence/Sequence.jsx
```diff
@@ -27,27 +28,64 @@ function Sequence({
...
+    }
+    sendTrackEvent(eventName, payload);
+  };
+
   const handleNext = () => {
     const nextIndex = unitIds.indexOf(activeUnitId) + 1;
     if (nextIndex < unitIds.length) {
       const newUnitId = unitIds[nextIndex];
-      handleNavigate(newUnitId);
+      onNavigateUnit(newUnitId);
```
stvstnfrd: See note below about this change.

stvstnfrd (commented):
@abutterworth @davidjoy Updated and consolidated into `Sequence.jsx`.

See additional comments inline..

src/courseware/sequence/Sequence.jsx
abutterworth: This looks good to me. If we think it's cleaner to read and not have to name these handlers maybe we just put them inline below?
```
<Button
  className="btn-outline-primary next-button w-75"
  onClick={() => {
    logEvent('edx.ui.lms.sequence.next_selected', 'bottom');
    handleNext();
  }}
>
```

src/courseware/sequence/Sequence.jsx
stvstnfrd: Good call, comment added above ^

src/courseware/sequence/Sequence.jsx
stvstnfrd: Done, handlers moved inline.

src/courseware/sequence/Sequence.jsx
stvstnfrd: Now irrelevant.

PR #17, opened 2020-03-04
Custom alerts for anonymous and unenrolled users.

davidjoy:
This PR implements user alerts we discovered as part of https://openedx.atlassian.net/browse/TNL-7073.

Currently there's no way to view these alerts because we automatically redirect to the course outline page if the user doesn't have access to the course.

If we allow anonymous access to the course blocks API and get the MFE working with the `COURSE_ENABLE_UNENROLLED_ACCESS_FLAG` waffle flag, then these will become visible.

That said, they'll both become necessary and used as soon as we implement the course outline page in the micro-frontend, as anonymous and unenrolled users are NOT redirected away from the course home page, and instead see these messages.  

This PR also adds a `customAlerts` property to `AlertList` allowing applications to specify custom alert components for given message codes.  This mechanism does not currently support adding arbitrary data into the custom component from the message source.  i.e., if a server wants to display a custom alert component with data from the server, there's no way to get that into the component as part of the message.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   "dependencies": {
     "@edx/frontend-component-footer": "^10.0.6",
     "@edx/frontend-component-header": "^2.0.3",
-    "@edx/frontend-platform": "^1.1.11",
+    "@edx/frontend-platform": "^1.3.1",
     "@edx/paragon": "^7.2.0",
     "@fortawesome/fontawesome-svg-core": "^1.2.26",
     "@fortawesome/free-brands-svg-icons": "^5.12.0",
--- a/src/courseware/CourseContainer.jsx
+++ b/src/courseware/CourseContainer.jsx
@@ -66,6 +66,7 @@ function CourseContainer(props) {
       unitId={unitId}
       models={models}
       tabs={props.metadata.tabs}
+      enrollmentIsActive={props.metadata.enrollmentIsActive}
       verifiedMode={props.metadata.verifiedMode}
     />
   );
@@ -90,6 +91,7 @@ CourseContainer.propTypes = {
       type: PropTypes.string,
       url: PropTypes.string,
     })),
+    enrollmentIsActive: PropTypes.bool,
     verifiedMode: PropTypes.shape({
       price: PropTypes.number.isRequired,
       currency: PropTypes.string.isRequired,
--- a/src/courseware/CourseContainer.jsx
+++ b/src/courseware/CourseContainer.jsx
@@ -66,7 +66,7 @@ function CourseContainer(props) {
       unitId={unitId}
       models={models}
       tabs={props.metadata.tabs}
-      enrollmentIsActive={props.metadata.enrollmentIsActive}
+      isEnrolled={props.metadata.isEnrolled}
       verifiedMode={props.metadata.verifiedMode}
     />
   );
@@ -91,7 +91,7 @@ CourseContainer.propTypes = {
       type: PropTypes.string,
       url: PropTypes.string,
     })),
-    enrollmentIsActive: PropTypes.bool,
+    isEnrolled: PropTypes.bool,
     verifiedMode: PropTypes.shape({
       price: PropTypes.number.isRequired,
       currency: PropTypes.string.isRequired,
--- a/src/courseware/course/Course.jsx
+++ b/src/courseware/course/Course.jsx
@@ -22,7 +22,7 @@ export default function Course({
   courseName,
   courseOrg,
   courseUsageKey,
-  enrollmentIsActive,
+  isEnrolled,
   models,
   sequenceId,
   tabs,
@@ -52,7 +52,7 @@ export default function Course({
   });
 
   useLogistrationAlert();
-  useEnrollmentAlert(enrollmentIsActive);
+  useEnrollmentAlert(isEnrolled);
 
   return (
     <>
@@ -110,7 +110,7 @@ Course.propTypes = {
   courseId: PropTypes.string.isRequired,
   sequenceId: PropTypes.string.isRequired,
   unitId: PropTypes.string,
-  enrollmentIsActive: PropTypes.bool,
+  isEnrolled: PropTypes.bool,
   models: PropTypes.objectOf(PropTypes.shape({
     id: PropTypes.string.isRequired,
     displayName: PropTypes.string.isRequired,
@@ -135,6 +135,6 @@ Course.propTypes = {
 
 Course.defaultProps = {
   unitId: undefined,
-  enrollmentIsActive: false,
+  isEnrolled: false,
   verifiedMode: null,
 };
--- a/src/courseware/course/Course.jsx
+++ b/src/courseware/course/Course.jsx
@@ -10,9 +10,24 @@ import CourseHeader from './CourseHeader';
 import CourseSock from './course-sock';
 import CourseTabsNavigation from './CourseTabsNavigation';
 import InstructorToolbar from '../InstructorToolbar';
+import { useLogistrationAlert, useEnrollmentAlert } from '../../hooks';
+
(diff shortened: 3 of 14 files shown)


package.json
```diff
@@ -36,7 +36,7 @@
   "dependencies": {
     "@edx/frontend-component-footer": "^10.0.6",
     "@edx/frontend-component-header": "^2.0.3",
-    "@edx/frontend-platform": "^1.1.11",
+    "@edx/frontend-platform": "^1.3.1",
```
davidjoy, line 39: Needed the `getLoginRedirectUrl` function from the latest version of the platform.

src/courseware/course/Course.jsx
```diff
@@ -52,7 +70,14 @@ export default function Course({
       <main className="d-flex flex-column flex-grow-1">
         <div className="container-fluid">
           <CourseTabsNavigation tabs={tabs} className="mb-3" activeTabSlug="courseware" />
-          <AlertList topic="course" className="mb-3" />
+          <AlertList
+            topic="course"
+            className="mb-3"
+            customAlerts={{
+              clientEnrollmentAlert: EnrollmentAlert,
```
davidjoy, line 77: The keys here can be whatever we want - I put the word 'client' in front of them to ensure they're highly unlikely to collide with anything from the backend.

src/hooks.js
```diff
@@ -0,0 +1,50 @@
+import { useContext, useState, useEffect } from 'react';
+import { AppContext } from '@edx/frontend-platform/react';
+import UserMessagesContext from './user-messages/UserMessagesContext';
+
+export function useLogistrationAlert() {
```
davidjoy, line 5: This is a custom hook because we'll need it in two places once the outline page goes in.

src/hooks.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+    return () => {
+      if (alertId !== null) {
+        remove(alertId);
+      }
+    };
+  }, [authenticatedUser]);
+}
+
+export function useEnrollmentAlert(enrollmentIsActive) {
```
davidjoy: Ditto.

src/user-messages/AlertList.jsx
```diff
@@ -14,26 +18,38 @@ export default function AlertList({ topic, className }) {
...
-          dismissible={message.dismissible}
-          onDismiss={() => remove(message.id)}
-        >
-          {message.text}
-        </Alert>
-      ))}
+      {topicMessages.map(message => {
+        const AlertComponent = getAlertComponent(message.code);
+        return (
+          <Suspense key={message.id} fallback={null}>
```
davidjoy, line 24: We don't know where the AlertComponent is coming from.  It may be a component bundled with the main application, or (as it is in our current use case), one that we lazy-load as necessary.  This `<Suspense/>` around the component doesn't hurt the former, but is necessary to use `React.lazy` for the latter.  We have no fallback because we don't want something else flashing in while the component (quickly) loads.

src/user-messages/UserMessagesProvider.jsx
```diff
@@ -7,21 +7,35 @@ export default function UserMessagesProvider({ children }) {
   const [messages, setMessages] = useState([]);
   const [nextId, setNextId] = useState(1);
 
+  const refMessages = useRef(messages);
```
davidjoy, line 10: The changes to this file fix a particular bug that can happen if you call `add` or `remove` multiple times during one rendering of a component.  Because of lexical scoping, the second call acts on the previous version of the messages, and effectively undoes the work of the first call.  A ref is not part of the component's re-render cycle, and so is a safe place to store the "real" list across calls to the add/remove/clear handlers.

src/courseware/course/Course.jsx
```diff
@@ -10,9 +10,24 @@ import CourseHeader from './CourseHeader';
...
+  courseNumber,
+  courseName,
+  courseOrg,
+  courseUsageKey,
+  enrollmentIsActive,
+  models,
+  sequenceId,
+  tabs,
+  unitId,
+  verifiedMode,
```
stvstnfrd, line 30: This is alphabetizing and adding `enrollmentIsActive`?

src/enrollment-alert/EnrollmentAlert.jsx
```diff
@@ -0,0 +1,24 @@
...
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+
+import Alert from '../user-messages/Alert';
+import messages from './messages';
+
+function EnrollmentAlert({ intl }) {
+  return (
+    <Alert type="error">
+      {intl.formatMessage(messages['learning.enrollment.alert'])}
+      {' '}
```
stvstnfrd, line 12: A literal whitespace character?
What's this?

src/user-messages/UserMessagesProvider.jsx
```diff
@@ -7,21 +7,35 @@ export default function UserMessagesProvider({ children }) {
...
+
   const add = ({
     code, dismissible, text, type, topic, ...others
   }) => {
     const id = nextId;
-    setMessages([...messages, {
+    refMessages.current = [...refMessages.current, {
```
abutterworth, line 16: Should we update to use this form of of setState and drop our use of useRef? https://reactjs.org/docs/hooks-reference.html#functional-updates

stvstnfrd (commented):
Just a couple of clarifying questions, but I'll keep looking over this for React reference. Thanks!

src/courseware/CourseContainer.jsx
```diff
@@ -90,6 +91,7 @@ CourseContainer.propTypes = {
       type: PropTypes.string,
       url: PropTypes.string,
     })),
+    enrollmentIsActive: PropTypes.bool,
```
abutterworth: Not sure if there's a better name or not, but I don't exactly know what `enrollmentIsActive` means

abutterworth (approved):

src/courseware/CourseContainer.jsx
ormsbee: Yeah, I might be clearer to say `isEnrolled`? In the beginning, enrollments either existed or not. But later on, we made it so that there's an `is_active` flag to course enrollments so that unenrolling doesn't delete the association (which is really important for things like revenue tracking). I suspect this is an implementation detail that somehow bubbled up?

src/user-messages/UserMessagesProvider.jsx
davidjoy: Yeah, I punted on trying it cause I didn't want to get bogged down by proving it worked, but yes, I think we should.

src/courseware/course/Course.jsx
davidjoy: Yeppers - added `enrollmentIsActive`, and after a certain width, the linter suggests putting them one on each line. And I figured I might as well alphabetize 'em while I was there.

src/enrollment-alert/EnrollmentAlert.jsx
davidjoy: Without it, the "Enroll Now" button is stuck right up against the message because any whitespace in the JSX itself is stripped out.  By putting it in `{ }` it basically treats it as important and preserves it when it gets rendered out to the actual DOM.  

The alternative would be a `&nbsp;` but I didn't think of it at the time, and also my post-explanation is that I wouldn't want it to be non-breaking necessarily. ;)

src/courseware/CourseContainer.jsx
davidjoy: Dave has it right - the field is:

```
enrollment: {
  is_active: true
}
```

... from the server. :). I flattened it because in the client it felt like arbitrary nesting. 

Happy to change the name to isEnrolled, but I was mostly trying to keep consistent naming where I didn't have a good reason to change it.

PR #18, opened 2020-03-04
Quick styling updates

abutterworth:
image

### DIFF
--- a/src/courseware/CourseContainer.jsx
+++ b/src/courseware/CourseContainer.jsx
@@ -47,28 +47,29 @@ function CourseContainer(props) {
     }
   }, [metadataLoaded]);
 
-  if (!courseId || !sequenceId) {
-    return (
-      <PageLoading
-        srMessage={intl.formatMessage(messages['learn.loading.learning.sequence'])}
-      />
-    );
-  }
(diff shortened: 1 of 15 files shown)


davidjoy (approved):

marcotuts:
Two minor nits / changes  
1- I would switch to font awesome's "edit" icon for assessment pages instead of "pencil", as this icon is likely to also be added back into the outline with the Always Available squad efforts, and it separates the pencil to be our edit icon (which isn't a problem for learners but would be overloaded for educators). 
2- Next / Previous don't show angle-left / angle-right currently is that intended? I would add but I didn't know if that was an intentional shift perhaps? 
3 - The blue seems off here, that might be a general comment about our blue in general across paragon, is that the case? (I seem to remember maybe discussing this third one in person at some point...) 
4 - Is the thought to still shift to light gray for background around content frame where breadcrumbs or should we leave as is?

abutterworth:
Marco:
1 - Will do
2 - Will add
3 - It is Paragon's blue. I can change it in this project pretty easily, but unsure if we should or not. Wdyt?
4 - Eugene suggested keeping it white.

marcotuts (approved):

PR #20, opened 2020-03-05
Initial version of Course Home page

davidjoy:
Screen Shot 2020-03-05 at 5 05 51 PM

This page is not complete.

- It contains the ‘outline’ itself with links to the Sequences in the course.
- It contains a very basic stab at displaying dates - they’re not even formatted.
- It shows logistration and enrollment alerts for anonymous and unenrolled users.

It does not include any other content in the right-hand sidebar.  It also doesn’t include a welcome message, or perhaps any number of other features on the page.  This is effectively an initial implementation for discovering how much data we’re missing from our APIs.  It should not be used as-is by any means.

We're checking this in because we expect we'll continue work on it in the near future, and we don't want this implementation to drift too far from the underlying MFE code.  Since it's not linked to, it's effectively hidden and won't be used.

Data needed in our APIs to finish implementing this page:

- Welcome message
- Prerequisite data for sequences shown in the outline.
- Any additional date information that Always Available may need/be adding.
- Course tools
- Course handouts
- Anything else that should appear in the right-hand sidebar
- Course search


src/courseware/CourseContainer.jsx
```diff
@@ -80,28 +80,7 @@ CourseContainer.propTypes = {
...
-    isEnrolled: PropTypes.bool,
-    verifiedMode: PropTypes.shape({
-      price: PropTypes.number.isRequired,
-      currency: PropTypes.string.isRequired,
-      currencySymbol: PropTypes.string.isRequired,
-      sku: PropTypes.string.isRequired,
-      upgradeUrl: PropTypes.string.isRequired,
-    }),
-  }),
+  metadata: courseMetadataShape,
```
abutterworth, line 83: Yay

src/data/course-blocks/slice.js
```diff
@@ -132,3 +133,10 @@ export const {
 } = blocksSlice.actions;
 
 export const { reducer } = blocksSlice;
+
+export const courseBlocksShape = PropTypes.objectOf(PropTypes.shape({
```
abutterworth, line 137: Thoughts around another `shapes.js` file instead of adding here?

src/index.jsx
```diff
@@ -39,6 +40,7 @@ subscribe(APP_READY, () => {
       <UserMessagesProvider>
         <Switch>
           <Route exact path="/" render={courseLinks} />
+          <Route path="/outline/:courseUsageKey" component={OutlineContainer} />
```
abutterworth, line 43: What do you think about url structure like this instead?

- `/course/:courseUsageKey/outline`
- `/course/:courseUsageKey/: sequenceId /:unitId`

src/outline/Chapter.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Collapsible } from '@edx/paragon';
+import { faChevronRight, faChevronDown } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import SequenceLink from './SequenceLink';
+import { courseBlocksShape } from '../data/course-blocks';
+
+export default function Chapter({ id, courseUsageKey, models }) {
```
abutterworth, line 9: Do we prefer the name Section? @ormsbee

src/outline/Outline.jsx
```diff
@@ -0,0 +1,115 @@
...
+            </div>
+          </div>
+        </div>
+      </main>
+    </>
+  );
+}
+
+Outline.propTypes = {
+  courseOrg: PropTypes.string.isRequired,
```
abutterworth, line 92: Is is possible to use the course meta shape here?

abutterworth (approved):

src/data/course-blocks/slice.js
davidjoy: I thought it was worthwhile to keep it in the same file so it was easy to remember to update it - adding the index.js file means that importers don't need to guess or wonder what file it's in, so that felt pretty good.

src/index.jsx
davidjoy: I made it the way it is so that we could avoid any errors around `outline` accidentally being treated as a sequenceId... which is maybe over paranoid, but, yeah.  I think when we pick up this work again, we should intentionally pick a URL we like.

src/outline/Outline.jsx
davidjoy: I don't think so... these are passed as individual props and there may be other, non-course metadata props added in too.

src/outline/Chapter.jsx
ormsbee: @abutterworth, @davidjoy: Sorry, this was buried in my mentions. I think we want to move towards "Section" as the terminology for this in general. Chapters are named that way for historical reasons in the code. Unfortunately, "section" is bizarrely overloaded in the system, and I'm not sure what the least confusing thing to do here is.

I'm going to reproduce a comment I made in TNL-7021, in case anyone stumbles across this PR later:

--------

It’s worth noting that the values for chapter and section are currently looked up using the chapter and sequence url_name attributes passed in via URL. Technically, a course is a DAG and could potentially have different routes to the same sequence (though in practice this never happens at the sequence level):

https://github.com/edx/edx-platform/blob/master/lms/djangoapps/courseware/views/index.py#L322-L354

Also, the terminology used in this view is extremely confusing:

https://github.com/edx/edx-platform/blob/master/lms/djangoapps/courseware/views/index.py#L94-L111

The above `get` method passes in `chapter` and `section`. This is extremely confusing because “section” shows up as multiple, very different things in courseware.

Authoring Term | XBlock OLX Tag | Class
-- | -- | --
Section | `<chapter>` | `SectionDescriptor`
Subsection | `<sequential>`<br>Aliases (non-Studio):<br>`<problemset>`<br>  `<videosequence>`<br>Compatibility Shim:<br> `<section>` | `SequenceDescriptor`

So what this view is really talking about when it says chapter/section is Section/Subsection or `SectionDescriptor`/`SequenceDescriptor`, because the original tag for a sequence was `<section>`.

The difference between an alias and a compatibility shim with respect to the table above is that if you put `<problemset>` in your OLX it should reserialize back to `<problemset>` but map to the `SequenceDescriptor`. The `<section>` tag is translated to `<sequential>` as part of the import process. At least that was the intent. Those aliases are almost never used, and I don’t know if it’s broken at this point.

PR #21, opened 2020-03-06
Add preview message for instructors

abutterworth:
See TNL-7106

image

### DIFF
--- a/src/courseware/InstructorToolbar.jsx
+++ b/src/courseware/InstructorToolbar.jsx
@@ -10,8 +10,15 @@ function InstructorToolbar(props) {
 
   return (
     <div className="bg-primary text-light">
-      <div className="container-fluid py-2 d-flex justify-content-end">
-        <a className="text-light" href={props.activeUnitLmsWebUrl}>View unit in the existing experience</a>
+      <div className="container-fluid py-3 d-flex justify-content-end align-items-center">
+        <div>
+          <p className="mb-0 mr-5">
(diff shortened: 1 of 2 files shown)


davidjoy (approved):

src/courseware/InstructorToolbar.jsx
```diff
@@ -10,8 +10,15 @@ function InstructorToolbar(props) {
 
   return (
     <div className="bg-primary text-light">
-      <div className="container-fluid py-2 d-flex justify-content-end">
-        <a className="text-light" href={props.activeUnitLmsWebUrl}>View unit in the existing experience</a>
+      <div className="container-fluid py-3 d-flex justify-content-end align-items-center">
+        <div>
+          <p className="mb-0 mr-5">
+          You are currently previewing the edX learning sequence experience. This preview is to allow for early content testing, especially for custom content blocks, with the goal of ensuring it renders as expected in the next experience. You can learn more through the following <a className="text-white" style={{ textDecoration: 'underline' }} href="https://partners.edx.org/announcements/author-preview-learning-sequence-experience-update" target="blank" rel="noopener">Partner Portal post</a>.  Please report any issues or provide <a className="text-white" style={{ textDecoration: 'underline' }} target="blank" rel="noopener" href="https://forms.gle/R6jMYJNTCj1vgC1D6">feedback using the linked form</a>.
```
stvstnfrd, line 16: Community pet-peeve: over-branding and hard-coding the word `edX` all over the place...

Can we use a different word or remove it completely?
Nine times out of ten, the specifier isn't particularly helpful [1].

Suggestion:

> You are currently previewing the _new_ learning sequence experience.



- [1] This burnt us at Stanford countless times over the years. In particular, I remember when the (then) new Learner Profile stuff went out; we had to edit sooo many instances/variations of "your edX learner profile", as opposed to just saying "your learner profile". The former is confusing if it shows up on the wrong platform, while I don't think the latter sacrifices any useful detail.

abutterworth: Makes sense, I'll make another PR. FWIW, we'll remove this message after we roll out to all learners and verify we have caused any negative impacts.

PR #22, opened 2020-03-06
WIP: Menu component for course breadcrumb jump nav

davidjoy:
This is work in progress for https://openedx.atlassian.net/browse/TNL-7107

We wanted to use the Menu component from frontend-component-header-edx to display our jump nav dropdowns.  I started by converting it to a functional component, but then we decided to deprioritize TNL-7107 for now.  I didn't want to lose the work, though, so here's a branch!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6556,9 +6556,9 @@
       }
     },
     "csstype": {
-      "version": "2.6.7",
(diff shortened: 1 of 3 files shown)


davidjoy:
Going to close this, I don't think we need it.

PR #23, opened 2020-03-06
Disabling next/previous buttons at course beginning and end

davidjoy:
Task: https://openedx.atlassian.net/browse/TNL-7111

At the end of a course, the “next” button at the bottom of the page is replaced with a friendly message. 🤗 

This PR also alphabetizes some props for SequenceNavigation and Sequence, as I was adding two new ones - isFirst and isLast.

### DIFF
--- a/src/courseware/course/SequenceContainer.jsx
+++ b/src/courseware/course/SequenceContainer.jsx
@@ -10,6 +10,7 @@ import PageLoading from '../../PageLoading';
 import Sequence from '../sequence/Sequence';
 import AlertList from '../../user-messages/AlertList';
 import { fetchSequenceMetadata, checkBlockCompletion, saveSequencePosition } from '../../data/course-blocks';
+import { createSequenceIdList } from '../utils';
 
 function SequenceContainer(props) {
   const {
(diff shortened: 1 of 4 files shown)


src/courseware/course/SequenceContainer.jsx
```diff
@@ -68,7 +71,9 @@ function SequenceContainer(props) {
   }, [isTimeLimited]);
 
   const isLoading = !loaded || !unitId || isTimeLimited;
-
+  const isFirst = sequenceIds.indexOf(sequenceId) === 0 && unitIds.indexOf(unitId) === 0;
```
abutterworth: `isFirstUnit`?

src/courseware/course/SequenceContainer.jsx
```diff
@@ -30,10 +31,12 @@ function SequenceContainer(props) {
     position,
     items,
     lmsWebUrl,
+    models,
   } = props;
   const loaded = fetchState === 'loaded';
 
   const unitIds = useMemo(() => items.map(({ id }) => id), [items]);
+  const sequenceIds = useMemo(() => createSequenceIdList(models, courseId), [models, courseId]);
```
abutterworth, line 39: What do you think about doing this work in mapStateToProps?

src/courseware/course/SequenceContainer.jsx
davidjoy: Hm - eh, sure. :) Will change.

abutterworth (approved):

src/courseware/course/SequenceContainer.jsx
davidjoy: Eh, we haven't been doing much derived state in there, so I've been avoiding it.  We also do something very similar in Course.jsx, and it's not connected to redux, so it doesn't have a similar opportunity.  

We were talking about whether or not we wanted to refactor anything for readability - totally open to tweaking our approach a bit when we do that.  For now I'm not sure which is really better.

PR #24, opened 2020-03-09
Log api request errors to New Relic and display load error

abutterworth:
TNL-7114 Logs any failed api request to New Relic. `CourseContainer` now shows an error message instead of a spinner when a course fails to load.

image

### DIFF
--- a/src/courseware/CourseContainer.jsx
+++ b/src/courseware/CourseContainer.jsx
@@ -51,25 +51,39 @@ function CourseContainer(props) {
 
   return (
     <main className="flex-grow-1 d-flex flex-column">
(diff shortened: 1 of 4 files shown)


marcotuts:
Should this prompt users to reload and try again? Anything we should do here beyond the current message?

abutterworth:
Reloading is unlikely to fix an issue like this one. We could update the message to say that our engineers have been notified (via new relic logging). But really there's nothing they can do I think. We can also iterate on this in the future.

ormsbee (approved):

stvstnfrd (approved):

PR #26, opened 2020-03-10
Mobile layout for breadcrumbs and tabs

abutterworth:
Part of mobile layout updates TNL-7072. Breadcrumbs truncate section and subsection titles with ellipsis. Tabs that would overflow are tucked under a "more" dropdown.

image

image


src/tabs/useWindowSize.js
```diff
@@ -0,0 +1,26 @@
...
+
+  useEffect(() => {
+    if (!isClient) {
+      return false;
+    }
+
+    const handleResize = () => {
+      setWindowSize(getSize());
+    };
+    window.addEventListener('resize', handleResize);
```
davidjoy: `window` -> `global`

davidjoy (approved):

src/tabs/useWindowSize.js
abutterworth: Thank you, good catch.

PR #27, opened 2020-03-11
Improve sequence padding and containers on mobile

abutterworth:
TNL-7072. 
- refactors some of the css class container naming, for example: `sequence-container` `sequence` `unit-container` `unit` now provide more control/granularity around padding and styling.
- moved `UnitNavigation` out of the `Sequence` and into its own component because it seemed sensible.
- Also fixes an issue with course tabs where multi-word titles would wrap text.

image

### DIFF
--- a/src/courseware/course/SequenceContainer.jsx
+++ b/src/courseware/course/SequenceContainer.jsx
@@ -8,7 +8,6 @@ import { history } from '@edx/frontend-platform';
 import messages from '../messages';
 import PageLoading from '../../PageLoading';
 import Sequence from '../sequence/Sequence';
-import AlertList from '../../user-messages/AlertList';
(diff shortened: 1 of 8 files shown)


src/courseware/course/SequenceContainer.jsx
```diff
@@ -75,35 +74,32 @@ function SequenceContainer(props) {
   const isLastUnit = sequenceIds.indexOf(sequenceId) === sequenceIds.length - 1
     && unitIds.indexOf(unitId) === unitIds.length - 1;
   return (
-    <>
-      <AlertList topic="sequence" />
```
davidjoy, line 79: This seems like a good change - this AlertList shouldn't be in the container, presumably!

davidjoy (approved):

PR #28, opened 2020-03-13
Mobile responsive sequence navigation

abutterworth:
Relates to TNL-7072. When a sequence navigation would overflow, convert it to a dropdown.
image
image

### DIFF
--- a/src/courseware/sequence/SequenceNavigationTabs.jsx
+++ b/src/courseware/sequence/SequenceNavigationTabs.jsx
@@ -1,0 +1,83 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Dropdown } from '@edx/paragon';
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
+
+import UnitButton from './UnitButton';
+import useIndexOfLastVisibleChild from '../../tabs/useIndexOfLastVisibleChild';
+
+const invisibleStyle = {
+  position: 'absolute',
+  left: 0,
+  pointerEvents: 'none',
+  visibility: 'hidden',
+};
+
+export default function SequenceNavigationTabs({
+  unitIds, activeUnitId, showCompletion, onNavigate,
+}) {
+  const [
+    indexOfLastVisibleChild,
+    containerRef,
+  ] = useIndexOfLastVisibleChild();
+  const shouldDisplayDropdown = indexOfLastVisibleChild === -1;
+
+  return (
+    <div style={{ flexBasis: '100%', minWidth: 0 }}>
+      <div className="sequence-navigation-tabs-container" ref={containerRef}>
+        <div
+          className="sequence-navigation-tabs"
+          style={shouldDisplayDropdown ? invisibleStyle : null}
+        >
+          {unitIds.map(unitId => (
+            <UnitButton
+              key={unitId}
+              unitId={unitId}
+              isActive={activeUnitId === unitId}
+              showCompletion={showCompletion}
+              onClick={onNavigate}
+            />
+          ))}
+
+        </div>
+      </div>
+      {shouldDisplayDropdown && (
+        <Dropdown className="sequence-navigation-dropdown">
+          <Dropdown.Button className="dropdown-button font-weight-normal w-100 border-right-0">
+            <FormattedMessage
+              id="learn.course.sequence.navigation.mobile.menu"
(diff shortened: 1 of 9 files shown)


src/courseware/sequence/SequenceNavigationTabs.jsx
```diff
@@ -0,0 +1,83 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Dropdown } from '@edx/paragon';
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
+
+import UnitButton from './UnitButton';
+import useIndexOfLastVisibleChild from '../../tabs/useIndexOfLastVisibleChild';
+
+const invisibleStyle = {
```
davidjoy: Sine you have two uses of this invisible style thing, is it worth making this the 'official' list of things you need to do to get this sorta Tabs behavior to work?

src/courseware/sequence/SequenceNavigationTabs.jsx
```diff
@@ -0,0 +1,83 @@
...
+};
+
+export default function SequenceNavigationTabs({
+  unitIds, activeUnitId, showCompletion, onNavigate,
+}) {
+  const [
+    indexOfLastVisibleChild,
+    containerRef,
+  ] = useIndexOfLastVisibleChild();
+  const shouldDisplayDropdown = indexOfLastVisibleChild === -1;
```
davidjoy, line 16: Not sure I understand this... does -1 mean, in this case, that one of the children isn't visible so we should switch?  Or that they're _all_ invisible?  Seems like the use case in this component is different than in Tabs, right?  In that at some point we want to just convert to using the dropdown instead of slowly each child to the dropdown as it goes invisible?

src/courseware/sequence/SequenceNavigationTabs.jsx
```diff
@@ -0,0 +1,83 @@
...
+
+  return (
+    <div style={{ flexBasis: '100%', minWidth: 0 }}>
+      <div className="sequence-navigation-tabs-container" ref={containerRef}>
+        <div
+          className="sequence-navigation-tabs"
+          style={shouldDisplayDropdown ? invisibleStyle : null}
```
davidjoy, line 23: Why is it we leave these UnitButtons in the DOM if we're showing the dropdown?  Is that for accessibility?  Naively it feels like it'd be confusing to have two sets of the same nav in the DOM...

Or is it because you need them to be there to know whether or not the dropdown is even needed?  (In which case, should they be rendered as simple divs/containers that can be measured instead of a full fledged UnitButton?

davidjoy (approved):
A few questions, but otherwise looks good.

src/courseware/sequence/SequenceNavigationTabs.jsx
abutterworth: Good call out. -1 means no children fit. So in our case we are just swapping to a dropdown entirely if they don't all fit. I'll rename to `doChildrenOverflowContainer` or something

src/courseware/sequence/SequenceNavigationTabs.jsx
abutterworth: It's the second. They're left in the DOM so that they can be measured. invisible style removes them from screen-readers seeing them, visual display, and pulls them out of the "flow" to no long affect the surround layout. If we render something different our measurements might be off and the logic could get busted

src/tabs/useIndexOfLastVisibleChild.js
```diff
@@ -0,0 +1,49 @@
+import { useLayoutEffect, useRef, useState } from 'react';
+import useWindowSize from './useWindowSize';
+
+export default function useIndexOfLastVisibleChild() {
```
stvstnfrd, line 32: Is it worth a quick note about what this does, like a docstring equivalent?

stvstnfrd (approved):
A question about documentation, but otherwise +1 

This was long a frustration w/ staff at Stanford (in the original app), so yay!

src/tabs/useIndexOfLastVisibleChild.js
abutterworth: Yea good call

src/courseware/sequence/SequenceNavigationTabs.jsx
abutterworth: Added it to the hook

PR #32, opened 2020-03-18
Extensive refactor of application data management.

davidjoy:
- “course-blocks” and “course-meta” are replaced with “courseware” module.  This obscures the difference between the two from the application itself.
- a generic “model-store” module is used to store all course, section, sequence, and unit data in a normalized way, agnostic to the metadata vs. blocks APIs.
- SequenceContainer has been removed, and it’s work is just done in CourseContainer instead.
- UI components are - in general - more responsible for deciding their own behavior during data loading.  If they want to show a spinner or nothing, it’s up to their discretion.
- The API layer is responsible for normalizing data into a form the app will want to use, prior to putting it into the model store.


src/courseware/CourseContainer.jsx
```diff
@@ -1,103 +1,248 @@
...
+  fetchSequence,
+  checkBlockCompletion,
+  saveSequencePosition,
+} from '../data/courseware';
+import { useModel } from '../data/model-store';
+
 import Course from './course/Course';
 
-function CourseContainer(props) {
+function firstSequenceIdSelector(state) {
```
davidjoy: All the hooks and selectors here work, but I'd kind of like to go back over them and make sure 1) they have the right memoization parameters and 2) that the things that are hooks should be hooks, and the things that are selectors should be selectors (vs being the opposite).  Also, maybe these go in a separate file?

src/courseware/CourseContainer.jsx
```diff
@@ -1,103 +1,248 @@
...
-    sequenceId,
     unitId,
   } = match.params;
 
-  useEffect(() => {
-    props.fetchCourseMetadata(courseUsageKey);
-    props.fetchCourseBlocks(courseUsageKey);
-  }, [courseUsageKey]);
+  const dispatch = useDispatch();
+
```
davidjoy: The comment below this was an interesting discovery.  I kept having components further down the tree blow up when I navigated because the sequence wasn't fully loaded yet.  The URL updated immediately and the tree re-rendered before the sequence's loading state was able to even turn back to "loading" from the previous "loaded" - which meant it thought the sequence was fully loaded even though it hadn't even started yet.

src/courseware/CourseContainer.jsx
```diff
@@ -1,103 +1,248 @@
...
+    sequence,
+    match.params.sequenceId,
+    unitId,
+  );
+  useExamRedirect(sequenceId);
 
-  // Whether or not the container is ready to render the Course.
-  const ready = metadataLoaded && courseId && sequenceId;
+  useSavedSequencePosition(courseUsageKey, sequenceId, unitId);
+  const { isFirstUnit, isLastUnit } = useSequenceNavigationMetadata(sequenceId, unitId);
```
davidjoy: Couldn't decide what to call this hook - this one may be better as a selector.

src/courseware/CourseContainer.jsx
```diff
@@ -1,103 +1,248 @@
...
+  const status = {
+    course: courseStatus,
+    section: courseStatus,
+    sequence: sequenceStatus,
+    unit: sequenceStatus,
+  };
 
   return (
     <main className="flex-grow-1 d-flex flex-column">
-      {(() => {
```
davidjoy, line 54: All the different render states are the responsibility of the course itself now, which should make it easier for it to decide what parts of a potential skeleton get shown, etc.

src/courseware/CourseContainer.jsx
```diff
@@ -1,103 +1,248 @@
...
-        return (
-          <PageLoading
-            srMessage={intl.formatMessage(messages['learn.loading.learning.sequence'])}
-          />
-        );
-      })()}
+      <Course
+        status={status}
+        course={course}
+        section={section}
```
davidjoy: The section, sequence, and unit here might be better passed through via connecting components to redux.  Thinking of changing this, since Course itself doesn't really need them all.

src/courseware/InstructorToolbar.jsx
```diff
@@ -51,7 +51,7 @@ const mapStateToProps = (state, props) => {
     return {};
   }
 
-  const activeUnit = state.courseBlocks.blocks[props.unitId];
+  const activeUnit = state.models.units[props.unitId];
```
davidjoy, line 54: The new home for blocks data, normalized all pretty.

src/courseware/course/Course.jsx
```diff
@@ -1,146 +1,126 @@
...
+  unit,
+  status,
+  isFirstUnit,
+  isLastUnit,
+  nextSequenceHandler,
+  previousSequenceHandler,
+  unitNavigationHandler,
+  intl,
 }) {
-  const nextSequenceHandler = useCallback(() => {
```
davidjoy, line 33: These handlers moved up to the container, which I think makes sense.  It's the one that knows about how we load the next sequence/unit.

src/courseware/course/Course.jsx
```diff
@@ -1,146 +1,126 @@
...
+          <CourseBreadcrumbs
+            status={status}
+            course={course}
+            section={section}
+            sequence={sequence}
+          />
+          <AlertList topic="sequence" />
+        </div>
+        <div className="flex-grow-1 d-flex flex-column">
+          <Sequence
```
davidjoy, line 84: Another example of components being responsible for their own spinners/loading indicators - we render the Sequence component immediately... it decides how it wants to load itself if `sequence` is null.  Again, might just pass an ID through here instead and connect `Sequence` to redux.

src/courseware/course/CourseBreadcrumbs.jsx
```diff
@@ -30,27 +33,28 @@ CourseBreadcrumb.defaultProps = {
...
+  section,
+  sequence,
+  status,
 }) {
   const links = useMemo(() => {
-    const sectionId = models[sequenceId].parentId;
-    return [sectionId, sequenceId].map((nodeId) => {
-      const node = models[nodeId];
-      return {
+    if (status.course === 'loaded' && status.sequence === 'loaded') {
```
davidjoy: This should all be about the same, except it needs to be a bit smarter about making sure it doesn't try to render if the course/sequence aren't loaded.

src/courseware/course/CourseBreadcrumbs.jsx
```diff
@@ -30,27 +33,28 @@ CourseBreadcrumb.defaultProps = {
...
+    }
+    return [];
+  }, [status]);
 
   return (
     <nav aria-label="breadcrumb" className="my-4">
       <ol className="list-unstyled d-flex m-0">
         <CourseBreadcrumb
-          url={`${getConfig().LMS_BASE_URL}/courses/${courseUsageKey}/course/`}
+          url={`${getConfig().LMS_BASE_URL}/courses/${course.id}/course/`}
```
davidjoy, line 61: Note - the course block ID is no longer used at all, so courseUsageKey and course.id are identical now.

src/courseware/course/shapes.js
```diff
@@ -0,0 +1,59 @@
+import PropTypes from 'prop-types';
+
+export const courseShape = PropTypes.shape({
```
davidjoy: The whole shape thing is interesting - I couldn't make any of these `isRequired` because the shape object itself may not even exist, and they complain too, regardless of if the shape itself was required.  Kind of a bummer.  

Also, if I start pulling these things out of redux with `useSelector`, these shapes will mostly become unused.

davidjoy: It's a good question of if there's value in validating our model schemas somewhere if we can't do it with PropTypes.

src/courseware/messages.js
```diff
@@ -1,21 +1,11 @@
 import { defineMessages } from '@edx/frontend-platform/i18n';
 
 const messages = defineMessages({
-  'learn.loading.learning.sequence': {
```
davidjoy, line 4: These moved into other messages.js files.

src/courseware/sequence/Sequence.jsx
```diff
@@ -75,119 +72,135 @@ function Sequence({
...
-    if (bannerText) {
-      id = add({
-        code: null,
-        dismissible: false,
-        text: bannerText,
-        type: 'info',
-        topic: 'sequence',
-      });
+    if (status.sequence === 'loaded') {
+      if (sequence.bannerText) {
```
davidjoy: Could collapse these two nested `if`s

src/courseware/sequence/Sequence.jsx
```diff
@@ -75,119 +72,135 @@ function Sequence({
...
+  }, [status.sequence, sequence]);
 
   const [unitHasLoaded, setUnitHasLoaded] = useState(false);
   const handleUnitLoaded = () => {
     setUnitHasLoaded(true);
   };
   useEffect(() => {
-    setUnitHasLoaded(false);
-  }, [activeUnitId]);
+    if (unit) {
```
davidjoy: This component can now render prior to the unit being known.  So we need to check here.

src/courseware/sequence/SequenceNavigation.jsx
```diff
@@ -33,7 +33,7 @@ export default function SequenceNavigation({
       </Button>
 
 
-      {isLocked ? <UnitButton type="lock" isActive /> : (
+      {isLocked ? <UnitButton unitId={activeUnitId} title="" contentType="lock" isActive onClick={() => {}} /> : (
```
davidjoy: Had to satisfy the required props here.

src/data/courseware/slice.js
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable no-param-reassign */
+import { createSlice } from '@reduxjs/toolkit';
+
+export const LOADING = 'loading';
+export const LOADED = 'loaded';
+export const FAILED = 'failed';
+
+const slice = createSlice({
+  name: 'courseware',
+  initialState: {
```
davidjoy: So I separated out all the loading stuff from the models themselves, which made the reducers for each much simpler.  Later, when working with bookmarks, I realized I could just... add a key to the unit for the "bookmarkUpdateState" and update it as the thunk progressed, and that it didn't really need to be anything that got special handling.  You'll see that change in the thunks for handling bookmarks.  In a way that was nice, and makes me want to perhaps move the status stuff here into the models themselves.  Any sequence that hasn't had it's metadata loaded yet, for instance, could default to a status of "loading".

src/data/courseware/thunks.js
```diff
@@ -0,0 +1,212 @@
...
+      {
+        courses, sections, sequences, units,
+      },
+      course,
+    ]) => {
+      dispatch(addModelsMap({
+        modelType: 'courses',
+        modelsMap: courses,
+      }));
+      dispatch(updateModel({
```
davidjoy: I could have merged this into the addModelsMap above, but... eh.

src/data/courseware/thunks.js
```diff
@@ -0,0 +1,212 @@
...
+export function fetchSequence(sequenceId) {
+  return async (dispatch) => {
+    dispatch(fetchSequenceRequest({ sequenceId }));
+    try {
+      const { sequence, units } = await getSequenceMetadata(sequenceId);
+      dispatch(updateModel({
+        modelType: 'sequences',
+        model: sequence,
+      }));
+      dispatch(updateModels({
```
davidjoy: So after getSequenceMetadata is normalized, the units fall out of the sequence data and we just add them separately.  No more 'related models' in the reducers.

src/data/courseware/thunks.js
```diff
@@ -0,0 +1,212 @@
...
+}
+
+export function checkBlockCompletion(courseUsageKey, sequenceId, unitId) {
+  return async (dispatch, getState) => {
+    const { models } = getState();
+    if (models.units[unitId].complete) {
+      return; // do nothing. Things don't get uncompleted after they are completed.
+    }
+
+    try {
```
davidjoy: So this API didn't need loading state or optimistic updates to `complete`, so it only updates the model once.  See saveSequencePosition for optimistic updating, and addBookmark/removeBookmark for keeping track of loading state.

src/data/courseware/thunks.js
```diff
@@ -0,0 +1,212 @@
...
+
+export function addBookmark(unitId) {
+  return async (dispatch) => {
+    // Optimistically update the bookmarked flag.
+    dispatch(updateModel({
+      modelType: 'units',
+      model: {
+        id: unitId,
+        bookmarked: true,
+        bookmarkedUpdateState: 'loading',
```
davidjoy: My slight innovation - the update state is just another property.  We don't need any special handling for property update states this way.

src/data/model-store/slice.js
```diff
@@ -0,0 +1,77 @@
...
+
+function remove(state, modelType, id) {
+  if (state[modelType] === undefined) {
+    state[modelType] = {};
+  }
+
+  delete state[modelType][id];
+}
+
+const slice = createSlice({
```
davidjoy: Generic model handling ahoy!

src/hooks.js
```diff
@@ -26,25 +26,27 @@ export function useLogistrationAlert() {
   }, [authenticatedUser]);
 }
 
-export function useEnrollmentAlert(isEnrolled) {
+export function useEnrollmentAlert(course) {
```
davidjoy: You'll see this sorta change around... we have to often check whether the model exists before accessing properties on it.  I'm not thrilled with this.

src/courseware/course/shapes.js
abutterworth: you can't make the innards required without making the top level shape({}) required?

src/data/courseware/api.js
```diff
@@ -0,0 +1,194 @@
...
+    org: metadata.org,
+    enrollmentStart: metadata.enrollment_start,
+    enrollmentEnd: metadata.enrollment_end,
+    end: metadata.end,
+    start: metadata.start,
+    enrollmentMode: metadata.enrollment.mode,
+    isEnrolled: metadata.enrollment.is_active,
+    userHasAccess: metadata.user_has_access,
+    isStaff: metadata.user_has_staff_access,
+    verifiedMode: camelCaseObject(metadata.verified_mode),
```
abutterworth: thoughts on camelcasing metadata to start with?

src/data/courseware/api.js
```diff
@@ -0,0 +1,194 @@
...
+
+function normalizeBlocks(courseUsageKey, blocks) {
+  const models = {
+    courses: {},
+    sections: {},
+    sequences: {},
+    units: {},
+  };
+  Object.values(blocks).forEach(block => {
+    if (block.type === 'course') {
```
abutterworth: switch?

src/data/courseware/thunks.js
```diff
@@ -0,0 +1,212 @@
...
+  fetchSequenceFailure,
+} from './slice';
+
+
+export function fetchCourse(courseUsageKey) {
+  return async (dispatch) => {
+    dispatch(fetchCourseRequest({ courseUsageKey }));
+    Promise.all([
+      getCourseBlocks(courseUsageKey),
+      getCourseMetadata(courseUsageKey),
```
abutterworth: Just worth noting that block and metadata are parallel but we wait until both complete. May or may not slow us down. You could imagine showing the Courseware header before blocks return for example.

src/data/courseware/thunks.js
abutterworth: very nice!

src/hooks.js
abutterworth: not bad i think

abutterworth (approved):
Nice stuff. It's a lot to fully parse but in general I like the direction. My only concern is a vague feeling that all the hooks in CourseContainer may be unwieldy or rigid as we do things like add a "Section overview" screen before starting the first sequence (not sure if this is happen fwiw).

I think we could prefer using the redux store further down the tree enabling more granular loading states (skeletons and such).

Models stuff is cool. ADR?

src/courseware/course/shapes.js
davidjoy: It's freakin' annoying, yeah.  If the top-level shape isn't required - and is undefined - it still tries to evaluate all the `isRequired`'s inside the shape and complains if they aren't present, when... of course they're not present.  It makes no sense.

src/data/courseware/api.js
davidjoy: I was thinking of it as this being the place where all of that happens in one pass, rather than having the 1) server's representation, 2) an initial camelCasing, and then 3) translation into other names.  Felt like just doing 2 and 3 together would make it easier to reason?

davidjoy: I don't feel terribly strongly about this.

src/data/courseware/api.js
davidjoy: Sure!  🎈

src/data/courseware/thunks.js
davidjoy: Truth... I'm a bit concerned about the added complexity for the UI layer, managing another level of "loading"... hmm.  I wonder how many places we explicitly need _just one_ of the two.

davidjoy:
Cool - yeah, the number of hooks in CourseContainer and using redux more are both things I was a bit concerned about.  Going to take a pass at seeing if I can improve them a bit.

PR #33, opened 2020-03-20
Added calculator component

davestgermain:
[TNL-7138](https://openedx.atlassian.net/browse/TNL-7138)

calc2 2020-03-27 09_22_23

### DIFF
--- a/src/courseware/course/calculator/index.jsx
+++ b/src/courseware/course/calculator/index.jsx
@@ -4,8 +4,11 @@ import { getConfig } from '@edx/frontend-platform';
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 import { FormattedMessage } from '@edx/frontend-platform/i18n';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { faCalculator, faQuestion } from '@fortawesome/free-solid-svg-icons';
+import {
+  faCalculator, faQuestionCircle, faTimesCircle, faEquals,
+} from '@fortawesome/free-solid-svg-icons';
 
+import './calculator.scss';
 
 export default class Calculator extends Component {
   constructor(props) {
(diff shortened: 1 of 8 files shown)


davestgermain:
Calculator icon fixed to bottom of page:
Screen Shot 2020-03-20 at 1 28 52 PM

Calculator uncollapsed:
Screen Shot 2020-03-20 at 1 29 00 PM

Hint popover:
Screen Shot 2020-03-20 at 1 29 21 PM

Equation results:
Screen Shot 2020-03-20 at 1 30 15 PM

davestgermain:
@marcotuts Could you sanity check this new UI? I know it's not perfect...

marcotuts:
This might be an opportunity for us to move the calculator app visually into the sidebar format even if we aren't yet at a v1 of that extension boundary. My sense is that the time it would take to adjust current paragon base button / other patterns for this might be better spent moving this to the future location anyways. 

FYI @abutterworth  on this ^ - If we'd like to do this, perhaps we should merge this but flag it to be not visible for educators until we've done another separate pass at UI styling as a sidebar experience? thoughts on that?

src/courseware/course/calculator/index.jsx
```diff
@@ -0,0 +1,257 @@
+import React, { Component } from 'react';
+import Button from 'react-bootstrap/Button';
+import Collapse from 'react-bootstrap/Collapse';
+import OverlayTrigger from 'react-bootstrap/OverlayTrigger';
+import Popover from 'react-bootstrap/Popover';
```
abutterworth: Button and Collapse can be used from paragon: https://edx.github.io/paragon/components/collapsible

OverlayTrigger and Popover are not things we have in paragon yet... I wonder if we should add them now

src/courseware/course/calculator/index.jsx
```diff
@@ -0,0 +1,257 @@
...
+                </li>
+              </ul>
+            </li>
+            <li className="hint-item" id="hint-list">
+              <table className="table">
+                <thead>
+                  <tr>
+                    <th scope="col">To Use</th>
+                    <th scope="col">Type</th>
+                    <th scope="col">Examples</th>
```
abutterworth: I think we need translations here?

davidjoy (approved):

PR #34, opened 2020-03-23
Allow loading with no unit, and of sequences with no unitIds.

davidjoy:
This requires some defensive programming here and there to let things load in a minimal state.

### DIFF
--- a/src/data/api.js
+++ b/src/data/api.js
@@ -41,14 +41,14 @@ function normalizeBlocks(courseUsageKey, blocks) {
         models.courses[block.id] = {
           id: courseUsageKey,
           title: block.display_name,
-          sectionIds: block.children,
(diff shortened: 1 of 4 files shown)


src/data/api.js
```diff
@@ -41,14 +41,14 @@ function normalizeBlocks(courseUsageKey, blocks) {
         models.courses[block.id] = {
           id: courseUsageKey,
           title: block.display_name,
-          sectionIds: block.children,
+          sectionIds: block.children ? block.children : [],
```
abutterworth: this preferred over `block.children || []`?

abutterworth (approved):

src/data/api.js
davidjoy: Duh - thanks.  I forget that syntax. :)

PR #37, opened 2020-03-25
Only redirect if the user has no access and isn’t staff

davidjoy:
TNL-7129

This adds a third clause to our useAccessDeniedRedirect hook, which makes sure the user doesn’t have staff access (instead of normal, enrolled access) prior to redirecting.

As an aside, this redirection approach - irrespective of this PR - is not great.  The UI mostly renders prior to this redirect happening.  It would be better of this hook returned something that would help prevent the UI from rendering while the redirect is in progress.  As it stands, a redirected user will see a flash of the page content prior to being booted.  Not wonderful.


stvstnfrd (approved):
I distinctly remember you saying this was 15 chars, not 19!
jk, +1 ;)

PR #38, opened 2020-04-01
fix: Fix UserMessagesProvider state references

davidjoy:
See details in code, but this causes UserMessagesProvider to always use the most “recent” version of its messages and nextId state when its callbacks are called.

### DIFF
--- a/src/user-messages/UserMessagesProvider.jsx
+++ b/src/user-messages/UserMessagesProvider.jsx
@@ -4,32 +4,40 @@ import PropTypes from 'prop-types';
 import UserMessagesContext from './UserMessagesContext';
 
 export default function UserMessagesProvider({ children }) {
+  // Note: The callbacks (add, remove, clear) below interact with useState in very subtle ways.
+  // When we call setMessages, we always do so with the function-based form of the handler, making
+  // use of the "current" state and not relying on lexical scoping to access the state exposed
+  // above with useState.  This is very important and allows us to call multiple "add", "remove",
+  // or "clear" functions in a  single render.  Without it, each call to one of the callbacks
+  // references back to the -original- value of messages instead of the most recent, causing them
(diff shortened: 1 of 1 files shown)


src/user-messages/UserMessagesProvider.jsx
```diff
@@ -4,32 +4,40 @@ import PropTypes from 'prop-types';
...
+  // all to override each other.  Last one in would win.
   const [messages, setMessages] = useState([]);
   const [nextId, setNextId] = useState(1);
 
-  const refMessages = useRef(messages);
+  // Because the add, remove, and clear handlers also need to access nextId, we have to do
+  // something a bit different.  There's no way to wait for the "currentNextId" in a setMessages
+  // handler.  The alternative is to update a ref, which will always point to the current value by
+  // its very nature.
+  const refId = useRef(nextId);
```
abutterworth, line 21: We can avoid using useRef and do `setNextId(currentId => currentId++)` below right?

davidjoy: We can't - that's what the comment is about... we need the most recent version of nextId _in the setMessages_ handler... unless we use a ref, there's no way for the closure in `setMessages` to access the "current" nextId.  

Unless I'm missing something. :)

davidjoy: I mean, I could do that for the call to `setNextId`, but that wouldn't get rid of the ref because of the above.  Since I had it, I figured I'd just use it.

abutterworth: *ignores comment describing the exact thing he comments about*

abutterworth (approved):

abutterworth:
FYI the travis failures are caused by an i18n thing that I'm fixing in #39

PR #39, opened 2020-04-01
Improve access control behavior

abutterworth:
**Fixes TNL-7175 - Redirect to course home if a user is not unenrolled and the course is private**
- Require authentication to use the app while course blocks api requires it
- Gracefully handle course blocks api request failures allowing app to proceed to it redirection logic

**Notable changes:**
- selectors related to sequences are more resilient to missing models. In the case the course blocks api returns successfully but empty (in this case of enrolled but course not yet started).
- `fetchCourse` thunk handles failures for fetchCourseMeta and fetchCourseBlocks separately using `Promise.allSettled` instead of `Promise.all`
- `denied` is a new `courseStatus`
- Access denied redirect is done using a component at a new route `redirect/course-home/:courseId`

**Now handles cases** 
- User is unauthenticated > redirect to login
- User is authenticated but not enrolled > redirects to lms course home
- [_Requires server side update_] Current result when an enrolled user attempts to access courseware before the course start date: 
image

### DIFF
--- a/src/courseware/data/selectors.js
+++ b/src/courseware/data/selectors.js
@@ -1,22 +1,20 @@
 /* eslint-disable import/prefer-default-export */
 export function sequenceIdsSelector(state) {
-  try {
-    const { sectionIds } = state.models.courses[state.courseware.courseId];
-    const sequenceIds = sectionIds.flatMap(sectionId => state.models.sections[sectionId].sequenceIds);
-    return sequenceIds;
-  } catch (e) {
+  if (state.courseware.courseStatus !== 'loaded') {
     return [];
   }
+  const { sectionIds } = state.models.courses[state.courseware.courseId];
+  let sequenceIds = [];
+  sectionIds.forEach(sectionId => {
+    sequenceIds = [...sequenceIds, ...state.models.sections[sectionId].sequenceIds];
+  });
+  return sequenceIds;
 }
 
 export function firstSequenceIdSelector(state) {
   if (state.courseware.courseStatus !== 'loaded') {
     return null;
   }
-  try {
-    const sectionId = state.models.courses[state.courseware.courseId].sectionIds[0];
-    return state.models.sections[sectionId].sequenceIds[0];
-  } catch (e) {
-    return null;
-  }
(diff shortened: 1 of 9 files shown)


src/courseware/data/selectors.js
```diff
@@ -1,20 +1,22 @@
 /* eslint-disable import/prefer-default-export */
 export function sequenceIdsSelector(state) {
-  if (state.courseware.courseStatus !== 'loaded') {
+  try {
```
davidjoy: These catch-all try/catch blocks make me nervous... both because they assume an error will be thrown, and because they swallow all errors that may occur.

src/data/slice.js
```diff
@@ -10,7 +10,7 @@ const slice = createSlice({
   initialState: {
     courseStatus: 'loading',
     courseId: null,
-    sequenceStatus: 'loading',
+    sequenceStatus: null,
```
davidjoy: I think I made this 'loading' so that there wouldn't potentially be a flash of content as it's changing from null to loading.  Why's it need to change, out of curiosity?

abutterworth: When a user can't see course content the sequence component is still rendered – in this case it's not actually loading, it's not loaded or failed either.

src/courseware/data/selectors.js
abutterworth: Fair enough about the swallowing of errors. I did this rather than adding more checks. I had half a mind to add lodash.get so we don't need to check that everything exists first. Should I take the check that things are not null route?

davidjoy: The `state.courseware.courseStatus !== 'loaded'` check was a way of avoiding all those "not null" checks, in that if the course isn't loaded, we just avoid the whole thing because we know the data won't be there.  Is the problem that we can now say the course is "loaded" even though the course blocks API has failed?  I think that might be a problem if so... it makes "loaded" kinda meaningless, rather than definitive.  If both of those APIs aren't successful, we should be dead in the water and just fail fast, I think.

davidjoy (approved):

PR #41, opened 2020-04-01
Adding a separate StaffEnrollmentAlert

davidjoy:
This is a separate component because we have no mechanism for passing context/state into these alerts right now, and I’m not sure it’s worth building.  Easier to just use different codes for different situations.

Screen Shot 2020-04-01 at 11 46 28 AM


abutterworth (approved):
lgtm

PR #44, opened 2020-04-03
Adds notes visibility toggle

davestgermain:
[TNL-7162](https://openedx.atlassian.net/browse/TNL-7162)

This adds a notes visibility toggle button, which posts a message to the unit iframe in order to trigger the visibility on the LMS side. Requires https://github.com/edx/edx-platform/pull/23614

This also rearranges the calculator and notes inside a `ContentTools` component.

output


src/courseware/course/tools/data/thunks.js
```diff
@@ -0,0 +1,6 @@
+import { getConfig } from '@edx/frontend-platform';
+
+export default function toggleNotes() {
```
davidjoy: I'm pretty certain I suggested we should add this as a thunk, but now that I read it, I realize this isn't a thunk. :)  I think I'd move this to an `api.js` file instead, just for clarity. Thunks in the frontend apps sorta imply async/promise-based redux actions, neither of which this ended up being in the end.

src/courseware/course/tools/index.jsx
```diff
@@ -0,0 +1,38 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+import Calculator from './calculator';
+import NotesVisibility from './notes';
+import './tools.scss';
+
+
+export default class ContentTools extends Component {
```
davidjoy: FYI - our best practices would be to have this file be named `ContentTools.jsx` and for the `index.js` file of this directory to merely be a place to consolidate all the directory/module's exports, i.e., the things the rest of the app need to use.  

Take a look at src/courseware/course/bookmark/index.js as an example.

davidjoy: Also, this component is a great candidate for a functional component, since it's so simple!

src/courseware/course/tools/notes/index.jsx
```diff
@@ -0,0 +1,67 @@
...
+  injectIntl, intlShape,
+} from '@edx/frontend-platform/i18n';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faPencilAlt } from '@fortawesome/free-solid-svg-icons';
+import toggleNotes from '../data/thunks';
+import messages from './messages';
+import './notes.scss';
+
+
+class NotesVisibility extends Component {
```
davidjoy: Same comment as course/tools/index.jsx above.

src/courseware/course/tools/notes/notes.scss
```diff
@@ -0,0 +1,17 @@
+ .notes-toggle {
+    cursor: pointer;
+    display: inline-block;
```
davidjoy: To the extent that it's possible, it'd be good to use our utility CSS classes on the tag, which would let us get rid of a good chunk of this custom CSS, I think.  https://edx.github.io/paragon/foundations/css-utilities

davidjoy (changes requested):
Looks good!  Just have a few stylistic suggestions, mostly, to help bring it in line with some best practices and all that.

src/courseware/course/tools/data/thunks.js
davestgermain: Ah, got it. I moved it to `api.js`

davestgermain:
@davidjoy I've refactored the modules. I don't know how to totally avoid custom css for those bottom buttons, but I made a common class for both the calculator and notes buttons.

davidjoy (approved):

PR #47, opened 2020-04-09
Support can_load_courseware as legacy boolean and future object

abutterworth:
Relates to https://github.com/edx/edx-platform/pull/23651

### DIFF
--- a/src/data/api.js
+++ b/src/data/api.js
@@ -15,7 +15,8 @@ function normalizeMetadata(metadata) {
     start: metadata.start,
     enrollmentMode: metadata.enrollment.mode,
     isEnrolled: metadata.enrollment.is_active,
-    canLoadCourseware: metadata.can_load_courseware,
+    canLoadCourseware: typeof metadata.can_load_courseware === 'boolean'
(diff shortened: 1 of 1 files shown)


src/data/api.js
```diff
@@ -15,7 +15,8 @@ function normalizeMetadata(metadata) {
     start: metadata.start,
     enrollmentMode: metadata.enrollment.mode,
     isEnrolled: metadata.enrollment.is_active,
-    canLoadCourseware: metadata.can_load_courseware,
+    canLoadCourseware: typeof metadata.can_load_courseware === 'boolean'
+      ? metadata.can_load_courseware : metadata.can_load_courseware.has_access,
```
stvstnfrd, line 19: Is the idea that the API will sometimes return the boolean or rather is this a compatibility/transitional change and the boolean will ultimately be removed?

abutterworth: The latter. This will enable us to deploy our new change without breaking the app in the meantime.

stvstnfrd: :+1:

stvstnfrd (approved):

PR #48, opened 2020-04-09
Redirect users when they cannot access content

abutterworth:
TNL-7171, TNL-7172, TNL-7173, TNL-7174: When a user is denied access to load courseware, redirect them to the appropriate location based upon the error code returned. If the error code is unknown they will be redirected to course home.

Depends on: https://github.com/edx/edx-platform/pull/23651

### DIFF
--- a/src/courseware/CoursewareContainer.jsx
+++ b/src/courseware/CoursewareContainer.jsx
@@ -2,7 +2,7 @@ import React, { useEffect, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import { useSelector, useDispatch } from 'react-redux';
 import { history } from '@edx/frontend-platform';
-
+import { getLocale } from '@edx/frontend-platform/i18n';
 import { useRouteMatch, Redirect } from 'react-router';
 import {
   fetchCourse,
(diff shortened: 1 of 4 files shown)


src/courseware/CoursewareContainer.jsx
```diff
@@ -2,7 +2,7 @@ import React, { useEffect, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import { useSelector, useDispatch } from 'react-redux';
 import { history } from '@edx/frontend-platform';
-
+import { getLocale } from '@edx/frontend-platform/i18n';
```
davidjoy, line 5: This is also available in `AppContext` I believe, FYI.

abutterworth: Will switch

davidjoy (approved):

src/courseware/CoursewareContainer.jsx
abutterworth: Will not switch after discussing

abutterworth:
Waiting for https://github.com/edx/edx-platform/pull/23651 before merge

PR #50, opened 2020-04-14
Add audit access locked banner

stvstnfrd:
when Content Type Gating, aka Feature Based Enrollment is enabled.

This depends on:
- https://github.com/edx/edx-platform/pull/23710


src/courseware/course/sequence/Unit.jsx
```diff
@@ -1,12 +1,24 @@
-import React, { useRef, useEffect, useState } from 'react';
+import React, {
+  Suspense,
```
stvstnfrd, line 2: I'll be honest; I don't fully understand why I had to add all of this `Suspense` logic (is it because the `LockPaywall` is lazily loaded?), but it my component errored without it...

src/courseware/course/sequence/lock-paywall/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './LockPaywall';
```
stvstnfrd, line 1: This component is very similar to the `ContentLock`;
I'm open to thoughts about abstracting/consolidating, now or later.

src/data/api.js
```diff
@@ -111,7 +113,7 @@ export async function getCourseBlocks(courseId) {
   url.searchParams.append('course_id', courseId);
   url.searchParams.append('username', username);
   url.searchParams.append('depth', 3);
-  url.searchParams.append('requested_fields', 'children,show_gated_sections');
+  url.searchParams.append('requested_fields', 'children,show_gated_sections,graded');
```
stvstnfrd, line 117: The API already supported this, as long as we request it.

src/index.scss
```diff
@@ -272,6 +272,31 @@ $primary: #1176B2;
     padding-left: 40px;
     padding-right: 40px;
   }
+  .content-paywall {
```
stvstnfrd: I was largely "just getting this to work", so open to feedback/optimization/cleanup.

src/courseware/course/sequence/Unit.jsx
abutterworth: I'm not sure why that would be needed – I wonder what kind of errors you found without it.

src/courseware/course/sequence/lock-paywall/index.js
abutterworth: I think repeating ourselves here might be ideal since this upgrade messaging is a likely place for future innovation

src/index.scss
abutterworth: In general I think we have been aiming to use as much utility-first CSS as we can to avoid making class names for things that aren't reusable visual objects. In this case even if we use the component more than once, these styles will still likely only ever apply to that component. The [bootstrap utility classes](https://getbootstrap.com/docs/4.4/utilities/spacing/) we can use instead:

on the .content-paywall element
`border border-gray`
`rounded`
`d-flex`
`justify-content-between`
`mt-2`
`px-3 py4`

.fa-lock element
`text-black`
`mr-2`
(font-size, inline style in rems 2rem should work)
`ml-1`

h4 element
`font-weight-bold`
`mb-2`

p element
`mb-0`

img element
`border-0`
(height, use inline style)

It's not critical that we do things this way, but it helps us from bloating our scss into something hard to maintain. We're aware that this creates a challenge for more in-depth theming, and it's something we need to make time to consider how to address.

abutterworth (approved):
Some minor lint errors and take-it-or-leave-it feedback on scss. LGTM.

src/courseware/course/sequence/Unit.jsx
stvstnfrd: I dug a little and confirmed that this is, indeed, caused by the `React.lazy` import.

Swapping that back to a regular import removes the need for the `Suspense`.

The error I get (when lazily loading w/out `Suspense`) is:

```
Error: "A React component suspended while rendering, but no fallback UI was specified.

Add a <Suspense fallback=...> component higher in the tree to provide a loading indicator or placeholder to display.
```

@abutterworth I guess I don't know enough about how React lazily loads components and we appear to use a combination of lazy/non-lazy imports for "optional" components (eg: Alerts are lazily-loaded, while `CourseSock` is not).

I can make it work either way (lazy or not); preferences?

src/index.scss
stvstnfrd: This was all helpful; thank you @abutterworth !
I'll spend some time on the side getting a better grasp of the bootstrap util classes.

I'm not looking to rock the boat (on this point, for now ;) ), so I've adopted all of your recommendations; consistency is king!

src/courseware/course/sequence/Unit.jsx
abutterworth: Lazy sounds good.

davidjoy: Just confirming everything here - Suspense is for fallback rendering while code gets lazy-loaded, basically.  It works hand-in-hand with React.lazy.  We've been trying to lazy load components in here that don't need to be loaded except in certain circumstances.  Practically speaking, most of the ones we've been lazy loading are small things that probably save us a few kilobytes, so not exactly high-value yet. 

If we have multiple pages in the MFE, though, we might choose to lazy-load something like `Course.jsx` or CourseHome.jsx`, which would make more of a difference.

davidjoy (approved):

PR #52, opened 2020-04-21
Add new upgrade offer banner alert

stvstnfrd:
This depends on:
- https://github.com/edx/edx-platform/pull/23779

### DIFF
--- a/src/offer-alert/OfferAlert.jsx
+++ b/src/offer-alert/OfferAlert.jsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Alert from '../user-messages/Alert';
+
+function OfferAlert(props) {
+  const {
+    rawHtml,
+  } = props;
+  return rawHtml && (
+    <Alert type="info">
+      <div dangerouslySetInnerHTML={{ __html: rawHtml }} />
+    </Alert>
+  );
+}
+
+OfferAlert.propTypes = {
+  rawHtml: PropTypes.string.isRequired,
+};
+
+export default OfferAlert;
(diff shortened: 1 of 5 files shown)


src/offer-alert/OfferAlert.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Alert from '../user-messages/Alert';
+
+function OfferAlert(props) {
```
stvstnfrd, line 6: This component is nearly identical to `AccessExpiredAlert`.
While we could consolidate (and have like a `RawHtmlAlert`),
we'll soon want to build this HTML directly in the MFE, at which point,
we'll need our own components anyway...

abutterworth (approved):
LGTM. A thought maybe for a minor future refactor: What do you think about making a `<CourseAlerts courseId={courseId} />` component that encapsulates all the `useXxxxAlert` hooks and the `<AlertList />`?

davidjoy:
I kinda like that idea. 👍

stvstnfrd:
> LGTM. A thought maybe for a minor future refactor: What do you think about making a `<CourseAlerts courseId={courseId} />` component that encapsulates all the `useXxxxAlert` hooks and the `<AlertList />`?

Yeah, I like it!

I also dislike the proliferation of `*-alert` directories popping up in the base `src` directory (of which I'm now 40% responsible!), so those could be grouped together in an `alerts` subdirectory of some sort too.

davidjoy:
Agreed - one thing worth noting is that course home will eventually use some of these alerts, which is why they’re so high up the tree.

PR #53, opened 2020-04-29
TNL-7164, Enroll Now button fix, flash messages, and custom message props

davidjoy:
enrollment_button 2020-04-29 15_05_04

Above:

1. Unenrolled staff user.
2. Clicking the "Enroll Now" button.
3. A spinner is shown while the enrollment request is made.
4. After success, the page refreshes.
5. A "Success" flash message is displayed after refresh.
6. The Success message can be dismissed.
7. After another manual page refresh, the Success message is gone (It's only shown once regardless of whether it's dismissed or not)

This PR fixes TNL-7164 by properly hooking up the Enroll Now button in the EnrollmentAlert and StaffEnrollmentAlert components.

It also adds two new features to user-messages:

1. "flash" messages.  Adding a message with the `addFlash` function will cause it to be shown after page refresh as a flash message.  This effectively lets us pass messages between page refreshes.  It uses localStorage.
2. An AlertList can be handed `customProps` to be passed to its Alerts.  This is useful for passing data to custom alerts.

### DIFF
--- a/src/access-expiration-alert/AccessExpirationAlert.jsx
+++ b/src/access-expiration-alert/AccessExpirationAlert.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import Alert from '../user-messages/Alert';
+import { Alert } from '../user-messages';
 
 function AccessExpirationAlert(props) {
   const {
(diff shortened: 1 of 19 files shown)


abutterworth (approved):

stvstnfrd (approved):

marcotuts:
Curious here about full page level refresh versus the refresh icon / message being just in-message. Is the full app reload essential? Not a critical different / issue here just curious what your thoughts on that are @davidjoy

davidjoy:
Yeah, the full page refresh is a safe way of making sure the user has all the content for enrolled students, and that the iframe isn’t showing a public view. We could refresh all that manually via APIs and refreshing the iframe, but it’d be significantly more error prone than a simple full page refresh.

PR #55, opened 2020-05-01
Refactor iframe message handler

stvstnfrd:
TNL-7187

Original Issue:
>  Refactored as discussed. Sadly, I'm not sure this fixes the issue.
>
> Bittersweetly, I think I've found a way to consistently reproduce the issue locally;
navigating across the boundaries of a subsection.
>
> With and without this patch, if I click "Next" while at the final unit in a subsection, or "Previous" while at the first unit, the navigation seems to fail. Not only does the iframe fail to resize itself, but unless I'm misreading things, it appears that filmstrip _also_ does not update properly; it appears "stuck" on the previous subsection's filmstrip (despite the unit title updating). It's "stuck" until you click next/previous for a second time.
>
> Can others repro this (consistently)?

This was part of a larger issue, resolved in https://github.com/edx/frontend-app-learning/pull/56 .

This now just contains some cleanup found along the way.

### DIFF
--- a/src/courseware/course/sequence/Unit.jsx
+++ b/src/courseware/course/sequence/Unit.jsx
@@ -1,6 +1,7 @@
 import React, {
   Suspense,
   useEffect,
+  useRef,
   useState,
   useLayoutEffect,
 } from 'react';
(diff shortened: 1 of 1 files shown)


stvstnfrd:
Platform companion: https://github.com/edx/edx-platform/pull/23869

src/courseware/course/sequence/Unit.jsx
```diff
@@ -67,27 +68,24 @@ function Unit({
...
     }
-    return () => {
-      global.onmessage = null;
-    };
   }, [id]);
 
+  useEffect(() => {
+    global.addEventListener('message', receiveMessage);
+    return () => global.removeEventListener('message', receiveMessage);
+  }, [receiveMessage]);
```
stvstnfrd: Does this work how I want/expect it to? Or should I be using `id` instead of `receiveMessage`? Or...?

stvstnfrd (commented):
Original issues described above were resolved in:
- https://github.com/edx/frontend-app-learning/pull/56

This PR now contains some event handler refactoring performed during discovery.

@davidjoy @abutterworth

src/courseware/course/sequence/Unit.jsx
```diff
@@ -67,27 +68,24 @@ function Unit({
...
+        if (onLoaded) {
+          onLoaded();
         }
-      };
+      }
     }
-    return () => {
-      global.onmessage = null;
-    };
   }, [id]);
```
davidjoy, line 82: So I think technically this `[id]` array needs to be: 

```
[id, setIframeHeight, hasLoaded, iframeHeight, setHasLoaded, onLoaded]
```

Basically any props or state/callbacks created in earlier hooks should be in that array so that it knows to give you a new copy of the callback with the right lexical scope.

src/courseware/course/sequence/Unit.jsx
```diff
@@ -67,27 +68,24 @@ function Unit({
...
-
-        if (type === 'plugin.resize') {
-          setIframeHeight(payload.height);
-          if (!hasLoaded && iframeHeight === 0 && payload.height > 0) {
-            setHasLoaded(true);
-            if (onLoaded) {
-              onLoaded();
-            }
-          }
+  const receiveMessage = useCallback((event) => {
```
davidjoy:
```
// We use this ref so that we can hold a reference to the currently active event listener.
const messageEventListenerRef = useRef(null);
useEffect(() => {
  function receiveMessage(event) {
    const { type, payload } = event.data;
    if (type === 'plugin.resize') {
      setIframeHeight(payload.height);
      if (!hasLoaded && iframeHeight === 0 && payload.height > 0) {
        setHasLoaded(true);
        if (onLoaded) {
          onLoaded();
        }
      }
    }
  }
  // If we currently have an event listener, remove it.
  if (messageEventListenerRef.current !== null) {
    global.removeEventListener('message', messageEventListenerRef.current)
    messageEventListenerRef.current = null;
  }
  // Now add our new receiveMessage handler as the event listener.
  global.addEventListener('message', receiveMessage);
  // And then save it to our ref for next time.
  messageEventListenerRef.current = receiveMessage;
  // When the component finally unmounts, use the ref to remove the correct handler.
  return () => global.removeEventListener('message', messageEventListenerRef.current);
```

davidjoy: I didn't test this, but think it's correct.  No memory leaks. 🤞

src/courseware/course/sequence/Unit.jsx
davidjoy: See messy code snippet above - working with global data in functional components is hard!

davidjoy (approved):

PR #56, opened 2020-05-04
Use layout effect to avoid iframe pausing React lifecycle

davidjoy:
Fixes TNL-7187 - Adds a no-op useLayoutEffect hook to Unit.jsx to prevent the unit iframe from pausing React’s rendering lifecycle.  Very strange bug - see comments in that file for more detail.

### DIFF
--- a/src/courseware/course/sequence/Unit.jsx
+++ b/src/courseware/course/sequence/Unit.jsx
@@ -1,8 +1,8 @@
 import React, {
   Suspense,
-  useRef,
   useEffect,
   useState,
+  useLayoutEffect,
 } from 'react';
 import PropTypes from 'prop-types';
 import { getConfig } from '@edx/frontend-platform';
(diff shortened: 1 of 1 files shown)


src/courseware/course/sequence/Unit.jsx
```diff
@@ -14,13 +14,44 @@ import PageLoading from '../../../PageLoading';
...
+ * when navigating between sequences, as the UI partially updates leaving the user in a nebulous
+ * state.
+ *
+ * We were able to solve this error by using a layout effect to update some component state, which
+ * executes synchronously on render.  Somehow this forces React to continue it's lifecycle
+ * immediately, rather than waiting for user interaction.  This layout effect could be anywhere in
+ * the parent tree, as far as we can tell - we chose to add a conspicuously 'load bearing' (that's
+ * a joke) one here so it wouldn't be accidentally removed elsewhere.
+ *
+ * If we remove this hook when one of these happens:
```
abutterworth, line 29: It's possible some clever refactor could solve the problem too? Someday we may want to create a minimal reproducible case so we can make a ticket in the react project.

src/courseware/course/sequence/Unit.jsx
```diff
@@ -33,25 +64,29 @@ function Unit({
     enrollmentMode,
   } = course;
 
+  // Do not remove this hook.  See function description.
```
abutterworth, line 67: 🤪

src/courseware/course/sequence/Unit.jsx
```diff
@@ -33,25 +64,29 @@ function Unit({
...
+  useLoadBearingHook(id);
+
   useEffect(() => {
-    global.onmessage = (event) => {
-      const { type, payload } = event.data;
+    if (!global.onmessage) {
+      global.onmessage = (event) => {
```
abutterworth, line 72: Maybe in a subsequent PR it would be good to use the `global.addEventListener('message', fn)` syntax to avoid collisions of multiple iframe listening components?

abutterworth (approved):

src/courseware/course/sequence/Unit.jsx
davidjoy: Good call, yes - I'd forgotten we wanted to do that.

src/courseware/course/sequence/Unit.jsx
davidjoy: That's option number 5 below. 😉

src/courseware/course/sequence/Unit.jsx
```diff
@@ -78,7 +113,6 @@ function Unit({
         <iframe
           id="unit-iframe"
           title={unit.title}
-          ref={iframeRef}
```
stvstnfrd, line 81: Yeah, I had played with removing this locally;
doesn't look like we were using it?

stvstnfrd (approved):

src/courseware/course/sequence/Unit.jsx
stvstnfrd: Oh, I had done this locally; I can try to reapply the change after this lands?

src/courseware/course/sequence/Unit.jsx
davidjoy: Yeah, I think we were once upon a time, but aren't now.

PR #61, opened 2020-05-05
Refactor containers to share more code

mikix:
Specifically, make sure that the header, footer, and tabs are all shared code so that they look the same and don't need to be redefined as we add more tab pages.


davidjoy:
Doesn't the Makefile need to be 755 so it can be executed or something like that?  This is ringing a bell for some reason...

mikix:
> Doesn't the Makefile need to be 755 so it can be executed or something like that? This is ringing a bell for some reason...

You don't generally execute Makefiles. Other repos don't have executable ones (e.g. edx-platform).

davidjoy:
👍 on Makefiles.  Wonder how it ended up that way... the frontend-template-application may need an update too.

src/index.jsx
```diff
@@ -29,7 +30,10 @@ subscribe(APP_READY, () => {
       <UserMessagesProvider>
         <Switch>
           <Route path="/redirect" component={CoursewareRedirect} />
-          <Route path="/course/:courseId/home" component={CourseHomeContainer} />
+          <Route
+            path="/course/:courseId/home"
+            render={(props) => <TabContainer tab="courseware" component={CourseHome} {...props} />}
```
davidjoy: Hmm, can't see a reason for this to be a render-props situation... 

```
<Route path="/course/:courseId/home">
  <TabContainer tab="courseware">
    <CourseHome />
  </TabContainer>
</Route>
```

Would something like that work and be a bit cleaner?  The route match and such can be used via `useRouteMatch` fwiw

mikix: Oh, I forgot that was an option. Can retool.

mikix: OK, retooled.

davidjoy (approved):

davidjoy:
Pre-emptive approval - want to just verify that the enrollment/logistration alerts are called in the right places.

PR #64, opened 2020-05-07
Ensure "Audit Access Expires" banner shows when expected

stvstnfrd:
Make the two rawHtml alerts look near-identical.

### DIFF
--- a/src/access-expiration-alert/hooks.js
+++ b/src/access-expiration-alert/hooks.js
@@ -7,15 +7,15 @@ export function useAccessExpirationAlert(courseId) {
   const course = useModel('courses', courseId);
   const { add, remove } = useContext(UserMessagesContext);
   const [alertId, setAlertId] = useState(null);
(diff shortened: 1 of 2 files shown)


src/access-expiration-alert/hooks.js
```diff
@@ -7,15 +7,15 @@ export function useAccessExpirationAlert(courseId) {
...
-    if (expiredMessage && alertId === null) {
+    if (rawHtml && alertId === null) {
       setAlertId(add({
         code: 'clientAccessExpirationAlert',
         topic: 'course',
-        rawHtml: expiredMessage,
+        rawHtml,
       }));
-    } else if (alertId !== null) {
+    } else if (!rawHtml && alertId !== null) {
```
stvstnfrd, line 18: In making this file near-identical to `src/offer-alert/hooks.js `, this was the difference that jumped out.
We don't want to ever hide the `rawHtml`, as long as it's present.

davidjoy (approved):

PR #65, opened 2020-05-07
Add studio/insights links to staff toolbar

stvstnfrd:
Screenshot_2020-05-07_14-37-21


.env
```diff
@@ -1,6 +1,8 @@
 NODE_ENV='production'
 ACCESS_TOKEN_COOKIE_NAME=null
+ANALYTICS_DASHBOARD_URL=null
 BASE_URL=null
+CMS_BASE=null
```
stvstnfrd: For these two settings, I'm using the equivalent setting name from `edx-platform`.

src/courseware/course/InstructorToolbar.jsx
```diff
@@ -1,37 +1,54 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
+
+import { getConfig } from '@edx/frontend-platform';
```
stvstnfrd, line 5: Is this how I want to be fetching `edx-platform`-like settings?

src/courseware/course/InstructorToolbar.jsx
```diff
@@ -1,37 +1,54 @@
...
+
+  // const url_studio_base = getConfig().CMS_BASE;
+  const url_studio_base = 'http://localhost:18010';
+  let url_studio = url_studio_base + '/course/' + courseId;
+  if (unitId) {
+    url_studio = url_studio_base + '/container/' + unitId;
+  }
+  url_studio = null;
+
+  if (!url_studio && !url_insights) {
```
stvstnfrd: If we aren't displaying either link, bail, since there's nothing else to show yet (no masquerade dropdown).

.env
```diff
@@ -1,6 +1,8 @@
 NODE_ENV='production'
 ACCESS_TOKEN_COOKIE_NAME=null
+ANALYTICS_DASHBOARD_URL=null
```
stvstnfrd: Is this what I need to do for `getConfig`?
And if so, how do these values get set during deployment?

src/courseware/course/InstructorToolbar.jsx
```diff
@@ -1,37 +1,59 @@
...
+  let url_insights;
+  if (url_insights_base) {
+      url_insights = url_insights_base + '/courses';
+      if (courseId) {
+        url_insights += '/' + courseId;
+      }
+  }
+
+  const url_studio_base = getConfig().CMS_BASE;
+  // const url_studio_base = 'http://localhost:18010';
```
stvstnfrd: I have these hard-coded sections here because the `getConfig` doesn't seem to be plumbing through...

stvstnfrd (commented):
Here's a first take at TNL-7152.

The one thing that isn't nailed down yet is the getting/setting of the config variables (for the Studio and Insights URL bases), but everything else works around that. @davidjoy Am I on the right track?

src/courseware/course/InstructorToolbar.jsx
```diff
@@ -1,37 +1,59 @@
...
+  let url_studio;
+  if (url_studio_base) {
+    if (unitId) {
+      url_studio = url_studio_base + '/container/' + unitId;
```
stvstnfrd: Is there a blessed way to build cross-service links?
Within the same django service, we could call `reverse`, but I'm not clear on how, if at all, we do so reproducibly in MFEs.

We could also build the URL on the backend and spit it out via the API, instead of letting the frontend build it :shrug:

.env
stvstnfrd: Ah, it looks like I need a PR to frontend-platform, in the spirit of https://github.com/edx/frontend-platform/pull/77 ?

davidjoy: The frontend-platform PR is only necessary if it's a thing we want/desire all micro-frontends to have by default.  

As for production, environment variables are setup in `edx-internal` here: https://github.com/edx/edx-internal/blob/master/frontends/frontend-app-learning/prod_config.yml

(Note that that's the prod one - stage is a sibling of it)

If we _do_ want ANALYTICS_DASHBOARD_URL to be available to all MFEs, then it can instead be added here:

https://github.com/edx/edx-internal/blob/master/frontends/common/prod_config.yml

... in which case we'll want that frontend-platform PR.

.env
davidjoy: Hmm, I'm not sure which is the greater virtue - making it consistent with the other frontend values, or with another system... for what it's worth, they don't actually share config (see the above comment).  I think I'd lean towards making this .env file consistent/making the values predictable when writing MFEs.

.env
davidjoy: As per the below, maybe this is `ANALYTICS_DASHBOARD_BASE_URL`?

src/courseware/course/InstructorToolbar.jsx
```diff
@@ -1,48 +1,81 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
-import { Collapsible } from '@edx/paragon';
+
+// import { getConfig } from '@edx/frontend-platform';
+
+function getInsightsUrl(courseId) {
+  // const urlBase = getConfig().ANALYTICS_DASHBOARD_URL;
```
davidjoy: Without the frontend-platform PR, you can get custom config into `getConfig()` like this, as demonstrated in the account settings MFE:

https://github.com/edx/frontend-app-account/blob/master/src/index.jsx#L59

src/courseware/course/InstructorToolbar.jsx
```diff
@@ -1,48 +1,81 @@
...
+  const urlBase = 'http://localhost:18000';
+  let urlFull;
+  if (urlBase) {
+    urlFull = `${urlBase}/courses`;
+    if (courseId) {
+      urlFull += `/${courseId}`;
```
davidjoy, line 16: This function should always have a courseId provided to it, I'd guess.

In which case I think the whole function collapses down to:

```
return `${getConfig().ANALYTICS_DASHBOARD_URL}/courses/${courseId}`;
```

I don't feel strongly about it at all, but in general I've just been putting strings like that inline and not bothering with a function or constant unless I'm reusing it. ¯\_(ツ)_/¯

src/courseware/course/InstructorToolbar.jsx
```diff
@@ -1,48 +1,81 @@
...
+  }
+  return urlFull;
+}
+
+function getStudioUrl(courseId, unitId) {
+  // const urlBase = getConfig().CMS_BASE;
+  const urlBase = 'http://localhost:18010';
+  let urlFull;
+  if (urlBase) {
+    if (unitId) {
```
davidjoy, line 26: Unlike the above, `unitId` doesn't always exist yet, so this one may continue to need some conditionals like this. 👍

src/courseware/course/InstructorToolbar.jsx
```diff
@@ -1,48 +1,81 @@
...
   }
-
+  const {
+    courseId,
+    unitId,
+  } = props;
+  const urlInsights = getInsightsUrl(courseId);
+  const urlStudio = getStudioUrl(courseId, unitId);
+  if (!urlStudio && !urlInsights) {
+    return null;
```
davidjoy: I _think_ we agreed at standup yesterday that we still want the links to the legacy experience to hang around for now, because instructors outside the canary are still testing/using both.

davidjoy (changes requested):
See ze comments!  Nothing crazy, I hope. :)

src/courseware/course/InstructorToolbar.jsx
marcotuts: what do we do for the "view unit in studio" button, is that any help at all? (maybe not in MFE land?)

src/courseware/course/InstructorToolbar.jsx
marcotuts: I was going to ask if this PR would remove the existing "view unit in studio" buttons. 

One option is - these buttons could also be added via an edx-platform PR to the instructor bar and removed from all content views for both legacy and MFE? (thoughts on that? )

.env
stvstnfrd: So it looks like we already have a variable `STUDIO_BASE_URL` in the common internal vars.
So I'll reuse that and punt on the moral dilemma :)

src/courseware/course/InstructorToolbar.jsx
stvstnfrd: Thanks for the sample!
Implemented.

.env.development
```diff
@@ -1,5 +1,4 @@
 NODE_ENV='development'
-PORT=2000
```
stvstnfrd, line 2: Just alphabetized.

src/courseware/course/InstructorToolbar.jsx
```diff
@@ -1,48 +1,80 @@
...
-            </Collapsible.Body>
-          </Collapsible.Advanced>
-        </div>
-        <div className="flex-shrink-0">
-          <a className="btn d-block btn-outline-light" href={props.activeUnitLmsWebUrl}>View unit in the existing experience</a>
+          &nbsp;
         </div>
+        {urlLms && (
+          <div className="flex-shrink-0">
+            <a className="btn d-block btn-outline-light" href={urlLms}>View in the existing experience</a>
```
stvstnfrd, line 51: Back by popular demand!

.env
```diff
@@ -13,4 +14,5 @@ ORDER_HISTORY_URL=null
 REFRESH_ACCESS_TOKEN_ENDPOINT=null
 SEGMENT_KEY=null
 SITE_NAME=null
+STUDIO_BASE_URL=
```
stvstnfrd, line 17: Just looping back here from the "outdated" thread:
This variable appears to already exist in our prod/stage secrets, so using this name.

.env
```diff
@@ -4,6 +4,7 @@ BASE_URL=null
 CREDENTIALS_BASE_URL=null
 CSRF_TOKEN_API_PATH=null
 ECOMMERCE_BASE_URL=null
+INSIGHTS_BASE_URL=
```
stvstnfrd, line 7: Since we're not tied to the `edx-platform` convention, I renamed this to be more in line with the existing pattern of `<service>_BASE_URL`.

.env.test
```diff
@@ -1,5 +1,6 @@
+NODE_ENV='test'
 ACCESS_TOKEN_COOKIE_NAME='edx-jwt-cookie-header-payload'
-BASE_URL='localhost:1995'
+BASE_URL='localhost:2000'
```
stvstnfrd, line 3: Updating, per template thread, to keep in sync.

src/courseware/course/InstructorToolbar.jsx
stvstnfrd: #### re: function or not

Here, we do need at least the first `if` check;
we don't want to show any link if the `*_BASE_URL` isn't set.
I'm not a fan of ternary operators, especially with computed/formatted values (harder to debug), so I'll keep the function in use :)

#### Re: the `courseId` check:
Yeah, it should always be set now, but...
Maybe I was over-future-proofing; I figured covering that edge case would also make this "widget" work in a course-unaware situation (and be slightly more resilient).

#### As is, if called without a `courseId`, it will link to your list of all courses on your Insights Dashboard


Alternatively, if we do _not_ want to support this behavior (ie the button not be used/displayed in a course-unaware situation), we could return `falsey` to:

#### "hide" the button if called without a `courseId`.

```js
let urlFull;
if (urlBase) {
  if (courseId) {
    urlFull = `${urlBase}/courses/${courseId}`;
  }
}
```

If we don't check for `courseId` at all (and just interpolate directly) and this is called without `courseId`, this will fail silently with links like:
```
https://localhost/courses/null
```

which is the kind of thing we probably won't notice until we notice ;)

FWIW, the Studio link is _not_ displayed if called without either `courseId` or `unitId`..

My preference would be leave as-is, but I'm also open to hiding the button entirely if misconfigured (called without `courseId` (see above) or even with an empty/`falsey` value. I'd prefer to not leave this edge uncovered :)

### TL;DR: Should we fail or fall-back if this is called without a `courseId`?
- fail: do not show the link/button
- fall-back: show a link/button to the general Insights dashboards (list of all courses)

src/courseware/course/InstructorToolbar.jsx
stvstnfrd: @marcotuts Note: I did remove `unit` from `View [unit] in the existing experience`, to match the text on the other two buttons `View in Studio`, `View in Insights`.
It also makes the long button slightly shorter ;)

Happy to restore the full text or change to something else.

.env
stvstnfrd: Oh, also worth pointing out @davidjoy , defaulting these to `null` appears to set them to the _string_ `null`;
I couldn't figure out why it wasn't coming back `falsey`...

So we may want to circle back here and revisit how we're using default `null` values, in this repo and through the ecosystem (starting w/ the template?).

- Maybe I did something weird?
  - Can you repro?
- Maybe this is just the first it's come up?

src/courseware/course/InstructorToolbar.jsx
stvstnfrd: Screenshot_2020-05-11_14-50-14

stvstnfrd (commented):
Comments addressed.

.env.test
```diff
@@ -9,7 +10,8 @@ LOGIN_URL='http://localhost:18000/login'
 LOGOUT_URL='http://localhost:18000/logout'
 MARKETING_SITE_BASE_URL='http://localhost:18000'
 ORDER_HISTORY_URL='localhost:1996/orders'
+PORT=2000
 REFRESH_ACCESS_TOKEN_ENDPOINT='http://localhost:18000/login_refresh'
-SEGMENT_KEY=null
```
davidjoy, line 13: Huh, looks like this disappeared here and in the dev one?  (I only nit about this because keeping the three files in sync with the same keys is the only way to stay sane when you actually have to debug something related to them. :)

src/courseware/course/InstructorToolbar.jsx
davidjoy: Regarding the existence of `*_BASE_URL` - we should be able to take for granted in our application code that these have been defined.  Until recently it was a fatal error in frontend-platform if one of them came through as `undefined`, today you'll get a console warning for each one that hasn't been set properly.  

As for the courseId, sounds a _bit_ over-future-proofy to me, but I don't feel very strongly.  Might be worth noting that it won't happen in the current client-side implementation, but we want to leave the check in as an accurate representation of what it's capable of.

davidjoy: And if you do the fallback, that'll be the 'correct' behavior and would only mildly inconvenience a user if there was a bug in getting the courseId in, right?  Seems better than hiding it, which stops them dead.  I vote for fallback!

src/courseware/course/InstructorToolbar.jsx
davidjoy: @marcotuts - Do we need the paragraph of text still?  Just sanity-checking.

.env.test
stvstnfrd: I may have subconsciously done that :)

I didn't actually mean to merge it like that, though I think I was thinking:
> I bet we don't need to redefine those since they _should_ inherit from `.env`

:)

Anyway, it's since been restored.

src/courseware/course/InstructorToolbar.jsx
stvstnfrd: Re: fallback

Yeah, exactly. The idea is that worst case (assuming `courseId` is missing in error, not by intent) you're inconvenienced by going to the dashboard and picking out your course, instead of a getting a broken link that doesn't take you where you want (and probably 404s).

src/courseware/course/InstructorToolbar.jsx
stvstnfrd: @davidjoy and @marcotuts Ah yeah, good catch. I think I nuked it in preparation of the masquerade toolbar and didn't revisit. I can restore it for now and then pull it when masquerade bar lands?

src/index.jsx
```diff
@@ -51,6 +52,14 @@ subscribe(APP_INIT_ERROR, (error) => {
 });
 
 initialize({
+  handlers: {
+    config: () => {
+      mergeConfig({
+        'INSIGHTS_BASE_URL': process.env.INSIGHTS_BASE_URL || '',
```
davidjoy: I think, for me, I'd only use an empty string if I expected to do some string manipulation on it even though it's "empty"... otherwise I think I'd expect these to just be set to null.

davidjoy (approved):
I think we still have a small decision to make about allowing INSIGHTS_BASE_URL and STUDIO_BASE_URL to be optional, but in principle this PR is good to go, I think.

src/courseware/course/InstructorToolbar.jsx
marcotuts: meh, ok to remove starting now. leave as is :) thanks!

marcotuts: (if you already restored it that's fine too. it'll go away at the latest with the masquerade work.)

stvstnfrd: Nope, still gone, will stay gone then!

src/index.jsx
stvstnfrd: `null` it is!

PR #66, opened 2020-05-11
Resume from last completed unit

davestgermain:
[TNL-7203](https://openedx.atlassian.net/browse/TNL-7203)

Depends on https://github.com/edx/edx-platform/pull/23956

### DIFF
--- a/src/data/api.js
+++ b/src/data/api.js
@@ -121,6 +121,7 @@ export async function getCourseBlocks(courseId) {
   return normalizeBlocks(courseId, data.blocks);
 }
 
+
 function normalizeSequenceMetadata(sequence) {
   return {
     sequence: {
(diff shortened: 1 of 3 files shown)


src/data/thunks.js
```diff
@@ -105,3 +106,7 @@ export function fetchSequence(sequenceId) {
     }
   };
 }
+
+export function fetchResumeBlock(courseId) {
```
davidjoy: I'd remove this and call `getResumeBlock` directly in the component.  A thunk is only necessary/appropriate for actions being `dispatch`'d to redux.  Since we're not changing redux's state with this request, we don't need a thunk.

src/data/api.js
```diff
@@ -152,3 +153,9 @@ export async function getSequenceMetadata(sequenceId) {
 
   return normalizeSequenceMetadata(data);
 }
+
+export async function getResumeBlock(courseId) {
+  const url = new URL(`${getConfig().LMS_BASE_URL}/api/courseware/resume/${courseId}`);
+  const { data } = await getAuthenticatedHttpClient().get(url.href, {});
+  return data;
```
davidjoy: If you change the keys from `section` and `unit` to `section_id` and `unit_id` in the edx-platform PR, it'd be appropriate to camelCase them here so that the rest of the app sees them as `sectionId` and `unitId.

davidjoy (changes requested):
Few small cleanup things, otherwise looks great!

src/data/thunks.js
davestgermain: Ok, I got rid of the `thunks` indirection.

davidjoy (approved):

PR #67, opened 2020-05-12
Show course-sock only when the API says so

stvstnfrd:
See: https://github.com/edx/edx-platform/pull/23979


davidjoy (approved):

stvstnfrd:
This is ready to merge _after_ https://github.com/edx/edx-platform/pull/23979 goes live.

PR #70, opened 2020-05-19
Finish up basic dates tab support

mikix:
- Drop mock data, call real API instead
- Call course metadata API for general info, not the dates API
- Mark text as translatable
- Add badges and timeline dots, group same-day items

https://openedx.atlassian.net/browse/AA-116


src/dates-tab/utils.jsx
```diff
@@ -0,0 +1,16 @@
...
+  if (a.getMonth() < b.getMonth()) { return -1; }
+  if (a.getMonth() > b.getMonth()) { return 1; }
+  if (a.getDate() < b.getDate()) { return -1; }
+  if (a.getDate() > b.getDate()) { return 1; }
+  return 0;
+}
+
+// item is a date block returned from the API
+function isLearnerAssignment(item) {
+  return item.learnerHasAccess && item.dateType === 'assignment-due-date';
```
Dillon-Dumesnil, line 13: Don't think we need it now, but might want to keep in mind having a constants file at some point if we see this happening too much.

src/dates-tab/Timeline.jsx
```diff
@@ -4,21 +4,69 @@ import { useSelector } from 'react-redux';
...
 
+  // Group date items by day (assuming they are sorted in first place) and add some metadata
+  const groupedDates = [];
+  const now = new Date();
+  let foundNextDue = false;
+  let foundToday = false;
+  courseDateBlocks.forEach(dateInfo => {
+    const parsedDate = new Date(dateInfo.date);
+
+    let hasNextAssignment = false;
```
Dillon-Dumesnil: Can we rename this to `hasDueNextAssignment`? I was confused for a while trying to figure out what the next assignment was.

src/dates-tab/Day.jsx
```diff
@@ -1,34 +1,160 @@
...
+      className: 'text-white',
+    },
+  ];
+  let dotColor = null;
+  const badges = (
+    <>
+      {badgesInfo.map(b => {
+        if (b.shown) {
+          if (!dotColor && !isInFuture) {
+            dotColor = b.bg;
```
Dillon-Dumesnil, line 80: Might be worth a comment here that there is an implicit rule here that the ordering of the `badgesInfo` affects the dot color and the earlier it shows up in `badgesInfo`, the higher its priority for setting the color.

src/dates-tab/Badge.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import './Badge.scss';
+
+export default function Badge({ children, className }) {
```
Dillon-Dumesnil, line 7: Nit: have the second parameter named `classNames` since it technically could be specifying more than one.

Dillon-Dumesnil (approved):
Honestly very minor comments. This is awesome MT. Fantastic job. I appreciate you laying some great groundwork here that we will be able to continue to build on!

src/dates-tab/Badge.jsx
mikix: That's just a react thing. They use the singular.

src/dates-tab/Day.jsx
mikix: Done.

src/dates-tab/Timeline.jsx
mikix: Done.

PR #71, opened 2020-05-20
Create masquerade widget component

stvstnfrd:
on the Staff Instructor Toolbar

### DIFF
--- a/src/courseware/course/sequence/Unit.jsx
+++ b/src/courseware/course/sequence/Unit.jsx
@@ -103,7 +103,7 @@ function Unit({
         isBookmarked={unit.bookmarked}
         isProcessing={unit.bookmarkedUpdateState === 'loading'}
       />
-      { contentTypeGatingEnabled && unit.graded && enrollmentMode === 'audit' && (
+      { contentTypeGatingEnabled && unit.graded && (
         <Suspense
           fallback={(
             <PageLoading
--- a/src/courseware/course/sequence/Unit.jsx
+++ b/src/courseware/course/sequence/Unit.jsx
@@ -62,7 +62,6 @@ function Unit({
   const course = useModel('courses', courseId);
   const {
     contentTypeGatingEnabled,
-    enrollmentMode,
   } = course;
 
   // Do not remove this hook.  See function description.
--- a/src/courseware/course/InstructorToolbar.jsx
+++ b/src/courseware/course/InstructorToolbar.jsx
@@ -4,6 +4,8 @@ import { connect } from 'react-redux';
 
 import { getConfig } from '@edx/frontend-platform';
 
+import MasqueradeWidget from './masquerade-widget';
+
 function getInsightsUrl(courseId) {
   const urlBase = getConfig().INSIGHTS_BASE_URL;
   let urlFull;
@@ -44,7 +46,7 @@ function InstructorToolbar(props) {
     <div className="bg-primary text-light">
       <div className="container-fluid py-3 d-md-flex justify-content-end align-items-center">
         <div className="flex-grow-1">
-          &nbsp;
+          <MasqueradeWidget courseId={courseId} />
         </div>
         {urlLms && (
           <div className="flex-shrink-0">
--- a/src/courseware/course/masquerade-widget/MasqueradeWidget.jsx
+++ b/src/courseware/course/masquerade-widget/MasqueradeWidget.jsx
@@ -1,0 +1,73 @@
+import React, {
+  Component,
+} from 'react';
+import PropTypes from 'prop-types';
+
+import { Dropdown } from '@edx/paragon';
+
+import { getMasqueradeOptions } from '../../../data/api';
+import MasqueradeWidgetOption from './MasqueradeWidgetOption';
+
+class MasqueradeWidget extends Component {
+  constructor(props) {
+    super(props);
+    this.courseId = props.courseId;
+    this.state = {
+      options: [],
+    };
+  }
+
+  componentDidMount() {
+    getMasqueradeOptions(this.courseId).then((data) => {
+      if (data.success) {
+        const options = this.parseAvailableOptions(data);
+        this.setState({
+          options,
+        });
+      } else {
+        console.warn('Unable to get masquerade options', data);
+      }
+    }).catch((response) => {
+      console.error('Unable to get masquerade options', response);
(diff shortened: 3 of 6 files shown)


src/courseware/course/masquerade-widget/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+  Component,
+} from 'react';
+import PropTypes from 'prop-types';
+
+import { Dropdown } from '@edx/paragon';
+
+import { getMasqueradeOptions } from '../../../data/api';
+import parseAvailableOptions from './utils';
+
+class MasqueradeWidget extends Component {
```
davidjoy: Fyi - the convention we've been following is for the index.jsx file for a module to merely define the public API for that module, and for the component to be in a file where the filename matches the component name - i.e., MasqueradeWidget.jsx

src/courseware/course/masquerade-widget/utils.jsx
```diff
@@ -0,0 +1,85 @@
+import React from 'react';
```
davidjoy: Hmm, I think what we might want to do here is to just create a `<MasqueradeOptions>` component that we pass some props into, rather than moving these into a utils file.  The handleClick function, iterating over the available options, rendering the items, that's all classic component stuff.  In MasqueradeWidget, where you're putting `{options}` in the dropdown menu, you'd just put <MasqueradeOptions {...theRightProps} /> in there instead.  Would be much more canonical.  I think we tend to use 'utils' files for random helper functions that are reused in a few places and otherwise don't fit anywhere else - i.e., they're not an api call, or a thunk, a reducer, or a selector, etc.

davidjoy (changes requested):
A few code organization things, mostly... I think there are some more familiar patterns we can use.  Otherwise looks good.

src/courseware/course/masquerade-widget/index.jsx
stvstnfrd: Renamed in upcoming commit.

src/courseware/course/masquerade-widget/utils.jsx
stvstnfrd: Heh, so funny story here...

I totally agree that it's all classic component stuff, so much so that I originally wrote those functions as instance methods on `MasqueradeWidget`!
But after some refactoring, I was never referencing instance state (`this`) in the methods, so I just made them state-less functions (in the `MasqueradeWidget.jsx` file) (I don't know if `eslint` enforces this, but `pylint` _does_ check for instance methods that don't touch `self`). Then I got to thinking, since the class doesn't need to share its state with these functions, and it even only needs one of them, I figured I'd move the helpers to their own module and then just import the single one we need.

That said, I'd prefer to _not_ create a new component just for this; `MasqueradeWidget` is thin as it is.
Let me know if you feel strongly otherwise.

So I can move these back as stateless in `MasqueradeWidget.jsx`.
We _could_ add them to the `class`, but that seems a bit dirty :shrug:

stvstnfrd (commented):
@davidjoy Addressed the file naming issue, left a comment about the `utils` file; let me know what you think.

src/courseware/course/masquerade-widget/utils.jsx
davidjoy: Hmm, a component that doesn't reference `this` is actually a prime example of an old school functional component before hooks. :D  

Put another way - `parseAvailableOption` is already a functional component because it's returning JSX that's getting rendered - it just happens to be in a file named utils.jsx instead of something like "AvailableOptions.jsx".  So... you _did_ create a component just for it! :) 

Now that I'm reading a bit closer, the `handleClick` is relying on data props, which isn't generally done in React.  If it were declared inside the functional component as a handler (with or without the `useCallback` hook), it'd have access to all that via lexical scoping, which is the canonical way that's done.  

Even `isSelected` could do its work via lexical scoping, rather than by passing props in, but I don't feel strongly about that/it may be clearer how it is currently.

But, er, yeah, now that I'm reading, I feel pretty strongly that you already have a component here, masquerading as utils. :)

stvstnfrd: > Put another way - `parseAvailableOption` is already a functional component because it's returning JSX that's getting rendered - it just happens to be in a file named utils.jsx instead of something like "AvailableOptions.jsx". So... you _did_ create a component just for it! :)

Alright, alright, I see it now ;)
 
> Now that I'm reading a bit closer, the `handleClick` is relying on data props, which isn't generally done in React. If it were declared inside the functional component as a handler (with or without the `useCallback` hook), it'd have access to all that via lexical scoping, which is the canonical way that's done.

Soooo, the reason I wanted to remove `handleClick` from the closure was that we'd need to iterate over that in a loop and I'd generally prefer to avoid creating `n` copies of a function with separate closures, than to just have a single function/method.

But I agree, `data-*` props is not the way to do this, old habits and all, so I've turned this into an official component.
I opted for a class-based component, for consistency w/ the other new component here, plus to avoid creating `handleClick` in a loop ;)

> Even `isSelected` could do its work via lexical scoping, rather than by passing props in, but I don't feel strongly about that/it may be clearer how it is currently.

It's now an instance method :)

> But, er, yeah, now that I'm reading, I feel pretty strongly that you already have a component here, masquerading as utils. :)

Reimplemented as `MasqueradeWidgetOption.jsx`.

stvstnfrd (commented):
@davidjoy Alright, no more utils ;)

davidjoy (approved):

src/courseware/course/sequence/Unit.jsx
```diff
@@ -104,7 +104,7 @@ function Unit({
         isBookmarked={unit.bookmarked}
         isProcessing={unit.bookmarkedUpdateState === 'loading'}
       />
-      { contentTypeGatingEnabled && unit.graded && enrollmentMode === 'audit' && (
+      { contentTypeGatingEnabled && unit.graded && (
```
stvstnfrd, line 106: @davidjoy This is the one other change needed here (since your original +1).

The backend now determines entirely whether or not content type gating should be enabled for this course for this user, accounting for masquerade settings.
We no longer want to check the user's enrollment track here because during masquerade it's either:
- misleading (eg: a staff user in the audit track), or
- missing (eg: "Limited Access")

PR #73, opened 2020-05-27
AA-120: Course Tools Widget

ciduarte:


src/data/thunks.js
```diff
@@ -81,15 +91,18 @@ export function fetchCourse(courseId) {
       if (!fetchedMetadata) {
         logError(courseMetadataResult.reason);
       }
+      if (!fetchedTools) {
+        logError(courseToolsResult.reason);
+      }
 
       if (fetchedMetadata) {
-        if (courseMetadataResult.value.canLoadCourseware.hasAccess && fetchedBlocks) {
+        if (courseMetadataResult.value.canLoadCourseware.hasAccess && fetchedBlocks && fetchedTools) {
```
ciduarte: I wasn't sure if this should be blocking to loading the whole page, I assumed yes.

src/course-home/CourseTools.jsx
```diff
@@ -0,0 +1,58 @@
...
+
+
+export default function CourseTools(
+  { courseId },
+) {
+  const {
+    courseTools,
+  } = useModel('courseTools', courseId);
+
+  const renderIcon = (iconClasses) => {
```
ciduarte, line 18: This conversion had to happen somewhere to make the fontawesome class names that get passed in from the Course Tools Plugin Manager compatible with the fontawesome SVG class names we use in the MFE. I went back and forth between basing the icon assignment on the icon class names that came from each course tool (ex. `fa fa-certificate`), or on the course tool id itself (ex. `edx.tool.verified_upgrade`). The first option relies on those icon class names not changing, or else we'd have to update this conversion. The second renders the icon_classes field slightly useless for MFE purposes.

src/course-home/CourseHome.jsx
```diff
@@ -5,11 +5,12 @@ import { Button } from '@edx/paragon';
 import { AlertList } from '../user-messages';
 
 import CourseDates from './CourseDates';
+import CourseTools from './CourseTools';
 import Section from './Section';
 import { useModel } from '../model-store';
 
 // Note that we import from the component files themselves in the enrollment-alert package.
-// This is because Reacy.lazy() requires that we import() from a file with a Component as it's
+// This is because React.lazy() requires that we import() from a file with a Component as it's
```
Dillon-Dumesnil: 👏

src/course-home/CourseTools.jsx
Dillon-Dumesnil: I'll let MT/more front-end savvy people comment on what the best approach is here. Both have tradeoffs that you considered, but I don't have a strong preference one  way or the other.

src/data/api.js
```diff
@@ -55,6 +55,27 @@ export async function getTabData(courseId, tab, version) {
   return camelCaseObject(data);
 }
 
+function normalizeCourseToolData(courseId, courseToolData) {
+  const courseTools = [];
+  if (courseToolData) {
+    courseToolData.forEach(courseTool => {
+      courseTools.push({
+        analytics_id: courseTool.analytics_id,
+        icon_classes: courseTool.icon_classes,
```
Dillon-Dumesnil: camelCase

src/data/api.js
```diff
@@ -55,6 +55,27 @@ export async function getTabData(courseId, tab, version) {
...
+      });
+    });
+  }
+  return { id: courseId, courseTools };
+}
+
+export async function getCourseTools(courseId) {
+  const url = `${getConfig().LMS_BASE_URL}/api/course_home/v1/outline/${courseId}`;
+  const { data } = await getAuthenticatedHttpClient().get(url, {});
+  return normalizeCourseToolData(courseId, data.course_tools);
```
Dillon-Dumesnil: Eh, probably just me, but it feels a little weird to me to pass in `courseId` just to return it. But it might be better than `return { id: courseId, courseTools: normalizeCourseToolData(data.course_tools);`. I dunno, just wanted to mention it in case others share my weirdness

src/data/api.js
```diff
@@ -55,6 +55,27 @@ export async function getTabData(courseId, tab, version) {
...
+        icon_classes: courseTool.icon_classes,
+        title: courseTool.title,
+        url: courseTool.url,
+      });
+    });
+  }
+  return { id: courseId, courseTools };
+}
+
+export async function getCourseTools(courseId) {
```
Dillon-Dumesnil: This is not `getCourseTools` anymore. It is the entire outline.

src/data/thunks.js
Dillon-Dumesnil: Yes, but this should really be `fetchedOutline` which we do want to block on. See my above comment about updating the api call function name as well.

Dillon-Dumesnil (commented):
Solid start. One thing that needs to change before merging is updating the language form courseTools to courseOutline. The API is for the whole outline and we want to treat the returned data as such.

src/data/api.js
ciduarte: Yeah I agree. This was my workaround for the addModel call in the thunks.js file because each model needs an ID. But this will change when we use fetchTab for the Outline Tab instead of this current implementation.

src/data/api.js
ciduarte: I was going to wait to make this change for when we refactor the MFE to rely on the Outline Tab API for all of the Outline Tab data. Thought it might be best for clarity sake to leave it as course tools because that's the only thing being returned at the moment. I'm also happy to change it now!

src/data/api.js
Dillon-Dumesnil: In regards to both your points listed here, is there a reason to not `fetchTab` starting now? I'm perfectly fine only showing the tools data (or potentially temporarily adding in the course blocks call into the `getOutlineTab` function), but I think we should use this PR to lay more of the groundwork of what this is.

src/data/api.js
Dillon-Dumesnil: Yes, I think it should be treated as Outline data and for now, it only has to deal with the tools data that is being returned. But this will set the groundwork for future PRs and being able to add to this more clearly.

src/course-home/CourseHome.jsx
davidjoy: Oops.  Thank you. 😀

src/course-home/CourseTools.jsx
mikix: Oh yeah, I forgot that we have the tool id name... Maybe that is easier to just use and not send `icon_classes`? (especially in light of my comment on your other PR)

src/data/api.js
mikix: Can you basically replace most/all of this method with a call to `camelCaseObject`?

src/course-home/CourseTools.jsx
```diff
@@ -0,0 +1,58 @@
...
+  };
+
+  return (
+    <section>
+      <h4>Course Tools</h4>
+      {courseTools.map((courseTool) => (
+        <div id={courseTool.analytics_id}>
+          <a data-analytics-id={courseTool.analytics_id} href={courseTool.url}>
+            <FontAwesomeIcon icon={renderIcon(courseTool.icon_classes)} className="mr-2" />
+            {courseTool.title}
```
mikix, line 44: The horizontal alignment of the titles is off. Maybe the icons need a constant width.

Screenshot from 2020-06-01 13-04-36

mikix: While you're there, there isn't much vertical space between the Course Tools section and the Upcoming Dates -- maybe add enough there to feel less cramped.

Screenshot from 2020-06-01 13-10-40

src/data/thunks.js
```diff
@@ -27,7 +28,8 @@ export function fetchCourse(courseId) {
     Promise.allSettled([
       getCourseMetadata(courseId),
       getCourseBlocks(courseId),
-    ]).then(([courseMetadataResult, courseBlocksResult]) => {
+      getCourseTools(courseId),
```
davidjoy: Is there a reason this can't be done in the `allSettled` with the above two calls?  Making the `getCourseTools` in the `then` will slow down loading for the whole app by serializing 'em... `allSettled` will happily make all the requests in parallel! 🎉

src/data/thunks.js
```diff
@@ -70,8 +72,16 @@ export function fetchCourse(courseId) {
         }));
       }
 
+      if (courseToolsResult.status === 'fulfilled') {
+        dispatch(addModel({
+          modelType: 'courseTools',
+          model: courseToolsResult.value,
```
davidjoy: Just curious/sanity checking, what does this return value look like?  Is the `value` an object with an `id` key in it?  Looks like it uses a `courseId` as its ID, from reading the `useModel` call above?

davidjoy: (If so that's fine!)

src/data/thunks.js
davidjoy: Further, is this data we envision being used on the courseware page?  If so, cool, if not, we may need a larger refactor to help course home / courseware each get their data separately, rather than adding more to `fetchCourse`.

src/course-home/CourseTools.jsx
ciduarte: Yeah I agree, let's go with using the tool id name and nix sending `icon_classes` through the API.

src/data/thunks.js
ciduarte: `getCourseTools` is being called in the `allSettled`!

We have a ticket that I will be moving on to that addresses just that. We've written `fetchTab` so that we can call `fetchDatesTab`, `fetchOutlineTab` as necessary, rather than calling `fetchCourse` for the Outline Tab. We just don't have an API that returns all of the Outline Tab's data yet, so we will be refactoring this shortly. (FYI: I have another PR open on edx-platform for AA-120 that is the Outline Tab API that we plan to add course blocks to, among other features for the Outline Tab. It currently only returns Course Tool data.)

src/data/thunks.js
ciduarte: Yeah it's currently `{ id: courseId, courseTools: [{}, {}, ...] }`

src/data/api.js
ciduarte: We've briefly talked about this, but if I go forward with refactoring things now without our Outline API returning course blocks, I'll have to push a bandaid solution to pull together all of the necessary pieces for the Outline Tab, just to undo those changes when the API is done. Whereas if I wrap up this Course Tools work, I can start a new ticket to focus adding course blocks to our Outline API and refactor the MFE in concert.

I'm happy to show you what I mean on a call. I've already walked through what changes would be necessary for this to work with MT and we agreed that the refactor can wait.

Dillon-Dumesnil: Okay, I trust that you've done your due diligence here. Thanks for the clarity. Maybe just link the ticket in this PR/thread so we can track. Thanks Carla!

ciduarte: Good call!

src/data/api.js
ciduarte: Yes 🙃

src/course-home/CourseTools.jsx
```diff
@@ -0,0 +1,58 @@
...
+      default:
+        return null;
+    }
+  };
+
+  return (
+    <section className="mb-3">
+      <h4>Course Tools</h4>
+      {courseTools.map((courseTool) => (
+        <div id={courseTool.analyticsId} key={courseTool.analyticsId}>
```
Dillon-Dumesnil: What are these used for?

Dillon-Dumesnil (approved):
Very nice!

src/course-home/CourseTools.jsx
ciduarte: I added these in to satisfy a React warning that said children need a key. Revisiting this, I think I added in the ID by mistake.

PR #74, opened 2020-05-28
Miscellaneous small refactorings

davidjoy:
Strongly suggest reviewing commit-by-commit.

- Merges "courseInfo" model-store modelType back into "courses" - since they're the same object, this keeps them normalized.
- Renaming some files to be more consistent with best practices around sub-modules
- Moving our four alert types into an `alerts` subdirectory.
- Refactoring alert creation into a `useAlert` hook that's part of user-messages.

A lot of this is just stuff I found while preparing to write jest tests for the repository, so it's a bit of a mixed bag, and I don't think it's all worth having separate PRs.


src/data/thunks.js
```diff
@@ -33,17 +33,6 @@ export function fetchCourse(courseId) {
           modelType: 'courses',
           model: courseMetadataResult.value,
         }));
-        dispatch(addModel({
```
davidjoy, line 36: @mikix - FYI.  Happy to discuss if this doesn't make sense, but given that "courseInfo" is just a subset of the data in "courses", it doesn't feel like we need to break this out into a separate model type.  Keeping them in the same model type means the data stays normalized and prevents any potential bugs around them drifting apart.  Not likely anytime soon, but probably good to avoid going down that road unless we need to.

src/user-messages/hooks.js
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable import/prefer-default-export */
+import { useContext, useState, useEffect } from 'react';
+import UserMessagesContext from './UserMessagesContext';
+
+export function useAlert(isVisible, {
```
davidjoy, line 5: This is the common code from all our alert creation hooks.  Those hooks now exist just to determine the value of a "isVisible" parameter which decides if the alert should be shown.  They also pass in any custom data as part of a `payload` parameter.

src/data/thunks.js
mikix: I guess it felt weird to be filling in courses with an incomplete data set -- don't these models get cached so we don't re-request the same course data as we move around the MFE? In which case putting a half model in the store would create a problem later? But maybe things aren't that intelligent yet and it's not a problem.

davidjoy: Ah, yeah, there's a lot here... lemme try to unpack it:

These models are for sharing amongst any part of the app that wants to use it, rather than caching.  When any given page loads, it's going to hit the server for what it knows it needs regardless of what's already in the store - there isn't really any sort of cache expiration policy, though we could build something like that in the future if it mattered.  In lieu of caching, the "top of the tree" (i.e., pages), are generally responsible for ensuring data exists for their children.  

That said, all the course-level data is always being loaded from the course metadata and course blocks endpoints right now, regardless of which page is using it.  (i.e., this `fetchCourse` thunk).  For sequence-level data, for instance, we can take for granted that this fetchCourse thunk has already run so we have basic sequence data (from the blocks), but as we load sequences we always hit the sequence metadata endpoint to ensure we've got what we need.  That additional, detailed data is merged into what's already in "sequences" via the `updateModels` action.

When a component wants to _use_ what's in the model store, it can just pull out whatever it needs - if it needs a subset of what's in the course, it can either just pull the whole object and ignore the parts it doesn't care about, or use a mapStateToProps/selector function to whittle it down to just the stuff it cares about - either is totally fine.  

In the case of courses/courseInfo, courseInfo is a strict subset of what's being put into courses (especially since they're both from the same endpoints).  It doesn't buy us much to denormalize it when we put it into redux - redux best practices say we should try to keep data in the store as normalized as possible.  Instead, we can just snag the parts we care about when we pull it out, a la the above.

mikix: So the MFE doesn't have any single-page-app stuff to avoid re-requesting data? Is that planned? Redux makes that easy, yah? We did that in Publisher, but the Learning MFE does redux a bit different.

davidjoy: We don't have any caching/expiration logic planned. In general I don't think there's too much frequent re-requesting of the same data. So far, anyway. I think the worst would be if a user goes back and forth between two sequences with the next/previous buttons - we'd hit the sequence metadata endpoint on each navigation, and re-load the same two iframes. 

Otherwise, the course blocks and course metadata endpoints are loaded once and a user could navigate through the whole course without it hitting those endpoints again.  

That said, now that we have both the course home and courseware pages in the same MFE, we could pretty easily 'cache' the course metadata and course blocks hits between pages if we determine it's too much of a performance hit or is putting too much load on the server.  That wouldn't be hard, but it _would_ raise the question of when we bother to make that request again... we could save a timestamp of when it happened and decide to re-request after some amount of time has passed.  

@ormsbee has a new "Course Outline API" he's working on to replace course blocks/metadata for the MFE, which promises to be much faster - I think if we were going to do performance optimizations like this, it'd probably be after we had a sense of the load we're putting on the server.  

Anyway, yeah, not planned, but wouldn't be too hard.

davidjoy: Oh, I just realized I misread `fetchTab`.  I now understand what you mean by 'incomplete data set' - I didn't notice it wasn't hitting the blocks endpoint.  Hm.  In courseware we have a similar situation with sequences... the blocks API returns a limited set of sequence data, and the sequence metadata endpoint fleshes it out when we actually try to load that sequence in particular.

It's not a perfect mental model, agreed.  My short answer is that it felt worse to have to pass around twice as many models (course info, course blocks, sequence info, sequence detail or something) than the flaws in the current situation.

davidjoy: Might think about this a bit more... your version is definitely "more correct", but I'd want to be consistent and update courseware to function that way too, and I'm not 100% on whether that's worth the refactoring. 🤔

stvstnfrd (approved):
Thanks for cleaning up the alerts!

src/data/thunks.js
davidjoy: Okay, now that I'm looking, I see that the blocks API only provides `sectionIds` beyond what the course metadata endpoint already provides.  It feels pretty reasonable to leave it as `courses` and add in a dummy `sectionIds: []` array for completeness sake, and if we wanted to be super-awesome we could include a status flag indicating the model hasn't been fully hydrated.  Not that anyone needs it right now.

PR #75, opened 2020-05-29
AA-117 mfe dates tab waffle flag

xnick421:
- redirect back to platform dates tab if waffle flags are not
  enable for the mfe dates tab

### DIFF
--- a/src/data/api.js
+++ b/src/data/api.js
@@ -3,14 +3,11 @@ import { getConfig, camelCaseObject } from '@edx/frontend-platform';
 import { getAuthenticatedHttpClient, getAuthenticatedUser } from '@edx/frontend-platform/auth';
 import { logError } from '@edx/frontend-platform/logging';
 
-function overrideTabUrls(id, tabs) {
-  // "LMS tab slug" to "MFE URL slug" for overridden tabs
-  const tabOverrides = {};
+function normalizeTabUrls(id, tabs) {
+  // If api doesn't return the mfe base url, change tab url to point back to LMS
   return tabs.map((tab) => {
-    let url;
-    if (tabOverrides[tab.slug]) {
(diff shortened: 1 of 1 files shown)


src/data/api.js
```diff
@@ -51,8 +51,14 @@ export async function getCourseMetadata(courseId) {
...
 export async function getTabData(courseId, tab, version) {
   const url = `${getConfig().LMS_BASE_URL}/api/course_home/${version}/${tab}/${courseId}`;
-  const { data } = await getAuthenticatedHttpClient().get(url);
-  return camelCaseObject(data);
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(url);
+    return camelCaseObject(data);
+  } catch (error) {
+    const { httpErrorRequestUrl } = error.customAttributes;
+    const platformBase = httpErrorRequestUrl.split('/api')[0];
```
mikix: This feels unnecessary? The MFE has `${getConfig().LMS_BASE_URL}` (which it uses right above) for this.

mikix (approved):
I was thinking we wouldn't need this. But I guess if we enable the waffle then turn it off, anybody with the old URL would keep seeing the MFE instead of the LMS. So makes sense.

mikix:
Oh... another thing you'll probably want to do MFE-side is support accepting absolute URLs in the `tabs` field from the backend. Right now it always prefixes the `LMS_BASE_URL`

mikix:
Look at `api.js` and in particular the `overrideTabUrls` method which I put it, and which doesn't need to exist (I was thinking we'd have override code on the MFE side, but we won't -- it will all be controlled by backend like you have). So you can refactor that method to just delete the `tabOverrides` object and instead just see if the URL is absolute or relative.

mikix (approved):
Love it

PR #79, opened 2020-06-10
AA-133 mfe dates banner

xnick421:
- dates banner on dates tab

### DIFF
--- a/src/data/api.js
+++ b/src/data/api.js
@@ -40,10 +40,28 @@ function normalizeMetadata(metadata) {
   };
 }
 
+function normalizeCourseHomeCourseMetadata(metadata) {
+  const data = camelCaseObject(metadata);
+  return {
+    ...data,
+    tabs: data.tabs.map(tab => ({
+      slug: tab.tabId,
+      title: tab.title,
+      url: tab.url,
+    })),
+  };
+}
+
 export async function getCourseMetadata(courseId) {
   const url = `${getConfig().LMS_BASE_URL}/api/courseware/course/${courseId}`;
   const { data } = await getAuthenticatedHttpClient().get(url);
   return normalizeMetadata(data);
+}
+
+export async function getCourseHomeCourseMetadata(courseId) {
+  const url = `${getConfig().LMS_BASE_URL}/api/course_home/v1/course_metadata/${courseId}`;
+  const { data } = await getAuthenticatedHttpClient().get(url);
+  return normalizeCourseHomeCourseMetadata(data);
 }
 
 export async function getDatesTabData(courseId, version) {
@@ -207,4 +225,9 @@ export async function getResumeBlock(courseId) {
   const url = new URL(`${getConfig().LMS_BASE_URL}/api/courseware/resume/${courseId}`);
   const { data } = await getAuthenticatedHttpClient().get(url.href, {});
   return camelCaseObject(data);
+}
+
+export async function updateCourseDeadlines(courseId) {
+  const url = new URL(`${getConfig().LMS_BASE_URL}/api/course_experience/v1/reset_course_deadlines`);
+  await getAuthenticatedHttpClient().post(url.href, { course_key: courseId });
 }
--- a/src/data/thunks.js
+++ b/src/data/thunks.js
@@ -5,6 +5,8 @@ import {
   getSequenceMetadata,
   getDatesTabData,
   getOutlineTabData,
+  getCourseHomeCourseMetadata,
+  updateCourseDeadlines,
 } from './api';
 import {
   addModelsMap, updateModel, updateModels, updateModelsMap, addModel,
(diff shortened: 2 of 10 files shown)


src/dates-banner/messages.js
```diff
@@ -0,0 +1,66 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  'datesBanner.datesTabInfoBanner.header': {
```
mikix, line 4: I know you are copying from the TNL team's messages file. But look at what I do in the dates-tab message file for example (and this is also what the react-intl docs do).

Repeating the id as the message key is pointless and makes it harder to use. Instead, you can simply do:

```
defineMessags({
  header: {
    id: 'datesBanner.datesTabInfoBanner.header',
    ...
  }, ...
}
```

Then simply `intl.formatMessage(messages.header)` instead of an akward `intl.formatMessage(messages['datesBanner.datesTabInfoBanner.header'])`.

But after writing all that, I see that you dynamically generate the lookup id. So less good advice for your use case. But leaving this here as general advice.

src/dates-banner/DatesBannerContainer.jsx
```diff
@@ -0,0 +1,68 @@
...
+  const datesBanners = [
+    { name: 'datesTabInfoBanner', shouldDisplay: model === 'dates' && !missedDeadlines && isSelfPaced },
+    { name: 'upgradeToCompleteGradedBanner', shouldDisplay: upgradeToCompleteGraded, upgradeLink: verifiedUpgradeLink },
+    { name: 'upgradeToResetBanner', shouldDisplay: upgradeToReset, upgradeLink: verifiedUpgradeLink },
+    {
+      name: 'resetDatesBanner',
+      shouldDisplay: resetDates,
+      model,
+      courseId,
+    },
```
mikix: Why not this one in one line like the others? And why specify model and courseId?

src/dates-banner/DatesBanner.jsx
```diff
@@ -0,0 +1,61 @@
...
+    intl,
+    name,
+    model,
+    upgradeLink,
+    courseId,
+  } = props;
+
+  const dispatch = useDispatch();
+  const clickHandler = () => {
+    if (name === 'resetDatesBanner') {
```
mikix: Checking name like this is kinda breaking component boundaries. Now this lower down component has to know the name of an upper component.

Could you have maybe... an `onClick` prop passed to `DatesBanner` that gets handled in `DatesBannerContainer`? Maybe as a property in the banner dictionaries.

src/dates-banner/DatesBanner.jsx
```diff
@@ -0,0 +1,61 @@
...
+  return (
+    <div className="banner rounded my-2 p-3 container-fluid">
+      <div className="row justify-content-around">
+        <div className={name === 'datesTabInfoBanner' ? 'col-12' : 'col-9'}>
+          <strong>
+            {intl.formatMessage(messages[`datesBanner.${name}.header`])}
+          </strong>
+          {intl.formatMessage(messages[`datesBanner.${name}.body`])}
+        </div>
+        {name !== 'datesTabInfoBanner' && (
```
mikix: Same comment here. Maybe if you added an onClick prop, you could use whether that's defined or not to decide whether to show a button.

src/data/api.js
```diff
@@ -199,3 +205,9 @@ export async function getResumeBlock(courseId) {
   const { data } = await getAuthenticatedHttpClient().get(url.href, {});
   return camelCaseObject(data);
 }
+
+export async function updateCourseDeadlines(courseId) {
+  const url = new URL(`${getConfig().LMS_BASE_URL}/api/course_experience/v1/reset_course_deadlines`);
+  const { data } = await getAuthenticatedHttpClient().post(url.href, { course_key: courseId });
+  return camelCaseObject(data);
```
mikix: Is there actually data returned from this? If not, might just not want to return anything.

src/data/api.js
```diff
@@ -46,6 +46,12 @@ export async function getCourseMetadata(courseId) {
   return normalizeMetadata(data);
 }
 
+export async function getCourseHomeCourseMetadata(courseId) {
```
mikix, line 61: Oh I thought we were already using this endpoint. I guess not...

src/data/thunks.js
```diff
@@ -109,9 +111,11 @@ export function fetchTab(courseId, tab, version) {
     Promise.allSettled([
       getCourseMetadata(courseId),
       getTabData(courseId, tab, version),
-    ]).then(([courseMetadataResult, tabDataResult]) => {
+      getCourseHomeCourseMetadata(courseId),
```
mikix, line 100: This was originally intended as a replacement for `getCourseMetadata` for tabs I think. You should be able to just one-line replace it above and then skip the rest of the changes to this method.

src/data/thunks.js
```diff
@@ -172,3 +188,13 @@ export function fetchSequence(sequenceId) {
     }
   };
 }
+
+export function resetDeadlines(courseId, tab) {
+  return async (dispatch) => {
+    updateCourseDeadlines(courseId).then(() => {
+      if (tab === 'dates') {
+        dispatch(fetchDatesTab(courseId));
+      }
```
mikix: Rather than having to have a switch statement here, why not just pass down the `fetch` prop to `TabContainer` all the way down to this `resetDeadlines` call? Then it can just call the right fetch method without any knowledge of what the various tabs are.

src/data/api.js
ciduarte: nit: Could we rename this to `getCourseHomeMetadata`? Or is that not specific enough

ciduarte (commented):
Just passing through.. one small nit. MT brought up some good points.

I didn't know that `messages.js` file existed, or that any of that implementation existed for i18n, so glad I browsed. Good learning moment.

src/dates-banner/messages.js
davidjoy: Nice, I hadn't fully thought through that you could simplify it this way.

src/dates-banner/DatesBanner.scss
```diff
@@ -0,0 +1,11 @@
+.banner {
```
davidjoy: If at all possible, we should try to use the Bootstrap utility classes here, especially for borders/background colors/margins.  One, it keeps it all standardized and lookin' nice, and two, it's easily update-able when we do a rebrand or it's themed in Open edX instances.

davidjoy: I'm not sure exactly which colors these are, but hopefully they're in our existing color palette!  It also helps make sure we're meeting our a11y requirements.

davidjoy (changes requested):

davidjoy (approved):

PR #80, opened 2020-06-10
AA-181: Outline Tab Refactor

ciduarte:
- Updated the Outline Tab to fetch course blocks from the Outline API.
- Changed naming conventions to more accurately portray the tab naming scheme
(ex. Outline Tab, Dates Tab, etc.)
- Removed logic from `fetchCourses` that was specific to the Outline Tab

### DIFF
--- a/src/data/api.js
+++ b/src/data/api.js
@@ -46,8 +46,8 @@ export async function getCourseMetadata(courseId) {
   return normalizeMetadata(data);
 }
 
-export async function getTabData(courseId, tab, version) {
-  const url = `${getConfig().LMS_BASE_URL}/api/course_home/${version}/${tab}/${courseId}`;
+export async function getDatesTabData(courseId, version) {
+  const url = `${getConfig().LMS_BASE_URL}/api/course_home/${version}/dates/${courseId}`;
   try {
     const { data } = await getAuthenticatedHttpClient().get(url);
     return camelCaseObject(data);
@@ -61,17 +61,6 @@ export async function getTabData(courseId, tab, version) {
     // the "unexpected error try again" screen to the user.
     return true;
   }
-}
-
-function normalizeOutlineTabData(courseId, courseToolData) {
-  const courseTools = camelCaseObject(courseToolData);
-  return { id: courseId, courseTools };
-}
-
-export async function getOutlineTabData(courseId) {
-  const url = `${getConfig().LMS_BASE_URL}/api/course_home/v1/outline/${courseId}`;
-  const { data } = await getAuthenticatedHttpClient().get(url, {});
-  return normalizeOutlineTabData(courseId, data.course_tools);
 }
 
 function normalizeBlocks(courseId, blocks) {
@@ -161,6 +150,26 @@ export async function getCourseBlocks(courseId) {
   return normalizeBlocks(courseId, data.blocks);
 }
 
+export async function getOutlineTabData(courseId, version) {
+  const url = `${getConfig().LMS_BASE_URL}/api/course_home/${version}/outline/${courseId}`;
+  let { tabData } = {};
+  try {
+    tabData = await getAuthenticatedHttpClient().get(url);
+  } catch (error) {
+    const { httpErrorStatus } = error && error.customAttributes;
+    if (httpErrorStatus === 404) {
+      return window.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/home`);
+    }
+  }
+
+  const {
+    data,
+  } = tabData;
+  const courseBlocks = normalizeBlocks(courseId, data.course_blocks.blocks);
+  const courseTools = camelCaseObject(data.course_tools);
+
+  return { courseTools, courseBlocks };
+}
 
 function normalizeSequenceMetadata(sequence) {
   return {
--- a/src/data/thunks.js
+++ b/src/data/thunks.js
@@ -3,7 +3,7 @@ import {
   getCourseMetadata,
   getCourseBlocks,
   getSequenceMetadata,
-  getTabData,
+  getDatesTabData,
   getOutlineTabData,
 } from './api';
 import {
@@ -28,8 +28,7 @@ export function fetchCourse(courseId) {
     Promise.allSettled([
       getCourseMetadata(courseId),
       getCourseBlocks(courseId),
-      getOutlineTabData(courseId),
-    ]).then(([courseMetadataResult, courseBlocksResult, outlineTabResult]) => {
+    ]).then(([courseMetadataResult, courseBlocksResult]) => {
       if (courseMetadataResult.status === 'fulfilled') {
         dispatch(addModel({
           modelType: 'courses',
@@ -62,16 +61,8 @@ export function fetchCourse(courseId) {
         }));
       }
 
-      if (outlineTabResult.status === 'fulfilled') {
-        dispatch(addModel({
-          modelType: 'outline',
-          model: outlineTabResult.value,
-        }));
-      }
-
       const fetchedMetadata = courseMetadataResult.status === 'fulfilled';
       const fetchedBlocks = courseBlocksResult.status === 'fulfilled';
-      const fetchedOutline = outlineTabResult.status === 'fulfilled';
 
       // Log errors for each request if needed. Course block failures may occur
       // even if the course metadata request is successful
@@ -81,18 +72,15 @@ export function fetchCourse(courseId) {
       if (!fetchedMetadata) {
         logError(courseMetadataResult.reason);
       }
-      if (!fetchedOutline) {
-        logError(outlineTabResult.reason);
-      }
 
       if (fetchedMetadata) {
-        if (courseMetadataResult.value.canLoadCourseware.hasAccess && fetchedBlocks && fetchedOutline) {
+        if (courseMetadataResult.value.canLoadCourseware.hasAccess && fetchedBlocks) {
           // User has access
           dispatch(fetchCourseSuccess({ courseId }));
           return;
         }
         // User either doesn't have access or only has partial access
-        // (can't access course blocks or course outline)
+        // (can't access course blocks)
         dispatch(fetchCourseDenied({ courseId }));
         return;
       }
@@ -103,12 +91,12 @@ export function fetchCourse(courseId) {
   };
 }
 
-export function fetchTab(courseId, tab, version) {
+export function fetchTab(courseId, tab, version, getTabData) {
   return async (dispatch) => {
     dispatch(fetchTabRequest({ courseId }));
     Promise.allSettled([
       getCourseMetadata(courseId),
-      getTabData(courseId, tab, version),
+      getTabData(courseId, version),
     ]).then(([courseMetadataResult, tabDataResult]) => {
       const fetchedMetadata = courseMetadataResult.status === 'fulfilled';
       const fetchedTabData = tabDataResult.status === 'fulfilled';
@@ -149,7 +137,11 @@ export function fetchTab(courseId, tab, version) {
 }
 
 export function fetchDatesTab(courseId) {
-  return fetchTab(courseId, 'dates', 'v1');
+  return fetchTab(courseId, 'dates', 'v1', getDatesTabData);
+}
+
+export function fetchOutlineTab(courseId) {
+  return fetchTab(courseId, 'outline', 'v1', getOutlineTabData);
 }
 
 export function fetchSequence(sequenceId) {
--- a/src/course-home/Section.jsx
+++ b/src/course-home/Section.jsx
@@ -6,9 +6,13 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import SequenceLink from './SequenceLink';
 import { useModel } from '../model-store';
 
-export default function Section({ id, courseId }) {
-  const section = useModel('sections', id);
-  const { title, sequenceIds } = section;
+export default function Section({ courseId, title, sequenceIds }) {
+  const {
+    courseBlocks: {
+      sequences,
+    },
+  } = useModel('outline', courseId);
+
   return (
     <Collapsible.Advanced className="collapsible-card mb-2">
       <Collapsible.Trigger className="collapsible-trigger d-flex align-items-start">
@@ -31,6 +35,7 @@ export default function Section({ id, courseId }) {
             key={sequenceId}
             id={sequenceId}
             courseId={courseId}
+            title={sequences[sequenceId].title}
           />
         ))}
       </Collapsible.Body>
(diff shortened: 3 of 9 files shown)


src/data/thunks.js
```diff
@@ -103,12 +91,12 @@ export function fetchCourse(courseId) {
   };
 }
 
-export function fetchTab(courseId, tab, version) {
+export function fetchTab(courseId, tab, version, getCustomTabData) {
```
ciduarte: Hoping someone can think of a better name than `getCustomTabData`. We already define a `getTabData`, so I couldn't use that here.

src/course-home/OutlineTab.jsx
```diff
@@ -29,9 +29,17 @@ export default function CourseHome() {
...
   } = useModel('courses', courseId);
 
+  const {
+    courseBlocks: {
+      courses,
+      sections,
+    },
+  } = useModel('outline', courseId);
+
+  const { sectionIds } = courses[Object.keys(courses)[0]];
```
Dillon-Dumesnil: This line is wild to me. Can you add a comment or something of what's actually happening here? I know you're pulling out `sectionIds`, but it is very unclear to me how you're doing that.

If this is just my javascript lack of knowledge, maybe just respond to this comment instead with the explanation.

src/data/api.js
```diff
@@ -46,7 +46,7 @@ export async function getCourseMetadata(courseId) {
   return normalizeMetadata(data);
 }
 
-export async function getTabData(courseId, tab, version) {
+export const getTabData = async (courseId, tab, version) => {
```
Dillon-Dumesnil: Is this just the preferred way of creating these functions? I assume it continues to function in the same way.

src/data/api.js
```diff
@@ -161,6 +150,24 @@ export async function getCourseBlocks(courseId) {
   return normalizeBlocks(courseId, data.blocks);
 }
 
+export const getOutlineTabData = async (courseId, tab, version) => {
+  const url = `${getConfig().LMS_BASE_URL}/api/course_home/${version}/${tab}/${courseId}`;
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(url);
+    const courseBlocks = normalizeBlocks(courseId, data.course_blocks.blocks);
+    const courseTools = camelCaseObject(data.course_tools);
+    return { courseTools, courseBlocks };
```
Dillon-Dumesnil: Does all of this need to be inside of the `try`? Maybe you or MT can correct me if I'm wrong, but I think you typically want to keep the `try` block as small as possible. If the part we are concerned about is the request, let's keep that in here, but pull out the normalizing of data since we wouldn't expect that to fail if we have good data.

src/data/api.js
```diff
@@ -161,6 +150,24 @@ export async function getCourseBlocks(courseId) {
...
+  } catch (error) {
+    const { httpErrorStatus } = error && error.customAttributes;
+    if (httpErrorStatus === 404) {
+      return window.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/dates`);
```
Dillon-Dumesnil: This doesn't feel like the right spot to redirect to when trying to load the outline. ;P Be careful of copy-pasta

We should also think about when we get a 404 because the course simply does not exist.. I.e. trying to hit `courses.edx.org/api/course_home/v1/outline/course-v1:Dillons+hella+fake+course`. I wonder if we could end up in a loop here of us trying to redirect to the platform frontend which then tries to come back to the MFE. Might honestly be better to use a different status code (403 maybe) if the MFE is not active. Not a question for you to answer in this PR, but something I want to think about.

src/data/thunks.js
```diff
@@ -149,7 +137,11 @@ export function fetchTab(courseId, tab, version) {
 }
 
 export function fetchDatesTab(courseId) {
-  return fetchTab(courseId, 'dates', 'v1');
+  return fetchTab(courseId, 'dates', 'v1', getTabData);
```
Dillon-Dumesnil: It feels weird that this is passing in `getTabData` rather than a `getDatesTabData`. Maybe that will also fix your above comment about having to use `getCustomTabData`?

Dillon-Dumesnil (commented):
I have a couple comments, but overall is looking very nice. I'm hype to have this work done!

While we are renaming files, it might be good to start following the format we mention in AA-186 ( course-home/outline-tab, course-home/outline-tab/widgets, etc.). In this case, something like `course-home/outline-tab/OutlineTab.jsx`.

src/data/api.js
ciduarte: I can't comment on the _preferred_ way, but this is the way I knew how to do this (ie. pass these functions as parameters). I'm also curious if there's a better syntax.

src/data/api.js
ciduarte: This is also the result of some serious copy-pasta from getTabData. I'm not sure what the answer to this is.

src/course-home/OutlineTab.jsx
mikix: Yeah, I'll second that. Object.keys()[0] is kind of unusual. You're just trying to get any single element back from a dictionary object?

src/data/api.js
ciduarte: Very good catch, shame on me for copying & pasting getTabData over!! I'll change that dates -> home for now while we think through the second piece of this comment.

src/data/thunks.js
ciduarte: Yeah so I wasn't sure about changing `getTabData` to `getDatesTabData` only because I could see us potentially reusing `getTabData` for the other tabs (in the case where just camel casing field names is good enough as normalization). Not sure if MT has any thoughts on that.

src/data/api.js
Dillon-Dumesnil: Until we have an answer on if there is a preferred way, I think consistency is the best way. Let's update these to match the way the rest of the functions are defined in this file

src/data/thunks.js
Dillon-Dumesnil: Maybe rename it to something more specific to that then? `getCamelCaseTabData`? That will also make it more clear to other people without (hopefully) having to dig into what `getTabData` does to understand oh, it's just going to camelCase the response.

src/data/api.js
```diff
@@ -161,6 +150,24 @@ export async function getCourseBlocks(courseId) {
   return normalizeBlocks(courseId, data.blocks);
 }
 
+export const getOutlineTabData = async (courseId, tab, version) => {
```
mikix: This is written with generic arguments (taking a `tab` arg) but specific name (`getOutlineTabData`). That feels at odds.

src/data/api.js
mikix: I would agree with Dillon.

src/data/thunks.js
ciduarte: I like that idea!! And I'll change the `getCustomTabData` to `getTabData`. Thanks

src/data/api.js
```diff
@@ -161,6 +150,24 @@ export async function getCourseBlocks(courseId) {
...
+    return { courseTools, courseBlocks };
+  } catch (error) {
+    const { httpErrorStatus } = error && error.customAttributes;
+    if (httpErrorStatus === 404) {
+      return window.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/dates`);
+    }
+    // async functions expect return values. to satisfy that requirement
+    // we return true here which in turn continues with the normal flow of displaying
+    // the "unexpected error try again" screen to the user.
+    return true;
```
mikix: The return values are all over the place here. You return a dictionary object in the happy path. An undefined object in the 404 case (window.location.replace doesn't return anything). And a bool in the error case. Can you bubble up the error here instead of returning anything? Or at least return an empty object to keep the types constant.

src/data/api.js
ciduarte: I didn't know I could just pass these functions in as parameters how they were originally declared, I'm now realizing I was passing them in incorrectly! So much learning going on. Thanks Dillon

src/data/thunks.js
mikix: So my original suggestion/vision was:
1. Have `getTabData` in `api.js` stop normalizing.
2. Add a `getDatesTabData` that calls `getTabData` and does its own normalizing.
3. Add a `getOutlineTabData` that calls `getTabData` and does its own normalizing.

Now each tab has its own normalizer, but shares the actual make-a-request code. And `fetchDatesTab` and `fetchOutlineTab` just each pass in their own method.

mikix: But I guess you're worried that as we add more tabs, they might *just* need camel casing and it's a lot of boiler to have per-tab methods for each just to do that? I can see that point of view and leave it to your judgement about which is cleaner long term.

mikix: Maybe `getSimpleTabData` or something -- or `generic or something that makes it sound like it's the default/common case.

src/course-home/OutlineTab.jsx
ciduarte: So this line is an intermediate variable which replaces all of that mapping that was happening on line 60. I'm doing the `Object.keys()[0]` here because courses is a list of course block objects, but it only ever has one object which is the root course. Instead of `courses[Object.keys()[0]]`, I tried to do `courses[courseId]` but the syntax of the courseId & key didn't match up.

If it helps, the object looks like this:
`courses: [ { 'insert-course-key-here': { sectionIds: [] }}]`

Now I'm wondering if my concern about returning that root value in the API would apply here, if I could substitute `courses[Object.keys()[0]]` with `courses[root]` for clarity.

mikix: Is it reasonable to reorganize the api data to something that makes more sense then? Like a `course_id` key and a `sections:` key. A `courses` key with only ever one course feels like a mis-typing.

But maybe you are sticking with existing code that generates that format?

Dillon-Dumesnil: Yeah, I wonder if that was just a miss. If these only ever return a single course, I would definitely prefer being able to treat it that way rather than a list of courses, but we only ever expect one.

ciduarte: I realized I lied about courses being a list, it's an object, but the Object.keys() call returns a list of keys. And since courses only has one object, it felt safe calling [0]. Just a clarification, not a justification for that line.

So the API returns a flat structure of blocks, where there's one object of type course and however many of type chapter, sequential, etc. (I changed the depth of this so I think it only goes down to sequential, but it might go down to vertical too... not sure):
```
blocks: [
    {id: 'insert-course-key-here', type='course'},
    {id: 'insert-course-key-here', type='chapter'},
    {id: 'insert-course-id-here', type='sequential'}, ..., {}
]
```

The `normalizeBlocks` method restructures that data before we pass it into the models. When `normalizeBlocks` is done, the data looks like:
```
courses: {},
sections: {},
sequences: {},
units: {},
```

Knowing that, would we want to write our own `normalizeBlocks`? Something that returns:
```
sectionIds: [],
sections: {},
sequences: {},
```

mikix: If normalizeBlocks is giving us a `sections` value, can we just use that?

ciduarte: We have the sections value, but the blocks aren't in the order in which they should be displayed. The sectionIds within a course block maintain that order.

src/data/api.js
ciduarte: I'm realizing this redirect only exists for this case: https://github.com/edx/frontend-app-learning/pull/75 (redirecting back to the platform dates tab if waffle flags aren't enabled for the MFE dates tab)

Is this redirect even necessary for the outline tab? (ie. should I nix the try/catch logic)

Dillon-Dumesnil: I think we will probably still want some sort of error catching and handling, and a redirect to the standard LMS might not be the worse. But yeah, it will be more applicable once we add in the outline waffle and are actually checking that condition in the backend

src/data/api.js
ciduarte: I'm realizing this redirect only exists for this case: https://github.com/edx/frontend-app-learning/pull/75 (redirecting back to the platform dates tab if waffle flags aren't enabled for the MFE dates tab)

Is this redirect even necessary for the outline tab? (ie. should I nix the try/catch logic)

src/data/thunks.js
ciduarte: Update: I went with just changing `fetchTabData` to `fetchDatesTabData`. The optimistic path would be that all the tabs could just use return camelCased data, but we saw that wasn't the case with the Outline Tab. My hunch is to not be optimistic. :slightly_smiling_face:

src/data/api.js
```diff
@@ -161,6 +150,26 @@ export async function getCourseBlocks(courseId) {
...
+  try {
+    tabData = await getAuthenticatedHttpClient().get(url);
+  } catch (error) {
+    const { httpErrorStatus } = error && error.customAttributes;
+    if (httpErrorStatus === 404) {
+      return window.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/home`);
+    }
+  }
+
+  const {
```
ciduarte, line 165: Pulled that logic out of the try block and it resulted in this (due to the scope of the try block). I'd appreciate suggestions on how to make this better.

src/course-home/OutlineTab.jsx
ciduarte: @mikix @Dillon-Dumesnil circling back to this convo, what do you guys think is the best solution for `line 41`?
- Leave it as is & add a comment for clarity
- Pass the root value through the API to replace `courses[Objects.keys()[0]]` with `courses[root]` (this is the value that we talked about removing because we had no use for it at the time)
- Rewrite `normalizeBlocks` to restructure the data for this use case

Dillon-Dumesnil: I'm not super familiar with `normalizeBlocks`, but is there a way to preserve order there so we don't have to rewrite any of it? It feels weird going into it with some sort of order and then losing that in the function.
I don't love passing in root because we do technically already have the data and sending it again just for access feels off.
If we can't preserve order, I would then say either the restructuring of `normalizeBlocks` is fine (assuming it is not being used by anyone but us because we can't be changing that on others) or just using an intermediate value with comments to help make it more clear.

davidjoy (approved):

src/data/api.js
```diff
@@ -161,6 +150,26 @@ export async function getCourseBlocks(courseId) {
...
 
+export async function getOutlineTabData(courseId, version) {
+  const url = `${getConfig().LMS_BASE_URL}/api/course_home/${version}/outline/${courseId}`;
+  let { tabData } = {};
+  try {
+    tabData = await getAuthenticatedHttpClient().get(url);
+  } catch (error) {
+    const { httpErrorStatus } = error && error.customAttributes;
+    if (httpErrorStatus === 404) {
+      return window.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/home`);
```
Dillon-Dumesnil, line 161: @ciduarte I think this URL is incorrect. the `/home` is MFE and the edx-platform frontend uses `/course/`

PR #82, opened 2020-06-15
AA122: Created Dates Widget on course home page

daphneli-chen:
Includes upcoming dates and link to dates tab, widget in sidebar of course outline. 
Screen Shot 2020-06-19 at 1 35 09 PM
Screen Shot 2020-06-19 at 1 35 13 PM


src/course-home/CourseDates.jsx
```diff
@@ -1,41 +1,42 @@
...
-      <div><strong>Course Start:</strong><br />{start}</div>
-      <div><strong>Course End:</strong><br />{end}</div>
-      <div><strong>Enrollment Start:</strong><br />{enrollmentStart}</div>
-      <div><strong>Enrollment End:</strong><br />{enrollmentEnd}</div>
-      <div><strong>Mode:</strong><br />{enrollmentMode}</div>
-      <div>{isEnrolled ? 'Active Enrollment' : 'Inactive Enrollment'}</div>
+      {courseDateBlocks.map((courseDateBlock) => (
+          <DateSummary
+              key = {courseDateBlock.date.toString()}//TODO: find a different unique identifier
+              date = {courseDateBlock.date}
```
Dillon-Dumesnil: Remove the spaces before and after the `=`

src/course-home/CourseDates.jsx
```diff
@@ -1,41 +1,42 @@
...
 export default function CourseDates({
-  start,
-  end,
-  enrollmentStart,
-  enrollmentEnd,
-  enrollmentMode,
-  isEnrolled,
+    courseId,
 }) {
+    //retrieves the courseDataBlocks field received in fetchCourse
```
Dillon-Dumesnil: I don't think we need this comment here. The code in this case is self-documenting

src/course-home/CourseDates.jsx
```diff
@@ -1,41 +1,42 @@
...
+    <section className="mb-3">
       <h4>Upcoming Dates</h4>
-      <div><strong>Course Start:</strong><br />{start}</div>
-      <div><strong>Course End:</strong><br />{end}</div>
-      <div><strong>Enrollment Start:</strong><br />{enrollmentStart}</div>
-      <div><strong>Enrollment End:</strong><br />{enrollmentEnd}</div>
-      <div><strong>Mode:</strong><br />{enrollmentMode}</div>
-      <div>{isEnrolled ? 'Active Enrollment' : 'Inactive Enrollment'}</div>
+      {courseDateBlocks.map((courseDateBlock) => (
+          <DateSummary
```
Dillon-Dumesnil: Is it possible to just pass in the whole `courseDateBlock` object? It might be able to help with having to specify all these parameters here and potentially help make this a bit easier to read. Then you only pass in `courseDateBlock` and `userTimezone`

src/course-home/CourseHome.jsx
```diff
@@ -68,6 +68,7 @@ export default function CourseHome() {
             enrollmentEnd={enrollmentEnd}
             enrollmentMode={enrollmentMode}
             isEnrolled={isEnrolled}
+            courseId={courseId}
```
Dillon-Dumesnil: With the addition of this, you can drop all the other info that used to be passed in.

src/course-home/DateSummary.jsx
```diff
@@ -0,0 +1,69 @@
+import {FontAwesomeIcon} from "@fortawesome/react-fontawesome";
```
Dillon-Dumesnil: Be careful of your indentation. You're switching between 2 spaces and 4 spaces in this file. I believe we just want 2 everywhere.

src/course-home/DateSummary.jsx
```diff
@@ -0,0 +1,69 @@
...
+import {faCalendarAlt} from "@fortawesome/free-regular-svg-icons";
```
Dillon-Dumesnil: Use consistent quotes. I believe we tend to prefer single quotes.

src/course-home/DateSummary.jsx
```diff
@@ -0,0 +1,69 @@
...
+              {...timezoneFormatArgs}
+          /></p>
+        </div>
+            <div className="row ml-4">
+              {titleHtml &&
+              <a className="font-weight-bold mt-2" href={titleHtml}>{title}</a>
+              }
+              {!titleHtml &&
+              <div className="font-weight-bold mt-2">{title}</div>
+              }
```
Dillon-Dumesnil: This would be a good spot for a ternary operator. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator

Given my above comment, I don't think it will be necessary anymore,  but still cool to read about!

src/course-home/DateSummary.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+export default function DateSummary({
+    date,
+    title,
+    description,
+    link,
+    linkText,
+    titleHtml,
+    userTimezone
+                                    }) {
```
Dillon-Dumesnil: Spacing here needs to be fixed.

I also left a comment about this DateSummary parameters above. It's to your discretion though if you want to switch the format.

src/course-home/DateSummary.jsx
```diff
@@ -0,0 +1,69 @@
...
+        <div className="row">
+          <FontAwesomeIcon icon={faCalendarAlt} className="ml-3 mt-1 mr-2" style={{ width: '20px' }}/>
+          <p className="m-0 font-weight-bold"><FormattedDate
+              value={date}
+              day="numeric"
+              month="short"
+              weekday="short"
+              year="numeric"
+              {...timezoneFormatArgs}
+          /></p>
```
Dillon-Dumesnil:
```
  <p className="m-0 font-weight-bold">
    <FormattedDate
      value={date}
      day="numeric"
      month="short"
      weekday="short"
      year="numeric"
      {...timezoneFormatArgs}
    />
  </p>

File: src/course-home/CourseDates.jsx [hunk 10]
```diff
@@ -1,41 +1,42 @@
...
+              key = {courseDateBlock.date.toString()}//TODO: find a different unique identifier
```
Dillon-Dumesnil: Mike Terry used `{item.title + item.date}` to set the key if you'd like to borrow that.

src/course-home/DateSummary.jsx
```diff
@@ -0,0 +1,69 @@
...
+        </div>
+            <div className="row ml-4">
+              {titleHtml &&
+              <a className="font-weight-bold mt-2" href={titleHtml}>{title}</a>
```
Dillon-Dumesnil: `titleHtml` is actually a field we don't want to pull in to the MFE. You should be able to just use logic like this to determine the title to display

```
          const showLink = item.link && isLearnerAssignment(item);
          const title = showLink ? (<u><a href={item.link} className="text-reset">{item.title}</a></u>) : item.title;

File: src/course-home/DateSummary.jsx [hunk 12]
```diff
@@ -0,0 +1,69 @@
...
+
+DateSummary.defaultProps = {
+  date: null,
+  title: null,
+  description: null,
+  link: null,
+  linkText: null,
+  titleHtml: null,
+  userTimezone: null,
+}
```
Dillon-Dumesnil: Include new lines at the end of files. You should be able to configure your editor to automatically do this :)

src/course-home/DateSummary.jsx
```diff
@@ -0,0 +1,69 @@
...
+              <div className="font-weight-bold mt-2">{title}</div>
+              }
+            {description &&
+            <p className="m-0 mt-1">{description}</p>
+            }
+            {link &&
+            <div className="date-summary-link">
+              <a href={link}>{linkText}</a>
+            </div>
+            }
```
Dillon-Dumesnil: I haven't run this locally, but I think we aren't going to want this in the case that we are looking at an assignment because we are using the link for the title in that case.

src/data/api.js
```diff
@@ -63,15 +63,18 @@ export async function getTabData(courseId, tab, version) {
...
+  const userTimezone = camelCaseObject(userTimezoneData)
+  const datesTabLink = camelCaseObject(datesTabLinkData)
+  return { id: courseId, courseTools, courseDateBlocks, userTimezone, datesTabLink };
 }
 
 export async function getOutlineTabData(courseId) {
   const url = `${getConfig().LMS_BASE_URL}/api/course_home/v1/outline/${courseId}`;
   const { data } = await getAuthenticatedHttpClient().get(url, {});
-  return normalizeOutlineTabData(courseId, data.course_tools);
+  return normalizeOutlineTabData(courseId, data.course_tools, data.course_date_blocks, data.user_timezone, data.dates_tab_link);
```
Dillon-Dumesnil: I don't love the way this is growing, but we can revisit this once you have rebased on the refactor Carla did (I can't remember if this changed during that).

I'm thinking we might want to return all of your dates information inside a single key in the backend though and then you will pull it out from there instead of three distinct keys.

Dillon-Dumesnil (commented):
Couple things that we are going to want to change for consistency throughout the MFE, but certainly a good start. 

We will also have to see how the refactor Carla did will impact your changes here, but for the most part, this is looking to be in good shape!

src/data/api.js
```diff
@@ -185,8 +185,10 @@ export async function getOutlineTabData(courseId, version) {
   } = tabData;
   const courseBlocks = normalizeBlocks(courseId, data.course_blocks.blocks);
   const courseTools = camelCaseObject(data.course_tools);
-
-  return { courseTools, courseBlocks };
+  const datesWidget = camelCaseObject(data.dates_widget);
```
Dillon-Dumesnil, line 188: :chefkiss: Yes, love it.

src/course-home/DateSummary.jsx
```diff
@@ -0,0 +1,60 @@
...
+      <div className="row ml-4">
+        {linkedTitle
+        && <a className="font-weight-bold mt-2" href={dateBlock.link}>{dateBlock.title}</a>}
+        {!linkedTitle
+        && <div className="font-weight-bold mt-2">{dateBlock.title}</div>}
+        {dateBlock.description
+        && <p className="m-0 mt-1">{dateBlock.description}</p>}
+        {!linkedTitle
+        && dateBlock.link
+        && <a href={dateBlock.link}>{dateBlock.linkText}</a>}
```
Dillon-Dumesnil: Nit: Can we one or two line these instead? I think eslint has a 100 character limit per line so we need to respect that, but if we can one line, I think that would look nicer. If not, separate nit for moving the `&&` to the top line instead of the bottom. And maybe even indenting the tag too. So something like
```
        {linkedTitle && <a className="font-weight-bold mt-2" href={dateBlock.link}>{dateBlock.title}</a>}
```
Or
```
        {linkedTitle &&
          <a className="font-weight-bold mt-2" href={dateBlock.link}>{dateBlock.title}</a>}

File: src/course-home/DateSummary.jsx [hunk 17]
```diff
@@ -0,0 +1,60 @@
...
+      </div>
+
```
Dillon-Dumesnil: Nit: remove this space.

Dillon-Dumesnil (approved):
Everything looks great! Woooooo!!!!

src/course-home/outline-tab/DateSummary.scss
```diff
@@ -0,0 +1,17 @@
...
+}
+
+.small-shift {
+  margin-left: 1px;
+}
+
+.small-shift-and-basis {
+  margin-left: 1px;
+  flex-basis: 100%;
+}
```
Dillon-Dumesnil: New line at the end of file please

Dillon-Dumesnil (approved):
Looking good! small thing to fix with newlines

src/course-home/outline-tab/DateSummary.scss
```diff
@@ -0,0 +1,17 @@
+.upcoming-title {
```
Dillon-Dumesnil: What is the "upcoming" part of this supposed to mean? I think it might be more clear to just say `date-summary-tittle` and `date-summary-description` or something similar.

src/course-home/outline-tab/DateSummary.scss
```diff
@@ -0,0 +1,17 @@
...
+  margin-left: 1px;
+}
+
+.upcoming-description {
+  flex-basis: 100%;
+}
+
+.small-shift {
+  margin-left: 1px;
+}
```
Dillon-Dumesnil: It looks like you could probably combine all these scss changes into the title and description classes. 
Title just becomes 
```
.upcoming-title {
  flex-basis: 100%;
  margin-left: 2px;
}
```
and description becomes
```
.upcoming-description {
  flex-basis: 100%;
  margin-left: 1px;
```
and then you might just need to add in one for your link as well.

If the styling is off with those changes, then I would at least say that these class names shouldn't be so specific to what they are doing and rather to the element they are working on.
}

src/course-home/outline-tab/DateSummary.jsx
```diff
@@ -0,0 +1,67 @@
...
+        </div>
+        <div className="small-shift-and-basis">
+          {dateBlock.description
+          && <div className="small-shift m-0 mt-1">{dateBlock.description}</div>}
+        </div>
+        <div>
+          {!linkedTitle
+          && dateBlock.link
+          && <a href={dateBlock.link} className="small-shift upcoming-link">{dateBlock.linkText}</a>}
+        </div>
```
Dillon-Dumesnil: For these wrapping divs now, I think we would actually want them inside of our checking condition (`dateBlock.description &&...` and `!linkedTitle && dateBlock.link &&...`). Otherwise they'll show in the HTML and just be empty if those conditions aren't true.

I would also take this time to advocate for `!linkedTitle && dateBlock.link` being on the same line

Dillon-Dumesnil (commented):
Couple of comments about the styling classes and code formatting, but woo! It's super close! :)

Ahernkb (approved):
Looks great!

src/course-home/outline-tab/DateSummary.jsx
```diff
@@ -0,0 +1,64 @@
...
+          {linkedTitle
+          && <div className="font-weight-bold mt-2"><a href={dateBlock.link}>{dateBlock.title}</a></div>}
+          {!linkedTitle
+          && <div className="font-weight-bold mt-2">{dateBlock.title}</div>}
+        </div>
+        {dateBlock.description
+          && <div className="date-summary-description"><div className="m-0 mt-1">{dateBlock.description}</div></div>
+        }
+        {!linkedTitle && dateBlock.link
+          && <a href={dateBlock.link} className="description-link">{dateBlock.linkText}</a>
```
Dillon-Dumesnil: Nit: Move the `&&` to the top line

src/course-home/outline-tab/DateSummary.jsx
```diff
@@ -0,0 +1,64 @@
...
+      </div>
+      <div className="row ml-4 px-2">
+        <div className="date-summary-title">
+          {linkedTitle
+          && <div className="font-weight-bold mt-2"><a href={dateBlock.link}>{dateBlock.title}</a></div>}
+          {!linkedTitle
+          && <div className="font-weight-bold mt-2">{dateBlock.title}</div>}
+        </div>
+        {dateBlock.description
+          && <div className="date-summary-description"><div className="m-0 mt-1">{dateBlock.description}</div></div>
```
Dillon-Dumesnil: Much cleaner class names!

Dillon-Dumesnil (approved):
Fix the quality error and I think we are looking good!

src/course-home/outline-tab/DateSummary.jsx
```diff
@@ -0,0 +1,64 @@
...
+        }
+        {!linkedTitle && dateBlock.link
```
ciduarte, line 39: nit: Is there a reason for this logic to check `!linkedTitle`? This is technically asking for this to be true `dateBlock.link && !isLearnerAssignment(dateBlock) && dateBlock.link` (ie. checking `dateBlock.link == true` twice). For clarity, I wonder if we should switch this to be `!isLearnerAssignment(dateBlock) && dateBlock.link`

src/course-home/outline-tab/DateSummary.jsx
```diff
@@ -0,0 +1,64 @@
...
+            day="numeric"
+            month="short"
+            weekday="short"
+            year="numeric"
+            {...timezoneFormatArgs}
+          />
+        </div>
+      </div>
+      <div className="row ml-4 px-2">
+        <div className="date-summary-title">
```
ciduarte: I also made a comment in the scss but it seems like we're only using this div and the `date-summary-description` div to apply the `margin-left: 2px; flex-basis: 100%`. What would happen if we condensed that css into `line 30`? Either with inline styling (`style={{margin-left: 2px; flex-basis: 100%}}`) or by condensing that css class to be called `date-summary-text` and adding it to the className.

Maybe you did it this way on purpose because the margin-left and flex-basis didn't play nice with Bootstrap.

src/course-home/outline-tab/DateSummary.scss
```diff
@@ -0,0 +1,13 @@
+.date-summary-title {
+  margin-left: 2px;
+  flex-basis: 100%;
+}
+
+.date-summary-description {
+  margin-left: 2px;
+  flex-basis: 100%;
```
ciduarte: Curious if we can condense these into one class to avoid duplicate styling declarations. If the functionality is the same, maybe it would be better to name these `.date-summary-text`.

ciduarte (commented):
Looks good! 😄Just a few small styling considerations.

another nit: Curious about the files that have been moved into `outline-tab` that aren't relevant to this ticket. If this ticket is going to cover some refactoring, `CourseTools.jsx` and `CourseDates.jsx` should be inside of `outline-tab/widgets`. Not a huge deal, because they'll be moved over anyways in my refactor (AA-186). Just calling that out.

PR #87, opened 2020-06-16
fix: Ensure lmsWebUrl is loaded in useExamRedirect

davidjoy:
The sequence.lmsWebUrl variable is loaded as part of the course blocks API.  The status of that API’s request is stored in courseStatus.

The useEffect hook in useExamRedirect didn’t ensure that courseStatus was equal to “loaded”.  This meant that if the sequence loaded first, it might attempt to redirect to sequence.lmsWebUrl even though that variable is still undefined.

When global.location.assign() is given `undefined` as a value, it tacks it onto the end of the URL and calls it a day.  After that, we’ve got a badly formed URL.

### DIFF
--- a/src/courseware/CoursewareContainer.jsx
+++ b/src/courseware/CoursewareContainer.jsx
@@ -89,7 +89,7 @@ function useExamRedirect(sequenceId) {
   const sequence = useModel('sequences', sequenceId);
   const sequenceStatus = useSelector(state => state.courseware.sequenceStatus);
   useEffect(() => {
-    if (sequenceStatus === 'loaded' && sequence.isTimeLimited) {
+    if (sequenceStatus === 'loaded' && sequence.isTimeLimited && sequence.lmsWebUrl !== undefined) {
       global.location.assign(sequence.lmsWebUrl);
     }
   }, [sequenceStatus, sequence]);


stvstnfrd (approved):

src/courseware/CoursewareContainer.jsx
```diff
@@ -87,12 +87,15 @@ function usePreviousSequenceHandler(courseId, sequenceId) {
 
 function useExamRedirect(sequenceId) {
   const sequence = useModel('sequences', sequenceId);
+  const courseStatus = useSelector(state => state.courseware.courseStatus);
   const sequenceStatus = useSelector(state => state.courseware.sequenceStatus);
   useEffect(() => {
-    if (sequenceStatus === 'loaded' && sequence.isTimeLimited) {
+    // We check both courseStatus and sequenceStatus here because lmsWebUrl is returned with the
+    // former, but isTimeLimited is returned by the latter.
+    if (courseStatus === 'loaded' && sequenceStatus === 'loaded' && sequence.isTimeLimited) {
```
stvstnfrd: +1 to fix the issue for now.

It still feels a little weird that we need to "know" this much about the internal data fetching process, especially this high up the call stack (outside of the datastore, really). Specifically, to know that the model depends on two different loading calls.

Curious though, do we _really_ need to check the `loaded` status? Won't the required properties just be `undefined`?

Since we're checking if `sequence.isTimeLimited` is `truthy`, can we just omit the status check and rely on the property being `undefined`, thus `falsey`?

Same thing with `sequence.lmsWebUrl`, can't we just check for its `truthiness` in the `if` check?

```js
useEffect() => {
  if (sequence.isTimeLimited && sequence.lmsWebUrl) {
    global.location.assign(sequence.lmsWebUrl);
  }
}, [sequence]);
```

Or am I misunderstanding the data model?

davidjoy: I think you're right here - it felt a bit dirty when I was writing it.  The split where we have different sequence data coming from multiple APIs is a bit of a problem... it might also make sense if `sequenceStatus` simply wasn't `loaded` until both APIs came back, just to hide that detail from the rest of the frontend.  Going to play with it a bit...

src/courseware/CoursewareContainer.jsx
```diff
@@ -89,7 +89,7 @@ function useExamRedirect(sequenceId) {
   const sequence = useModel('sequences', sequenceId);
   const sequenceStatus = useSelector(state => state.courseware.sequenceStatus);
   useEffect(() => {
-    if (sequenceStatus === 'loaded' && sequence.isTimeLimited) {
+    if (sequenceStatus === 'loaded' && sequence.isTimeLimited && sequence.lmsWebUrl !== undefined) {
```
stvstnfrd, line 92: Nit: I'd be more inclined to just check for broad truthy-ness, `if (sequence.lmsWebUrl)`. It's not only shorter, but prevents the case where an empty string would pass this check but fail to redirect as intended. 

I've been bit by APIs/libraries returning or defaulting to empty string to represent empty values (instead of `undefined`) enough times that I just to stay a bit defensive.

stvstnfrd (approved):

src/courseware/CoursewareContainer.jsx
davidjoy: The problem is actually that `sequence` itself may come back as `null` if sequenceStatus isn't `loaded` yet.  I started doing it the way you described and then realized I'd have to do a null check on sequence anyway.  It seemed better to check a semantically meaningful variable than to just check for non-null on sequence.

stvstnfrd: Gotcha. I _personally_ like the `null` check there, because it feels more like "duck typing" (sort of) for me.

My mental model is, "I don't actually care what this is or if it's actually even fully loaded; I just want to know if it has the data I need".

```js
// do we have sequence?
if (sequence)  {
  // does the sequence have time limiting enabled?
  if (sequence.isTimeLimited) {
    // does the sequence have an LMS web URL?
    if (sequence.lmsWebUrl) {
      // We have everything we need, whether or not the whole sequence has loaded.
    }
  }
}
```

But we're already doing these checks elsewhere, so +1 :)

davidjoy: Yeah, the reason this evolved this way is that the truthiness of these models isn't as granular as their state, which can be `loaded`, `failed`, `denied`, or `loading`.  There are things we want to do on `failed`/`denied` that are very different than `loading`, for instance, but in all of those states the model will be `null` or `undefined`.  

In this particular case, yeah, you could get by without it because there's only one state in which all of those conditions can be fulfilled, and it's the right one.  I'm sorta leaning on the status state variable for consistency, and for saying "I only want to do this check if everything has successfully loaded, otherwise it's not appropriate to even proceed."

PR #89, opened 2020-06-19
Don't show celebration modal too much

mikix:
Make sure we only show it for the first unit of a sequence.


Dillon-Dumesnil (approved):
Sure?

PR #93, opened 2020-06-22
AA-186: Refactoring to separate Course Home logic from Courseware

ciduarte:
- Pulled Course Home specific components into `course-home`
- Created a courseHome reducer (and all necessary data files - api, thunks, slice)
- Removed Course Home logic from Courseware's data files (api, thunks, slice, etc.)
- Renamed Outline Tab URL to end in `/home` rather than `/outline` again (per Product)
- Freebie: Changed the reset dates banner styling to mirror the other banners (I couldn't help it)
Previous styling:
Screen Shot 2020-06-24 at 4 36 27 PM

Updated styling:
Screen Shot 2020-06-24 at 4 54 08 PM


src/course-header/CourseTabsNavigation.jsx
```diff
@@ -36,7 +36,6 @@ CourseTabsNavigation.propTypes = {
   className: PropTypes.string,
   tabs: PropTypes.arrayOf(PropTypes.shape({
     title: PropTypes.string.isRequired,
-    priority: PropTypes.number.isRequired,
```
Dillon-Dumesnil, line 39: :chefkiss:

src/course-home/data/api.js
```diff
@@ -0,0 +1,64 @@
+import { camelCaseObject, getConfig } from '@edx/frontend-platform';
+import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
+import { normalizeBlocks } from '../../data';
+
+function normalizeCourseHomeCourseMetadata(metadata) {
```
Dillon-Dumesnil, line 5: Something we might want to think about, but when should we call out to a normalize function vs just normalizing in the call? Maybe just a feeling? Or how big the original is? I don't know. Also depends on what people find more clear. I don't think I have any sort of preference right now, but just wanted to bring it up to think about.

src/course-home/data/api.js
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export async function getDatesTabData(courseId, version) {
+  const url = `${getConfig().LMS_BASE_URL}/api/course_home/${version}/dates/${courseId}`;
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(url);
+    return camelCaseObject(data);
+  } catch (error) {
+    const { httpErrorStatus } = error && error.customAttributes;
+    if (httpErrorStatus === 404) {
```
Dillon-Dumesnil, line 30: I can't remember where else I had this conversation, but I think we might want to reconsider using a 404 as our status code for the MFE is off as it confuses the cases of `is the MFE off` and `does this course run not exist`. @mikix any thoughts on this? I believe you also brought up the oddities of 3 different return types in this function, but I have fewer ideas there.

src/course-home/data/api.js
```diff
@@ -0,0 +1,64 @@
...
+      return window.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/dates`);
+    }
+    // async functions expect return values. to satisfy that requirement
+    // we return true here which in turn continues with the normal flow of displaying
+    // the "unexpected error try again" screen to the user.
+    return true;
+  }
+}
+
+export async function getOutlineTabData(courseId, version) {
```
Dillon-Dumesnil: The above function states how it returns `true` at the end of the `error` block because async functions expect it, but this one doesn't do it. So does it need it or not? If we can avoid having it, I think that would be preferred as it's less lines of code, but we should be consistent.

src/course-home/data/api.js
```diff
@@ -0,0 +1,64 @@
...
+  const url = `${getConfig().LMS_BASE_URL}/api/course_home/${version}/outline/${courseId}`;
+  let { tabData } = {};
+  try {
+    tabData = await getAuthenticatedHttpClient().get(url);
+  } catch (error) {
+    const { httpErrorStatus } = error && error.customAttributes;
+    if (httpErrorStatus === 404) {
+      return window.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/home`);
```
davidjoy: Nit: `window` is not available in Gatsby-related environments, so as a best practice we try to use `global` instead.  Not that this is likely to be in gatsby anytime soon, but it's a good habit. :)

src/course-home/data/thunks.js
```diff
@@ -0,0 +1,76 @@
...
+  addModel,
+} from '../../model-store';
+
+import {
+  fetchTabFailure,
+  fetchTabRequest,
+  fetchTabSuccess,
+} from './slice';
+
+export function fetchTab(courseId, tab, version, getTabData) {
```
Dillon-Dumesnil: Not that this needs to be you, but I think we should start adding in some doc strings for these functions. It's luckily pretty readable as is, but I think we could improve it by stating what the output is.

davidjoy (approved):
Very nice! :shipit:

Dillon-Dumesnil (approved):
Honestly looks great! I have a couple comments to spark discussions, but nothing that needs to be done as part of your work. Great job here Carla!

src/course-home/data/api.js
davidjoy: I originally just found it helpful to package it up to keep the actual API call function clean and make it clear that it was all unrelated to making the actual call, but just there as an anti-corruption layer.  I think "a feeling" is fine. ;)

davidjoy: I've also done that work in thunks in the past, just to shove it over onto the other side of the fence.  But in this MFE the thunks were all kinda long to begin with, so I brought it back into api.js just cause it felt like I wanted it 'clean' by the time the thunks saw it.

src/course-home/data/thunks.js
davidjoy: (I'd be down with this too - this MFE is likely to be well-used by the community, one hopes, so it'd be nice if it felt really approachable)

src/course-home/data/api.js
```diff
@@ -0,0 +1,64 @@
...
+  };
+}
+
+export async function getCourseHomeCourseMetadata(courseId) {
+  const url = `${getConfig().LMS_BASE_URL}/api/course_home/v1/course_metadata/${courseId}`;
+  const { data } = await getAuthenticatedHttpClient().get(url);
+  return normalizeCourseHomeCourseMetadata(data);
+}
+
+export async function getDatesTabData(courseId, version) {
```
mikix: For these tab-specific calls, you can probably drop the version, yah? We're not going to be calling multiple versions of the dates API inside the MFE.

src/course-home/data/api.js
mikix: Something like 501 (Not Implemented) would probably make more sense. 4xx is client error anyway, and that's not the case really? I don't know how in-scope that change is for this PR. But something that would be nice for us to clean up.

And in terms of return values:
- Instead of `return true` and having that end up throwing an exception because the reading code was expecting an object, you can simply `throw error;` to re-throw the original error.
- Instead of returning the value of `replace` (which does not return anything), maybe just return an empty object? To at least conform to expected return types.

(I realize you are just moving this stuff around. But still.)

mikix (approved):

src/course-home/data/api.js
ciduarte: I switched this (and `getOutlineTabData`) to return an empty object after the redirect. Then I added in the `throw error;`. Let me know if that looks okay.

hilaryg123 (approved):
Awesome updates! Thank you for your attention to detail here.

src/course-home/data/api.js
```diff
@@ -0,0 +1,65 @@
...
+      global.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/dates`);
+      return {};
+    }
+    throw error;
+  }
+}
+
+export async function getOutlineTabData(courseId) {
+  const url = `${getConfig().LMS_BASE_URL}/api/course_home/v1/outline/${courseId}`;
+  let { tabData } = {};
```
Dillon-Dumesnil, line 40: What's the point of this vs just having `const tabData = await get...` from line 42?

src/course-home/data/api.js
Dillon-Dumesnil: I'll let MT respond to this, but yeah, both of those changes seem fine to me.

mikix: :+1:

Dillon-Dumesnil (approved):
One small question, but LGTM. Woo!

src/course-home/data/api.js
ciduarte: You can't access the variables created in the try/catch block outside of that scope. In my previous ticket, it was suggested that we move the code that normalizes the data & returns it to outside of the try block, so this was the solution.

Dillon-Dumesnil: Ah okay. Yeah, that makes snese

PR #96, opened 2020-06-25
dlichen/AA-203-Progress-FE-Setup

daphneli-chen:
Screen Shot 2020-07-08 at 10 58 40 AM

note: may remove user information in final page

### DIFF
--- a/src/course-home/data/api.js
+++ b/src/course-home/data/api.js
@@ -36,6 +36,21 @@ export async function getDatesTabData(courseId) {
   }
 }
 
+export async function getProgressTabData(courseId) {
+  const url = `${getConfig().LMS_BASE_URL}/api/course_home/v1/progress/${courseId}`;
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(url);
+    return camelCaseObject(data);
+  } catch (error) {
+    const { httpErrorStatus } = error && error.customAttributes;
+    if (httpErrorStatus === 404) {
+      global.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/progress`);
+      return {};
+    }
+    throw error;
+  }
+}
+
 export async function getOutlineTabData(courseId) {
   const url = `${getConfig().LMS_BASE_URL}/api/course_home/v1/outline/${courseId}`;
   let { tabData } = {};
--- a/src/course-home/progress-tab/ProgressTab.jsx
+++ b/src/course-home/progress-tab/ProgressTab.jsx
@@ -1,0 +1,27 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+import { getAuthenticatedUser } from '@edx/frontend-platform/auth';
+import { useModel } from '../../generic/model-store';
+
+export default function ProgressTab() {
+  const {
+    courseId,
+  } = useSelector(state => state.courseHome);
+
+  const { administrator, username } = getAuthenticatedUser();
+
+  const {
+    enrollmentMode,
+  } = useModel('progress', courseId);
+
+
+  return (
+    <section>
+      <h2 className="mb-4">
+        the user is {username} and they are enrolled as an {enrollmentMode} learner
+        {administrator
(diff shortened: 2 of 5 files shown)


src/course-home/data/api.js
```diff
@@ -35,6 +35,23 @@ export async function getDatesTabData(courseId) {
...
 
+export async function getProgressTabData(courseId) {
+  const url = `${getConfig().LMS_BASE_URL}/api/course_home/v1/progress/${courseId}`;
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(url);
+    return camelCaseObject(data);
+  } catch (error) {
+    const { httpErrorStatus } = error && error.customAttributes;
+    if (httpErrorStatus === 404) {
+      return window.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/dates`);
```
ciduarte: - We'll replace `window` here with `global`.
- Remove `return` from this line, and return an empty object below it.
- `dates` at the end of this line should probably be `progress`, or whatever the equivalent is in the LMS.

MT explained the reasoning, if you're curious: https://github.com/edx/frontend-app-learning/pull/93/files#r444391454

src/course-home/data/api.js
```diff
@@ -35,6 +35,23 @@ export async function getDatesTabData(courseId) {
...
+    }
+    // async functions expect return values. to satisfy that requirement
+    // we return true here which in turn continues with the normal flow of displaying
+    // the "unexpected error try again" screen to the user.
+    return true;
```
ciduarte: We'll want to `throw error;` here instead of returning true. You can also remove the three line comment above this!

src/course-home/progress-tab/ProgressTab.jsx
```diff
@@ -0,0 +1,24 @@
...
+
+  const {
+    user,
+    hasAccess,
+  } = useModel('progress', courseId);
+
+  return (
+    <section>
+      {hasAccess
+      && <div>hello</div>}
```
ciduarte: nit: move the `&&` to `line 17`.

src/progress-tab/ProgressTab.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+import { useModel } from '../model-store';
+
+export default function ProgressTab() {
+  const {
+    courseId,
+  } = useSelector(state => state.courseware);
```
ciduarte: We want to use the courseHome reducer here, so this would look like `state.courseHome` instead!

ciduarte (commented):
Awesome start! Looks great. Just a few comments that have to do with the refactor.

src/course-home/data/api.js
ciduarte: I thought I added this in but you can reference `getDatesTabData`! This is just how we updated these commands in the refactor :)

src/course-home/data/api.js
```diff
@@ -35,6 +35,21 @@ export async function getDatesTabData(courseId) {
...
 
+export async function getProgressTabData(courseId) {
+  const url = `${getConfig().LMS_BASE_URL}/api/course_home/v1/progress/${courseId}`;
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(url);
+    return camelCaseObject(data);
+  } catch (error) {
+    const { httpErrorStatus } = error && error.customAttributes;
+    if (httpErrorStatus === 404) {
+      global.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/dates`);
```
ciduarte: Don't forget to change this to `/progress` rather than `/dates`

ciduarte (approved):
Looks good to me. Nice work!

Dillon-Dumesnil (approved):
All looks good to me!

Dillon-Dumesnil:
If you don't mind, I would like seeing an updated screenshot! I noticed you're pulling in more data now than the original one :)

PR #97, opened 2020-06-26
 [BD-29] [TNL-7268] Add high priority tests

Agrendalath:
This adds tests for:
- src/courseware/course/sequence/sequence-navigation/SequenceNavigation.jsx
- src/courseware/course/sequence/sequence-navigation/SequenceNavigationDropdown.jsx
- src/courseware/course/sequence/sequence-navigation/SequenceNavigationTabs.jsx
- src/courseware/course/sequence/sequence-navigation/UnitButton.jsx
- src/courseware/course/sequence/sequence-navigation/UnitNavigation.jsx
- src/courseware/course/sequence/Unit.jsx
- src/courseware/course/sequence/Sequence.jsx
- src/courseware/course/sequence/SequenceContent.jsx

## Testing instructions:
1. Run `npm i && npm test`.

## Reviewers
- [x] @pcockwell 
- [ ] @davidjoy


openedx-webhooks:
Thanks for the pull request, @Agrendalath! I've created [BLENDED-394](https://openedx.atlassian.net/browse/BLENDED-394) to keep track of it in Jira.   More details are on the [BD-29](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1543995805/%5BBD-29%5D+Learning+Sequence+Experience+Improvements) project page.

pcockwell (commented):
@Agrendalath 
I've never come across any of this, so I definitely feel a little out of place with reviews, but ill go back and take a look through some `jest` documentation before I come back for a full review round. So far this is looking good though. I did notice you have some snapshots that seem unused, and `npm test` fails, listing those obsolete snapshots as the reason.
```
 › 5 snapshots obsolete.
   • Unit Navigation displays loading message 1
   • Unit Navigation displays message for no content 1
   • Unit Navigation displays message for the locked content 1
   • Unit Navigation displays messages for the locked content 1
   • Unit Navigation displays messages for the locked content 2
```
Also, `package-lock.json` is conflicting

src/courseware/course/sequence/Sequence.test.jsx
```diff
@@ -0,0 +1,240 @@
...
+    window.postMessage(messageEvent, '*');
+    await waitFor(() => expect(screen.queryByText('Loading learning sequence...')).not.toBeInTheDocument());
+    const unitPreviousButton = screen.getAllByRole('button', { name: /previous/i })
+      .filter(button => button !== sequencePreviousButton)[0];
+    fireEvent.click(unitPreviousButton);
+    expect(previousSequenceHandler).toHaveBeenCalledTimes(2);
+    expect(sendTrackEvent).toHaveBeenCalledTimes(2);
+    expect(sendTrackEvent).toHaveBeenNthCalledWith(2, 'edx.ui.lms.sequence.previous_selected', {
+      current_tab: Number(unitId), id: unitId, tab_count: testUnits.length, widget_placement: 'bottom',
+    });
```
pcockwell: Why are you testing the previous unit button if the test description is this?
```
navigates to the previous sequence if the unit is the first in the sequence
```

Would it make sense for this to be a separate test?

src/courseware/course/sequence/Sequence.test.jsx
```diff
@@ -0,0 +1,240 @@
...
+    window.postMessage(messageEvent, '*');
+    await waitFor(() => expect(screen.queryByText('Loading learning sequence...')).not.toBeInTheDocument());
+    const unitNextButton = screen.getAllByRole('button', { name: /next/i })
+      .filter(button => button !== sequenceNextButton)[0];
+    fireEvent.click(unitNextButton);
+    expect(nextSequenceHandler).toHaveBeenCalledTimes(2);
+    expect(sendTrackEvent).toHaveBeenCalledTimes(2);
+    expect(sendTrackEvent).toHaveBeenNthCalledWith(2, 'edx.ui.lms.sequence.next_selected', {
+      current_tab: Number(unitId), id: unitId, tab_count: testUnits.length, widget_placement: 'bottom',
+    });
```
pcockwell: Ditto on this testing the next unit button as well

src/courseware/course/sequence/Sequence.test.jsx
```diff
@@ -0,0 +1,240 @@
...
+    const nextSequenceHandler = jest.fn();
+    render(<Sequence {...mockData} {...{ unitNavigationHandler, previousSequenceHandler, nextSequenceHandler }} />);
+
+    fireEvent.click(screen.getByRole('button', { name: /previous/i }));
+    expect(previousSequenceHandler).not.toHaveBeenCalled();
+    expect(unitNavigationHandler).toHaveBeenCalledWith(String(Number(mockData.unitId) - 1));
+
+    fireEvent.click(screen.getByRole('button', { name: /next/i }));
+    expect(nextSequenceHandler).not.toHaveBeenCalled();
+    expect(unitNavigationHandler).toHaveBeenNthCalledWith(2, String(Number(mockData.unitId) + 1));
```
pcockwell: I take it that the value it is called with is `Number(mockData.unitId) + 1` rather than just `Number(mockData.unitId)` because the `unitNavigationHandler` is mocked and doesn't actually do anything to change the active `unitId` in the component? Maybe include a comment to that effect?

src/courseware/course/sequence/sequence-navigation/SequenceNavigationDropdown.test.jsx
```diff
@@ -0,0 +1,65 @@
...
+  it('handles the clicks', () => {
+    const onNavigate = jest.fn();
+
+    render(<SequenceNavigationDropdown
+      {...mockData}
+      onNavigate={onNavigate}
+    />, { initialState });
+
+    screen.getAllByText(/^\d+$/).forEach(element => fireEvent.click(element));
+    expect(onNavigate).toHaveBeenCalledTimes(testUnits.length);
```
pcockwell: Would it make sense to change this so that you loop through the `testUnits`, find the button for it, `fireEvent.click` on the button, and then `expect(onNavigate).toHaveBeenNthCalledTimes` to check it was called with the correct `unitId`? Or would that be overkill?

src/setupTest.js
```diff
@@ -1,2 +1,148 @@
...
+function render(
+  ui,
+  {
+    initialState = baseInitialState,
+    store = configureStore({
+      reducer: {
+        models: modelsReducer,
+        courseware: coursewareReducer,
+      },
+      preloadedState: initialState,
```
pcockwell: I assume you are attempting to default this to the content of `baseInitialState` defined above. As a result, shouldn't this be `preloadedState: baseInitialState` instead of `preloadedState: initialState`? Or is this a syntax thing that defines it as equal to the contents of the `initialState` argument defined on L110?

src/courseware/course/sequence/sequence-navigation/SequenceNavigationTabs.test.jsx
```diff
@@ -0,0 +1,57 @@
...
+        (acc, unitId) => Object.assign(acc, {
+          [unitId]: {
+            contentType: 'other',
+            title: unitId,
+          },
+        }),
+        {},
+      ),
+    },
+  };
```
pcockwell: Couldn't these both be imported from `setupTest`? At the minimum, `testUnits` is identical.

src/courseware/course/sequence/sequence-navigation/UnitNavigation.test.jsx
```diff
@@ -0,0 +1,72 @@
...
+    onClickNext: () => {},
+  };
+
+  it('renders correctly without units', () => {
+    const { asFragment } = render(<UnitNavigation
+      {...mockData}
+      sequenceId=""
+      unitId=""
+      onClickPrevious={() => {}}
+      onClickNext={() => {}}
```
pcockwell, line 34: Given that you override all the keys in `mockData` anyways, does it make sense to even include `{...mockData}`? Same question applies to other locations in this file.

pcockwell (approved):
Several questions and a few nitpicks, but overall, it looks good. Would like to hear from you regarding the comments though. 

Leaving this on `Approved` (meant to do it with this, review, but messed that up, so it was done separately below) so you can respond and address concerns and not be blocked on me though since none of the comments are blocking.

- [x] I tested this: ran `npm i && npm test` and received passing tests
```
Test Suites: 9 passed, 9 total
Tests:       73 passed, 73 total
Snapshots:   34 passed, 34 total
```
- [x] I read through the code

pcockwell (approved):

src/courseware/course/sequence/sequence-navigation/UnitNavigation.test.jsx
Agrendalath: In this case it doesn't make any difference, but if we end up adding more required [props](https://github.com/edx/frontend-app-learning/blob/9cbb765f8a2669a8740e9f85a5111ff9d19e9211/src/courseware/course/sequence/sequence-navigation/UnitNavigation.jsx#L61-L66) we will be able to override a prop for all tests at once, without checking each test to see if it needs `{...mockData}` back.

Does it sound reasonable?

src/courseware/course/sequence/sequence-navigation/SequenceNavigationTabs.test.jsx
Agrendalath: Ah, good catch! I missed this one.

src/setupTest.js
Agrendalath: The latter - `initialState` has `baseInitialState` [as the default value](https://github.com/open-craft/frontend-app-learning/blob/01bf23145542b2b2577e414060482ab08a264516/src/setupTest.js#L110), but it's possible to override it [like this](https://github.com/open-craft/frontend-app-learning/blob/ada07504fdd6bf5ab4a4f2f68491f7f53a160416/src/courseware/course/sequence/Sequence.test.jsx#L49-L51).

src/courseware/course/sequence/sequence-navigation/SequenceNavigationDropdown.test.jsx
Agrendalath: Sure, I'll add this. I would still keep the `expect(onNavigate).toHaveBeenCalledTimes(testUnits.length);` check, though - its error message is pretty clear, so it would be good to have both for debugging in case of any test failures.

src/courseware/course/sequence/Sequence.test.jsx
Agrendalath: Right, this might be a bit confusing. I added the comment.

src/courseware/course/sequence/Sequence.test.jsx
Agrendalath: I wrapped these tests in `describe('sequence and unit navigation buttons')` to indicate that they're related to navigation buttons. Does it look better or is it still confusing?

> Would it make sense for this to be a separate test?

I felt that adding bigger fragmentation level to `Sequence.test` would make it harder to read, so I grouped both buttons in these tests to avoid code repetition by reusing mocks' setup.

src/courseware/course/sequence/Sequence.test.jsx
Agrendalath: I suppose the previous answer would be applicable here too.

pcockwell (approved):
:+1:

- [x] I tested this: ran tests again
- [x] I read through the code

natabene:
@Agrendalath Thank you for your contribution. Please let me know once this is ready for review.

davidjoy:
It's on me for not getting this note into the TNL-7268 ticket, but I'm wondering if we can focus on specific assertions rather than snapshot tests - the latter are very brittle and difficult to maintain.   We should be mindful not to over-assert on implementation/styling details in component tests, which is basically what a snapshot does.  Doing so makes it much more likely to break on inconsequential changes.  

(Similar comment over here: https://github.com/edx/frontend-app-course-authoring/pull/2#discussion_r447741731)

Making specific assertions also helps with maintainability by signaling intent.  It helps us understand what specific things we consider our acceptance criteria - i.e., the sequence nav appears with the right number of buttons.  Put another way, these tests shouldn't fail if we add another `<div>` into the DOM to attach some styling to it.

Let me know what you think - thanks!

davidjoy:
Note, I just merged https://github.com/edx/frontend-app-learning/pull/102 which does some code organization.  I don't expect it should affect these tests _too_ much beyond needing to adjust some imports, perhaps.  The contents of the files didn't change for the most part, I just moved some things around. Let me know if you have any issues.

davidjoy:
FYI, I've added some tests for the CoursewareContainer here:

https://github.com/edx/frontend-app-learning/pull/108

As part of that, I used the rosie factories from the data integration tests to set up a redux store, and modified them so that they could allow for multiple units (useful for testing sequence navigation, presumably).  Not all components will need the full store, but it was pretty nice being able to set up the store and simply mock the API requests.

I noticed in this PR here there's a bunch of manual set up of redux store states... it might be more resilient to take the approach pioneered in those data int tests.  Manually setup stores have a tendency to diverge from what gets set up by actions.  (i.e., if someone changes the store, they don't know to go update the manual testing state)

Agrendalath:
@davidjoy, as mentioned on Slack, it totally makes sense to use factories here to have more reliable and consistent approach. I rebased the branch and made some amends (e.g. global.MutationObserver issue for async functions has been fixed after updating dependencies (yay!) and the FontAwesome icon mock turned out to be redundant, as we're not using the snapshot tests eventually).
Given that there's plenty of tests, I'll continue rewriting this tomorrow, as it's still in a WIP state for me.

Agrendalath:
@davidjoy, a quick update: it seems to be a bit more tricky than I expected. For now I'm using something similar to [this snippet](https://gist.github.com/Agrendalath/b2a8759a540240d758f5334f9581d39c#file-setuptest-js-L82-L152) for providing a generic way to create stores for tests. I'm invoking it this way:
```js
  beforeEach(async () => {
    store = await initializeTestStore();

    const sequence = Object.values(store.getState().models.sequences)[0];
    unit = store.getState().models.units[sequence.unitIds[0]];
    mockData = {
      unitId: unit.id,
      onClick: () => {},
    };
  });
```
The solution from #108 is very verbose, but if we end up using it for all existing tests, we'll end up with tons of boilerplate code that will make them very hard to read. Currently I'm looking for a way to customize the generated metadata without duplicating too many lines fo code.
What do you think about this?

davidjoy:
I think the above `initializeTestStore` snippet makes sense.  My only caution would be that we try to use it where we need it, and not just have it initialize on all tests whether they need it or not.  The test suite isn't going to be massive, but running all that on tests that don't need it feels a bit overkill. :)   

We should be able to update the store to put it in specific states by dispatching actions/calling thunks.  I added that "buildSimpleCourseBlocks" function to try to reduce some of the boilerplate I was seeing... I think it'd be reasonable for the factories to have helpers like that that abstract away a little of the complexity of modifying the factory build.  Because all the factories need to interact in very specific ways, it'd be helpful for future test writers to have some helpers they can use easily.

Long story short, I think adding this setup makes sense.

Agrendalath:
@davidjoy, I pushed the refactored version that uses factories and `api.js` for generating test data. It initializes new store at the beginning of each test with `beforeAll` (instead of `beforeEach`, which I was using before). Travis reports that it's working even a bit faster than an approach without factories.

There are some minor changes to be done (e.g. renaming `sequence` to `sequences` in `courseBlocks` factory) related to adding support for courses with multiple sections/sequences/units, but except that it's ready for the review. Please let me know if you have any questions about this approach :)

src/courseware/data/redux.test.js
```diff
@@ -25,17 +24,17 @@ describe('Data layer integration tests', () => {
   // building minimum set of api responses to test all thunks
   const courseMetadata = Factory.build('courseMetadata');
   const courseId = courseMetadata.id;
-  const { courseBlocks, unitBlock, sequenceBlock } = buildSimpleCourseBlocks(courseId);
+  const { courseBlocks, unitBlocks, sequenceBlock } = buildSimpleCourseBlocks(courseId);
```
davidjoy, line 27: Seems like `sequenceBlock` is expected to come back as an array?  But I don't see that the thing being returned from `buildSimpleCourseBlocks` has an array for `sequenceBlock` - it's just the block.  A bit confused how it's working.

Agrendalath: It's [an array](https://github.com/edx/frontend-app-learning/pull/97/files#diff-b3ef81408a4e781ef23ee18fd4026c92R62-R66), if the `options.sequenceBlock` is `undefined`. I'll reword this to plural to make it less confusing.

Edit: linking the related snippet here, because GitHub doesn't seem to be scrolling to the code from the link for some reason:
```js
const sequenceBlock = options.sequenceBlock || [Factory.build(
  'block',
  { type: 'sequential', children: unitBlocks.map(block => block.id) },
  { courseId },
)];
```

davidjoy:
FYI - I've rebased and merged this into `djoy/high_priority_tests` so I can merge it with some recent changes I just made to CoursewareContainer.test.jsx that I don't want you to have to deal with. :) I've caused enough conflicts on your work as it is!

If this is successful, I should be able to merge that in shortly.

davidjoy:
Don't worry about resolving these conflicts for the moment - going to give it a shot myself.

Agrendalath:
Cool, thank you for tackling this, @davidjoy!
Btw. there are some options for reducing code duplication in `CoursewareContainer.test` by reusing the `initializeTestStore` approach from this PR, though I'm not sure if we want to do this.

davidjoy:
Heyo - so I resolved the merge conflicts here: https://github.com/edx/frontend-app-learning/pull/136

I merged that in, so we're good here!  Thanks @Agrendalath!  I'm going to close this one.  We'll have to make sure the JIRA tickets and all that get resolved properly.

PR #98, opened 2020-06-29
[BD-29] [TNL-7288] Fix front-end behavior when the course has no sections or no subsections in the first section

taniwha:
This PR fixes the behavior of the front end when the course has no sections or the first section of the course has no subsections. It now displays the same "no content" message as when the subsection has no units.

**JIRA tickets**: https://tasks.opencraft.com/browse/BB-2596

**Dependencies**: https://github.com/edx/edx-platform/pull/24340

**Screenshots**: N/A?

**Sandbox URL**: TBD - sandbox is being provisioned.

**Merge deadline**: ASAP

**Testing instructions**:

1. In studio, create a new course but do no add any sections
2. In lms, go to the blank course and start it.
3. Then click on "View In The New Experience"
4. without this PR, the browser will be redirected to a url ending in `.../null` and an "unexpected error" message will be displayed, breaking the interface. After this patch, no redirection will occur and a "no content" message will be displayed without breaking the interface.
5. In studio, add a section to the new course.
6. back in the lms, reload the page (remove `null` from the url if necessary). Before this PR, the url will be redirected such that it ends in `.../undefined` and the unexpected error message will once again be shown, breaking the interface. After the PR, again no redirection and the *no content* message will be shown.
7. In stuido, add a subsection to the course's section.
8. back in the lms, reload the page (remove `undefined` if necessary). Both with and without this patch, the url is redirected to the full sequence id and the "no content" message is shown without breaking the interface. That is, this behavior is unchanged.
9. Adding a unit causes the unit to be shown as epected. Adding subsequent sections, even ones with no subsections, does not seem to cause breakage.

**Author notes and concerns**: none at this time

**Reviewers**
- [ ] @Agrendalath
- [ ] edX reviewer[s] TBD

### DIFF
--- a/src/courseware/course/sequence/Sequence.jsx
+++ b/src/courseware/course/sequence/Sequence.jsx
@@ -100,6 +100,9 @@ function Sequence({
   }, [unit]);
 
   if (sequenceStatus === 'loading') {
+    if (!sequence) {
+      return (<div> {intl.formatMessage(messages['learn.sequence.no.content'])} </div>);
+    }
     return (
       <PageLoading
         srMessage={intl.formatMessage(messages['learn.loading.learning.sequence'])}
--- a/src/courseware/course/sequence/Sequence.jsx
+++ b/src/courseware/course/sequence/Sequence.jsx
@@ -110,7 +110,7 @@ function Sequence({
     );
   }
 
-  const gated = sequence.gatedContent !== undefined && sequence.gatedContent.gated;
+  const gated = sequence && sequence.gatedContent !== undefined && sequence.gatedContent.gated;
 
   if (sequenceStatus === 'loaded') {
     return (
--- a/src/courseware/course/sequence/Sequence.jsx
+++ b/src/courseware/course/sequence/Sequence.jsx
@@ -100,7 +100,7 @@ function Sequence({
   }, [unit]);
 
   if (sequenceStatus === 'loading') {
-    if (!sequence) {
+    if (!sequenceId) {
       return (<div> {intl.formatMessage(messages['learn.sequence.no.content'])} </div>);
     }
     return (
--- a/src/courseware/course/sequence/Unit.jsx
+++ b/src/courseware/course/sequence/Unit.jsx
@@ -95,10 +95,6 @@ function Unit({
     return () => global.removeEventListener('message', messageEventListenerRef.current);
   }, [id, setIframeHeight, hasLoaded, iframeHeight, setHasLoaded, onLoaded]);
 
-  if (!unit) {
-    return (<div> {intl.formatMessage(messages['learn.sequence.no.content'])} </div>);
-  }
-
   return (
     <div className="unit">
       <h2 className="mb-0 h4">{unit.title}</h2>
--- a/src/courseware/course/sequence/Unit.jsx
+++ b/src/courseware/course/sequence/Unit.jsx
@@ -95,6 +95,10 @@ function Unit({
     return () => global.removeEventListener('message', messageEventListenerRef.current);
   }, [id, setIframeHeight, hasLoaded, iframeHeight, setHasLoaded, onLoaded]);
 
+  if (!unit) {
+    return (<div> {intl.formatMessage(messages['learn.sequence.no.content'])} </div>);
+  }
+
   return (
     <div className="unit">
       <h2 className="mb-0 h4">{unit.title}</h2>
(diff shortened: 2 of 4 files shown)


src/courseware/course/sequence/Sequence.jsx
```diff
@@ -100,6 +100,9 @@ function Sequence({
   }, [unit]);
 
   if (sequenceStatus === 'loading') {
+    if (!sequence) {
```
davidjoy: During initial page load, I expect there's a period of time before the course blocks API and the sequence metadata API come back where the `sequenceStatus` is `loading` but the `sequence` is still falsy, meaning that we'll see a flash of this 'no content' messaging for a moment before the data comes in.  

If we instead check whether `sequenceId` is `null` here, that may give us a more accurate condition.  The `sequenceId` in redux is only populated when we begin to request a sequence (`fetchSequence` thunk).  If we have no sequence ID in the URL route, then `fetchSequence` never happens and the `sequenceId` in redux stays `null`.

davidjoy (changes requested):
If you can, give that suggestion a shot... I think it may be slightly more 'correct.'

src/courseware/course/sequence/Sequence.jsx
taniwha: @davidjoy Thanks. I've made the suggested change. Seems to work nicely.

Agrendalath:
@davidjoy, during testing this I found two more cases when the UI doesn't handle errors:

### Non-existent unit
Steps to reproduce the problem:
1. Create new section, subsection and unit inside it.
1. Publish the new unit.
1. View this unit in the MFE.
1. Delete unit in studio and publish the section.
1. Refresh page with MFE and see the unhandled error.

(Or even simpler - type random unit ID in the URL.)

Possible fix: add [here](https://github.com/edx/frontend-app-learning/blob/73c74119f095fc743c199101320981c7950d0d40/src/courseware/course/sequence/Unit.jsx#L97) something like this:
```js
  if (!unit) {
    return (<div> {intl.formatMessage(messages['learn.sequence.no.content'])} </div>);
  }
```

### Non-existent subsection
Steps to reproduce the problem:
1. After doing the steps above remove the subsection via Studio.
1. Refresh the MFE page and see the unhandled error.

(Or again - type random subsection ID in the URL.)

Possible fix: add `sequence &&` to [this line](https://github.com/edx/frontend-app-learning/blob/73c74119f095fc743c199101320981c7950d0d40/src/courseware/course/sequence/Sequence.jsx#L110).

Agrendalath (approved):
:+1:
* [x]  I tested this: checked that the MFE no longer breaks while following the testing instructions
* [x]  I read through the code
* [x]  I checked for accessibility issues
* [ ]  <s>Includes documentation</s>

Agrendalath:
@taniwha, just two nits here regarding the PR:
1. This "Settings" section in the PRs description is redundant.
1. Please change the PR name to the following format: `[BD-29] [TNL-7288] Brief description of what this fixes`.

src/courseware/course/sequence/Unit.jsx
```diff
@@ -95,6 +95,10 @@ function Unit({
     return () => global.removeEventListener('message', messageEventListenerRef.current);
   }, [id, setIframeHeight, hasLoaded, iframeHeight, setHasLoaded, onLoaded]);
 
+  if (!unit) {
```
davidjoy: I'm seeing that in SequenceContent we have a similar check that's doing a `unitId === null` instead of this more general falsy-ness check on the `unit` itself. 

I'm wondering if we can combine those two or at least have the check in one place instead of both.  I _think_ the right place to do it is in `SequenceContent`, since the absence of a `unit` means we know we don't have the data to render a `<Unit />`.  That way we don't even set up all the global event listeners right above this unless we're really going to render an iframe.  I think what I'd do is if we have a `unitId`, get the `unit` in `SequenceContent` via:

```
const unit = useModel('units', id);
```

And then between `unitId` and `unit`, we can determine in there if we need to show the "no content" message.

davidjoy (changes requested):
Let me know if the suggestion about moving the `!unit` check to `SequenceContent` doesn't make sense, but I think it'd be better to combine these "no content" checks.

davidjoy:
@Agrendalath Thanks for catching those two cases - makes sense to fix them too.

src/courseware/course/sequence/Unit.jsx
taniwha: @davidjoy I've implemented this change. I hope it's as you intended.

davidjoy (approved):

openedx-webhooks:
Thanks for the pull request, @taniwha! I've created [BLENDED-731](https://openedx.atlassian.net/browse/BLENDED-731) to keep track of it in Jira.   More details are on the [BD-29](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1543995805/%5BBD-29%5D+Learning+Sequence+Experience+Improvements) project page.

PR #100, opened 2020-06-30
AA-195: Don't show schedule banner if no schedule

mikix:
In some cases (user schedule is close to or past the course end), we don't actually have dates for a course, even if we would normally have them.

When that happens, let's not show a banner saying we made a schedule for the learner.

Related to https://github.com/edx/edx-when/pull/58

### DIFF
--- a/src/course-home/dates-banner/DatesBannerContainer.jsx
+++ b/src/course-home/dates-banner/DatesBannerContainer.jsx
@@ -17,6 +17,7 @@ function DatesBannerContainer(props) {
   } = useSelector(state => state.courseHome);
 
   const {
+    courseDateBlocks,
     datesBannerInfo,
   } = useModel(model, courseId);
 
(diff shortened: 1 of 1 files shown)


src/course-home/dates-banner/DatesBannerContainer.jsx
```diff
@@ -32,13 +33,14 @@ function DatesBannerContainer(props) {
   } = useModel('courses', courseId);
 
   const dispatch = useDispatch();
+  const hasDeadlines = courseDateBlocks.some(x => x.dateType === 'assignment-due-date');
```
Dillon-Dumesnil, line 36: Nit: `hasDeadlines` -> `hasAssignmentDeadlines`

Dillon-Dumesnil (approved):
Sure

src/course-home/dates-banner/DatesBannerContainer.jsx
mikix: Eh, there's already `missedDeadlines` that doesn't talk about assignment. I'm inclined to leave as is, though I understand where you're coming from.

PR #101, opened 2020-06-30
Update 0004-model-store.md

davidjoy:
Explaining the rationale for model-store in a bit more detail.


stvstnfrd (commented):
+1, makes sense to me.

stvstnfrd (approved):

PR #102, opened 2020-06-30
Organizationing

davidjoy:
**This PR is best reviewed commit-by-commit.**

I created a `generic` sub-directory in `src` for components and code I wanted to note as being generic/reusable and otherwise unrelated to the actual domain of the MFE.  These are things I expect we'll want to extract to Paragon/frontend-platform in the longer-term.

I pulled the instructor toolbar and masquerade widget up to the top level, since they're shared by courseware and course home (through the loaded tab page).

I also pulled the thunks/APIs used by courseware _down_ into the courseware sub-directory.


src/course-home/data/api.js
```diff
@@ -1,6 +1,7 @@
 import { camelCaseObject, getConfig } from '@edx/frontend-platform';
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
-import { normalizeBlocks } from '../../data';
+// TODO: Pull this normalization function up so we're not reaching into courseware
+import { normalizeBlocks } from '../../courseware/data/api';
```
ciduarte, line 4: 👍

docs/decisions/0006-thunk-and-api-naming.md
```diff
@@ -0,0 +1,24 @@
+# Naming API functions and redux thunks
```
ciduarte, line 1: Yay for clear documentation 🎉 Thanks for cleaning up the code to align with this.

ciduarte (approved):
Looks great!

Question for my own understanding: I see `user-messages` was put into `generic/`. Would `i18n` fall into that same bucket of code that isn't specific to frontend-app-learning? It's not super clear to me how the different messages in `i18n` get populated/used.

Dillon-Dumesnil (commented):
Not approving because I haven't been working in the MFE closely and am likely not the best person to leave comments on the details, but just wanted to let you know I looked through the PR and it's all looking awesome! 

Big thanks for the tip about going through commit by commit and huge thanks for actually working on this in a way to make it so much easier to review!

davidjoy:
> Looks great!
> 
> Question for my own understanding: I see `user-messages` was put into `generic/`. Would `i18n` fall into that same bucket of code that isn't specific to frontend-app-learning? It's not super clear to me how the different messages in `i18n` get populated/used.

The i18n directory - while empty right now - will ultimately contain translated versions of the specific strings we've added to the repo once they're pushed up to Transifex, translated, and pulled back down.  So it's definitely specific to this application :D

PR #103, opened 2020-07-03
[BD-29] [TNL-7303] Add course license details

pcockwell:
Add `CourseLicense` component with a display

**JIRA tickets**:
BD-29
[TNL-7303](https://openedx.atlassian.net/browse/TNL-7303)
[BLENDED-391](https://openedx.atlassian.net/browse/BLENDED-391)

**Dependencies**:
https://github.com/edx/edx-platform/pull/24393

**Screenshots**:
All Rights Reserved License
all-rights-reserved-license

Creative Commons License
creative-commons-license

**Merge deadline**: None

**Testing instructions**:

1. Pull the `patrick/BB-2678-tnl-7303-add-course-license-details-to-course-l` branch of the `open-craft:edx-platform` repository locally
2. Create a course in Studio
3. Edit the course, and select Settings > Schedule & Details
4. Set a license at the bottom of the page
5. Pull the `patrick/BB-2678-tnl-7303-add-course-license-details` branch of the `open-craft:frontend-app-learning` repository locally and setup your local LMS.
6. Run `npm install && npm start` in your `frontend-app-learning` folder
7. Navigate to to any unit in the course
8. Observe the license changes

**Author notes and concerns**:

1. Some CSS was added for course license display that does not use Utility-First CSS classes. If desired, this can be modified.

**Reviewers**
- [ ] @Agrendalath
- [ ] edX reviewer[s] TBD

### DIFF
--- a/src/courseware/course/course-license/CourseLicense.jsx
+++ b/src/courseware/course/course-license/CourseLicense.jsx
@@ -1,0 +1,166 @@
+import React from 'react';
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import PropTypes from 'prop-types';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faCopyright } from '@fortawesome/free-regular-svg-icons';
+import {
+  faCreativeCommons,
+  faCreativeCommonsBy,
+  faCreativeCommonsNc,
+  faCreativeCommonsNd,
+  faCreativeCommonsSa,
+  faCreativeCommonsZero,
+} from '@fortawesome/free-brands-svg-icons';
+
+import messages from './messages';
+
+const CreativeCommonsLicenseTags = {
+  by: {
+    intlMessagesId: 'learn.course.license.creativeCommons.terms.by',
+    icon: faCreativeCommonsBy,
+  },
+  nc: {
+    intlMessagesId: 'learn.course.license.creativeCommons.terms.nc',
+    icon: faCreativeCommonsNc,
+  },
+  nd: {
+    intlMessagesId: 'learn.course.license.creativeCommons.terms.nd',
+    icon: faCreativeCommonsNd,
+  },
+  sa: {
+    intlMessagesId: 'learn.course.license.creativeCommons.terms.sa',
+    icon: faCreativeCommonsSa,
+  },
+  zero: {
+    intlMessagesId: 'learn.course.license.creativeCommons.terms.zero',
+    icon: faCreativeCommonsZero,
+  },
+};
+
+function parseLicense(license) {
+  if (!license) {
+    // Default to All Rights Reserved if no license
+    // is detected
+    return ['all-rights-reserved', {}];
+  }
+
+  // Search for a colon character denoting the end
+  // of the license type and start of the options
+  const colonIndex = license.indexOf(':');
+  if (colonIndex === -1) {
+    // no options, so the entire thing is the license type
+    return [license, {}];
+  }
+
+  // Split the license on the colon
+  const licenseType = license.slice(0, colonIndex).trim();
+  const optionStr = license.slice(colonIndex + 1).trim();
+
+  let options = {};
+  let version = '';
+
+  // Set the defaultVersion to 4.0
+  const defaultVersion = '4.0';
+  optionStr.split(' ').forEach(option => {
+    // Split the option into key and value
+    // Default the value to `true` if no value
+    let key = '';
+    let value = '';
+    if (option.indexOf('=') !== -1) {
+      [key, value] = option.split('=');
+    } else {
+      key = option;
+      value = true;
+    }
+
+    // Check for version
+    if (key === 'ver') {
+      version = value;
+    } else {
+      // Set the option key to lowercase to make
+      // it easier to query
+      options[key.toLowerCase()] = value;
+    }
+  });
+
+  // No options
+  if (Object.keys(options).length === 0) {
+    // If no other options are set for the
+    // license, set version to 1.0
+    version = '1.0';
+
+    // Set the `zero` option so the link
+    // works correctly
+    options = {
+      zero: true,
+    };
+  }
+
+  // Set the version to whatever was included,
+  // using `defaultVersion` as a fallback if unset
+  version = version || defaultVersion;
+
+  return [licenseType, options, version];
+}
+
+function CourseLicense({
+  license,
+  intl,
+}) {
+  const renderAllRightsReservedLicense = () => (
+    <div>
+      <FontAwesomeIcon aria-hidden="true" className="mr-1" icon={faCopyright} />
+      <span className="license-text">
+        {intl.formatMessage(messages['learn.course.license.allRightsReserved.text'])}
+      </span>
+    </div>
+  );
+
+  const renderCreativeCommonsLicense = (activeCreativeCommonsLicenseTags, version) => (
+    <a
+      className="text-decoration-none text-gray-700"
+      rel="license noopener noreferrer"
+      target="_blank"
+      href={`https://creativecommons.org/licenses/${activeCreativeCommonsLicenseTags.join('-')}/${version}/`}
+    >
+      <span className="sr-only">
+        {intl.formatMessage(messages['learn.course.license.creativeCommons.terms.preamble'])}&nbsp;
+      </span>
+      <FontAwesomeIcon aria-hidden="true" className="mr-1" icon={faCreativeCommons} />
+      {activeCreativeCommonsLicenseTags.map(tag => (
+        <span key={tag}>
+          <span className="sr-only">
+            {intl.formatMessage(messages[CreativeCommonsLicenseTags[tag].intlMessagesId])}&nbsp;
+          </span>
+          <FontAwesomeIcon aria-hidden="true" className="mr-1" icon={CreativeCommonsLicenseTags[tag].icon} />
+        </span>
+      ))}
+      <span className="license-text">
+        {intl.formatMessage(messages['learn.course.license.creativeCommons.text'])}
+      </span>
+    </a>
+  );
+
+  const [licenseType, licenseOptions, licenseVersion] = parseLicense(license);
+
+  return (
+    <div className="course-license text-right small">
+      {licenseType === 'all-rights-reserved' && renderAllRightsReservedLicense()}
+      {licenseType === 'creative-commons' && renderCreativeCommonsLicense(
+        Object.keys(licenseOptions),
+        licenseVersion,
+      )}
+    </div>
+  );
+}
+
+CourseLicense.propTypes = {
+  license: PropTypes.string,
+  intl: intlShape.isRequired,
+};
+
+CourseLicense.defaultProps = {
+  license: 'all-rights-reserved',
+};
+
+export default injectIntl(CourseLicense);
--- a/src/courseware/course/course-license/messages.js
+++ b/src/courseware/course/course-license/messages.js
@@ -1,0 +1,47 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  'learn.course.license.allRightsReserved.text': {
+    id: 'learn.course.license.allRightsReserved.text',
+    defaultMessage: 'All Rights Reserved',
+    description: 'License text shown when using All Rights Reserved license type.',
+  },
+  'learn.course.license.creativeCommons.terms.preamble': {
+    id: 'learn.course.license.creativeCommons.terms.preamble',
+    defaultMessage: 'Creative Commons licensed content, with terms as follows:',
+    description: 'Screen reader only text preamble before reading specific Creative Commons license terms.',
+  },
+  'learn.course.license.creativeCommons.terms.by': {
+    id: 'learn.course.license.creativeCommons.terms.by',
+    defaultMessage: 'Attribution',
+    description: 'Creative Commons license text for Attribution term.',
+  },
+  'learn.course.license.creativeCommons.terms.nc': {
+    id: 'learn.course.license.creativeCommons.terms.nc',
+    defaultMessage: 'Noncommercial',
+    description: 'Creative Commons license text for Noncommercial term.',
+  },
+  'learn.course.license.creativeCommons.terms.nd': {
+    id: 'learn.course.license.creativeCommons.terms.nd',
+    defaultMessage: 'No Derivatives',
+    description: 'Creative Commons license text for No Derivatives term.',
+  },
+  'learn.course.license.creativeCommons.terms.sa': {
+    id: 'learn.course.license.creativeCommons.terms.sa',
+    defaultMessage: 'Share Alike',
+    description: 'Creative Commons license text for Share Alike term.',
+  },
+  // No text for `zero` license
+  'learn.course.license.creativeCommons.terms.zero': {
+    id: 'learn.course.license.creativeCommons.terms.zero',
+    defaultMessage: 'No terms',
+    description: 'Creative Commons license text for license with no terms.',
+  },
+  'learn.course.license.creativeCommons.text': {
+    id: 'learn.course.license.creativeCommons.text',
+    defaultMessage: 'Some Rights Reserved',
+    description: 'License text shown when using all Creative Commons license types.',
+  },
+});
+
+export default messages;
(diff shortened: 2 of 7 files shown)


openedx-webhooks:
Thanks for the pull request, @pcockwell! I've created [BLENDED-391](https://openedx.atlassian.net/browse/BLENDED-391) to keep track of it in Jira.

src/courseware/course/sequence/Sequence.jsx
```diff
@@ -154,6 +156,9 @@ function Sequence({
             )}
           </div>
         </div>
+        <div className="sequence-footer">
+          <CourseLicense license={course.license} />
+        </div>
```
Agrendalath: This is going to break if we don't set a license, because the API is going to return `license: null`. We could do something like this to prevent it from breaking, but I'll ask on the upstream ticket what should be the default behavior for this.

```suggestion
        {course.license && (
          <div className="sequence-footer">
            <CourseLicense license={course.license} />
          </div>
        )}
```

src/courseware/course/CourseLicense.jsx
```diff
@@ -0,0 +1,114 @@
...
+  },
+  nd: {
+    text: 'No Derivatives',
+    icon: faCreativeCommonsNd,
+  },
+  sa: {
+    text: 'Share Alike',
+    icon: faCreativeCommonsSa,
+  },
+};
```
Agrendalath: Would it make sense to create an enum above to make this more readable?
E.g.:
```js
enum CreativeCommonsLicenseTagsEnum {
    Attribution = 'by',
    Noncommercial = 'nc',
    NoDerivatives = 'nd',
    ShareAlike = 'sa',
};

const CreativeCommonsLicenseTags = {
  [CreativeCommonsLicenseTagsEnum.Attribution]: {
    text: 'Attribution',
    icon: faCreativeCommonsBy,
  },
  [CreativeCommonsLicenseTagsEnum.Noncommercial]: {
    text: 'Noncommercial',
    icon: faCreativeCommonsNc,
  },
  [CreativeCommonsLicenseTagsEnum.NoDerivatives]: {
    text: 'No Derivatives',
    icon: faCreativeCommonsNd,
  },
  [CreativeCommonsLicenseTagsEnum.ShareAlike]: {
    text: 'Share Alike',
    icon: faCreativeCommonsSa,
  },
};
```

src/courseware/course/CourseLicense.jsx
```diff
@@ -0,0 +1,114 @@
...
+
+function parseLicense(license) {
+  if (license === undefined) {
```
Agrendalath: `undefined !== null` and the API can actually return `null`. The `!license` check should be sufficient here.

src/courseware/course/CourseLicense.jsx
```diff
@@ -0,0 +1,114 @@
...
+  const optionStr = license.slice(colonIndex + 1).trim();
+  let options = {};
+  let version = '';
+  optionStr.split(' ').forEach(option => {
+    if (option.indexOf('=') !== -1) {
+      const [key, value] = option.split('=');
+      if (key === 'ver') {
+        version = value;
+      } else {
+        options[key.toLowerCase()] = value.toLowerCase();
```
Agrendalath: What does this line relate to? Some more comments would be nice for this function.

src/courseware/course/CourseLicense.jsx
```diff
@@ -0,0 +1,114 @@
...
+      }
+    } else {
+      options[option.toLowerCase()] = true;
+    }
+  });
+
+  if (options.length === 0) {
+    version = version || '1.0';
```
Agrendalath: What's the use-case here? Can we run into something like `ver=` while parsing this? Couldn't we just use `let version = '1.0'` in line 46?

src/courseware/course/CourseLicense.jsx
```diff
@@ -0,0 +1,114 @@
...
+        version = value;
+      } else {
+        options[key.toLowerCase()] = value.toLowerCase();
+      }
+    } else {
+      options[option.toLowerCase()] = true;
+    }
+  });
+
+  if (options.length === 0) {
```
Agrendalath: Isn't this `undefined` for an `Object`? Shouldn't we use something like this instead?
```suggestion
  if (Object.keys(options).length === 0) {
```

src/courseware/course/CourseLicense.jsx
```diff
@@ -0,0 +1,114 @@
...
+  license,
+}) {
+  const renderAllRightsReservedLicense = () => (
+    <div>
+      <FontAwesomeIcon aria-hidden="true" className="mr-1" icon={faCopyright} />
+      <span className="license-text">All Rights Reserved</span>
+    </div>
+  );
+
+  const renderCreativeCommonsLicense = (activeCreativeCommonsLicenseTags = [], version) => (
```
Agrendalath: Nit: the default arg for `activeCreativeCommonsLicenseTags` will never be used, but sounds fine to keep it.

src/courseware/course/CourseLicense.jsx
```diff
@@ -0,0 +1,114 @@
...
+    >
+      <span className="sr-only">Creative Commons licensed content, with terms as follow:&nbsp;</span>
+      <FontAwesomeIcon aria-hidden="true" className="mr-1" icon={faCreativeCommons} />
+      {activeCreativeCommonsLicenseTags.map(tag => (
+        <span key={tag}>
+          <span className="sr-only">{CreativeCommonsLicenseTags[tag].text}&nbsp;</span>
+          <FontAwesomeIcon aria-hidden="true" className="mr-1" icon={CreativeCommonsLicenseTags[tag].icon} />
+        </span>
+      ))}
+      <span className="license-text">Some Rights Reserved</span>
```
Agrendalath: These texts (plus `sr-only` messages) should be wrapped with something like `{intl.formatMessage(messages['learn.course.load.failure'])` to allow translations.

src/courseware/course/CourseLicense.jsx
```diff
@@ -0,0 +1,114 @@
...
+      {licenseType === 'creative-commons' && renderCreativeCommonsLicense(
+        Object.keys(licenseOptions),
+        licenseVersion,
+      )}
+    </div>
+  );
+}
+
+CourseLicense.propTypes = {
+  license: PropTypes.string.isRequired,
```
Agrendalath: It might not be `required`, depending on how we'll handle the `null` values from the API.

Agrendalath (changes requested):

src/courseware/course/CourseLicense.jsx
pcockwell: I tried setting an `enum` in the code, and kept getting errors. Do you have any examples of where it has successfully been implemented? Documentation online seems sparse regarding the requirements to make this work.

src/courseware/course/CourseLicense.jsx
pcockwell: Done in next commit

src/courseware/course/CourseLicense.jsx
pcockwell: Cleaned up in next commit

src/courseware/course/CourseLicense.jsx
pcockwell: Comments added in next commit

src/courseware/course/CourseLicense.jsx
pcockwell: Done in next commit

src/courseware/course/CourseLicense.jsx
pcockwell: Removed in next commit

src/courseware/course/CourseLicense.jsx
pcockwell: Done in next commit

src/courseware/course/CourseLicense.jsx
Agrendalath: Ah, sorry, I got confused and thought we were using TypeScript for a moment…
It would look like this for JS:
```js
const CreativeCommonsLicenseTagsEnum = {
  Attribution: 'by',
  Noncommercial: 'nc',
  NoDerivatives: 'nd',
  ShareAlike: 'sa',
};

const CreativeCommonsLicenseTags = {
  [CreativeCommonsLicenseTagsEnum.Attribution]: {
    text: 'Attribution',
    icon: faCreativeCommonsBy,
  },
  [CreativeCommonsLicenseTagsEnum.Noncommercial]: {
    text: 'Noncommercial',
    icon: faCreativeCommonsNc,
  },
  [CreativeCommonsLicenseTagsEnum.NoDerivatives]: {
    text: 'No Derivatives',
    icon: faCreativeCommonsNd,
  },
  [CreativeCommonsLicenseTagsEnum.ShareAlike]: {
    text: 'Share Alike',
    icon: faCreativeCommonsSa,
  },
};
```

I guess it's fine to keep this as is, though.

src/courseware/course/course-license/messages.js
```diff
@@ -0,0 +1,47 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  'learn.courseLicense.allRightsReserved.license.text': {
+    id: 'learn.courseLicense.allRightsReserved.license.text',
+    defaultMessage: 'All Rights Reserved',
+    description: 'License text shown when using all All Rights Reserved license type.',
```
Agrendalath: Type: doubled "all"?

src/courseware/course/course-license/messages.js
```diff
@@ -0,0 +1,47 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  'learn.courseLicense.allRightsReserved.license.text': {
```
Agrendalath: Could we make this `learn.course.license.licenseType.text`, etc? It will clearly indicate the hierarchy.
E.g. `learn.course.license.allRightsReserved.text`?

src/courseware/course/course-license/messages.js
```diff
@@ -0,0 +1,47 @@
...
+    id: 'learn.courseLicense.allRightsReserved.license.text',
+    defaultMessage: 'All Rights Reserved',
+    description: 'License text shown when using all All Rights Reserved license type.',
+  },
+  'learn.courseLicense.creativeCommons.license.terms.preamble': {
+    id: 'learn.courseLicense.creativeCommons.license.terms.preamble',
+    defaultMessage: 'Creative Commons licensed content, with terms as follow:',
```
Agrendalath: Typo: shouldn't it be "follows"?

src/courseware/course/course-license/CourseLicense.jsx
```diff
@@ -8,91 +9,132 @@ import {
...
   });
 
-  if (options.length === 0) {
-    version = version || '1.0';
-    // No options
+  // No options
+  if (Object.keys(options).length === 0) {
+    // If no other options are set for the
+    // license, set defaultVersion to 1.0
+    defaultVersion = '1.0';
```
Agrendalath: I suppose we should set `version` here? It couldn't have been set before anyway, as there were no `options`.

Agrendalath (commented):
Thank you for addressing the comments. Looks good, @pcockwell!
I left come minor comments. Let's wait for Marco to clarify the empty license case.

src/courseware/course/course-license/messages.js
pcockwell: Fixed in next commit

src/courseware/course/course-license/messages.js
pcockwell: Fixed in next commit

src/courseware/course/course-license/messages.js
pcockwell: Fixed in next commit

src/courseware/course/course-license/CourseLicense.jsx
pcockwell: Fixed in next commit

src/index.scss
```diff
@@ -262,6 +262,23 @@ $primary: #1176B2;
   }
 }
 
+.sequence-footer {
```
davidjoy: Where possible, it'd be good to use our CSS utility classes rather than having a semantic class name like this one.

https://edx.github.io/paragon/foundations/typography
https://edx.github.io/paragon/foundations/css-utilities

I think everything but that padding should be possible via classes shown on those two pages.  

What's the idea behind the percentage horizontal padding, rather than just choosing a particular number?  Does it change that much/improve the look on smaller devices?

src/courseware/course/course-license/CourseLicense.jsx
```diff
@@ -0,0 +1,161 @@
...
+        licenseVersion,
+      )}
+    </div>
+  );
+}
+
+CourseLicense.propTypes = {
+  license: PropTypes.string.isRequired,
+  intl: intlShape.isRequired,
+};
```
Agrendalath: This is going to work, but will show an error in the console. Wouldn't it be better to use a default value like this instead of checking `!license` in `parseLicense`?
```suggestion
CourseLicense.propTypes = {
  license: PropTypes.string,
  intl: intlShape.isRequired,
};

CourseLicense.defaultProps = {
  license: 'all-rights-reserved',
};
```

However this means that we need to create this component this way, because the API would return `null`, not `undefined`:
```js
<CourseLicense license={course.license || undefined} />
```

Agrendalath (approved):
Works great! Leaving one additional thing, but this will not need another review round from my end after being changed. Also please address David's comment.

Btw. please don't squash the commits before finishing the review. GitHub falsely flags some comments as outdated in such case (plus it's hard to see the actual changes in particular files).

:+1:
* [x]  I tested this: tested all license types locally
* [x]  I read through the code
* [x]  I checked for accessibility issues
* [ ]  <s>Includes documentation</s>

src/index.scss
pcockwell: I've been able to reproduce all of this CSS using utility classes except for the `:hover` effect on the link, which I've kept in the `index.scss` file.

pcockwell:
@ormsbee I believe everything should be resolved as of the most recent commits. Could you take another pass? Once ready, I'll rebase and squash.

Agrendalath:
@pcockwell, Travis reports one linting issue and two outdated snapshots for integration tests. Could you please:
1. change the linting accordingly,
1. add `license` to the [course metadata factory](https://github.com/edx/frontend-app-learning/blob/master/src/courseware/data/__factories__/courseMetadata.factory.js),
3. update outdated snapshots?

pcockwell:
@Agrendalath Those issues should now be resolved.

src/courseware/course/sequence/Sequence.jsx
```diff
@@ -154,6 +156,9 @@ function Sequence({
             )}
           </div>
         </div>
+        <div className="sequence-footer px-4 py-1">
+          <CourseLicense license={course.license || undefined} />
```
davidjoy, line 163: This is a bit of a nit: If we handle coercing falsy course license values in the api.js layer, we can do away with this `|| undefined` here and create a more consistent value throughout the app (i.e., use our anti-corruption layer).  Also, we always expect the `license` field to be defined in the response, so my preference would be for it to be `null` instead of `undefined`.

src/courseware/data/api.js
```diff
@@ -127,6 +127,7 @@ function normalizeMetadata(metadata) {
     isEnrolled: metadata.enrollment.is_active,
     canLoadCourseware: camelCaseObject(metadata.can_load_courseware),
     isStaff: metadata.is_staff,
+    license: metadata.license,
```
davidjoy, line 130: In the same way that we coerce other values to be camelCase, have different names, etc., we could ensure license is truthy or `null` here.  That way we don't have to be protective later.

davidjoy: (This code is the "anti-corruption layer" mentioned above)

davidjoy (changes requested):
Small nit, but worth touching I think.  Otherwise looks great/will Approve ASAP

src/courseware/course/sequence/Sequence.jsx
Agrendalath: @davidjoy, I actually suggested this approach, because `null` [doesn't fall back](https://reactjs.org/docs/react-component.html#defaultprops) to [defaultProps](https://github.com/edx/frontend-app-learning/pull/103/files#diff-d3d8c9fa2454167844ad1ce727c81d04R162-R164). Should we define the default license on the api.js layer or somewhere in the `CourseLicense` component's code instead?

davidjoy: Ah - we can just go ahead with this then, if there were issues.  It's probably not worth having to rethink it for now.

davidjoy (approved):

pcockwell:
@davidjoy My understanding is that this and https://github.com/edx/edx-platform/pull/24393 are ready to merge. Is there anything further you need?

PR #104, opened 2020-07-09
Treat html-type XBlocks as units

kdmccormick:
```
Fixes bug where courses with html-type XBlocks as children
of sequences would ignore those children instead of treating them
as units. This caused the app to later just give up and redirect
to the course home in the old experience.

Also, handle that scenario where we have sections/sequences
with children of unexpected block types more gracefully by
logging an error instead of crashing.
```

https://openedx.atlassian.net/TNL-7305

### DIFF
--- a/src/courseware/data/api.js
+++ b/src/courseware/data/api.js
@@ -34,6 +34,7 @@ export function normalizeBlocks(courseId, blocks) {
           unitIds: block.children || [],
         };
         break;
+      case 'html':
       case 'vertical':
         models.units[block.id] = {
           graded: block.graded,
@@ -43,7 +44,7 @@ export function normalizeBlocks(courseId, blocks) {
         };
         break;
       default:
-        logError(`Unexpected course block type: ${block.type} with ID ${block.id}.  Expected block types are course, chapter, sequential, and vertical.`);
+        logError(`Unexpected course block type: ${block.type} with ID ${block.id}.  Expected block types are course, chapter, sequential, vertical, and html.`);
     }
   });
 
@@ -61,7 +62,11 @@ export function normalizeBlocks(courseId, blocks) {
   Object.values(models.sections).forEach(section => {
     if (Array.isArray(section.sequenceIds)) {
       section.sequenceIds.forEach(sequenceId => {
-        models.sequences[sequenceId].sectionId = section.id;
+        if (sequenceId in models.sequences) {
+          models.sequences[sequenceId].sectionId = section.id;
+        } else {
+          logError(`Section ${section.id} has child block ${sequenceId}, but that block is not in the list of sequences.`);
+        }
       });
     }
   });
@@ -69,7 +74,11 @@ export function normalizeBlocks(courseId, blocks) {
   Object.values(models.sequences).forEach(sequence => {
     if (Array.isArray(sequence.unitIds)) {
       sequence.unitIds.forEach(unitId => {
-        models.units[unitId].sequenceId = sequence.id;
+        if (unitId in models.units) {
+          models.units[unitId].sequenceId = sequence.id;
+        } else {
+          logError(`Sequence ${sequence.id} has child block ${unitId}, but that block is not in the list of units.`);
+        }
       });
     }
   });


src/courseware/data/api.js
```diff
@@ -34,6 +34,7 @@ export function normalizeBlocks(courseId, blocks) {
           unitIds: block.children || [],
         };
         break;
+      case 'html':
```
davidjoy, line 37: Yikes.  I'd feared there were other types I didn't know about... the question should probably be asked, is `html` the only one we were missing!?

davidjoy (approved):

src/courseware/data/api.js
kdmccormick: Good question. I'll pose it to the Slack channel.

ormsbee:
Late comment, but HTML is a leaf-level block. I'd prefer to talk about what we're trying to accomplish here, because I'm guessing we should either:

1. Assume that this is always a Unit of some sort, with a common set of parameters/metadata that we know to expect; or
2. Assume that the children of a Sequence can be _any_ XBlock.

But doing a one-off exception for Units like this strikes me as an invitation to complicate things down the road.

kdmccormick:
@ormsbee Ahhh gotcha. I agree that we shouldn't handle this as a special case. I'll roll a revert to everything except the part where we log an error instead of crashing+redirecting. Once deployed, most units in 7.00x should work in the MFE, although it'll probably fail on the mis-structured HTML blocks.

For the immediate fix, is it fair for us to tell the course authors that they need to fix the course structure, such that the HTML blocks are all nested under units?

For the longer term, let's talk about the question you posed. I don't have much perspective on it, but I'm interested to see how we answer it.

kdmccormick:
@ormsbee Threw this together: https://github.com/edx/frontend-app-learning/pull/105/

ormsbee:
@kdmccormick (and also @davidjoy, @davestgermain, @marcotuts, @JAAkana):

> For the immediate fix, is it fair for us to tell the course authors that they need to fix the course structure, such that the HTML blocks are all nested under units?

Yes, I think that's reasonable.

> For the longer term, let's talk about the question you posed. I don't have much perspective on it, but I'm interested to see how we answer it.

I'm going to dump a lot of context here, as well as some of the future direction that we're marching towards. I've got a personal backlog of ADRs that I need to synthesize some of these into, but for now I'm just going to dump my thoughts here.

### Historical Context

There was an explicit early decision to allow for more freeform structures in the LMS, even as we made Studio more rigid to simplify authoring. The vast majority of work around XModules/XBlocks has been around implementing new assessment types, but it was originally envisioned as a much more flexible platform. One early course had a chapter with a child unit, which had a video on top and a sequence on the bottom (Chapter -> Unit -> [Video, Sequence -> [Units]]), so that you could quickly refer back to the video even as you were doing assessments.

Open edX was originally envisioned as a platform for learning research and experimentation. XBlocks were built with a "turtles all the way down" sort of philosophy with the idea that people were going to experiment with different navigational structures. When Studio came around, the focus was very different: Adapt existing in-person courses as quickly as possible. That made sense for the needs at the time, but you can see that tension play out in a bunch of places in the guts of the platform as a result.

The other major thing that happened was that the open-endedness of XBlock and the lack of separation of the navigational/orchestration pieces of it from the leaf node portions made it so that it was extremely hard to make operational and performance guarantees about courseware as a whole. These stresses happened during a time of rapid growth for us, and we made a slew of short-term compromises that assumed certain structural patterns in order to do caching and optimization. We had parallel access patterns that would avoid instantiating student state if at all possible. This is why we have such byzantine caching layers for the XBlock runtime. This is also a major source of the complexity we have in the system today. The fundamental XBlock runtime interfaces are not complicated–you get services, view handlers, and state storage. But we have a nightmare of duct tape to keep it running.

The worst part of it is that all this duct tape both complicated the implementation as well as limited the expressive power of the system. Take the `max_score` implementation in `x_module.py` for example:

```python
def max_score(self):
    """ Maximum score. Two notes:
        * This is generic; in abstract, a problem could be 3/5 points on one
          randomization, and 5/7 on another
        * In practice, this is a Very Bad Idea, and (a) will break some code
          in place (although that code should get fixed), and (b) break some
          analytics we plan to put in place.
    """
    return None
```

### Where we are today

We have a bunch of things that are way more freeform than people think if they dig into the OLX instead of relying on Studio's UI. Want to set the due date at the root Course level and have everything due at once? The `due` date is an inherited field that will be percolated through every child, so you can do that. Want to have a doubly nested sequence? That's of extremely questionable pedagogical value, but you can do it.

But at the same time, the moment you step outside the well worn path of Studio, nothing's guaranteed to work. Yeah, that double nested sequence might render, but are the tracking events going to make any sense? Sure you can set due dates at the root course level, but will that work with due date extensions set in edx-when? Yes, you can create a piece of content that has multiple parents, but lol if you want to get a direct `jump_to` link to the second instance of it.

A good deal of that owes to the fact that there is almost nothing in the way of formalized APIs and contracts. XBlock was designed to be pretty opaque. You're supposed to say, "render this XBlock here", and let it figure out if it needs to render children or whatever. But we didn't have a precise hierarchy saying "The child of a sequence must be a container XBlock that has Unit-like interface with a progress() call that will return a tuple of modules completed vs. total modules..." It was basically: "Render your child element here." and everything else has been convention or mixins.

### Where I think we should go

We're moving forward on a lot of different fronts with courseware, but here are the big themes of work that I think this relates to:

#### Separate orchestration/navigation (XBlock-agnostic) from Unit-level content.

We want to eventually support multiple navigational types. We need to do this for some type of adaptive learning work. But that should be a separate API interface with stronger data model control. This is happening now in the form of the `learning_sequences` app that is the center of the Course Outline stories currently being worked on.

The top level things that we might want to compose in different ways are: 

* What learning sequences make up a course/pathway of instruction?
* What is the experience of this particular learning sequence (static, adaptive, whatever)?
* What powers an individual Unit within this learning sequence?

#### Create XBlock-independent systems.

XBlock's extremely open ended nature makes it too difficult to make any sort of runtime guarantees. So we create systems to push data into that XBlocks use, rather than querying the XBlocks themselves. This has already happened with grading and scheduling.

The end goal is to significantly shrink what the XBlock runtime is actually responsible for. It's not making XBlock performant, but reducing what we ask of it to be just the rendering of a single Unit at a time (or smaller). Inheritance, broad course-wide tree traversals–all of that goes away, making performance much more predictable.

#### Separate the LMS concept of a Unit from XBlock.

Right now, a Unit _is_ an XBlock (well, a VerticalBlock specifically). But there's no reason why QTI, LTI, and whatever we create as the next generation of XBlock needs to be stuffed inside the XBlock runtime. We can separate the LMS concept of a Unit meaning "a single molecule of content to deliver to the user to teach them something" from XBlock specifics. So again, you'd hit an LMS app (not yet created) with its own data model to determine what a Unit is and what metadata it possesses. That app knows how to dispatch you to the correct runtime to actually render the unit.

It's this last bit that's somewhat related to this topic. I _don't_ think it's necessary to introduce this kind of courseware refactoring work yet, but I would like to evaluate it after the new Course Outline work is done.

kdmccormick:
Thanks for the great run-down @ormsbee!

PR #106, opened 2020-07-10
Add masquerade as specific student support

stvstnfrd:
This must merge _before_ the platform PR.

Screenshot_2020-07-13_03-19-44

Screenshot_2020-07-13_03-20-35

Screenshot_2020-07-13_03-20-54

Screenshot_2020-07-13_03-21-24

Screenshot_2020-07-13_03-21-51

Screenshot_2020-07-13_03-22-36

Screenshot_2020-07-13_03-22-56

Screenshot_2020-07-13_03-29-32

Screenshot_2020-07-13_03-29-19

### DIFF
--- a/src/instructor-toolbar/InstructorToolbar.jsx
+++ b/src/instructor-toolbar/InstructorToolbar.jsx
@@ -51,23 +51,21 @@ export default function InstructorToolbar(props) {
   return (
     <div className="bg-primary text-light">
       <div className="container-fluid py-3 d-md-flex justify-content-end align-items-center">
-        <div className="flex-grow-1">
+        <div className="align-items-center flex-grow-1 d-md-flex mx-1 my-1">
           <MasqueradeWidget courseId={courseId} />
         </div>
         {urlLms && (
-          <div className="flex-shrink-0">
+          <div className="flex-shrink-0 mx-1 my-1">
             <a className="btn d-block btn-outline-light" href={urlLms}>View in the existing experience</a>
           </div>
         )}
-        &nbsp;
         {urlStudio && (
-          <div className="flex-shrink-0">
+          <div className="flex-shrink-0 mx-1 my-1">
             <a className="btn d-block btn-outline-light" href={urlStudio}>View in Studio</a>
           </div>
         )}
-        &nbsp;
         {urlInsights && (
-          <div className="flex-shrink-0">
+          <div className="flex-shrink-0 mx-1 my-1">
             <a className="btn d-block btn-outline-light" href={urlInsights}>View in Insights</a>
           </div>
         )}
--- a/src/instructor-toolbar/masquerade-widget/MasqueradeUserNameInput.jsx
+++ b/src/instructor-toolbar/masquerade-widget/MasqueradeUserNameInput.jsx
@@ -1,0 +1,60 @@
+import React, {
+  Component,
+} from 'react';
+import PropTypes from 'prop-types';
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import { Input } from '@edx/paragon';
+
+import messages from './messages';
+
+class MasqueradeUserNameInput extends Component {
+  onError(...args) {
+    return this.props.onError(...args);
+  }
+
+  onKeyPress(event) {
+    if (event.key === 'Enter') {
+      return this.onSubmit(event);
+    }
+    return true;
+  }
+
+  onSubmit(event) {
+    const payload = {
+      role: 'student',
+      user_name: event.target.value,
+    };
+    this.props.onSubmit(payload).then((data) => {
+      if (data && data.success) {
+        global.location.reload();
+      } else {
+        const error = (data && data.error) || '';
+        this.onError(error);
+      }
+    }).catch(() => {
+      const message = this.props.intl.formatMessage(messages['userName.error.generic']);
+      this.onError(message);
+    });
+    return true;
+  }
+
+  render() {
+    return (
+      <Input
+        autoFocus
+        className="flex-shrink-1"
+        defaultValue=""
+        label={this.props.intl.formatMessage(messages['userName.input.label'])}
+        onKeyPress={(event) => this.onKeyPress(event)}
+        placeholder={this.props.intl.formatMessage(messages['userName.input.placeholder'])}
+        type="text"
+      />
+    );
+  }
+}
+MasqueradeUserNameInput.propTypes = {
+  intl: intlShape.isRequired,
+  onError: PropTypes.func.isRequired,
+  onSubmit: PropTypes.func.isRequired,
+};
+export default injectIntl(MasqueradeUserNameInput);
--- a/src/instructor-toolbar/masquerade-widget/MasqueradeWidget.jsx
+++ b/src/instructor-toolbar/masquerade-widget/MasqueradeWidget.jsx
@@ -2,10 +2,21 @@ import React, {
   Component,
 } from 'react';
 import PropTypes from 'prop-types';
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import { Dropdown } from '@edx/paragon';
 
+import {
+  ALERT_TYPES,
+  UserMessagesContext,
+} from '../../generic/user-messages';
+
+import MasqueradeUserNameInput from './MasqueradeUserNameInput';
 import MasqueradeWidgetOption from './MasqueradeWidgetOption';
-import { getMasqueradeOptions } from './data/api';
+import {
+  getMasqueradeOptions,
+  postMasqueradeOptions,
+} from './data/api';
+import messages from './messages';
 
 class MasqueradeWidget extends Component {
   constructor(props) {
@@ -13,33 +24,100 @@ class MasqueradeWidget extends Component {
     this.courseId = props.courseId;
     this.state = {
       options: [],
+      shouldShowUserNameInput: false,
     };
   }
 
   componentDidMount() {
     getMasqueradeOptions(this.courseId).then((data) => {
       if (data.success) {
-        const options = this.parseAvailableOptions(data);
-        this.setState({
-          options,
-        });
+        this.onSuccess(data);
       } else {
+        // This was explicitly denied by the backend;
+        // assume it's disabled/unavailable.
         // eslint-disable-next-line no-console
-        console.warn('Unable to get masquerade options', data);
+        this.onError('Unable to get masquerade options');
       }
     }).catch((response) => {
+      // There's not much we can do to recover;
+      // if we can't fetch masquerade options,
+      // assume it's disabled/unavailable.
       // eslint-disable-next-line no-console
       console.error('Unable to get masquerade options', response);
     });
   }
 
-  parseAvailableOptions(payload) {
-    const data = payload || {};
+  onError(message) {
+    if (message) {
+      this.errorAlertId = this.context.add({
+        text: message,
+        topic: 'course',
+        type: ALERT_TYPES.ERROR,
+        dismissible: false,
+      });
+    }
+  }
+
+  async onSubmit(payload) {
+    this.context.remove(this.errorAlertId);
+    const options = await postMasqueradeOptions(this.courseId, payload);
+    return options;
+  }
+
+  onSuccess(data) {
+    const options = this.parseAvailableOptions(data);
+    this.setState({
+      options,
+    });
+    const active = data.active || {};
+    const message = this.getStatusMessage(active);
+    if (message) {
+      this.context.add({
+        text: message,
+        topic: 'course',
+        type: ALERT_TYPES.INFO,
+        dismissible: false,
+      });
+    }
+  }
+
+  getStatusMessage(active) {
+    const {
+      groupName,
+    } = active;
+    let message = '';
+    if (active.userName) {
+      message = this.props.intl.formatMessage(messages['status.userName'], {
+        userName: active.userName,
+      });
+    } else if (groupName) {
+      message = this.props.intl.formatMessage(messages['status.groupName'], {
+        groupName,
+      });
+    } else if (active.role === 'student') {
+      message = this.props.intl.formatMessage(messages['status.learner']);
+    }
+    return message;
+  }
+
+  toggle(show) {
+    let shouldShow;
+    if (show === undefined) {
+      shouldShow = !this.state.shouldShowUserNameInput;
+    } else {
+      shouldShow = show;
+    }
+    this.setState({
+      shouldShowUserNameInput: shouldShow,
+    });
+  }
+
+  parseAvailableOptions(postData) {
+    const data = postData || {};
     const active = data.active || {};
     const available = data.available || [];
     const options = available.map((group) => (
       <MasqueradeWidgetOption
-        courseId={this.courseId}
         groupId={group.groupId}
         groupName={group.name}
         key={group.name}
@@ -47,6 +125,9 @@ class MasqueradeWidget extends Component {
         selected={active}
         userName={group.userName}
         userPartitionId={group.userPartitionId}
+        userNameInput={this.userNameInput}
+        userNameInputToggle={(...args) => this.toggle(...args)}
+        onSubmit={(payload) => this.onSubmit(payload)}
       />
     ));
     return options;
@@ -57,18 +138,34 @@ class MasqueradeWidget extends Component {
       options,
     } = this.state;
     return (
-      <Dropdown>
-        <Dropdown.Button>
-          View this course as
-        </Dropdown.Button>
-        <Dropdown.Menu>
-          {options}
-        </Dropdown.Menu>
-      </Dropdown>
+      <>
+        <Dropdown
+          className="flex-shrink-1 mx-1 my-1"
+          style={{ textAlign: 'center' }}
+        >
+          <Dropdown.Button>
+            View this course as
+          </Dropdown.Button>
+          <Dropdown.Menu>
+            {options}
+          </Dropdown.Menu>
+        </Dropdown>
+        {this.state.shouldShowUserNameInput && (
+          <MasqueradeUserNameInput
+            className="flex-shrink-0 mx-1 my-1"
+            label="test"
+            onError={(errorMessage) => this.onError(errorMessage)}
+            onSubmit={(payload) => this.onSubmit(payload)}
+            ref={(input) => { this.userNameInput = input; }}
+          />
+        )}
+      </>
     );
   }
 }
 MasqueradeWidget.propTypes = {
   courseId: PropTypes.string.isRequired,
+  intl: intlShape.isRequired,
 };
-export default MasqueradeWidget;
+MasqueradeWidget.contextType = UserMessagesContext;
+export default injectIntl(MasqueradeWidget);
--- a/src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx
+++ b/src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx
@@ -4,18 +4,28 @@ import React, {
 import PropTypes from 'prop-types';
 import { Dropdown } from '@edx/paragon';
 
-import { postMasqueradeOptions } from './data/api';
-
 class MasqueradeWidgetOption extends Component {
-  handleClick() {
+  onClick(event) {
+    // TODO: Remove this hack when we upgrade Paragon
+    // Note: The current version of Paragon does _not_ close dropdown components
+    // automatically (or easily programmatically) when you click on an item.
+    // We can simulate this behavior by programmatically clicking the
+    // toggle button on behalf of the user.
+    // The newest version of Paragon already contains this behavior,
+    // so we can remove this when we upgrade to that point.
+    event.target.parentNode.parentNode.click();
     const {
-      courseId,
       groupId,
       role,
       userName,
       userPartitionId,
+      userNameInputToggle,
     } = this.props;
     const payload = {};
+    if (userName || userName === '') {
+      userNameInputToggle(true);
+      return false;
+    }
     if (role) {
       payload.role = role;
     }
@@ -23,21 +33,24 @@ class MasqueradeWidgetOption extends Component {
       payload.group_id = parseInt(groupId, 10);
       payload.user_partition_id = parseInt(userPartitionId, 10);
     }
-    if (userName) {
-      payload.user_name = userName;
-    }
-    postMasqueradeOptions(courseId, payload).then(() => {
+    this.props.onSubmit(payload).then(() => {
       global.location.reload();
     });
+    return true;
   }
 
   isSelected() {
-    const selected = this.props.selected || {};
-    const isEqual = (
-      selected.userPartitionId === (this.props.userPartitionId || null)
-      && selected.groupId === (this.props.groupId || null)
-      && selected.role === this.props.role
-    );
+    /* eslint-disable arrow-body-style */
+    const isEqual = [
+      'groupId',
+      'role',
+      'userName',
+      'userPartitionId',
+    ].reduce((accumulator, currentValue) => {
+      return accumulator && (
+        this.props[currentValue] === this.props.selected[currentValue]
+      );
+    }, true);
     return isEqual;
   }
 
@@ -57,7 +70,7 @@ class MasqueradeWidgetOption extends Component {
       <Dropdown.Item
         className={className}
         href="#"
-        onClick={(event) => this.handleClick(event)}
+        onClick={(event) => this.onClick(event)}
       >
         {groupName}
       </Dropdown.Item>
@@ -65,9 +78,9 @@ class MasqueradeWidgetOption extends Component {
   }
 }
 MasqueradeWidgetOption.propTypes = {
-  courseId: PropTypes.string.isRequired,
   groupId: PropTypes.number,
   groupName: PropTypes.string.isRequired,
+  onSubmit: PropTypes.func.isRequired,
   role: PropTypes.string,
   selected: PropTypes.shape({
     courseKey: PropTypes.string.isRequired,
@@ -77,6 +90,7 @@ MasqueradeWidgetOption.propTypes = {
     userPartitionId: PropTypes.number,
   }),
   userName: PropTypes.string,
+  userNameInputToggle: PropTypes.func.isRequired,
   userPartitionId: PropTypes.number,
 };
 MasqueradeWidgetOption.defaultProps = {
--- a/src/instructor-toolbar/masquerade-widget/data/api.js
+++ b/src/instructor-toolbar/masquerade-widget/data/api.js
@@ -7,8 +7,8 @@ export async function getMasqueradeOptions(courseId) {
   return camelCaseObject(data);
 }
 
-export async function postMasqueradeOptions(courseId, data) {
+export async function postMasqueradeOptions(courseId, payload) {
   const url = new URL(`${getConfig().LMS_BASE_URL}/courses/${courseId}/masquerade`);
-  const { response } = await getAuthenticatedHttpClient().post(url.href, data);
-  return camelCaseObject(response);
+  const { data } = await getAuthenticatedHttpClient().post(url.href, payload);
+  return camelCaseObject(data);
 }
--- a/src/instructor-toolbar/masquerade-widget/messages.js
+++ b/src/instructor-toolbar/masquerade-widget/messages.js
@@ -1,0 +1,36 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  'status.groupName': {
+    id: 'masquerade-widget.status.groupName',
+    defaultMessage: 'You are masquerading as a learner in the {groupName} group.',
+    description: 'Message when masquerading as a generic user in a specific track',
+  },
+  'status.learner': {
+    id: 'masquerade-widget.status.learner',
+    defaultMessage: 'You are masquerading as a learner.',
+    description: 'Message when masquerading as a specific user',
+  },
+  'status.userName': {
+    id: 'masquerade-widget.status.userName',
+    defaultMessage: 'You are masquerading as the following user: {userName}',
+    description: 'Message when masquerading as a specific user',
+  },
+  'userName.input.label': {
+    id: 'masquerade-widget.userName.input.label',
+    defaultMessage: 'Masquerade as this user',
+    description: 'Label for the masquerade user input',
+  },
+  'userName.error.generic': {
+    id: 'masquerade-widget.userName.error.generic',
+    defaultMessage: 'An error has occurred; please try again.',
+    description: 'Message shown after a general error when attempting to masquerade',
+  },
+  'userName.input.placeholder': {
+    id: 'masquerade-widget.userName.input.placeholder',
+    defaultMessage: 'username or email',
+    description: 'Placeholder text to prompt for a user to masquerade as',
+  },
+});
+
+export default messages;


src/instructor-toolbar/masquerade-widget/data/api.js
```diff
@@ -7,8 +7,8 @@ export async function getMasqueradeOptions(courseId) {
   return camelCaseObject(data);
 }
 
-export async function postMasqueradeOptions(courseId, data) {
+export async function postMasqueradeOptions(courseId, payload) {
   const url = new URL(`${getConfig().LMS_BASE_URL}/courses/${courseId}/masquerade`);
-  const { response } = await getAuthenticatedHttpClient().post(url.href, data);
-  return camelCaseObject(response);
+  const { data } = await getAuthenticatedHttpClient().post(url.href, payload);
```
stvstnfrd, line 12: We have to unpack `data`, not `response`, else we get nothing.

This was "broken" before, but we weren't using the response yet, so it was irrelevant.

src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx
```diff
@@ -4,40 +4,49 @@ import React, {
...
-      && selected.groupId === (this.props.groupId || null)
-      && selected.role === this.props.role
-    );
+    /* eslint-disable arrow-body-style */
+    const isEqual = [
+      'groupId',
+      'role',
+      'userName',
+      'userPartitionId',
+    ].reduce((accumulator, currentValue) => {
```
stvstnfrd, line 49: Is there a more idiomatic way of doing this w/ JS/React?

This is essentially an equality check of the two objects, looking only at select properties.

src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx
```diff
@@ -4,40 +4,49 @@ import React, {
...
+      return accumulator && (
+        this.props[currentValue] === this.props.selected[currentValue]
+      );
+    }, true);
```
stvstnfrd, line 53: Initialize the state to `true`.

src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx
```diff
@@ -4,40 +4,49 @@ import React, {
...
-      && selected.role === this.props.role
-    );
+    /* eslint-disable arrow-body-style */
+    const isEqual = [
+      'groupId',
+      'role',
+      'userName',
+      'userPartitionId',
+    ].reduce((accumulator, currentValue) => {
+      return accumulator && (
```
stvstnfrd, line 50: Start short-circuiting once any check returns `false`.

src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx
```diff
@@ -4,40 +4,49 @@ import React, {
...
       global.location.reload();
     });
+    return true;
   }
 
   isSelected() {
-    const selected = this.props.selected || {};
-    const isEqual = (
-      selected.userPartitionId === (this.props.userPartitionId || null)
-      && selected.groupId === (this.props.groupId || null)
```
stvstnfrd, line 38: These two coalescing operators are unnecessary;
types are consistent for our purposes.

src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx
```diff
@@ -4,40 +4,49 @@ import React, {
...
-      && selected.role === this.props.role
-    );
+    /* eslint-disable arrow-body-style */
```
stvstnfrd, line 43: Intentionally ignoring this...

I get why it's complaining (the arrow function really is single-line), but this just feels too long for a single line;
I think readability would suffer.

Open to be overruled though.

src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx
```diff
@@ -4,40 +4,49 @@ import React, {
...
     const {
-      courseId,
       groupId,
       role,
       userName,
       userPartitionId,
+      userNameInput,
     } = this.props;
     const payload = {};
+    // TODO: Close the dropdown
```
stvstnfrd: I went through Paragon docs, but couldn't quite tell if this was directly supported or not.

The dropdown _does_ automatically close if you click outside of the element, but not when the `<input />` grabs `focus()`.

We should close the dropdown when the text input is displayed, but I hadn't got that working.

This wasn't an issue w/ the other masquerade options, since we reload the page, instead of showing an input.

src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx
```diff
@@ -4,40 +4,49 @@ import React, {
...
+    // return true;
     if (role) {
       payload.role = role;
     }
     if (groupId) {
       payload.group_id = parseInt(groupId, 10);
       payload.user_partition_id = parseInt(userPartitionId, 10);
     }
-    if (userName) {
-      payload.user_name = userName;
```
stvstnfrd, line 27: We don't need to include the `userName` in the payload, because we'll make specific-user masquerade requests separately (when they submit the text input, not click the dropdown option).

src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx
```diff
@@ -58,16 +66,17 @@ class MasqueradeWidgetOption extends Component {
...
         onClick={(event) => this.handleClick(event)}
+        data-toggle="dropdown"
       >
         {groupName}
       </Dropdown.Item>
     );
   }
 }
 MasqueradeWidgetOption.propTypes = {
-  courseId: PropTypes.string.isRequired,
```
stvstnfrd, line 68: We don't currently need this.

src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx
```diff
@@ -4,40 +4,48 @@ import React, {
 import PropTypes from 'prop-types';
 import { Dropdown } from '@edx/paragon';
 
-import { postMasqueradeOptions } from './data/api';
-
 class MasqueradeWidgetOption extends Component {
   handleClick() {
     const {
-      courseId,
```
stvstnfrd, line 12: We weren't using this any longer.

src/instructor-toolbar/masquerade-widget/MasqueradeWidget.jsx
```diff
@@ -55,14 +95,25 @@ class MasqueradeWidget extends Component {
...
+          </Dropdown.Button>
+          <Dropdown.Menu>
+            {options}
+          </Dropdown.Menu>
+        </Dropdown>
+        <MasqueradeUserNameInput
+          className="flex-shrink-0"
+          onError={() => this.onError.apply(this, arguments)}
+          onSubmit={(payload) => this.onSubmit(payload)}
+          ref={(input) => { this.userNameInput = input; }}
```
stvstnfrd: We create a `ref` to it, so we can call its `toggle` method from in the `MasqueradeWidgetOption`s.

src/instructor-toolbar/masquerade-widget/MasqueradeWidget.jsx
```diff
@@ -55,14 +95,25 @@ class MasqueradeWidget extends Component {
...
-        <Dropdown.Button>
-          View this course as
-        </Dropdown.Button>
-        <Dropdown.Menu>
-          {options}
-        </Dropdown.Menu>
-      </Dropdown>
+      <>
+        <Dropdown
+          className="flex-shrink-1"
```
stvstnfrd: I'll be totally honest, I haven't fully groked `flex` yet, so some of this was trial-and-error.
Let me know if it's off.

src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx
```diff
@@ -57,17 +65,18 @@ class MasqueradeWidgetOption extends Component {
       <Dropdown.Item
         className={className}
         href="#"
-        onClick={(event) => this.handleClick(event)}
+        onClick={(event) => this.onClick(event)}
```
stvstnfrd, line 73: Just renaming for consistency.

src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx
```diff
@@ -4,40 +4,48 @@ import React, {
...
     }
     if (groupId) {
       payload.group_id = parseInt(groupId, 10);
       payload.user_partition_id = parseInt(userPartitionId, 10);
     }
-    if (userName) {
-      payload.user_name = userName;
-    }
-    postMasqueradeOptions(courseId, payload).then(() => {
+    this.props.onSubmit(payload).then(() => {
```
stvstnfrd, line 36: Delegate the actual submission to the helper provided to us by the parent.

src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx
```diff
@@ -4,40 +4,48 @@ import React, {
 import PropTypes from 'prop-types';
 import { Dropdown } from '@edx/paragon';
 
-import { postMasqueradeOptions } from './data/api';
```
stvstnfrd, line 7: Let the parent handle all of the AJAX calls.

src/instructor-toolbar/masquerade-widget/MasqueradeUserNameInput.jsx
```diff
@@ -0,0 +1,80 @@
...
+    });
+  }
+
+  render() {
+    return (this.state.shouldShowUserNameInput
+      && (
+        <>
+          &nbsp;
+          <Input
+            autoFocus
```
stvstnfrd: Magically grab focus when displayed, yay!

src/alerts/masquerade-error-alert/hooks.js
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable import/prefer-default-export */
+import { useMemo } from 'react';
+import { useModel } from '../../generic/model-store';
+import { useAlert } from '../../generic/user-messages';
+
+export function useMasqueradeErrorAlert(courseId) {
+  const masquerade = useModel('masquerades', courseId);
```
stvstnfrd: Note: this data-store model does not exist. I wasn't able to get this all hooked up yet, so I wasn't able to use the built-in alert lists framework. IMO, the alerts look better than the current implementation (a message in the toolbar, similar to existing experience).

If we hook up the data-store, this _should_ just start working. If we don't do that though, let's remove these alerts.

src/alerts/masquerade-status-alert/MasqueradeStatusAlert.jsx
```diff
@@ -0,0 +1,29 @@
...
+  ALERT_TYPES,
+} from '../../generic/user-messages';
+
+function MasqueradeStatusAlert({ payload }) {
+  const {
+    message,
+  } = payload;
+  const isVisible = !!message;
+  return isVisible && (
+    <Alert type={ALERT_TYPES.INFO}>
```
stvstnfrd: It's unfortunate we need to spin up all of this machinery, just to say, "display a one-off alert message with arbitrary text"...

src/alerts/masquerade-status-alert/hooks.js
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable import/prefer-default-export */
+import { useMemo } from 'react';
+import { useModel } from '../../generic/model-store';
+import { useAlert } from '../../generic/user-messages';
+
+export function useMasqueradeStatusAlert(courseId) {
+  const masquerade = useModel('masquerades', courseId);
```
stvstnfrd: Per above, this doesn't exist.
The alert should be remove if unused.

src/instructor-toolbar/InstructorToolbar.jsx
```diff
@@ -51,23 +51,21 @@ export default function InstructorToolbar(props) {
...
           </div>
         )}
-        &nbsp;
         {urlStudio && (
-          <div className="flex-shrink-0">
+          <div className="flex-shrink-0 mx-1 my-1">
             <a className="btn d-block btn-outline-light" href={urlStudio}>View in Studio</a>
           </div>
         )}
-        &nbsp;
```
stvstnfrd, line 68: I learned my lesson, got rid of the spaces, and used proper margins :)

src/instructor-toolbar/masquerade-widget/MasqueradeUserNameInput.jsx
```diff
@@ -0,0 +1,56 @@
...
+    const payload = {
+      role: 'student',
+      user_name: event.target.value,
+    };
+    this.props.onSubmit(payload).then((data) => {
+      if (data && data.success) {
+        global.location.reload();
+      } else {
+        const error = (data && data.error) || '';
+        this.onError(error);
```
stvstnfrd, line 32: Proxy errors back up to the parent component, via the wrapper/prop function.

src/instructor-toolbar/masquerade-widget/MasqueradeUserNameInput.jsx
```diff
@@ -0,0 +1,56 @@
...
+      }
+    }).catch((error) => {
+      this.onError(error);
```
stvstnfrd: TODO: We probably don't want this displayed verbatim; should probably be replaced with a generic message.

src/instructor-toolbar/masquerade-widget/MasqueradeUserNameInput.jsx
```diff
@@ -0,0 +1,56 @@
...
+    return true;
+  }
+
+  render() {
+    return (
+      <Input
+        autoFocus
+        className="flex-shrink-1"
+        defaultValue=""
+        label="TODO"
```
stvstnfrd: TODO: whoops...

src/instructor-toolbar/masquerade-widget/MasqueradeUserNameInput.jsx
```diff
@@ -0,0 +1,56 @@
...
+        onKeyPress={(event) => this.onKeyPress(event)}
+        placeholder="username or email"
```
stvstnfrd: TODO: i18n?

src/instructor-toolbar/masquerade-widget/MasqueradeWidget.jsx
```diff
@@ -4,47 +4,111 @@ import React, {
...
       options: [],
+      error: null,
+      shouldShowUserNameInput: false,
+      status: null,
     };
   }
 
   componentDidMount() {
     getMasqueradeOptions(this.courseId).then((data) => {
+      // There are two failure states:
```
stvstnfrd: TODO: Remove comment block prior to merge.

src/instructor-toolbar/masquerade-widget/MasqueradeWidget.jsx
```diff
@@ -4,47 +4,111 @@ import React, {
...
+      // There's not much we can do to recover;
+      // if we can't fetch masquerade status,
+      // assume it's disabled/unavailable.
       console.error('Unable to get masquerade options', response);
     });
   }
 
-  parseAvailableOptions(payload) {
-    const data = payload || {};
+  onError(message) {
```
stvstnfrd, line 50: If we were to hook into the alert system, this is where we'd trigger the error.

src/instructor-toolbar/masquerade-widget/MasqueradeWidget.jsx
```diff
@@ -4,47 +4,111 @@ import React, {
...
+
+  async onSubmit(payload) {
+    this.setState({
+      shouldShowUserNameInput: false,
+    });
+    const options = await postMasqueradeOptions(this.courseId, payload);
+    return options;
+  }
+
+  onSuccess(data) {
```
stvstnfrd, line 67: If we were to hook into the alert system, this is where we'd trigger the status.

stvstnfrd (commented):
In the process of typing up hand-off notes, I was able to work out a few final kinks, so mostly everything left _should_ be superficial/optional.

TL;DR: Dropdown doesn't always autoclose and the alert/error/status styling could use some color/spacing tweaking.

Outstanding/unaddressed issues:
- Selecting an option from the dropdown does not close the dropdown.
  - This seems to be the default behavior from Paragon and it is how the existing experience works (in the MFE), specifically, that the dropdown _does_ close if you click outside the component, but not if you select an item from _within_ the dropdown. It doesn't even close when it blurs.
  - It wasn't an issue previously, since we refresh the entire page immediately (for the previously existing options). But now, it looks weird if the dropdown doesn't close and you select "view as a specific student".
  - Hopefully this is something easy enough (adding/removing a class, etc.), as it'll improve the UX.
- I couldn't get the messages hooked into the alerting framework.
  - This is probably just me not groking the data-store properly to configure/use it, but I couldn't get it working.
  - The relevant alerting components exist, they just don't do anything w/out the data-store backing. If we can't hook them up, we should revert that commit (and remove the alerting components entirely)
  - So I implemented them as messages directly w/in the instructor toolbar.
    - I'm using the bootstrap alert classes, but the default styling is a bit jarring; it's too light. This could use some tweaking.

src/instructor-toolbar/masquerade-widget/MasqueradeWidget.jsx
```diff
@@ -4,47 +4,111 @@ import React, {
...
-        const options = this.parseAvailableOptions(data);
-        this.setState({
-          options,
-        });
+        this.onSuccess(data);
       } else {
-        console.warn('Unable to get masquerade options', data);
+        // This was explicitly denied by the backend;
+        // assume it's disabled/unavailable.
+        console.error('Unable to get masquerade options', data);
```
stvstnfrd: Nit: I didn't _really_ mean to change this `s/warn/error`...

stvstnfrd:
Screenshot_2020-07-13_03-42-45

Screenshot_2020-07-13_03-49-47

stvstnfrd:
CC: @marcotuts

src/alerts/masquerade-error-alert/MasqueradeErrorAlert.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import {
+  Alert,
+  ALERT_TYPES,
+} from '../../generic/user-messages';
+
+function MasqueradeErrorAlert({ payload }) {
```
davidjoy: From our conversation: Because this error type is basically a generic error and isn't 'special' or custom in any way, we can do away with this error component and just use the default one.

src/alerts/masquerade-status-alert/MasqueradeStatusAlert.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import {
+  Alert,
+  ALERT_TYPES,
+} from '../../generic/user-messages';
+
+function MasqueradeStatusAlert({ payload }) {
```
davidjoy: Same goes for this one.

src/alerts/masquerade-status-alert/MasqueradeStatusAlert.jsx
davidjoy: Agreed! :)

src/alerts/masquerade-error-alert/hooks.js
davidjoy: As we talked about, we can luckily dispense with this hook completely.  If we use the UserMessagesContext in MasqueradeWidget, we can use the `add` function it exposes directly at the time when we know we want to add the message.  

Background on this hook/class component code like it: A hook like this is primarily useful when we need to react to a change in properties to inform a message being added/removed, which is an ugly/difficult thing to do in React.  In a class component, we would use the `componentDidUpdate` lifecycle method and compare the current props with the previous props to understand if something changed, and then conditionally create an alert based on it.  In either case, it's annoying.  Luckily we don't have need of that here. :)

src/courseware/course/Course.jsx
```diff
@@ -5,6 +5,8 @@ import { getConfig } from '@edx/frontend-platform';
```
davidjoy: All the changes in this file can go away too, if we drop the above custom alerts!

src/instructor-toolbar/InstructorToolbar.jsx
```diff
@@ -51,23 +51,21 @@ export default function InstructorToolbar(props) {
   return (
     <div className="bg-primary text-light">
       <div className="container-fluid py-3 d-md-flex justify-content-end align-items-center">
-        <div className="flex-grow-1">
+        <div className="flex-grow-1 d-md-flex mx-1 my-1">
```
davidjoy: FWIW - I noticed when we were demoing it that the input box seemed to be aligned a bit higher than the other elements, probably because it's height is different.  Since these are all in a flexbox, you should be able to add `align-items-center` as a class here and it _should_ improve it.  I think.

src/instructor-toolbar/masquerade-widget/MasqueradeUserNameInput.jsx
```diff
@@ -0,0 +1,56 @@
+import React, {
+  Component,
+} from 'react';
+import PropTypes from 'prop-types';
+import { Input } from '@edx/paragon';
+
+class MasqueradeUserNameInput extends Component {
+  constructor(props) {
+    super(props);
+    this.onError = this.props.onError;
```
davidjoy: Saving this as an instance variable is a bit non-standard, fwiw.  I think in general you'd just call it from `this.props`.

Another thing people do if they want to reuse something in props a lot is to destructure it in the function they're in, i.e.:

```
const { onError } = this.props; // at the top of onSubmit, perhaps
```

src/instructor-toolbar/masquerade-widget/MasqueradeUserNameInput.jsx
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  onKeyPress(event) {
```
davidjoy, line 15: This is a naming nit, but in terms of method/prop naming scheme... my general preference and that shown in a lot of the react documentation is for handler methods to be named with a `handle` prefix, i.e., `handleKeyPress`.  

This is less documented in the community and feels a bit more subjective, but when passing a handler down to a sub component (like `onError` above), I usually name those something like `errorHandler` since it's a reference to the handler.  It also lets me know at a glance that it's coming from the outside of the component, as opposed to one defined in it.

And finally, doing the above makes it so that the "on" prefix only really gets used in built-in HTML mouse/keyboard attributes on the JSX itself.  

I kinda like the division of having those three separate schemes.  Makes it easy to tell what's coming from where.

src/instructor-toolbar/masquerade-widget/MasqueradeUserNameInput.jsx
```diff
@@ -0,0 +1,56 @@
...
+    if (event.key === 'Enter') {
+      return this.onSubmit(event);
+    }
+    return true;
+  }
+
+  onSubmit(event) {
```
davidjoy, line 22: Separate from the above, if you declare these handler methods like this:

```
onSubmit = (event) => {
}
```

You can avoid any potential binding issues without having to go put a `this.onSubmit = this.onSubmit.bind(this);` in the constructor.

It's only a real _problem_ if you're passing them to child components, but I try to either do the constructor binding or declare them as arrow functions to avoid it coming back to bite me later when it starts getting passed around as the code gets more complex. :)

src/instructor-toolbar/masquerade-widget/MasqueradeUserNameInput.jsx
davidjoy: I haven't thought through the ownership implications, but if the parent component handles the response - rather than this input component - then we don't have to worry about doing any of the proxying.  Arguably this component could just be "dumber" and let the parent be "smarter" w/r/t handling requests and such.

It's also mildly interesting that we're waiting for a promise to resolve on one of our props... I don't think I've done that before.  I don't see any reason why it wouldn't work, technically, fwiw. :)

src/instructor-toolbar/masquerade-widget/MasqueradeUserNameInput.jsx
davidjoy: Yeah, prolly. 

```
import messages from './messages.js';

....

this.props.intl.formatMessage(messages['masquerade.input.placeholder'])

...

MasqueradeUserNameInput.propTypes = {
  ...
  intl: intlShape,
}

export default injectIntl(MasqueradeUserNameInput);
```

Or thereabouts. :)

src/instructor-toolbar/masquerade-widget/MasqueradeWidget.jsx
davidjoy: Wonder if we should log it externally if this happens? 🤔

src/instructor-toolbar/masquerade-widget/MasqueradeWidget.jsx
davidjoy: Agreed!  If we add UserMessagesContext as a `contextType` on this component, we can then do `this.context.add()` here and give it a type of error, our message (or a code), etc.

src/instructor-toolbar/masquerade-widget/data/api.js
davidjoy: Yeah, this looks right now. :)

src/instructor-toolbar/InstructorToolbar.jsx
```diff
@@ -51,23 +51,21 @@ export default function InstructorToolbar(props) {
   return (
     <div className="bg-primary text-light">
       <div className="container-fluid py-3 d-md-flex justify-content-end align-items-center">
-        <div className="flex-grow-1">
+        <div className="align-items-center flex-grow-1 d-md-flex mx-1 my-1">
```
stvstnfrd, line 54: @davidjoy `align-items-center` added.

src/instructor-toolbar/masquerade-widget/MasqueradeUserNameInput.jsx
stvstnfrd: How "non-standard" are we talking here?! ;)
mostly jk, but curious...

Hrmm, my thinking was that I wanted the caller(s) to be implementation agnostic;
this allows an `onError` method to be defined directly on this class itself, if it wanted to handle the error with_out_ passing it back up (eg: `console.log` and swallow the error).

so we can always just call `this.onError` and never worry about where it's defined.

It's the architectural equivalent of saying "pretty please use my handler, but I respect that you could use your own";
here, we're playing along ;)

src/instructor-toolbar/masquerade-widget/MasqueradeWidget.jsx
stvstnfrd: Done.

src/instructor-toolbar/masquerade-widget/MasqueradeWidget.jsx
stvstnfrd: Done.

src/instructor-toolbar/masquerade-widget/MasqueradeUserNameInput.jsx
stvstnfrd: That is good to know; I appreciate consistency.

> And finally, doing the above makes it so that the "on" prefix only really gets used in built-in HTML mouse/keyboard attributes on the JSX itself.

Just to clarify that unless I'm misunderstanding, this here _is_ a built-in HTML attribute;
we're binding an `onKeyPress` event to the input element.

src/instructor-toolbar/masquerade-widget/MasqueradeUserNameInput.jsx
stvstnfrd: Re: above: I consider this to also be a native(ish) DOM event handler, as it takes a DOM `event` object as its `argument`. If we were writing semantic HTML (don't get me started today ;) ), it'd be a proper `FORM.onSubmit` :)

stvstnfrd (commented):
@davidjoy i18n and other feedback addressed.

The one thing I haven't got working yet is the dropdown hiding.
Should I just make this a native dropdown/select until the bootrap/paragon update lands?

Bummer, I liked this UI!

src/instructor-toolbar/masquerade-widget/MasqueradeUserNameInput.jsx
davidjoy: I meant that in the sense that the JSX would read:

```onKeyPress={handleKeyPress}```

File: src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx (hunk 2 above)
Comment (davidjoy) [reply src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx:49]: I think this fits your use case?  I'd try using https://lodash.com/docs/4.17.15#isEqual or something like it.

Comment (davidjoy) [reply src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx:49]: You could first filter the objects by the properties you want, maybe?  

`reduce` is confusing at the best of times, IMHO. ;)

**Review by davidjoy — APPROVED** [review]

File: src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx (hunk 2 above)
Comment (stvstnfrd) [reply src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx:49]: I took a look and yeah, between needing to filter the props anyway, plus the library calls, I think we'd have a final result that's at least as long/complicated as this one. And since this is natively supported _AND_ the function does one thing and is (I think!) well-named, I'm going to leave this alone for now.

File: src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx [hunk 37]
```diff
@@ -4,40 +4,47 @@ import React, {
...
-
 class MasqueradeWidgetOption extends Component {
-  handleClick() {
+  onClick(event) {
+    // TODO: Remove this hack when we upgrade to Paragon's Dropdown
+    event.target.parentNode.parentNode.click();
```
stvstnfrd, line 16: @davidjoy I'm not particularly proud of this, but it did save me from converting the component ;)
and since this functionality has been added to the newest Paragon release, we don't need to worry about swapping the UI later when we upgrade, assuming we can live with this event-crawling ;)

Let me know if you're totally against this, want more documentation, and/or can live w/ this until we upgrade Paragon :)

stvstnfrd (commented):
@davidjoy Let me know if you can live w/ my "close the dropdown" hack.

I think I'd like to punt on the naming nits for now, in part to avoid last-minute churn that could trigger breakage, but also to spare both of us from a naming conversation now :D 

I'm going to take the "out" that this isn't a full standard (yet?) in either the React community or this codebase, but I'm a big fan of predictable naming conventions, so happy to talk about adopting some convention(s) moving forward.

src/instructor-toolbar/masquerade-widget/MasqueradeWidgetOption.jsx
davidjoy: Shipppppit!

davidjoy (approved):

PR #110, opened 2020-07-16
fix: Use activeUnitIndex instead of position, and remove the latter

davidjoy:
We were inconsistently using “position” - a 1-indexed value - in JS arrays which are 0-indexed.  We had an existing, normalized property called “activeUnitIndex” which we now use everywhere instead.  The value is modified back to 1-indexed before being returned to the server.

### DIFF
--- a/src/courseware/CoursewareContainer.jsx
+++ b/src/courseware/CoursewareContainer.jsx
@@ -125,10 +125,8 @@ function useContentRedirect(courseStatus, sequenceStatus) {
 
   useEffect(() => {
     if (sequenceStatus === 'loaded' && sequenceId && !unitId) {
-      // The position may be null, in which case we'll just assume 0.
       if (sequence.unitIds !== undefined && sequence.unitIds.length > 0) {
-        const unitIndex = sequence.position || 0;
-        const nextUnitId = sequence.unitIds[unitIndex];
+        const nextUnitId = sequence.unitIds[sequence.activeUnitIndex];
         // This is a replace because we don't want this change saved in the browser's history.
         history.replace(`/course/${courseId}/${sequence.id}/${nextUnitId}`);
       }
--- a/src/courseware/data/api.js
+++ b/src/courseware/data/api.js
@@ -151,7 +151,6 @@ function normalizeSequenceMetadata(sequence) {
       title: sequence.display_name,
       gatedContent: camelCaseObject(sequence.gated_content),
       isTimeLimited: sequence.is_time_limited,
-      position: sequence.position || 1,
       // Position comes back from the server 1-indexed. Adjust here.
       activeUnitIndex: sequence.position ? sequence.position - 1 : 0,
       saveUnitPosition: sequence.save_position,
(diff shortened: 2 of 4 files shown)


src/courseware/CoursewareContainer.jsx
```diff
@@ -125,10 +125,8 @@ function useContentRedirect(courseStatus, sequenceStatus) {
 
   useEffect(() => {
     if (sequenceStatus === 'loaded' && sequenceId && !unitId) {
-      // The position may be null, in which case we'll just assume 0.
       if (sequence.unitIds !== undefined && sequence.unitIds.length > 0) {
-        const unitIndex = sequence.position || 0;
-        const nextUnitId = sequence.unitIds[unitIndex];
+        const nextUnitId = sequence.unitIds[sequence.activeUnitIndex];
```
davidjoy, line 129: activeUnitIndex will always be a valid index, based on the normalization in api.js

src/courseware/data/api.js
```diff
@@ -200,13 +199,13 @@ export async function getBlockCompletion(courseId, sequenceId, usageKey) {
...
 
-export async function postSequencePosition(courseId, sequenceId, position) {
+export async function postSequencePosition(courseId, sequenceId, activeUnitIndex) {
   // Post data sent to this endpoint must be url encoded
   // TODO: Remove the need for this to be the case.
   // TODO: Ensure this usage of URLSearchParams is working in Internet Explorer
   const urlEncoded = new URLSearchParams();
   // Position is 1-indexed on the server and 0-indexed in this app. Adjust here.
-  urlEncoded.append('position', position + 1);
+  urlEncoded.append('position', activeUnitIndex + 1);
```
davidjoy, line 208: This is where we coerce the value back to being 1-indexed for the server.  Everywhere in between is consistently 0-indexed now.

davestgermain (approved):

src/courseware/data/api.js
```diff
@@ -151,7 +151,6 @@ function normalizeSequenceMetadata(sequence) {
       title: sequence.display_name,
       gatedContent: camelCaseObject(sequence.gated_content),
       isTimeLimited: sequence.is_time_limited,
-      position: sequence.position || 1,
```
davidjoy, line 154: We no longer pass this position through - it can only cause problems... like the one we're fixing.

src/courseware/data/api.js
```diff
@@ -151,7 +151,6 @@ function normalizeSequenceMetadata(sequence) {
...
       // Position comes back from the server 1-indexed. Adjust here.
       activeUnitIndex: sequence.position ? sequence.position - 1 : 0,
```
davidjoy, line 155: Note that activeUnitIndex is set to 0 here if we didn't have a truthy sequence.position.

src/courseware/data/thunks.js
```diff
@@ -131,27 +131,27 @@ export function checkBlockCompletion(courseId, sequenceId, unitId) {
...
     dispatch(updateModel({
       modelType: 'sequences',
       model: {
         id: sequenceId,
-        position,
+        activeUnitIndex,
       },
     }));
     try {
-      await postSequencePosition(courseId, sequenceId, position);
```
davidjoy, line 147: We pass the 0-indexed value into this API function - it changes it to 1-indexed inside.

src/courseware/data/thunks.js
```diff
@@ -131,27 +131,27 @@ export function checkBlockCompletion(courseId, sequenceId, unitId) {
...
+      await postSequencePosition(courseId, sequenceId, activeUnitIndex);
```
davidjoy, line 147: We pass the 0-indexed value into this API function - it changes it to 1-indexed inside.

src/courseware/data/thunks.js
```diff
@@ -131,27 +131,27 @@ export function checkBlockCompletion(courseId, sequenceId, unitId) {
...
     const { models } = getState();
-    const initialPosition = models.sequences[sequenceId].position;
+    const initialActiveUnitIndex = models.sequences[sequenceId].activeUnitIndex;
     // Optimistically update the position.
     dispatch(updateModel({
       modelType: 'sequences',
       model: {
         id: sequenceId,
-        position,
+        activeUnitIndex,
```
davidjoy, line 143: All of the usage of the index in this file is supposed to be 0-based, so we can consistently use activeUnitIndex.

PR #111, opened 2020-07-16
AA-230: Dates Tab styling

ciduarte:
Addressed a11y violations as well as general styling for parity with the proposed [mockup](https://app.zeplin.io/project/5e1cdb1ea87828a8c7934b71/screen/5eb19a5fa497f1b249faf92e).

Changes include:
- Dates banner responsiveness (see image 2)
- Size of date dots/left-hand border
- Spacing of overall timeline
- (a11y) Removed use of `<h6>` tags, incorporated `<h1>` tag per Jeff Witt's direction

Image 1: Dates Tab
mfe_datesTab

Image 2: Dates banner
reset_dates_banner_responsive

### DIFF
--- a/src/course-home/dates-tab/DatesTab.jsx
+++ b/src/course-home/dates-tab/DatesTab.jsx
@@ -8,9 +8,9 @@ import DatesBannerContainer from '../dates-banner/DatesBannerContainer';
 function DatesTab({ intl }) {
   return (
     <>
-      <h2 className="mb-4">
+      <div role="heading" aria-level="1" className="h4 my-3">
         {intl.formatMessage(messages.title)}
-      </h2>
+      </div>
       <DatesBannerContainer model="dates" />
       <Timeline />
     </>
--- a/src/course-home/dates-tab/Day.scss
+++ b/src/course-home/dates-tab/Day.scss
@@ -1,7 +1,6 @@
 $dot-radius: 0.3rem;
 $dot-size: $dot-radius * 2;
-$top-offset: 0.45rem;
-$left-offset: 0.225rem + $dot-radius;
+$offset: $dot-radius * 1.5;
 
 .dates-day {
   position: relative;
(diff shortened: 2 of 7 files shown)


src/course-home/dates-tab/DatesTab.jsx
```diff
@@ -8,9 +8,9 @@ import DatesBannerContainer from '../dates-banner/DatesBannerContainer';
 function DatesTab({ intl }) {
   return (
     <>
-      <h2 className="mb-4">
+      <h4 aria-level="1" className="h4 my-3">
```
ciduarte: This looks a little funky. But it's for good cause.

What we _wanted_ here was `<h1 className="h4 my-3">Title</h1>`. For clarity, we want this to be an `h1` level element for a11y, but we want the styling to be an `h4`. Unfortunately, using the `h4` className would not override the `h1` tag's styling because of the way Paragon is currently configured. The fedx recommendation was to use `aria-level` to override the heading tag. The design team will be working on uprooting Paragon in the next few months, so this will change in the future.

Another thing to note: Around edX's services, some headers have logos wrapped in `<h1>`, and we want to move away from doing that in the future (see picture). Jeff Witt's guidance is to tag the first heading in main as an `<h1>`.

coursehome_headers

xnick421 (approved):

src/course-home/dates-tab/Day.scss
```diff
@@ -1,7 +1,7 @@
 $dot-radius: 0.3rem;
 $dot-size: $dot-radius * 2;
 $top-offset: 0.45rem;
-$left-offset: 0.225rem + $dot-radius;
+$left-offset: $dot-radius + 0.1rem;
```
Dillon-Dumesnil: It doesn't look like this is being used anywhere.

src/course-home/dates-tab/Day.scss
```diff
@@ -10,7 +10,7 @@ $left-offset: 0.225rem + $dot-radius;
 .dates-line-top {
   display: inline-block;
   position: absolute;
-  left: $left-offset;
+  left: $dot-radius * 1.5;
```
Dillon-Dumesnil: It feels like we are never using the base `$dot-radius` and are always adding a modifier to it. Should we maybe  just alter the default? I don't have a strong opinion either way, just curious.

Dillon-Dumesnil (approved):
Small comment in the scss, but yeah, looks good. I don't know

PR #115, opened 2020-07-17
Change CoursewareContainer into a class component.

davidjoy:
I find it much more legible this way.

Some thoughts… as part of refactoring it, I made some of the redux selectors more formal, and made use of `reselect` more thoroughly.  this resulted in a reduction in re-renders from 16 to 12 on your average page load.  It’s also a bit more verbose, accounting for some of the increased line count.

I hadn’t tried it before, but found the `memoize` method of comparing previous props/state to current props/state to be very, very nice.  Much easier than manually comparing props, and much clearer to me than using react hooks’ dependency arrays.

The lack of dependency arrays feels really freeing in general to me.  They’ve been such a source of hard-to-track-down bugs, and the hooks linter does not always suggest the right solution for what belongs in and out of the array.

Function names are nice.  We had a ton of custom hooks in there so that we could put names to otherwise anonymous bits of functionality.

Also note: this component has a test suite.  It passed without any changes. 🥳 :

### DIFF
--- a/src/courseware/CoursewareContainer.jsx
+++ b/src/courseware/CoursewareContainer.jsx
@@ -1,88 +1,144 @@
-import React, { useEffect, useCallback } from 'react';
+import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { useSelector, useDispatch } from 'react-redux';
+import { connect } from 'react-redux';
 import { history } from '@edx/frontend-platform';
 import { getLocale } from '@edx/frontend-platform/i18n';
-import { useRouteMatch, Redirect } from 'react-router';
-
-import { useModel } from '../generic/model-store';
-import { TabPage } from '../tab-page';
+import { Redirect } from 'react-router';
+import memoize from 'lodash.memoize';
+import { createSelector } from '@reduxjs/toolkit';
 
 import {
+  checkBlockCompletion,
   fetchCourse,
   fetchSequence,
-  checkBlockCompletion,
-  saveSequencePosition,
   getResumeBlock,
-  sequenceIdsSelector,
-  firstSequenceIdSelector,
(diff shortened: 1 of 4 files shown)


src/courseware/CoursewareContainer.jsx
```diff
@@ -1,88 +1,147 @@
...
+  checkBlockCompletion,
+  saveSequencePosition,
+} from './data/thunks';
 
-function usePreviousSequence(sequenceId) {
-  const sequenceIds = useSelector(sequenceIdsSelector);
-  const sequences = useSelector(state => state.models.sequences);
-  if (!sequenceId || sequenceIds.length === 0) {
-    return null;
+const checkExamRedirect = memoize((sequenceStatus, sequence) => {
```
davidjoy, line 21: Using lodash.memoize, we can greatly simplify the process of comparing previous props with current props in `componentDidUpdate`.  By memoizing, we can let the memoization function handle keeping track of the previous props and `===` comparing them with the current iteration.

stvstnfrd (approved):
I really can only echo what you've already said;
which I'll do for dramatic effect ;)

Thanks for taking this on!

> I find it much more legible this way.

:+1: 

> this resulted in a reduction in re-renders from 16 to 12 on your average page load.

Nice!

> and much clearer to me than using react hooks’ dependency arrays.

This, 1000x this!

> The lack of dependency arrays feels really freeing in general to me. They’ve been such a source of hard-to-track-down bugs, and the hooks linter does not always suggest the right solution for what belongs in and out of the array.

:+1: 

> Function names are nice.

:)

> Also note: this component has a test suite. It passed without any changes. partying_face :

Listen up folks, because _this_ is how you refactor code!
Yay, test suites!

PR #118, opened 2020-07-20
AA-218: Course Tool Analytics

ciduarte:
Tracking analytics for onClick events in the Course Tool widget.
Extra: Fixed intl error in the Course Dates widget.

### DIFF
--- a/src/course-home/outline-tab/widgets/CourseDates.jsx
+++ b/src/course-home/outline-tab/widgets/CourseDates.jsx
@@ -14,7 +14,7 @@ function CourseDates({ courseId, intl }) {
 
   return (
     <section className="mb-3">
-      <h4>{intl.formatMessage(messages.tools)}</h4>
+      <h4>{intl.formatMessage(messages.dates)}</h4>
       {datesWidget.courseDateBlocks.map((courseDateBlock) => (
         <DateSummary
           key={courseDateBlock.title + courseDateBlock.date}
--- a/src/course-home/outline-tab/widgets/CourseTools.jsx
+++ b/src/course-home/outline-tab/widgets/CourseTools.jsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import { sendTrackEvent } from '@edx/frontend-platform/analytics';
+import { getAuthenticatedUser } from '@edx/frontend-platform/auth';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import {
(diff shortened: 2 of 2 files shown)


src/course-home/outline-tab/widgets/CourseTools.jsx
```diff
@@ -40,7 +50,7 @@ function CourseTools({ courseId, intl }) {
       <h4>{intl.formatMessage(messages.tools)}</h4>
       {courseTools.map((courseTool) => (
         <div key={courseTool.analyticsId}>
-          <a data-analytics-id={courseTool.analyticsId} href={courseTool.url}>
+          <a href={courseTool.url} onClick={logClick.bind(null, courseTool.analyticsId)}>
```
ciduarte: Without this `.bind(null, ...)` here, `logClick()` gets executed when the page is rendered rather than onClick. Curious if there's a more correct way to do this, or if this approach is fine.

src/course-home/outline-tab/widgets/CourseTools.jsx
```diff
@@ -16,6 +18,14 @@ function CourseTools({ courseId, intl }) {
     courseTools,
   } = useModel('outline', courseId);
 
+  const logClick = (analyticsId) => {
+    const { administrator } = getAuthenticatedUser();
+    sendTrackEvent(analyticsId, {
+      course_id: courseId,
+      is_staff: administrator,
+    });
+  };
```
Dillon-Dumesnil: This feels like it should live outside of this file and in a utils or analytics/utils file rather than in here. It just needs to grab the course id, but then we can use this logClick everywhere we want to track clicks (which I imagine is more than just course tools)

ciduarte: @davidjoy This code has changed very slightly, but wanted to highlight this comment to hear if you have any insight on the future of analytics within frontend-app-learning/think this would be useful?

I personally don't think this is necessary. Components would still need to build their own payloads in a helper method similar to this one, and a general use `logClick(eventName, payload)` would practically just become a wrapper for `sendTrackEvent(eventName, payload)`. 🤷

src/course-home/outline-tab/widgets/CourseDates.jsx
```diff
@@ -14,7 +14,7 @@ function CourseDates({ courseId, intl }) {
 
   return (
     <section className="mb-3">
-      <h4>{intl.formatMessage(messages.tools)}</h4>
+      <h4>{intl.formatMessage(messages.dates)}</h4>
```
mikix, line 17: :bless:

mikix (approved):

PR #120, opened 2020-07-21
fix: Use reselect’s defaultMemoize instead of lodash.memoize

davidjoy:
Fixes: https://openedx.atlassian.net/browse/TNL-7387

Lodash memoize doesn’t examine all parameters when deciding to memoize, apparently, meaning it doesn’t re-call the function if any parameter except the first changes.

More here: https://dev.to/nioufe/you-should-not-use-lodash-for-memoization-3441

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
     "@reduxjs/toolkit": "^1.2.3",
     "classnames": "^2.2.6",
     "core-js": "^3.6.2",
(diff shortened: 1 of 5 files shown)


stvstnfrd (approved):

stvstnfrd:
Hrmm, tests are failing though...

PR #121, opened 2020-07-21
AA-123 welcome message

xnick421:


src/course-home/data/api.js
```diff
@@ -72,16 +72,23 @@ export async function getOutlineTabData(courseId) {
...
 }
 
 export async function postCourseDeadlines(courseId) {
   const url = new URL(`${getConfig().LMS_BASE_URL}/api/course_experience/v1/reset_course_deadlines`);
   await getAuthenticatedHttpClient().post(url.href, { course_key: courseId });
 }
+
+export async function postDismissWelcomeMessage(courseId) {
+  const url = new URL(`${getConfig().LMS_BASE_URL}/api/course_home/v1/dismiss_welcome_message`);
+  await getAuthenticatedHttpClient().post(url.href, { course_id: courseId });
```
Dillon-Dumesnil, line 93: Out of curiosity, what's the difference between these lines and 
```
await getAuthenticatedHttpClient().post(`${getConfig().LMS_BASE_URL}/api/course_home/v1/dismiss_welcome_message`, { course_id: courseId });
```
So I guess really, the question is, why make the URL object?

src/course-home/outline-tab/widgets/WelcomeMessage.jsx
```diff
@@ -0,0 +1,68 @@
...
+  const [display, setDisplay] = useState(true);
+
+  const shortWelcomeMessageHtml = welcomeMessageHtml.length > 200 && `${welcomeMessageHtml.substring(0, 199)}...`;
+  const [showShortMessage, setShowShortMessage] = useState(!!shortWelcomeMessageHtml);
+  const dispatch = useDispatch();
+
+  return (
+    display
+    && (
+      <Alert
```
Dillon-Dumesnil: Have the `&&` on the first line.
```
display && (
  <Alert...
```
And doing this on lines 44-46 too

src/course-home/outline-tab/widgets/WelcomeMessage.jsx
```diff
@@ -0,0 +1,68 @@
...
+        type="info"
+        dismissible
+        onDismiss={() => {
+          setDisplay(false);
+          dispatch(dismissWelcomeMessage(courseId));
+        }}
+      >
+        <LmsHtmlFragment
+          html={showShortMessage ? shortWelcomeMessageHtml : welcomeMessageHtml}
+          title={intl.formatMessage(messages.welcomeMessage)}
```
Dillon-Dumesnil: We don't usually have a title for updates. I'm not exactly sure how this renders, but we don't want to show any title with this update.
Speaking on that, we definitely do not want to show a message that always says `Welcome Message` because this is not always the welcome message. It can just be the most recent course update.

Dillon-Dumesnil (commented):
Pretty solid, just a comment on some of the messaging we have here.

src/course-home/outline-tab/widgets/WelcomeMessage.jsx
xnick421: the title is just used for the iframe `title` attribute.  its not actually rendered.

src/course-home/outline-tab/widgets/WelcomeMessage.jsx
```diff
@@ -0,0 +1,68 @@
...
+        {
+          shortWelcomeMessageHtml
+          && (
+            <div className="d-flex justify-content-center">
+              <button
+                type="button"
+                className="btn rounded align-self-center border border-primary bg-white mt-3 mt-md-0 font-weight-bold"
+                onClick={() => setShowShortMessage(!showShortMessage)}
+              >
+                {showShortMessage ? 'Show More' : 'Show Less'}
```
ciduarte: Do we need intl on `'Show More'` and `'Show Less'`?

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -55,6 +56,7 @@ function OutlineTab({ intl }) {
           clientLogistrationAlert: LogistrationAlert,
         }}
       />
+      <WelcomeMessage courseId={courseId} />
```
ciduarte: I think this should display below the title and above the sections. Maybe after line 65, right before the sections get mapped?

ciduarte (commented):
Just two small comments. We should add a screenshot to this PR.

xnick421:
Screen Shot 2020-07-22 at 2 01 57 PM
Screen Shot 2020-07-22 at 2 01 21 PM
Screen Shot 2020-07-22 at 2 01 28 PM

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -59,6 +60,7 @@ function OutlineTab({ intl }) {
         <h2>{title}</h2>
         <Button className="btn-primary" type="button">{intl.formatMessage(messages.resume)}</Button>
       </div>
+      <WelcomeMessage courseId={courseId} />
```
ciduarte: I still think this should be under line 65 within that `col-8` so that it doesn't span over the right-hand side widgets.

src/course-home/outline-tab/widgets/WelcomeMessage.jsx
```diff
@@ -0,0 +1,68 @@
...
+        }}
+      >
+        <LmsHtmlFragment
+          html={showShortMessage ? shortWelcomeMessageHtml : welcomeMessageHtml}
+          title={intl.formatMessage(messages.welcomeMessage)}
+        />
+        {
+          shortWelcomeMessageHtml
+          && (
+            <div className="d-flex justify-content-center">
```
ciduarte: This should be justified to the end, not centered

src/course-home/outline-tab/widgets/WelcomeMessage.jsx
```diff
@@ -0,0 +1,68 @@
...
+          dispatch(dismissWelcomeMessage(courseId));
+        }}
+      >
+        <LmsHtmlFragment
+          html={showShortMessage ? shortWelcomeMessageHtml : welcomeMessageHtml}
+          title={intl.formatMessage(messages.welcomeMessage)}
+        />
+        {
+          shortWelcomeMessageHtml
+          && (
```
ciduarte: nit: I would make this one line, `{shortWelcomeMessageHtml && (`

ciduarte (commented):
I feel like we should base this styling on the Welcome Banner in the old experience (see image). Found [this comment in Jira](https://openedx.atlassian.net/browse/AA-46?focusedCommentId=444586).

I know we allow users to input their own HTML for updates in Studio, and that's probably causing the white text background on this blue info Alert (along with the circular "i" icon from the info Alert). Should we override the HTML provided to remove the white background? Should we override the Alert styling and make it white to match the old experience? (I would ask Eugene/Hilary)

Screenshot from 2020-04-06 14-23-35

src/course-home/outline-tab/widgets/WelcomeMessage.jsx
ciduarte: This makes me wonder if our language around this component should even be "welcome message"? If in Studio it's referred to as a "Course Update", and this component is being used to deliver... course updates... (which just happen to include a welcome message)?

src/course-home/outline-tab/widgets/WelcomeMessage.jsx
```diff
@@ -0,0 +1,68 @@
...
+    display && (
+      <Alert
+        type="welcome"
+        dismissible
+        onDismiss={() => {
+          setDisplay(false);
+          dispatch(dismissWelcomeMessage(courseId));
+        }}
+      >
+        <div className="mt-3">
```
ciduarte: One last thing!! I would switch this to `my-3`...

src/course-home/outline-tab/widgets/WelcomeMessage.jsx
```diff
@@ -0,0 +1,68 @@
...
+            html={showShortMessage ? shortWelcomeMessageHtml : welcomeMessageHtml}
+            title={intl.formatMessage(messages.welcomeMessage)}
+          />
+        </div>
+        {
+          shortWelcomeMessageHtml && (
+            <div className="d-flex justify-content-end">
+              <button
+                type="button"
+                className="btn rounded align-self-center border border-primary bg-white font-weight-bold my-3"
```
ciduarte: ... and switch this to `mb-3`.

ciduarte (approved):
Thanks for working through that styling!! Highly appreciated 😄 

Just one last comment on something I missed because I didn't see the version w/o the button in the Alert.

PR #123, opened 2020-07-24
Re-enable the celebration modal

mikix:
It got accidentally disabled during a refactor.

Also, try a little harder to make sure it doesn't re-appear during the same browsing session.

### DIFF
--- a/src/courseware/CoursewareContainer.jsx
+++ b/src/courseware/CoursewareContainer.jsx
@@ -17,6 +17,7 @@ import {
 import { TabPage } from '../tab-page';
 
 import Course from './course';
+import { handleNextSectionCelebration } from './course/celebration';
 
 const checkExamRedirect = memoize((sequenceStatus, sequence) => {
   if (sequenceStatus === 'loaded') {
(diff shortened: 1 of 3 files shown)


src/courseware/CoursewareContainer.jsx
```diff
@@ -125,15 +126,28 @@ class CoursewareContainer extends Component {
...
+      nextSequence,
+      sequence,
+      sequenceId,
+    } = this.props;
+
     if (nextSequence !== null) {
+      let nextUnitId = null;
       if (nextSequence.unitIds.length > 0) {
-        const nextUnitId = nextSequence.unitIds[0];
+        [nextUnitId] = nextSequence.unitIds;
```
Dillon-Dumesnil, line 140: I do not understand this, but okay!

Dillon-Dumesnil (approved):
Sure

src/courseware/CoursewareContainer.jsx
mikix: This is just a linting thing. You know how you can decompose a dictionary like `const { courseId } = props`?

Well this is how you decompose a list. So like `const [firstItem, seconItem] = arrayOfItems`. :shrug:

PR #126, opened 2020-07-27
AA-124: Refactor enrollment alerts

mikix:
- Place them only on the Outline page
- Support a few cases where enrollment isn't actually allowed

This moves the LogistrationAlert around to the Outline page too, but that's just theoretical, because the MFE always requires the user to be logged in right now.

Related to https://github.com/edx/edx-platform/pull/24620


src/alerts/enrollment-alert/EnrollmentAlert.jsx
```diff
@@ -5,24 +5,44 @@ import { Button } from '@edx/paragon';
...
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faSpinner } from '@fortawesome/free-solid-svg-icons';
-import { Alert } from '../../generic/user-messages';
+import { Alert, ALERT_TYPES } from '../../generic/user-messages';
 
 import messages from './messages';
 import { useEnrollClickHandler } from './hooks';
 
-function EnrollmentAlert({ intl, courseId }) {
+function EnrollmentAlert({ intl, payload }) {
```
Dillon-Dumesnil, line 13: `props`?

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -42,6 +39,9 @@ function OutlineTab({ intl }) {
     },
   } = useModel('outline', courseId);
 
+  const clientEnrollmentAlert = useEnrollmentAlert(courseId);
+  useLogistrationAlert();
```
Dillon-Dumesnil, line 43: How does this play with the LogistrationAlert we pass in on line 55 as a customAlert to the outlineTab?

Dillon-Dumesnil (approved):
Couple small questions in the PR, but otherwise looks good to me. 
Do we know how this would perform is a learner somehow directly accessed the courseware?

src/alerts/enrollment-alert/EnrollmentAlert.jsx
mikix: `payload` is the prop that the AlertsList passes to each alert component, from the alert's hook setting up the alert.

src/course-home/outline-tab/OutlineTab.jsx
mikix: The way alerts are set up in the MFE is that there is a hook that defines whether it's visible (to hide/show it) and various props for the alert, including a `code` for the alert.

Then when the AlertsList is looking to create the alert, it sees if any custom alert components have been defined for the `code` of the alert (otherwise it defaults to the default `Alert` component).

So the `useLogistrationAlert` defines an alert looking for the custom code. Then later, the custom code gets set to the LogistrationAlert component.

That felt a little too "implicit context" for me, so I started returning the custom component in my `useEnrollmentAlert`, but I didn't change the logistration alert.

mikix:
>  Do we know how this would perform is a learner somehow directly accessed the courseware?

Any other tab that tries to get accessed directly is redirected to the "about" page for the course, which on prod is the outline page (and on devstack is an actual `/about` page that looks like the marketing site a bit...)

PR #127, opened 2020-07-27
AA-214 social buttons

xnick421:
- social sharing buttons for linkedin and email

### DIFF
--- a/src/courseware/course/celebration/SocialIcons.jsx
+++ b/src/courseware/course/celebration/SocialIcons.jsx
@@ -1,8 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {
+  EmailIcon,
+  EmailShareButton,
   FacebookIcon,
   FacebookShareButton,
+  LinkedinIcon,
+  LinkedinShareButton,
   TwitterIcon,
   TwitterShareButton,
 } from 'react-share';
@@ -39,9 +43,17 @@ function SocialIcons({ courseId, intl }) {
 
   return (
     <div className="social-icons">
+      <LinkedinShareButton
+        beforeOnClick={() => logClick('linkedin')}
+        url={marketingUrl}
+      >
+        <LinkedinIcon round size={32} />
+        <span className="sr-only">{intl.formatMessage(messages.shareService, { service: 'LinkedIn' })}</span>
+      </LinkedinShareButton>
       { twitterAccount && (
         <TwitterShareButton
           beforeOnClick={() => logClick('twitter')}
+          className="ml-2"
           hashtags={['mooc']}
           title={intl.formatMessage(messages.social, { platform: `@${twitterAccount}`, title })}
           url={marketingUrl}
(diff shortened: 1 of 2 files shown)


xnick421:
So a couple notes here:

For linkedin, we are only able to pre populate a post with a url to our course.  Pre populating a message has been limited by linkedin to paid customers only, which doesn't suit our business needs.

As far as email goes, we originally wanted buttons for gmail and outlook specifically, but our library which we use for social sharing only has support for one email button which points to the user's default email client.  @Ahernkb confirmed this was an ok path to move forward with.

Additional question for @Ahernkb ...right now we have the body of emails being the same message that appears when sharing to facebook, or twitter.  However, the subject, I populated with "{course name} on edX".  Is this ok?  or do we have other language we want to populate the subject with?

src/courseware/course/celebration/SocialIcons.jsx
```diff
@@ -59,6 +71,16 @@ function SocialIcons({ courseId, intl }) {
         <FacebookIcon round size={32} />
         <span className="sr-only">{intl.formatMessage(messages.shareService, { service: 'Facebook' })}</span>
       </FacebookShareButton>
+      <EmailShareButton
+        beforeOnClick={() => logClick('email')}
+        body={intl.formatMessage(messages.emailBody, { platform: getConfig().SITE_NAME, title })}
+        className="ml-2"
+        subject={intl.formatMessage(messages.emailSubject, { platform: getConfig().SITE_NAME, title })}
+        url="www.google.com"
```
Dillon-Dumesnil: ?

Dillon-Dumesnil (approved):
Looks fine to me technically. Subject to Kaitlin's feedback.

And one small thing that I think should be removed.

src/courseware/course/celebration/SocialIcons.jsx
```diff
@@ -59,6 +71,16 @@ function SocialIcons({ courseId, intl }) {
...
+      >
+        <EmailIcon round size={32} />
+        <span className="sr-only">{intl.formatMessage(messages.shareService, { service: 'Email' })}</span>
```
mikix: This will format to `Share your progress on Email.` which sounds a little odd for English. I dunno if it's worth an extra string for the email case though. But if so, maybe just a version replacing `on` with `via` or `by` would work.

And probably email should be lowercase.

src/courseware/course/celebration/messages.js
```diff
@@ -13,6 +13,16 @@ const messages = defineMessages({
     id: 'learning.celebration.earned',
     defaultMessage: 'You earned it!',
   },
+  emailBody: {
+    id: 'learning.celebration.emailBody',
+    defaultMessage: 'I’m on my way to completing {title} online with {platform}. What are you spending your time learning?',
```
mikix: This is the same as `learning.celebration.social` -- why have a new id for this?

mikix (approved):

Ahernkb (approved):
Looks good @xnick421 ! Can we change the subject to: "I'm on my way to completing [Course Name} online with @edxonline!" and add "What are you spending your time learning?" to the body. Also want to confirm, are we include the celebratory image in the email as well?

src/courseware/course/celebration/SocialIcons.jsx
```diff
@@ -1,8 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {
+  EmailIcon,
+  EmailShareButton,
   FacebookIcon,
   FacebookShareButton,
+  LinkedinIcon,
+  LinkedinShareButton,
```
ciduarte, line 9: Super nit, but LinkedIn has a capital "i"

ciduarte (approved):
LGTM.

I know this was created before but can we get a screenshot of what this looks like with the additional buttons?

src/courseware/course/celebration/SocialIcons.jsx
xnick421: this comes from a third party library :(

daphneli-chen (approved):
I think it looks beautiful!

src/courseware/course/celebration/SocialIcons.jsx
ciduarte: 😞

PR #128, opened 2020-07-29
Fixes saving unit position and unit redirection bugs

davidjoy:
Fixes a few bugs in the courseware container:

- Position was not being saved because we weren’t reading “saveUnitPosition” correctly.
- We weren’t calling checkContentRedirect with the right arguments - it was using a non-existent unitId instead of the routeUnitId, meaning we would redirect to the active unit even if a unit was specified in the URL.

Adds tests in CoursewareContainer for various URL and data states.

Now explicitly tests:
- Exam redirects
- The resume block method when it has, and doesn’t have, a block to resume.
- The content redirect when a unit isn’t present on the URL (uses sequence.position)
- Loading a specific unit (not the first of a sequence!) by URL.

Updated some of the factories to be more flexible/allow multiple units.

### DIFF
--- a/src/courseware/CoursewareContainer.test.jsx
+++ b/src/courseware/CoursewareContainer.test.jsx
@@ -36,14 +36,13 @@ jest.mock(
 
 initializeMockApp();
 
-const axiosMock = new MockAdapter(getAuthenticatedHttpClient());
-
 describe('CoursewareContainer', () => {
   let store;
   let component;
+  let axiosMock;
 
   beforeEach(() => {
-    axiosMock.reset();
+    axiosMock = new MockAdapter(getAuthenticatedHttpClient());
 
     store = initializeStore();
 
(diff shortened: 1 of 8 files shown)


src/courseware/data/redux.test.js
```diff
@@ -66,9 +67,7 @@ describe('Data layer integration tests', () => {
           has_access: false,
         },
       });
-      const forbiddenCourseBlocks = Factory.build('courseBlocks', {
-        courseId: forbiddenCourseMetadata.id,
-      });
+      const { courseBlocks: forbiddenCourseBlocks } = buildSimpleCourseBlocks(forbiddenCourseMetadata.id, null);
```
stvstnfrd, line 70: Nit: Do we need to pass `null`? Or is that already the implied default behavior?

src/courseware/CoursewareContainer.test.jsx
```diff
@@ -76,62 +75,210 @@ describe('CoursewareContainer', () => {
...
+        );
+
+        // Re-call the mock setup now that sequenceMetadata is different.
+        setupMockRequests();
+        history.push(`/course/${courseId}/${sequenceBlock.id}/${unitBlocks[2].id}`);
+        render(component);
+
+        // This is an important line that ensures the spinner has been removed - and thus our main
+        // content has been loaded - prior to proceeding with our expectations.
+        await waitForElementToBeRemoved(screen.getByRole('status'));
```
stvstnfrd, line 277: :+1:

stvstnfrd: Just a thought, is it worth extracting this into a helper?
I know it's just a single line of _code_, but the longer _comment_ is useful and we seem to repeating this pattern a few times. I could imagine it being a bit easier to read if we were invoking, `ensureMainContentLoaded(screen)`.

Not a big deal if it's a wonky refactor though.

stvstnfrd (approved):
Looks good and love the tests! Just a question or two, but tentative +1

src/courseware/data/redux.test.js
davidjoy: Hm, I guess I could have just left the parameter out and it'd be `undefined`, but I figure it's better to obey the method signature.  It just needs to be falsy so that the default behavior happens, in this case.  ideally there'd be an `options` parameter here instead of this.  Once #97 merges, I think we can use options for this stuff.

src/courseware/CoursewareContainer.test.jsx
davidjoy: That's not a bad thought.

It's funny, I always find when I come into a test suite that has a lot of things extracted to helpers I find it very hard to find my way around, but when I'm _writing_ a test suite I want to do it... hard to know when to bother. :)

src/courseware/data/redux.test.js
stvstnfrd: I think the only reason it jumped out to me is that we're _not_ passing in the title in all invocations ;)

PR #131, opened 2020-07-29
Adding optimizely to header.

davidjoy:
This relies on a change to @edx/frontend-build that allows the OPTIMIZELY_PROJECT_ID to be passed through from configuration.

NOTE for edx.org: When this merges and deploys, Optimizely will immediately be enabled.  OPTIMIZELY_PROJECT_ID is in our common/prod_config.yml.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "reselect": "^4.0.0"
   },
   "devDependencies": {
-    "@edx/frontend-build": "^5.0.6",
+    "@edx/frontend-build": "^5.1.0",
     "@testing-library/dom": "^7.16.2",
     "@testing-library/jest-dom": "^5.10.1",
     "@testing-library/react": "^10.3.0",
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "reselect": "^4.0.0"
   },
   "devDependencies": {
-    "@edx/frontend-build": "^5.1.0",
+    "@edx/frontend-build": "^5.1.2",
     "@testing-library/dom": "^7.16.2",
     "@testing-library/jest-dom": "^5.10.1",
     "@testing-library/react": "^10.3.0",
--- a/public/index.html
+++ b/public/index.html
@@ -5,8 +5,8 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <link rel="shortcut icon" href="/favicon.ico" type="image/x-icon" />
-    <% if (htmlWebpackPlugin.options.optimizelyId) { %>
-      <script src="https://www.edx.org/optimizelyjs/<%= htmlWebpackPlugin.options.optimizelyId %>.js"></script>
+    <% if (htmlWebpackPlugin.options.OPTIMIZELY_PROJECT_ID) { %>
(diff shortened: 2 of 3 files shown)


public/index.html
```diff
@@ -1,10 +1,13 @@
...
   <head>
-      <title>Course | <%= process.env.SITE_NAME %></title>
-      <meta charset="utf-8">
-      <meta name="viewport" content="width=device-width, initial-scale=1.0">
-      <link rel="shortcut icon" href="/favicon.ico" type="image/x-icon" />
+    <title>Course | <%= process.env.SITE_NAME %></title>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="shortcut icon" href="/favicon.ico" type="image/x-icon" />
+    <% if (htmlWebpackPlugin.options.OPTIMIZELY_PROJECT_ID) { %>
```
stvstnfrd, line 8: Where does this `PROJECT_ID` come from?

public/index.html
```diff
@@ -1,10 +1,13 @@
...
+      <script src="https://www.edx.org/optimizelyjs/<%= htmlWebpackPlugin.options.OPTIMIZELY_PROJECT_ID %>.js"></script>
```
stvstnfrd, line 9: Pretty naive comment, but could/should we be using a library like this [1]?

- [1] https://docs.developers.optimizely.com/full-stack/docs/example-usage-react

public/index.html
davidjoy: I should add a comment.  It gets passed through by our frontend-build webpack configuration from the environment variables.

public/index.html
davidjoy: Oh, wow.  Cool.  Uh, never saw that before.  Who's naive now?! ;) 

Yes, I'll say we should probably be using that.  That said, I don't think we have expertise on how to use it yet, so it may be that for the moment we want to fall back on our old tricks...

stvstnfrd: Maybe we want to punt on this entirely then and leave this up to the team implementing the actual experiments?

Given that it:
- will _not_ be immediately useful
- but _will_ immediately impact performance
- the implementation will very likely change (given this SDK)
- and this change is fairly trivial

It may just make sense to leave this entire up to the implementation team. Thoughts?

package.json
```diff
@@ -60,7 +60,7 @@
     "reselect": "^4.0.0"
   },
   "devDependencies": {
-    "@edx/frontend-build": "^5.0.6",
+    "@edx/frontend-build": "^5.1.2",
```
stvstnfrd, line 63: Even if it makes sense to remove the `script` tag (per comment below), we could still keep in this bump :)

public/index.html
davidjoy: I think I'm pretty down with that, myself.

davidjoy: Note for posterity - our Optimizely plan does not include "full stack", which is what the above would require.  Thus, we can't use it.

stvstnfrd:
@davidjoy Looks like optimizely was added here [1],
so I guess we can close this, unless we want to keep the `frontend-build` bump
:shrug: 

- [1] d52aa324

davidjoy:
Yah, think we can just close this.

PR #133, opened 2020-07-30
AA-129: Resume Course button

ciduarte:
- Exposed Resume Course fields from API
- Made the Start/Resume Course button functional
- Freebies: fixed headers and alignment of a link in the Dates Widget

### DIFF
--- a/src/course-home/data/__factories__/outlineTabData.factory.js
+++ b/src/course-home/data/__factories__/outlineTabData.factory.js
@@ -22,4 +22,9 @@ Factory.define('outlineTabData')
     extra_text: 'Contact the administrator.',
   })
   .attr('handouts_html', [], () => '<ul><li>Handout 1</li></ul>')
-  .attr('offer_html', [], () => '<div>Great offer here</div>');
+  .attr('offer_html', [], () => '<div>Great offer here</div>')
+  .attr('resume_course', ['host', 'courseId'], (host, courseId) => ({
+    has_visited_course: false,
+    url: `${host}/courses/${courseId}/jump_to/block-v1:edX+Test+Block@12345abcde`,
+  }))
+  .attr('welcome_message_html', [], () => '<p>Welcome to this course!</p>');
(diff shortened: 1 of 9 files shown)


ciduarte:
**Outline Tab Before**
AA-129-Outline_Tab_Before

**Outline Tab After**
AA-129-Outline_Tab_After

src/course-home/data/__factories__/outlineTabData.factory.js
```diff
@@ -20,4 +20,8 @@ Factory.define('outlineTabData')
     can_enroll: true,
     extra_text: 'Contact the administrator.',
   })
-  .attr('handouts_html', [], () => '<ul><li>Handout 1</li></ul>');
+  .attr('handouts_html', [], () => '<ul><li>Handout 1</li></ul>')
+  .attr('resume_course', ['host', 'courseId'], (host, courseId) => ({
+    hasVisitedCourse: false,
```
mikix: In this mock, I think we use snake case, in the expectation that our camel casing code will later kick in.

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -37,11 +37,14 @@ function OutlineTab({ intl }) {
       courses,
       sections,
     },
+    resumeCourse,
   } = useModel('outline', courseId);
 
   const clientEnrollmentAlert = useEnrollmentAlert(courseId);
   useLogistrationAlert();
 
+  const { hasVisitedCourse } = resumeCourse;
```
mikix: You *could* do that decomposition above with:
```
resumeCourse: {
   hasVisitedCourse,
}
```

But I guess you use the resumeCourse object for its url later? Which feels slightly inconsistent. You could also do:

```
resumeCourse: {
   hasVisitedCourse,
   url: resumeUrl,
}
```

To rename the url object, if you were worried about such a generic variable name (I can never remember whether the new name is on left or right of the colon, but something like that).

mikix (approved):
Beautiful

src/course-home/outline-tab/OutlineTab.jsx
ciduarte: Yess you read my mind. I was trying to do that last suggestion but couldn't figure out the syntax. Thank you!!

PR #144, opened 2020-08-03
AA-277: Add certificate available alert on outline

mikix:
### Old:

Screenshot from 2020-08-03 12-07-38

### New:

Screenshot from 2020-08-03 12-07-43

### DIFF
--- a/src/alerts/certificate-available-alert/CertificateAvailableAlert.jsx
+++ b/src/alerts/certificate-available-alert/CertificateAvailableAlert.jsx
@@ -1,0 +1,62 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { getConfig } from '@edx/frontend-platform';
+import { FormattedMessage, FormattedRelative } from '@edx/frontend-platform/i18n';
+import { Hyperlink } from '@edx/paragon';
+
+import { Alert, ALERT_TYPES } from '../../generic/user-messages';
+
+function CertificateAvailableAlert({ payload }) {
+  const {
+    certDate,
+    username,
+    userTimezone,
+  } = payload;
+
+  const timezoneFormatArgs = userTimezone ? { timeZone: userTimezone } : {};
+
+  return (
+    <Alert type={ALERT_TYPES.INFO}>
+      <strong>
+        <FormattedMessage
+          id="learning.outline.alert.cert.title"
+          defaultMessage="We are working on generating course certificates."
+        />
+      </strong>
+      <br />
+      <FormattedMessage
+        id="learning.outline.alert.cert.when"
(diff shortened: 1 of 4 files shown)


src/alerts/certificate-available-alert/hooks.js
```diff
@@ -0,0 +1,42 @@
...
+  const {
+    datesWidget: {
+      courseDateBlocks,
+      userTimezone,
+    },
+  } = useModel('outline', courseId);
+  const { username } = getAuthenticatedUser();
+
+  const certBlock = courseDateBlocks.find(b => b.dateType === 'certificate-available-date');
+  const endBlock = courseDateBlocks.find(b => b.dateType === 'course-end-date');
```
Dillon-Dumesnil, line 22: Is this how we should be finding this information? Or are we just using the fact that these are also returned in the outline to our advantage?

src/alerts/certificate-available-alert/CertificateAvailableAlert.jsx
```diff
@@ -0,0 +1,62 @@
...
+  } = payload;
+
+  const timezoneFormatArgs = userTimezone ? { timeZone: userTimezone } : {};
+
+  return (
+    <Alert type={ALERT_TYPES.INFO}>
+      <strong>
+        <FormattedMessage
+          id="learning.outline.alert.cert.title"
+          defaultMessage="We are working on generating course certificates."
```
Dillon-Dumesnil, line 23: Should all these messages be internationalized?

Dillon-Dumesnil (commented):
Couple questions. As a more general question, should these alerts maybe live under the Course Home folder? `src/course-home/alerts/...`?

src/alerts/certificate-available-alert/hooks.js
mikix: The LMS alerts piggy back off the date blocks too, so I was kind of reproducing that logic here (at least for the cert available date).

For the course end date... mostly just using the fact that it's already there in the course date blocks to my advantage.

mikix:
> Couple questions. As a more general question, should these alerts maybe live under the Course Home folder? src/course-home/alerts/...?

Possibly...?  These alerts are scoped to the outline tab (topic of "course-outline-alerts" for example), so they can only reasonably be used there. So maybe they should live there.

But some alerts will eventually be shared? And I was frankly being lazy in not thinking about how they should all be organized.

Dillon-Dumesnil (approved):

src/alerts/certificate-available-alert/CertificateAvailableAlert.jsx
Dillon-Dumesnil: Didn't realize FormattedMessage is for internationalization

PR #164, opened 2020-08-04
AA-278 & AA-279: Add offer and course expired alerts to outline

mikix:
It was previously only used in the courseware. But to match the LMS, we also want to show it on the outline tab.

There is a change of design in the alert, but this is approved by Hilary and Eugene. They are looking at a more radical redesign and don't want to bother with focusing on this alert much right now.

### Current alerts
Screenshot from 2020-08-04 14-11-44

### New alerts
Screenshot from 2020-08-04 14-11-26


mikix:
Bundled both offer and course expired alerts into this one PR.

Dillon-Dumesnil (approved):

PR #167, opened 2020-08-05
AA-275: Persist if the original user was staff for Instructor Toolbar

Dillon-Dumesnil:
We want to be able to know if both the original user is a Staff user
as well as if the user being masqueraded as is staff. This updates
to accept both of these fields

### DIFF
--- a/src/course-home/data/__factories__/courseHomeMetadata.factory.js
+++ b/src/course-home/data/__factories__/courseHomeMetadata.factory.js
@@ -9,6 +9,7 @@ Factory.define('courseHomeMetadata')
   .option('host', 'http://localhost:18000')
   .attrs({
     is_staff: false,
+    original_user_is_staff: false,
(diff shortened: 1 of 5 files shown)


stvstnfrd:
Just to explicitly mention, the other place(s) I'm seeing `isStaff` in here is w/ `src/alerts/enrollment-alert`.
Do these references need updated too, or were they "part of the problem" (using the wrong `staff` indicator)?

davidjoy (approved):

Dillon-Dumesnil:
Yes, thank you for pointing that out. I believe we would want to maintain that reference on the current `isStaff` value because you should be viewing the course as that learner. If it turns out that learner is not enrolled or viewing a not enrolled kind of bug, we would want to try and make sure we don't bypass that by escaping that check with staff access.

mikix (approved):

PR #170, opened 2020-08-06
activeCourse reducer and loading state constants

davidjoy:
**Best reviewed commit-by-commit**.

This is some pre-work I decided to do prior to adding the 'kill switch' to the MFE.  Admittedly I decided to do some general scouts-principle-style cleanup as I went, so not all of this is directly related to that effort.  

One thing I did was to extract `courseStatus` and `courseId` into their own reducer - activeCourse - and now both course-home and courseware use it.  I found that we have a number of places where course-home and courseware need to pass a courseId into children like the header/instructor toolbar that are shared between them because in each page it has a different location in the redux store.  Given that all pages in the MFE rely on a course being loaded, I chose to consolidate that so that any given component would always know where to look for the active course.

Also glomming onto this PR - I created some constants for our various loading/loaded/failed states, which I've been meaning to do for a while.  I disambiguated the constant names just so we could import constants from the 'right' module in the app, rather than re-using them all from some constants file at the top.  Felt like the right thing to do, even though it was a bit more work.


src/course-home/data/redux.test.js
```diff
@@ -41,7 +42,8 @@ describe('Data layer integration tests', () => {
   });
 
   it('Should initialize store', () => {
-    expect(store.getState()).toMatchSnapshot();
+    expect(store.getState().activeCourse.courseId).toBeNull();
```
davidjoy, line 43: The snapshots for this test were very verbose, which can be a problem in general with snapshots.  I replaced them with specific assertions about the data we care about for a given test, or at least a representative sample of it.  You'll see more of that below.

src/course-home/data/slice.js
```diff
@@ -8,33 +8,16 @@ export const FAILED = 'failed';
 const slice = createSlice({
   name: 'course-home',
   initialState: {
-    courseStatus: 'loading',
```
davidjoy, line 11: So I folded courseStatus and courseId into their own reducer - activeCourse - and now both course-home and courseware use it.

More rationale in the PR description.

src/course/data/selectors.js
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable import/prefer-default-export */
+import { createSelector } from 'reselect';
+
+export const activeCourseSelector = createSelector(
```
davidjoy: Reusable selector to get the active course.

src/courseware/CoursewareContainer.jsx
```diff
@@ -280,21 +281,15 @@ CoursewareContainer.defaultProps = {
   sequence: null,
 };
 
-const currentCourseSelector = createSelector(
```
davidjoy, line 283: This was basically moved into the course/data/selectors.js file and renamed "activeCourseSelector" - I don't know why I changed the adjective, it felt more... active.

src/courseware/CoursewareRedirect.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { Switch, Route, useRouteMatch } from 'react-router';
 import { getConfig } from '@edx/frontend-platform';
 import { FormattedMessage } from '@edx/frontend-platform/i18n';
-import PageLoading from './generic/PageLoading';
+import PageLoading from '../generic/PageLoading';
```
davidjoy, line 5: I moved CoursewareRedirect into the courseware subdirectory, since it's specific to redirects from that page.

docs/decisions/0004-model-store.md
```diff
@@ -4,6 +4,6 @@ Because we have a variety of models in this app (course, section, sequence, unit
...
+A common pattern when loading data from an API endpoint is to use the model-store's redux actions (addModel, updateModel, etc.) to load the "models" themselves into the model store by ID, and then dispatch another action to save references elsewhere in the redux store to the data that was just added.  When adding courses, sequences, etc., to model-store, we also save the courseId and sequenceId in the 'courseware' part of redux.  This means the courseware React Components can extract the data from the model-store quickly by using the courseId as a key: `state.models.courses[state.activeCourse.courseId]`.  For an array, it iterates once over the ID list in order to extract the models from model-store.  This iteration is done when React components' re-render, and can be done less often through memoization as necessary.
```
davidjoy, line 9: This is just updating this comment to use 'activeCourse.courseId' instead of 'courseware.courseId'

stvstnfrd: Meta-comment: Thank you for this; I find explanatory comments like this to be helpful.
It saves me from asking what changed and/or trying to tease out which subtle little thing changed.

A thing I'd like to cover in a personal "How I work" includes this kind of explicit, pre-review code review, where I go through and explain the things I know people are going to ask about :)

src/course-home/data/redux.test.js
```diff
@@ -70,8 +72,30 @@ describe('Data layer integration tests', () => {
       await executeThunk(thunks.fetchDatesTab(courseId), store.dispatch);
 
       const state = store.getState();
-      expect(state.courseHome.courseStatus).toEqual('loaded');
-      expect(state).toMatchSnapshot();
+      expect(state.activeCourse.courseStatus).toEqual('loaded');
```
stvstnfrd: Do we also have a constant for `LOADED` like `LOADING` above?

*edit*: Oh, I think this comes in a later commit?

src/course-home/data/redux.test.js
```diff
@@ -86,7 +110,7 @@ describe('Data layer integration tests', () => {
       await executeThunk(thunks.fetchOutlineTab(courseId), store.dispatch);
 
       expect(loggingService.logError).toHaveBeenCalled();
-      expect(store.getState().courseHome.courseStatus).toEqual('failed');
+      expect(store.getState().activeCourse.courseStatus).toEqual('failed');
```
stvstnfrd: Constant value?

*edit*: Oh, I think this comes in a later commit?

src/course-home/data/redux.test.js
```diff
@@ -70,8 +72,30 @@ describe('Data layer integration tests', () => {
       await executeThunk(thunks.fetchDatesTab(courseId), store.dispatch);
 
       const state = store.getState();
-      expect(state.courseHome.courseStatus).toEqual('loaded');
-      expect(state).toMatchSnapshot();
+      expect(state.activeCourse.courseStatus).toEqual('loaded');
+      expect(state.activeCourse.courseId).toEqual(courseId);
+
+      // Validate course
```
stvstnfrd, line 77: I think these two blocks (`Validate course` and `Representative sample...`) are identical between the two tests, so at >10 lines, I'd be inclined to move this into a helper method and call it in both places.

Especially since they're using magic numbers (field count, tab count), a helper (and/or constants) will make this less annoying to maintain.

`validateCourseAndFields` or some such.

src/course-home/data/redux.test.js
```diff
@@ -70,8 +72,30 @@ describe('Data layer integration tests', () => {
...
+      expect(course.number).toEqual(courseHomeMetadata.number);
+      expect(Array.isArray(course.tabs)).toBe(true);
+      expect(course.tabs.length).toBe(5); // Weak assertion, but proves the array made it through.
+
+      // This proves the tab type came through as a modelType.  We don't need to assert much else
+      // here because the shape of this data is not passed through any sort of normalization scheme,
+      // it just gets camelCased.
+      const dates = state.models.dates[courseId];
+      expect(dates.id).toEqual(courseId);
+      expect(dates.verifiedUpgradeLink).toBe(datesTabData.verified_upgrade_link);
```
kdmccormick, line 97: Thanks for explaining why we are and aren't making certain assertions here :100:

src/course-home/data/redux.test.js
```diff
@@ -126,13 +127,14 @@ describe('Data layer integration tests', () => {
...
+      expect(state.courseHome.displayResetDatesToast).toBe(false);
 
       // Validate course
       const course = state.models.courses[courseId];
       const expectedFieldCount = Object.keys(course).length;
       // If this breaks, you should consider adding assertions below for the new data.  If it's not
-      // an "interesting" addition, bump the number.
-      expect(expectedFieldCount).toBe(8);
+      // an "interesting" addition, just bump the number anyway.
+      expect(expectedFieldCount).toBe(9);
```
stvstnfrd, line 135: This isn't a hard-veto or anything (more commentary, really), but just to voice, I'm not a huge fan of magic numbers and asserting counts. As well-intentioned as they are, they can easily devolve into an anti-pattern of "oh the test failed? just change the magic number to whatever the actual result is"... or am I the only who does that w/ query-count assertions in `edx-platform`?! ;)

This also means tests will "pass" if you remove a field and add a new one.

On a philosophical level, it becomes less "verify this works" and more "verify I'm keeping these two numbers in sync when the test-runner yells at me" :D

That said, I don't have a great alternative, so carry on for now, I guess :P

src/courseware/data/thunks.js
```diff
@@ -60,9 +65,6 @@ export function fetchCourse(courseId) {
         }));
       }
 
-      const fetchedMetadata = courseMetadataResult.status === 'fulfilled';
```
stvstnfrd, line 61: Thank you for the constants!

kdmccormick (approved):
Looks solid to me :+1:

stvstnfrd (approved):
Some optional questions, but thank you for the comments and constants!

docs/decisions/0004-model-store.md
kdmccormick: ^ +1

src/course-home/data/redux.test.js
davidjoy: Yeaaah, I'm not thrilled with it either, but didn't like the idea of the test slowly testing less and less of the actual data.  Ugh.  Will think on this and maybe just get rid of it. :)

ciduarte (approved):
Spent most of my efforts sanity checking course home data changes, but looks like the consolidation went smoothly. Thanks for being so thorough!

davidjoy:
I'm going to close this - it was a decent refactor, but is now probably so far behind HEAD that it's not worth trying to keep around.

PR #176, opened 2020-08-10
Redirect to the current unit’s lmsWebUrl if the MFE is disabled

davidjoy:
If we receive an error_code of ‘microfrontend_disabled’ - go to the equivalent unit in the LMS experience.

### DIFF
--- a/src/CoursewareRedirect.jsx
+++ b/src/CoursewareRedirect.jsx
@@ -3,6 +3,7 @@ import { Switch, Route, useRouteMatch } from 'react-router';
 import { getConfig } from '@edx/frontend-platform';
 import { FormattedMessage } from '@edx/frontend-platform/i18n';
 import PageLoading from './generic/PageLoading';
+import { useModel } from './generic/model-store';
 
 export default () => {
   const { path } = useRouteMatch();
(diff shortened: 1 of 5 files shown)


src/CoursewareRedirect.jsx
```diff
@@ -18,6 +19,18 @@ export default () => {
       />
 
       <Switch>
+        <Route
+          path={`${path}/unit/:unitId`}
+          render={({ match }) => {
+            const { unitId } = match.params;
+            const unit = useModel('units', unitId);
```
stvstnfrd: Nit: If we coalesce to an empty object here, we can forgo the explicit `undefined` check below:

```js
const unit = useModel('units', unitId) || {};
if (unit.lmsWebUrl) {
    global.location.assign(unit.lmsWebUrl);
}
```

src/courseware/CoursewareContainer.jsx
```diff
@@ -186,6 +192,9 @@ class CoursewareContainer extends Component {
       case 'unfulfilled_milestones':
         url = '/redirect/dashboard';
         break;
+      case 'microfrontend_disabled':
+        url = `/redirect/unit/${routeUnitId}`;
```
stvstnfrd: Just checking because I find some of the complex destructuring syntax to be a little opaque:

```js
routeUnitId === this.props.match.params.unitId
```
???

stvstnfrd (approved):
Just a nit and a clarifying question, otherwise +1

src/CoursewareRedirect.jsx
davidjoy: It's not written down anywhere, but in general the MFE assumes that a model object is either well-formed or `null`/`undefined` - using an empty object to represent it would cut against that grain pretty hard.

src/courseware/CoursewareContainer.jsx
davidjoy: Yeppers!

davidjoy:
This PR was absorbed into https://github.com/edx/frontend-app-learning/pull/183 and merged to master - this can be closed unmerged.

PR #178, opened 2020-08-12
Format may not be defined. Provide a default and only add to URL if i…

Dillon-Dumesnil:
…t exists

### DIFF
--- a/src/courseware/course/sequence/Unit.jsx
+++ b/src/courseware/course/sequence/Unit.jsx
@@ -57,7 +57,10 @@ function Unit({
   id,
   intl,
 }) {
-  const iframeUrl = `${getConfig().LMS_BASE_URL}/xblock/${id}?show_title=0&show_bookmark_button=0&format=${format}`;
+  let iframeUrl = `${getConfig().LMS_BASE_URL}/xblock/${id}?show_title=0&show_bookmark_button=0`;
+  if (format) {
+    iframeUrl += `&format=${format}`;
+  }
 
   const [iframeHeight, setIframeHeight] = useState(0);
   const [hasLoaded, setHasLoaded] = useState(false);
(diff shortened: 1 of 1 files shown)


src/courseware/course/sequence/Unit.jsx
```diff
@@ -57,7 +57,10 @@ function Unit({
   id,
   intl,
 }) {
-  const iframeUrl = `${getConfig().LMS_BASE_URL}/xblock/${id}?show_title=0&show_bookmark_button=0&format=${format}`;
+  let iframeUrl = `${getConfig().LMS_BASE_URL}/xblock/${id}?show_title=0&show_bookmark_button=0`;
```
Dillon-Dumesnil, line 60: Is this proper syntax or should it be
```
let iframeUrl;
iframeUrl = ...
if (format)...

Comment (davidjoy) [reply src/courseware/course/sequence/Unit.jsx:60]: I like the version you have here in the PR.  For some reason I don't like declaring but not initializing variables. ¯\_(ツ)_/¯  Personal preference.

File: src/courseware/course/sequence/Unit.jsx [hunk 2]
```diff
@@ -151,13 +154,14 @@ function Unit({
...
   courseId: PropTypes.string.isRequired,
-  format: PropTypes.string.isRequired,
+  format: PropTypes.string,
   id: PropTypes.string.isRequired,
   intl: intlShape.isRequired,
   onLoaded: PropTypes.func,
 };
 
 Unit.defaultProps = {
+  format: '',
```
davidjoy: I'd let this stay as `null`, which is what it's coming through as from higher up in the app.  Means it has a consistent value until it's added to the URL.

davidjoy (approved):
One nit, but looks good.

src/courseware/course/sequence/Unit.jsx
Dillon-Dumesnil: Yeah, totally fair point. I questioned this myself. I went with the empty string so it matched the PropType, but I do like the consistency of null. Will update.

PR #179, opened 2020-08-12
TNL-7409 Fix checkBlockCompletion parameters

davidjoy:
We were assuming a prop named unitId existed in CoursewareContainer - it doesn’t.  unitId is not in redux.  What we do have, is the unitId in the route params - what we refer to as routeUnitId.  If we use this instead of the non-existent unitId, then life is good.

I wrote a test (that breaks!) prior to implementing the fix.  The fix satisfies the test. :tada:

### DIFF
--- a/src/courseware/course/sequence/sequence-navigation/UnitNavigation.jsx
+++ b/src/courseware/course/sequence/sequence-navigation/UnitNavigation.jsx
@@ -60,7 +60,11 @@ export default function UnitNavigation(props) {
 
 UnitNavigation.propTypes = {
   sequenceId: PropTypes.string.isRequired,
-  unitId: PropTypes.string.isRequired,
+  unitId: PropTypes.string,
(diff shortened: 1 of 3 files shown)


src/courseware/course/sequence/sequence-navigation/UnitNavigation.jsx
```diff
@@ -60,7 +60,11 @@ export default function UnitNavigation(props) {
 
 UnitNavigation.propTypes = {
   sequenceId: PropTypes.string.isRequired,
-  unitId: PropTypes.string.isRequired,
+  unitId: PropTypes.string,
```
davidjoy, line 63: I noticed that I was getting a warning about this unitId being null when navigating between sequences (which is correct), so I fixed this too.

stvstnfrd (approved):

PR #180, opened 2020-08-12
wip: Use deep equality when checking whether to re-render alerts

kdmccormick:
https://openedx.atlassian.net/browse/TNL-7400

@davidjoy


kdmccormick:
We determined that this is an attempt to fix a kludge with an even deeper kludge. We'll go for a simpler fix in https://github.com/edx/frontend-app-learning/pull/182 and write a ticket to clean up alerts in general after release.

PR #183, opened 2020-08-13
Implement kill-switch for non-staff users

stvstnfrd:
to redirect to the current unit’s `lmsWebUrl` if the MFE is disabled

If we receive an `error_code` of `'microfrontend_disabled'`,
go to the equivalent unit in the LMS experience.

Fixes: TNL-7362
Co-authored-by: stvn

### DIFF
--- a/src/courseware/CoursewareContainer.jsx
+++ b/src/courseware/CoursewareContainer.jsx
@@ -171,7 +171,15 @@ class CoursewareContainer extends Component {
   }
 
   renderDenied() {
-    const { courseId, course } = this.props;
+    const {
+      course,
+      courseId,
+      match: {
+        params: {
+          unitId: routeUnitId,
+        },
+      },
+    } = this.props;
     let url = `/redirect/course-home/${courseId}`;
     switch (course.canLoadCourseware.errorCode) {
       case 'audit_expired':
@@ -185,6 +193,9 @@ class CoursewareContainer extends Component {
       case 'survey_required': // TODO: Redirect to the course survey
       case 'unfulfilled_milestones':
         url = '/redirect/dashboard';
+        break;
+      case 'microfrontend_disabled':
+        url = `/redirect/courseware/${courseId}/unit/${routeUnitId}`;
         break;
       case 'authentication_required':
       case 'enrollment_required':
--- a/src/courseware/CoursewareRedirect.jsx
+++ b/src/courseware/CoursewareRedirect.jsx
@@ -1,0 +1,14 @@
+import { getConfig } from '@edx/frontend-platform';
+
+import { useModel } from '../generic/model-store';
+
+export default function CourseRedirect({ match }) {
+  const {
+    courseId,
+    unitId,
+  } = match.params;
+  const unit = useModel('units', unitId) || {};
+  const coursewareUrl = unit.lmsWebUrl || `${getConfig().LMS_BASE_URL}/courses/${courseId}/courseware/`;
+  global.location.assign(coursewareUrl);
+  return null;
+}
--- a/src/courseware/CoursewareRedirectLandingPage.jsx
+++ b/src/courseware/CoursewareRedirectLandingPage.jsx
@@ -2,7 +2,9 @@ import React from 'react';
 import { Switch, Route, useRouteMatch } from 'react-router';
 import { getConfig } from '@edx/frontend-platform';
 import { FormattedMessage } from '@edx/frontend-platform/i18n';
-import PageLoading from './generic/PageLoading';
+
+import PageLoading from '../generic/PageLoading';
+import CoursewareRedirect from './CoursewareRedirect';
 
 export default () => {
   const { path } = useRouteMatch();
@@ -18,6 +20,10 @@ export default () => {
       />
 
       <Switch>
+        <Route
+          path={`${path}/courseware/:courseId/unit/:unitId`}
+          component={CoursewareRedirect}
+        />
         <Route
           path={`${path}/course-home/:courseId`}
           render={({ match }) => {
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -20,7 +20,7 @@ import './index.scss';
 import './assets/favicon.ico';
 import OutlineTab from './course-home/outline-tab';
 import CoursewareContainer from './courseware';
-import CoursewareRedirect from './CoursewareRedirect';
+import CoursewareRedirectLandingPage from './courseware/CoursewareRedirectLandingPage';
 import DatesTab from './course-home/dates-tab';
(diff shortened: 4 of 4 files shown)


src/courseware/CoursewareContainer.jsx
```diff
@@ -171,7 +171,15 @@ class CoursewareContainer extends Component {
   }
 
   renderDenied() {
-    const { courseId, course } = this.props;
+    const {
+      course,
```
stvstnfrd, line 175: @davidjoy The only changes I made here were to split this up onto multiple lines.

src/courseware/CoursewareRedirect.jsx
```diff
@@ -0,0 +1,14 @@
+import { getConfig } from '@edx/frontend-platform';
+
```
stvstnfrd, line 2: I split the internal and external imports.

src/courseware/CoursewareRedirect.jsx
```diff
@@ -0,0 +1,14 @@
...
+import { useModel } from '../generic/model-store';
+
+export default function CourseRedirect({ match }) {
+  const {
+    courseId,
+    unitId,
+  } = match.params;
+  const unit = useModel('units', unitId) || {};
```
stvstnfrd, line 10: # TL;DR: I'm about to get _suuuper_ wonky here, but I find this approach to be a little easier to read (less branching) and I have a compromise approach proposed below, syntax permitting.

You've been warned; feel free to roll your eyes ;)

> It's not written down anywhere, but in general the MFE assumes that a model object is either well-formed or null/undefined - using an empty object to represent it would cut against that grain pretty hard.

I definitely agree that public interfaces, data stores, etc. strongly benefit from constraining their possible values; it'd be a nightmare if we had to check that everywhere, especially since empty objects are truth-y in JS (whereas they're false-y in Python, which provides consistency w/ empty/None/null).

That said, I don't feel this is the same situation, as we're not changing the behavior/expectation of this data anywhere _except_ localized here in this function. If local functions "can't" parse data in a way they see fit for consumption, that strikes me as code-smell (or unhelpful dogma).

I guess my thinking here is, we don't really care if there's truly an underlying unit model and whether or not its actually backed by the datastore; we just want to find an `lmsWebUrl`, if we have access to one, otherwise we'll build our own URL.

Disclosure: I'm openly not a fan of ternary operators (they're susceptible to bloat and quickly become unreadable) and I find coalescing operations to be cleaner [5] than a bunch of `if`/`else` guards around object access/indexing.

## Rubber-ducky-ing out my thought process here:

> What do we want to do?

```
Send the user to a corresponding page in the old experience
```

> What is "the corresponding page"?

```
If we have a unit and the unit has a URL, that page.
Else, use a generic courseware page for the course.
```

So the reason I like the convention of coalescing to an non-nullish [4] value (`{}`, `''`) is that we can leverage (what I personally find to be one of the biggest advantages of) truthy-operators in JS and Python, to coalesce our way down to the values we need, without ternary operators and `if`/`else` guards.

Assuming one understands boolean-short-circuiting in these languages [3], I find a natural translation between:

```
Try to get the unit's URL,
otherwise, fall back to a generic courseware page for the course.
```

and:

```js
let coursewareUrl;
try {
  coursewareUrl = unit.lmsWebUrl;
} catch(error) {
  coursewareUrl = `${getConfig().LMS_BASE_URL}/courses/${courseId}/courseware/`
}
```

which we refactor to remove the need for `try`/`catch` overhead to:

```js
// redirect to the    the unit's URL or    the generic courseware page
const coursewareUrl = unit.lmsWebUrl || `${getConfig().LMS_BASE_URL}/courses/${courseId}/courseware/`
```

with the implied understanding [3] that 'or' means "stop looking when you find a (truthy/non-nullish) value".

That all said, `iff` we support the optional chaining syntax [2], we can have the best of both worlds, where I can freely index into objects and coalesce their values, and we can maintain the "don't ever let a unit be 'empty'" convention :stuck_out_tongue_closed_eyes: 

```js
// See implementation notes below [1][2][6]
const unit = useModel('units', unitId);
const coursewareUrl = unit?.lmsWebUrl || `${getConfig().LMS_BASE_URL}/courses/${courseId}/courseware/`;
```

Thanks you for coming to my TED Talk!

### References

- [1] As I've mentioned before, I'm still catching up on some of the more recent ES syntax, so I didn't realize optional chaining had landed in the spec. There are a few browsers that don't support it [2] (Internet Explorer still, some mobile browsers, etc.).
  - How does our compilation/build-step work here?
  - Do we transpile down (ES3/5/etc)?
  - Can we use this or no?
- [2] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining
- [3] I guess this is a good check-in to see if this is a reasonable assumption or not... I've long held (up hill! both ways!) that these are patterns all devs should understand, that boolean/truthy short-circuiting is a "way of life" in JS/Python, given its presence in their ecosystems.
- [4] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_Coalescing_Operator
- [5] `:hand-wave:` :wave: 
- [6] I'm intentionally _not_ using the nullish coalescing operator here [4], because we should still treat an empty string as an "invalid" URL.

src/courseware/CoursewareRedirect.jsx
```diff
@@ -0,0 +1,14 @@
...
+  const coursewareUrl = unit.lmsWebUrl || `${getConfig().LMS_BASE_URL}/courses/${courseId}/courseware/`;
```
stvstnfrd, line 11: This is the only other change I made to logic.

stvstnfrd (commented):
This replaces https://github.com/edx/frontend-app-learning/pull/176

src/courseware/CoursewareRedirect.jsx
davidjoy: I'm honestly not quite sure if optional chaining and nullish coalescing operators are properly transpiled down in our babel config - would need to test it.  Because I'm unsure, I've implicitly avoided using them. :)  And also, I wasn't even aware of `??` to be honest - nice!  Helpful.

I think all this is fine, thank you for the TED Talk. ;)

I find that adding in another potential intermediate state for the data (i.e., the `{}` if the unit doesn't exist) doesn't help my mental model, so maybe I avoid doing that and prefer to have fewer types of data and more explicit checks, but that's purely a subconscious personal preference and I don't feel particularly strongly about it.

I do think that there's something to be said for avoiding idiosyncratic or unique language features when developing code that may be inspected by a wide variety of folks with various skill levels, but I'm also certainly guilty of using those features anyway, so I've no leg to stand on there!  

Anyway, long story short, this looks great. 👍

davidjoy (approved):

PR #184, opened 2020-08-13
Allow units to create a modal in the parent window

davestgermain:
This allows https://github.com/edx/xblock-lti-consumer/pull/97 to show an LTI component in a modal window.

It requires changes to paragon modals to support flexible sizing and scoping of the component (so that other modals can be styled differently)

### DIFF
--- a/src/courseware/course/sequence/Unit.jsx
+++ b/src/courseware/course/sequence/Unit.jsx
@@ -9,6 +9,7 @@ import { useDispatch } from 'react-redux';
 import PropTypes from 'prop-types';
 import { getConfig } from '@edx/frontend-platform';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import { Modal } from '@edx/paragon';
 import messages from './messages';
 import BookmarkButton from '../bookmark/BookmarkButton';
 import { useModel } from '../../../generic/model-store';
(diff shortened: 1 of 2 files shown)


src/courseware/course/sequence/Unit.jsx
```diff
@@ -130,6 +135,28 @@ function Unit({
...
+                  <iframe
+                    title={modalOptions.title}
+                    allow="microphone *; camera *; midi *; geolocation *; encrypted-media *"
+                    frameBorder="0"
+                    src={modalOptions.url}
+                  />
+                )}
+            </>
+          )}
+          width={modalOptions.width}
```
davidjoy: I think we should probably remove this width line - if we're able to implement this, I think it'll be by adding a CSS class to the modal, rather than an explicit property like this.

src/index.scss
```diff
@@ -311,6 +311,21 @@ $primary: #1176B2;
   }
 }
 
+.modal {
```
davidjoy, line 314: Note for posterity - this CSS is intended to be temporary so we can launch courseware.  Because it affects all modals, it'll affect the Celebration Modal as well.  

In the future we need to get this applied to ONLY this modal.

davidjoy (approved):

PR #188, opened 2020-08-18
AA-305: Update Masquerade Widget UI

mikix:
It's not exactly the same as the mocks (I didn't bother with the horizontal rule when the window gets narrow). But the labels and layout is a lot closer.

https://www.figma.com/file/eQCBHwrmnQcuUucb75n07V/Masquerade?node-id=0%3A1

Screenshot from 2020-08-18 14-24-41

Screenshot from 2020-08-18 14-25-04

Screenshot from 2020-08-18 14-25-24

Screenshot from 2020-08-21 11-08-10

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "npm run lint"
+      "pre-commit": ""
     }
   },
   "author": "edX",
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": ""
+      "pre-commit": "npm run lint"
     }
   },
   "author": "edX",
--- a/src/instructor-toolbar/InstructorToolbar.jsx
+++ b/src/instructor-toolbar/InstructorToolbar.jsx
@@ -54,27 +54,30 @@ export default function InstructorToolbar(props) {
   return (
     <div>
       <div className="bg-primary text-light">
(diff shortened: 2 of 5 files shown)


mikix:
@hilaryg123 -- I know we're missing the horizontal rule and the "Specific Student..." should be "Specific learner" (both required a bit more work, so I didn't bother here), but how do those screenshots look?

hilaryg123:
@mikix looking good - just a few piece of feedback on my end:

- In lieu of the horizontal rule between masquerade feature and links to other apps, can we just add some more padding between on mobile to indicate that those two elements of the banner aren't _really_ related?
- Can we limit the width of the buttons on mobile to just be a little beyond the text length so that they can sit side by side instead of stacking, per the wireframes? Not sure if that's possible but certainly preferable.
- Any reason Insights wouldn't also be a link here? More just curious

ciduarte (approved):

mikix:
@hilaryg123 OK. Fixed buttons to be on one line and added the horizontal rule. The Insights button was missing just because I didn't have it configured locally. Fixed that and you can see Insights in my last screenshot in description, along with the one-line and horizontal-rule.

Now I think the big delta with the mockups is that the "view course in" buttons are not horizontally aligned with the masquerade dropdown and text box. But that actually is a little hard to solve.

hilaryg123:
@mikix Thanks for those fixes. I'm not too bothered by the delta you described, especially because this is an internal tool. Looks good on my end!

PR #190, opened 2020-08-20
AA-125: Course Goals in Course Home Outline Tab

ciduarte:
Initial Goal Card
initial-select-goal-card

Edit Goal Selector and Success Toast
AA-125_edit-goal-selector-and-success-toast

Mobile View - Initial Goal Card
initial-select-goal-mobile-view

Mobile View - Edit Goal Selector and Success Toast
AA-125_edit-goal-mobile-view

### DIFF
--- a/src/course-home/data/__factories__/outlineTabData.factory.js
+++ b/src/course-home/data/__factories__/outlineTabData.factory.js
@@ -17,6 +17,10 @@ Factory.define('outlineTabData')
       blocks: courseBlocks.blocks,
     };
   })
+  .attr('course_goals', [], () => ({
+    goal_options: [],
+    selected_goal: {},
+  }))
   .attr('enroll_alert', {
     can_enroll: true,
     extra_text: 'Contact the administrator.',
--- a/src/course-home/data/api.js
+++ b/src/course-home/data/api.js
@@ -69,6 +69,7 @@ export async function getOutlineTabData(courseId) {
     data,
   } = tabData;
   const courseBlocks = normalizeBlocks(courseId, data.course_blocks.blocks);
+  const courseGoals = camelCaseObject(data.course_goals);
   const courseExpiredHtml = data.course_expired_html;
   const courseTools = camelCaseObject(data.course_tools);
   const datesWidget = camelCaseObject(data.dates_widget);
@@ -80,6 +81,7 @@ export async function getOutlineTabData(courseId) {
 
   return {
     courseBlocks,
+    courseGoals,
     courseExpiredHtml,
     courseTools,
     datesWidget,
@@ -94,6 +96,11 @@ export async function getOutlineTabData(courseId) {
 export async function postCourseDeadlines(courseId) {
   const url = new URL(`${getConfig().LMS_BASE_URL}/api/course_experience/v1/reset_course_deadlines`);
   return getAuthenticatedHttpClient().post(url.href, { course_key: courseId });
+}
+
+export async function postCourseGoals(courseId, goalKey) {
+  const url = new URL(`${getConfig().LMS_BASE_URL}/api/course_home/v1/save_course_goal`);
+  return getAuthenticatedHttpClient().post(url.href, { course_id: courseId, goal_key: goalKey });
 }
 
 export async function postDismissWelcomeMessage(courseId) {
--- a/src/course-home/data/redux.test.js
+++ b/src/course-home/data/redux.test.js
@@ -106,6 +106,18 @@ describe('Data layer integration tests', () => {
     });
   });
 
+  describe('Test saveCourseGoal', () => {
+    it('Should save course goal', async () => {
+      const goalUrl = `${getConfig().LMS_BASE_URL}/api/course_home/v1/save_course_goal`;
+      axiosMock.onPost(goalUrl).reply(200, {});
+
+      await thunks.saveCourseGoal(courseId, 'unsure');
+
+      expect(axiosMock.history.post[0].url).toEqual(goalUrl);
+      expect(axiosMock.history.post[0].data).toEqual(`{"course_id":"${courseId}","goal_key":"unsure"}`);
+    });
+  });
+
   describe('Test resetDeadlines', () => {
     it('Should reset course deadlines', async () => {
       const resetUrl = `${getConfig().LMS_BASE_URL}/api/course_experience/v1/reset_course_deadlines`;
--- a/src/course-home/data/slice.js
+++ b/src/course-home/data/slice.js
@@ -10,8 +10,9 @@ const slice = createSlice({
   initialState: {
     courseStatus: 'loading',
     courseId: null,
-    resetDatesToastBody: null,
-    resetDatesToastHeader: null,
+    toastBodyText: null,
+    toastBodyLink: null,
+    toastHeader: null,
   },
   reducers: {
     fetchTabRequest: (state, { payload }) => {
@@ -26,13 +27,15 @@ const slice = createSlice({
       state.courseId = payload.courseId;
       state.courseStatus = FAILED;
     },
-    setResetDatesToast: (state, { payload }) => {
+    setCallToActionToast: (state, { payload }) => {
       const {
-        body,
         header,
+        link,
+        linkText,
       } = payload;
-      state.resetDatesToastBody = body;
-      state.resetDatesToastHeader = header;
+      state.toastBodyLink = link;
+      state.toastBodyText = linkText;
+      state.toastHeader = header;
     },
   },
 });
@@ -41,7 +44,7 @@ export const {
   fetchTabRequest,
   fetchTabSuccess,
   fetchTabFailure,
-  setResetDatesToast,
+  setCallToActionToast,
 } = slice.actions;
 
 export const {
--- a/src/course-home/outline-tab/OutlineTab.jsx
+++ b/src/course-home/outline-tab/OutlineTab.jsx
@@ -1,14 +1,17 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { useSelector } from 'react-redux';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 
 import { AlertList } from '../../generic/user-messages';
 
 import CourseDates from './widgets/CourseDates';
+import CourseGoalCard from './widgets/CourseGoalCard';
 import CourseHandouts from './widgets/CourseHandouts';
 import CourseTools from './widgets/CourseTools';
+import LearningToast from '../../toast/LearningToast';
 import messages from './messages';
 import Section from './Section';
+import UpdateGoalSelector from './widgets/UpdateGoalSelector';
 import useAccessExpirationAlert from '../../alerts/access-expiration-alert';
 import useCertificateAvailableAlert from './alerts/certificate-available-alert';
 import useCourseEndAlert from './alerts/course-end-alert';
@@ -39,6 +42,10 @@ function OutlineTab({ intl }) {
       courses,
       sections,
     },
+    courseGoals: {
+      goalOptions,
+      selectedGoal,
+    },
     courseExpiredHtml,
     resumeCourse: {
       hasVisitedCourse,
@@ -46,6 +53,9 @@ function OutlineTab({ intl }) {
     },
     offerHtml,
   } = useModel('outline', courseId);
+
+  const [courseGoalToDisplay, setCourseGoalToDisplay] = useState(selectedGoal);
+  const [goalToastHeader, setGoalToastHeader] = useState(null);
 
   // Above the tab alerts (appearing in the order listed here)
   const logistrationAlert = useLogistrationAlert();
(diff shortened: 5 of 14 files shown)


Ahernkb (approved):
Looks great!

src/course-home/outline-tab/widgets/UpdateGoalSelector.jsx
```diff
@@ -0,0 +1,101 @@
...
+  goalOptions,
+  intl,
+  selectedGoal,
+  setGoalToDisplay,
+}) {
+  const [editingGoal, setEditingGoal] = useState(false);
+
+  const dispatch = useDispatch();
+
+  function selectGoalHandler(e) {
```
xnick421: i know `e` is like universal known to be short hand for `event`, but maybe being verbose here and writing out `event` is clearer idk

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -22,6 +26,7 @@ import WelcomeMessage from './widgets/WelcomeMessage';
 function OutlineTab({ intl }) {
   const {
     courseId,
+    courseGoalToastHeader,
```
xnick421: not sure if we need to leverage redux for this state.  i think this state can just live in this component only.

src/course-home/outline-tab/widgets/CourseGoalCard.jsx
```diff
@@ -0,0 +1,83 @@
...
+function CourseGoalCard({
+  courseId,
+  goalOptions,
+  intl,
+  title,
+  setGoalToDisplay,
+}) {
+  const dispatch = useDispatch();
+
+  function selectGoalHandler(e) {
```
xnick421: i know `e` is universally known to be short hand for `event`, but maybe being verbose here wouldnt hurt.  up to you

src/course-home/outline-tab/widgets/CourseGoalCard.jsx
```diff
@@ -0,0 +1,83 @@
...
+        <div className="row w-100 m-0 justify-content-between align-items-center">
+          <div className="col col-8 p-0">
+            <Card.Title className="h6 m-0">{intl.formatMessage(messages.goalWelcome)}&nbsp;{title}</Card.Title>
+          </div>
+          <div className="col col-auto p-0">
+            <Button
+              variant="sm btn-link btn-block p-0"
+              data-goal-key="unsure"
+              data-goal-text={`${intl.formatMessage(messages.goalUnsure)}`}
+              onClick={(e) => { selectGoalHandler(e); }}
```
xnick421: same here in regards to `event`

src/course-home/outline-tab/widgets/CourseGoalCard.jsx
```diff
@@ -0,0 +1,83 @@
...
+        <Card.Text className="my-2 mx-1">{intl.formatMessage(messages.setGoal)}</Card.Text>
+        <div className="row w-100 m-0">
+          {goalOptions.map((goal) => (
+            (goal.key !== 'unsure') && (
+              <div key={`goal-${goal.key}`} className="col-auto flex-grow-1 mx-1 my-2 p-0">
+                <Button
+                  variant="outline-primary btn-block"
+                  data-goal-key={goal.key}
+                  data-goal-text={goal.text}
+                  onClick={(e) => { selectGoalHandler(e); }}
```
xnick421: same here in regards to `event`

src/course-home/outline-tab/widgets/UpdateGoalSelector.jsx
xnick421: same here in regards to `event`

src/course-home/outline-tab/widgets/UpdateGoalSelector.jsx
```diff
@@ -0,0 +1,101 @@
...
+                    {!editingGoal && (
+                      <p className="m-0">{selectedGoal.text}</p>
+                    )}
+                    {editingGoal && (
+                      <Input
+                        id="edit-goal-selector"
+                        type="select"
+                        defaultValue={selectedGoal.key}
+                        onBlur={() => { setEditingGoal(false); }}
+                        onChange={(e) => { selectGoalHandler(e); }}
```
xnick421: same here in regards to `event`

src/toast/LearningToast.scss
```diff
@@ -1,3 +1,17 @@
+#learning-toast {
```
xnick421: how come we use id and not class?  id's shouldnt be rendered in the dom twice.  what would happen if we wanted to display two toasts at the same time?

xnick421 (changes requested):

src/toast/LearningToast.scss
ciduarte: Per the design spec we shouldn't be displaying two toasts at once. But that's a good thought, maybe this would fare better as a class..

This will all go away though once we apply styling to the Toast Paragon component!

xnick421 (approved):

src/course-home/data/__factories__/outlineTabData.factory.js
```diff
@@ -17,6 +17,10 @@ Factory.define('outlineTabData')
       blocks: courseBlocks.blocks,
     };
   })
+  .attr('course_goals', [], () => ({
+    goal_options: [],
+    selected_goal: {},
+  }))
```
ciduarte, line 23: I left these fields empty because I'm not sure how we test different scenarios (verified vs audit, enrolled vs unenrolled)

src/toast/LearningToast.jsx
```diff
@@ -14,36 +15,37 @@ export default function LearningToast({
...
-        {/* eslint-disable-next-line react/no-danger */}
-        <div dangerouslySetInnerHTML={{ __html: header }} />
+      <Toast.Header className="bg-gray-700 border-bottom-0 p-0 text-light justify-content-between align-items-center">
+        <p className="small m-0 mr-4">{header}</p>
       </Toast.Header>
-      <Toast.Body className="bg-gray-700 text-light">
-        {/* eslint-disable-next-line react/no-danger */}
-        <div dangerouslySetInnerHTML={{ __html: body }} />
-      </Toast.Body>
+      {bodyLink && bodyText && (
```
mikix, line 30: This means that every toast has to have a link. Can we make that optional?

src/course-home/data/slice.js
```diff
@@ -10,7 +10,8 @@ const slice = createSlice({
   initialState: {
     courseStatus: 'loading',
     courseId: null,
-    resetDatesToastBody: null,
+    resetDatesToastBodyText: null,
+    resetDatesToastBodyLink: null,
```
mikix: We should also rename this stuff to just `toastBody*` and drop the `resetDates` prefix.

src/course-home/outline-tab/messages.js
```diff
@@ -9,6 +9,23 @@ const messages = defineMessages({
...
+    id: 'learning.outline.goal',
+    defaultMessage: 'Goal',
+  },
+  goalUnsure: {
+    id: 'learning.outline.goalUnsure',
+    defaultMessage: 'Not sure yet',
+  },
+  goalWelcome: {
+    id: 'learning.outline.goalWelcome',
+    defaultMessage: 'Welcome to',
```
mikix, line 28: This might be worth adding a `description` for the translators' benefit. Like "this goes before a course name" or something. So they know the context is a course, not like a town.

Maybe same for the couple messages above too -- Soccer goal or life goal? Not sure about what? Not every message string needs a translator description. But sometimes context is not obvious.

src/tab-page/TabPage.jsx
```diff
@@ -37,10 +38,11 @@ function TabPage({
...
       <>
         <LearningToast
-          body={resetDatesToastBody}
+          bodyLink={resetDatesToastBodyLink}
+          bodyText={resetDatesToastBodyText}
           header={resetDatesToastHeader}
-          onClose={() => dispatch(setResetDatesToast({ body: null, header: null }))}
-          show={!!(resetDatesToastBody && resetDatesToastHeader)}
+          onClose={() => dispatch(setResetDatesToast({ header: null, link: null, link_text: null }))}
+          show={!!(resetDatesToastBodyLink && resetDatesToastBodyText && resetDatesToastHeader)}
```
mikix: Could this just be `!!resetDatesToastHeader`? The component has the link and text being optional.

src/course-home/data/api.js
```diff
@@ -96,6 +98,11 @@ export async function postCourseDeadlines(courseId) {
   return getAuthenticatedHttpClient().post(url.href, { course_key: courseId });
 }
 
+export async function postCourseGoals(courseId, goalKey) {
+  const url = new URL(`${getConfig().LMS_BASE_URL}/api/course_home/v1/save_course_goal`);
```
mikix, line 102: This probably needs to change if you're switching endpoints on backend, but you know that already.

src/course-home/data/redux.test.js
```diff
@@ -106,6 +106,18 @@ describe('Data layer integration tests', () => {
     });
   });
 
+  describe('Test saveCourseGoal', () => {
```
mikix, line 109: Nice.

src/course-home/outline-tab/widgets/CourseGoalCard.jsx
```diff
@@ -0,0 +1,90 @@
...
+  return (
+    <Card className="mb-3">
+      <Card.Body>
+        <div className="row w-100 m-0 justify-content-between align-items-center">
+          <div className="col col-8 p-0">
+            <Card.Title className="h6 m-0">{intl.formatMessage(messages.goalWelcome)}&nbsp;{title}</Card.Title>
+          </div>
+          <div className="col col-auto p-0">
+            <Button
+              variant="sm btn-link btn-block p-0"
```
mikix: This looks at first glance like a misuse of `variant`? Isn't it supposed to be like `link` or something and then the rest of these are `className` content?

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -71,6 +81,11 @@ function OutlineTab({ intl }) {
           ...logistrationAlert,
         }}
       />
+      <LearningToast
```
mikix, line 84: This is separate from the reset deadlines toast? Ah, OK. That's easier than having just one at the TabPage level? I'd still like to rename that one to be generic though, since it's actually a generic CallToAction toast, not just a reset deadlines toast.

src/course-home/outline-tab/widgets/UpdateGoalSelector.jsx
```diff
@@ -0,0 +1,105 @@
...
+  courseId,
+  goalOptions,
+  intl,
+  selectedGoal,
+  setGoalToDisplay,
+  setGoalToastHeader,
+}) {
+  const [editingGoal, setEditingGoal] = useState(false);
+
+  function selectGoalHandler(event) {
```
mikix, line 22: Ugh. This is not on you and you shouldn't change anything based on this comment. But I much prefer class components. I don't like React's push for function components. Like, you're using a hook here when we could just have class state (which feels more natural to this backend developer). And functions in functions... is fine I guess. But come on. Dang kids on my lawn.

mikix (commented):
Screenshots look pretty

src/course-home/outline-tab/widgets/CourseGoalCard.jsx
ciduarte: Yeah the class names in variant are very strange. Some of them need the `btn-` prefix and some of them don't... I went through and removed all the `btn-` prefixes and learned this the hard way

src/toast/LearningToast.jsx
ciduarte: I guess the naming is misleading here, `bodyText` should only exist if there is a `bodyLink` because it'll go into a button. The actual success message is used in the header. ie. The entire `<Toast.Body />` is optional.

Do you think this would do better as `link` and `linkText`?

(Also, whatever choice we make here will most likely be changed by the Toast Paragon component that I'm creating)

src/course-home/outline-tab/widgets/CourseGoalCard.jsx
ciduarte: They worked! Probably just needed your blessing

src/toast/LearningToast.jsx
mikix: Ah got it. :+1:

src/course-home/outline-tab/widgets/CourseGoalCard.jsx
```diff
@@ -0,0 +1,94 @@
...
+      setGoalToastHeader(header);
+    });
+  }
+
+  return (
+    <Card className="mb-3">
+      <Card.Body>
+        <div className="row w-100 m-0 justify-content-between align-items-center">
+          <div className="col col-8 p-0">
+            <Card.Title className="h6 m-0">{intl.formatMessage(messages.goalWelcome)}&nbsp;{title}</Card.Title>
```
mikix: This is fine, who cares. But why do you care so much about not breaking in between `to` and the first word of the course title?

mikix (approved):

src/course-home/outline-tab/widgets/CourseGoalCard.jsx
ciduarte: Ooh I have a habit of sprinkling `&nbsp;` wherever I go. This was me thinking if I put a space between `{} {}` that it wouldn't be preserved in the UI, but that's not true...

PR #193, opened 2020-08-21
Use ‘modal-lg’ for LTI modals.  Update Paragon.

davidjoy:
The new version of Paragon includes a “dialogClassName” property on Modal which lets us set the modal width to ‘modal-lg’ or ‘modal-sm’ - we’re using the former here.

LTI modal sizing 2020-08-21 15_41_11

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1443,9 +1443,9 @@
       }
     },
     "@edx/paragon": {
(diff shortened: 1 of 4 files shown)


ciduarte (approved):
🎉

PR #195, opened 2020-08-24
AA-326: hide dates banner if the upgrade link is null

mikix:
This will prevent a pointless & broken button when the upgrade deadline has passed for the course.


jmyatt (approved):
LGTM

PR #197, opened 2020-08-28
AA-196 course celebration cert

xnick421:
Screen Shot 2020-09-04 at 4 40 42 PM


src/instructor-toolbar/InstructorToolbar.jsx
```diff
@@ -36,6 +36,12 @@ function getStudioUrl(courseId, unitId) {
 }
 
 export default function InstructorToolbar(props) {
+  const [didMount, setDidMount] = useState(false);
+  useEffect(() => {
+    setDidMount(true);
+    return () => setDidMount(false);
+  }, []);
+
```
Dillon-Dumesnil, line 44: ?

src/index.jsx
```diff
@@ -49,6 +51,11 @@ subscribe(APP_READY, () => {
               <ProgressTab />
             </TabContainer>
           </Route>
+          <Route path="/course/:courseId/course-completion">
+            <TabContainer tab="outline" fetch={fetchCourse} isCompletion>
```
Dillon-Dumesnil: isCompletion? also tab="outline" seems wrong.

src/instructor-toolbar/masquerade-widget/MasqueradeWidget.jsx
```diff
@@ -16,6 +16,8 @@ import {
 import messages from './messages';
 
 class MasqueradeWidget extends Component {
+  isMounted = false;
```
Dillon-Dumesnil, line 19: ?

src/tab-page/TabContainer.jsx
```diff
@@ -21,10 +22,11 @@ export default function TabContainer(props) {
 
   // The courseId from the store is the course we HAVE loaded.  If the URL changes,
   // we don't want the application to adjust to it until it has actually loaded the new data.
+  const slice = isCompletion ? 'courseware' : 'courseHome';
```
Dillon-Dumesnil: This feels very specific. We were fine before without this. Does it have to come from here and not a lower level that already knows to use the courseware slice?

Dillon-Dumesnil (commented):
Leaving out style comments until this becomes a Real(TM) PR, but here's a first pass

src/course-home/course-completion/CourseCompletion.jsx
```diff
@@ -0,0 +1,118 @@
...
+    return <Redirect to={`/course/${courseIdFromUrl}`} />;
+  }
+
+  const { username } = getAuthenticatedUser();
+
+  const dashboardLink = (
+    <Hyperlink destination={`${getConfig().LMS_BASE_URL}/dashboard`}>
+      <FormattedMessage
+        id="courseCelebration.dashboardLink"
+        defaultMessage="Dashboard"
```
ciduarte: lowercase d

src/course-home/course-completion/CourseCompletion.jsx
```diff
@@ -0,0 +1,118 @@
...
+  );
+
+  const handleDownload = () => {
+    global.location.replace(downloadUrl);
+  };
+
+  return (
+    <div className="d-flex flex-column">
+      <h2 className="align-self-center">{ intl.formatMessage(messages.congratulationsHeader) }</h2>
+      <h4 className="align-self-center mb-3">{ intl.formatMessage(messages.shareHeader) }</h4>
```
ciduarte: We shouldn't be using an `<h4>` here due to a11y concerns.  I'm not sure if we consider this a header, but if so it should be `<h3>`. This text also isn't styled how the mock suggests it should be.

src/course-home/course-completion/CourseCompletion.jsx
```diff
@@ -0,0 +1,118 @@
...
+      <div className="align-self-center mb-3"><SocialIcons courseId={courseIdFromUrl} /></div>
+      <div className="align-self-center mb-5"><img src={celebration} alt="" className="img-fluid" /></div>
```
ciduarte: This picture looks like it should be bigger per the design spec?

src/course-home/course-completion/CourseCompletion.jsx
```diff
@@ -0,0 +1,118 @@
...
+      { downloadUrl && (
+        <div className="align-self-center d-flex flex-row p-3 mb-3" style={{ backgroundColor: '#f4f7ff' }}>
+          <div className="d-flex flex-column justify-content-between">
+            <h4>{intl.formatMessage(messages.certificateHeader)}</h4>
```
ciduarte: Why is this an `<h4>`? If that's how it should be styled, we can use `className="h4"`. Also the bottom margin is too large here.

src/course-home/course-completion/CourseCompletion.jsx
```diff
@@ -0,0 +1,118 @@
...
+  };
+
+  return (
+    <div className="d-flex flex-column">
+      <h2 className="align-self-center">{ intl.formatMessage(messages.congratulationsHeader) }</h2>
+      <h4 className="align-self-center mb-3">{ intl.formatMessage(messages.shareHeader) }</h4>
+      <div className="align-self-center mb-3"><SocialIcons courseId={courseIdFromUrl} /></div>
+      <div className="align-self-center mb-5"><img src={celebration} alt="" className="img-fluid" /></div>
+      { downloadUrl && (
+        <div className="align-self-center d-flex flex-row p-3 mb-3" style={{ backgroundColor: '#f4f7ff' }}>
```
ciduarte: this padding should be bigger

src/course-home/course-completion/CourseCompletion.jsx
```diff
@@ -0,0 +1,118 @@
+import React from 'react';
+
+import { FormattedMessage, injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import { Redirect, useParams } from 'react-router-dom';
+import { Button, Hyperlink } from '@edx/paragon';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faCalendar } from '@fortawesome/free-regular-svg-icons';
```
ciduarte: This should be `faCalendarAlt` per the mockup

src/courseware/course/sequence/sequence-navigation/SequenceNavigation.jsx
```diff
@@ -60,12 +67,38 @@ export default function SequenceNavigation({
...
+                id="learn.sequence.navigation.endOfCourse.button"
+                description="The end of course button in the sequence nav"
+              />
+            </div>
+            <div>
+              <FormattedMessage
+                defaultMessage="(end of course)"
+                id="learn.sequence.navigation.endOfCourse.button"
+                description="The end of course button in the sequence nav"
+              />
```
ciduarte: should these have the same ID?

src/course-home/course-completion/messages.js
```diff
@@ -0,0 +1,36 @@
...
+    description: 'Congratulation header on course completion page',
+  },
+  downloadButton: {
+    id: 'courseCompletion.downloadButton',
+    defaultMessage: 'Download my certificate',
+    description: 'Text for download button in certificate banner',
+  },
+  linkedinButton: {
+    id: 'courseCompletion.linkedinButton',
+    defaultMessage: 'Share on Linkedin',
```
ciduarte: `LinkedIn` - capital I :)

src/course-home/course-completion/CourseCompletion.jsx
```diff
@@ -0,0 +1,118 @@
...
+  const { downloadUrl } = certificateData;
+
+  if (!courseCompletionIsActive || !userHasPassingGrade) {
+    return <Redirect to={`/course/${courseIdFromUrl}`} />;
+  }
+
+  const { username } = getAuthenticatedUser();
+
+  const dashboardLink = (
+    <Hyperlink destination={`${getConfig().LMS_BASE_URL}/dashboard`}>
```
ciduarte: I would double check the styling on these links. They're gray and underlined in the mockup, but blue in your screenshot.

src/course-home/course-completion/CourseCompletion.jsx
```diff
@@ -0,0 +1,118 @@
...
+        <FontAwesomeIcon icon={faCalendar} className="mr-2" style={{ width: '20px' }} />
+        <FormattedMessage
+          id="courseCelebration.dashboardInfo"
+          defaultMessage="You can always access this course and its materials on your {dashboardLink}."
+          description="Text letting the user know they can view their dashboard"
+          values={{
+            dashboardLink,
+          }}
+        />
+      </div>
```
ciduarte: This looks like `gray-500` in the mockup

ciduarte (commented):
Mostly styling comments, I'll come back to look at the actual behavior

Looks like the mockup has a thin outline around the entire congrats area... can we add that in?

src/instructor-toolbar/InstructorToolbar.jsx
xnick421: This is recommended via the react docs to prevent memory leaks (which i was getting warnings about in the console)

src/instructor-toolbar/masquerade-widget/MasqueradeWidget.jsx
xnick421: This is recommended via the react docs to prevent memory leaks (which i was getting warnings about in the console)

src/tab-page/TabContainer.jsx
xnick421: the celebration page is rendered inside this component.  its gets its data from the courseware slice.  by default everything else thats rendered in this component uses the coursehome slice which is why we were ok.  But we determined that even though the celebration page gets its data from the courseware slice, its still better to render it here inside this tabcontainer component, rather than the coursewarecontainer. This is because the coursewarecontainer component runs alot of unnecessary logic we dont need for the celebration page.

src/index.jsx
Dillon-Dumesnil: I would still bump on the `tab="outline"` being wrong. Maybe `tab="course-completion"`? or even leaving tab as null since technically, you're not on a tab.

src/courseware/course/sequence/sequence-navigation/SequenceNavigation.jsx
```diff
@@ -20,13 +21,19 @@ export default function SequenceNavigation({
...
+  goToCourseCompletion,
 }) {
   const sequence = useModel('sequences', sequenceId);
   const { isFirstUnit, isLastUnit } = useSequenceNavigationMetadata(sequenceId, unitId);
-  const sequenceStatus = useSelector(state => state.courseware.sequenceStatus);
+  const { sequenceStatus } = useSelector(state => state.courseware);
   const isLocked = sequenceStatus === LOADED ? (
     sequence.gatedContent !== undefined && sequence.gatedContent.gated
   ) : undefined;
+  const { courseId } = useSelector(state => state.courseware);
```
Dillon-Dumesnil: You have `useSelector(state => state.courseware);` twice. You can combine them

src/courseware/course/sequence/sequence-navigation/SequenceNavigation.jsx
```diff
@@ -12,6 +12,7 @@ import SequenceNavigationTabs from './SequenceNavigationTabs';
 import { useSequenceNavigationMetadata } from './hooks';
 import { useModel } from '../../../../generic/model-store';
 import { LOADED } from '../../../data/slice';
+// import { useSelector } from "react-redux";
```
Dillon-Dumesnil: Delete if not needed

src/course-home/course-completion/CourseCompletion.jsx
```diff
@@ -0,0 +1,118 @@
...
+
+function CourseCompletion({ intl }) {
+  const { courseId: courseIdFromUrl } = useParams();
+  const {
+    courseCompletionIsActive,
+    userHasPassingGrade,
+    certificateData,
+  } = useModel('courses', courseIdFromUrl);
+
+  const { downloadUrl } = certificateData;
```
Dillon-Dumesnil: If all you care about with certificateData is the download url, why not only return that from the backend rather than the whole certificateData?

src/course-home/course-completion/CourseCompletion.jsx
Dillon-Dumesnil: Also is there a way to avoid hard-coding a hex value for the background color? Use some bootstrap defined color that's close?

This applies to all the spots there is hard-coded styles. I believe the convention we want to follow is using bootstrap to get as close as possible. (i.e. the hex codes / exact px defined lower)

src/courseware/course/course-exit/messages.js
```diff
@@ -0,0 +1,76 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  certificateHeaderDownloadable: {
+    id: 'courseCelebration.certificateHeader.downloadable',
+    defaultMessage: 'Your certificate is available!',
+    description: 'Header in certificate banner',
```
ciduarte: nit: I recently learned that these descriptions are used by the people doing the translating for context. My concern is that "Header in certificate banner" is too vague, and we might end up with the wrong version of a translated word like "available."

I wonder if it would be better to say "Text displayed when course certificate is ready to be downloaded"

src/courseware/course/course-exit/messages.js
```diff
@@ -0,0 +1,76 @@
...
+  },
+  certificateHeaderNotAvailable: {
+    id: 'courseCelebration.certificateHeader.notAvailable',
+    defaultMessage: 'Your certificate will be available soon!',
+    description: 'Header in certificate banner',
```
ciduarte: Following the above comment "Text displayed when the certificate is not ready to be viewed"

src/courseware/course/course-exit/messages.js
```diff
@@ -0,0 +1,76 @@
...
+  },
+  certificateHeaderUnverified: {
+    id: 'courseCelebration.certificateHeader.unverified',
+    defaultMessage: 'You must complete verification to receive your certificate.',
+    description: 'Header in certificate banner',
```
ciduarte: "Text displayed when a user has not verified their identity and cannot view their course certificate"

src/courseware/course/course-exit/messages.js
```diff
@@ -0,0 +1,76 @@
...
+  },
+  certificateHeaderRequestable: {
+    id: 'courseCelebration.certificateHeader.requestable',
+    defaultMessage: 'Congratulations, you qualified for a certificate!',
+    description: 'Header in certificate banner',
```
ciduarte: "Text displayed when a user has completed the course and can request a certificate"

src/courseware/course/sequence/sequence-navigation/UnitNavigation.jsx
```diff
@@ -4,17 +4,87 @@ import { Button } from '@edx/paragon';
...
+      ? goToCourseExitPage : onClickNext);
+    // AA-198: The userHasPassingGrade condition can be removed once we have a view for learners with failing grades
+    const disabled = isLastUnit && (!courseExitPageIsActive || !userHasPassingGrade);
+
+    // This is just to support what used to show while we are getting courseExitPageIsActive turned on.
+    // This should be good to remove once disabled goes away.
+    if (disabled) {
+      return (
+        <div className="m-2">
+          <span role="img" aria-hidden="true">&#129303;</span> {/* This is a hugging face emoji */}
```
ciduarte, line 39: Interesting

src/courseware/course/sequence/sequence-navigation/SequenceNavigation.jsx
```diff
@@ -49,6 +57,52 @@ export default function SequenceNavigation({
...
+      ? goToCourseExitPage : nextSequenceHandler);
+    // AA-198: The userHasPassingGrade condition can be removed once we have a view for learners with failing grades
+    const disabled = isLastUnit && (!courseExitPageIsActive || !userHasPassingGrade);
+
+    let buttonText = (
+      <FormattedMessage
+        defaultMessage="Next"
+        id="learn.sequence.navigation.next.button"
+        description="The Next button in the sequence nav"
+      />
```
ciduarte, line 72: Why not put these (line 68 & line 76) in `messages.js`?

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -0,0 +1,181 @@
...
+      <Helmet>
+        <title>{`Congratulations! | ${getConfig().SITE_NAME}`}</title>
+      </Helmet>
+      <div className="align-items-center d-flex flex-column border border-light p-3 mb-3 mt-5">
+        <h2>{ intl.formatMessage(messages.congratulationsHeader) }</h2>
+        <div className="mb-3 h4 font-weight-normal">{ intl.formatMessage(messages.shareHeader) }</div>
+        <div className="mb-5"><img src={celebration} alt="" className="img-fluid" /></div>
+        <div className="d-flex flex-row p-4 mb-3 bg-primary-100">
+          <div className="d-flex flex-column justify-content-between">
+            <div className="h4 mb-0">{title}</div>
```
ciduarte, line 142: I'm skipping over all of this because I'm assuming I'll be fixing whatever needs fixing

ciduarte (approved):
LGTM, few nits and questions

src/courseware/course/course-exit/CourseCelebration.jsx
ciduarte: But I'm sure you did a great job 🙂

src/courseware/course/sequence/sequence-navigation/UnitNavigation.jsx
Dillon-Dumesnil: This was already present. I just moved it here as being part of the nextButton. I've spoken to Hilary about it and the plan is to remove it once we actually have an exit course experience for all learners (and not just passing certificate learners)

src/courseware/course/course-exit/CourseCelebration.jsx
Dillon-Dumesnil: I would say I made an improvement or two ;) but yes, you'll be fixing so smart to skip.

src/courseware/course/sequence/sequence-navigation/SequenceNavigation.jsx
Dillon-Dumesnil: We spoke about this offline and Carla will handle it as part of the improvements she makes in a follow-up PR. Thanks Carla!

PR #204, opened 2020-09-09
AA-320: Adding in UTM parameters to sharing links

Dillon-Dumesnil:


src/courseware/course/celebration/SocialIcons.jsx
```diff
@@ -41,11 +41,13 @@ function SocialIcons({ courseId, intl }) {
     });
   };
 
+  const socialUtmMarketingUrl = `${marketingUrl}?utm_campaign=edxmilestone&utm_medium=social`;
```
davidjoy, line 44: Lacking some context on this, but is this social marketing URL something appropriate for Open edX, somehow?  Seems like we're effectively hard-coding our own campaign parameters in the app...

Dillon-Dumesnil: Ooof. Yes. Good catch... Will work on being more clever about this.

ciduarte (approved):
👨‍🍳💋

src/courseware/course/celebration/SocialIcons.jsx
Dillon-Dumesnil: I have created https://openedx.atlassian.net/browse/AA-356 to handle this in the future. We have several other spots to fix up as well and I think it would best be tackled by a separate ticket and work.

In the meantime, I will merge this work in and it is documented in the above ticket.

PR #205, opened 2020-09-10
ENT-3181: Enables custom branding for enterprise customers in logo and menu

jlajoie:


src/course-header/Header.jsx
```diff
@@ -34,21 +34,46 @@ export default function Header({
...
   const { authenticatedUser } = useContext(AppContext);
 
-  const { enterpriseLearnerPortalLink } = useEnterpriseConfig(
+  const { enterpriseLearnerPortalLink, enterpriseCustomerBrandingConfig } = useEnterpriseConfig(
     authenticatedUser,
     getConfig().ENTERPRISE_LEARNER_PORTAL_HOSTNAME,
     getConfig().LMS_BASE_URL,
   );
 
+  let dashboardMenuItem = <Dropdown.Item href={`${getConfig().LMS_BASE_URL}/dashboard`}>Dashboard</Dropdown.Item>;
```
ciduarte: Should we i18n `Dashboard`?

ciduarte (approved):
Looks good! 😁 one comment about i18n

src/course-header/Header.jsx
jlajoie: Your call, just moved that from elsewhere in this file 😄

ciduarte: Yeah let's throw that into the nearest messages and `intl` that sucker

PR #207, opened 2020-09-14
AA-265: Add more tests for the dates tab

mikix:
Add an ADR to talk about how we want to test in this repo.

And refactor the fake dates tab data used for debugging to also be used for tests.


docs/decisions/0007-testing.md
```diff
@@ -0,0 +1,41 @@
...
+This has some implications about how to test. You can read the React Testing Library's
+[Guiding Principles](https://testing-library.com/docs/guiding-principles), but the main
+takeaway is that you should be interacting with React as closely as possible to the way
+the user will interact with it.
+
+For example, they discourage using class or element name selectors to find components
+during a test. Instead, you should find them by user-oriented attributes like labels,
+text, or roles. As a last resort, by a `data-testid` tag.
+
+## What to Test
```
jmyatt, line 24: 👍  this seems like reasonable guidance to start with

src/course-home/data/__snapshots__/redux.test.js.snap
```diff
@@ -78,15 +59,209 @@ Object {
...
+            "date": "2020-05-01T17:59:41Z",
             "dateType": "course-start-date",
             "description": "",
-            "extraInfo": "",
+            "extraInfo": null,
             "learnerHasAccess": true,
             "link": "",
             "title": "Course Starts",
           },
+          Object {
```
davidjoy, line 70: One thing I was finding a bit weird (when writing my own tests, and I see a bit of it here too, I think) is when we do snapshot tests of generated data that isn’t really used in the code under test… it’s like testing that our factories produce the stuff we think we coded them to produce.  As an example, I’m sorta guessing that the “courseDateBlocks” array here isn't terribly useful as part of a snapshot.  There’s so many I think it’d be hard for a human to tell if they’re “right”, you know? 

Would it be better to make specific assertions about the store in the test and dispense with this snapshot?

src/course-home/dates-tab/Badge.jsx
```diff
@@ -4,7 +4,10 @@ import classNames from 'classnames';
 
 export default function Badge({ children, className }) {
   return (
-    <span className={classNames('dates-badge badge align-text-bottom font-italic ml-2 px-2 py-1', className)}>
+    <span
+      className={classNames('dates-badge badge align-text-bottom font-italic ml-2 px-2 py-1', className)}
+      data-testid="dates-badge"
```
davidjoy, line 9: I don't feel strongly about it, but it feels weird to me to add attributes to our code specifically for testing... seems like a weird thing that testing library asks us to do. 🤔

davidjoy (approved):

src/course-home/dates-tab/Badge.jsx
mikix: React Testing Library's response is along the lines of "what else are you going to do? use a class name? here's several reasons why that sucks".

I don't have a strong opinion either, but I understand their concerns and figured I'd just buy-in to the RTL way.

src/course-home/data/__snapshots__/redux.test.js.snap
mikix: Yeah. Specific assertions are good. These snapshots end up being "test our normalization" tests which are kind of a pain to do with specific assertions (easy to miss when we start using new data from backend and not update the tests for it).

So I personally still like the wholesale snapshot, but I recognize that it's value is small (just not zero, I'm saying). Maybe specific assertions in addition, for non-normalization tests.

src/course-home/dates-tab/Badge.jsx
mikix: These tests in particular were difficult to test the way RTL really wants, because it's just a bunch of repeating non-interactive text generated from data. So like, can't really search by role or label (we do, but then find nearby components from that).

src/course-home/data/__factories__/datesTabData.factory.js
```diff
@@ -9,19 +12,213 @@ Factory.define('datesTabData')
     },
     course_date_blocks: [
       {
-        assigment_type: 'Homework',
-        date: '2013-02-05T05:00:00Z',
+        date: '2020-05-01T17:59:41Z',
```
Dillon-Dumesnil, line 15: ?

src/course-home/dates-tab/DatesTab.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+      </AppProvider>
+    );
+
+    axiosMock = new MockAdapter(getAuthenticatedHttpClient());
+
+    const courseMetadata = Factory.build('courseHomeMetadata');
+    courseId = courseMetadata.courseId;
+    axiosMock.onGet(`${getConfig().LMS_BASE_URL}/api/course_home/v1/course_metadata/${courseId}`).reply(200, courseMetadata);
+
+    history.push(`/course/${courseId}/dates`); // so tab can pull course id from url
```
Dillon-Dumesnil, line 47: Thanks for the comment 👍

src/course-home/dates-tab/DatesTab.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  describe('when receiving a full set of dates data', () => {
+    beforeEach(() => {
+      const datesTabData = Factory.build('datesTabData');
+      axiosMock.onGet(`${getConfig().LMS_BASE_URL}/api/course_home/v1/dates/${courseId}`).reply(200, datesTabData);
+      render(component);
+    });
+
+    it('handles unreleased & complete', async () => {
```
Dillon-Dumesnil, line 79: Interesting that these are use cases that can happen. Content that is both unreleased to the learner and complete or unreleased and past due.

Or is this related to your earlier comment about how the dates tab factory will create unrealistic situations for the purpose of showing all the pills?

src/course-home/dates-tab/DatesTab.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+    it('handles verified only', async () => {
+      const { day } = await getDay('Sun, Aug 18, 2030');
+      const badge = within(day).getByTestId('dates-badge');
+      expect(badge).toHaveTextContent('Verified Only');
+    });
+
+    it('verified only has no link', async () => {
+      const { day } = await getDay('Sun, Aug 18, 2030');
+      expect(within(day).queryByRole('link')).toBeNull();
+    });
```
Dillon-Dumesnil, line 104: Why separate these tests?
Or perhaps they should live under their own describe?

src/course-home/dates-tab/DatesTab.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+    const courseMetadata = Factory.build('courseHomeMetadata');
+    courseId = courseMetadata.courseId;
+    axiosMock.onGet(`${getConfig().LMS_BASE_URL}/api/course_home/v1/course_metadata/${courseId}`).reply(200, courseMetadata);
+
+    history.push(`/course/${courseId}/dates`); // so tab can pull course id from url
+  });
+
+  // The dates tab is largely repetitive non-interactive static data. Thus it's a little tough to follow
+  // testing-library's advice around testing the way your user uses the site (i.e. can't find form elements by label or
+  // anything). Instead, we find elements by printed date (which is what the user sees) and data-testid. Which is
```
Dillon-Dumesnil, line 52: I like the use of printed date so if that ever inadvertently changes, this test suite will catch them.

src/course-home/dates-tab/DatesTab.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+    it('shows extra info', async () => {
+      const { items } = await getDay('Sat, Aug 17, 2030');
+      expect(items).toHaveLength(3);
+
+      const tipIcon = within(items[2]).getByTestId('dates-extra-info');
+      const tipText = "ORA Dates are set by the instructor, and can't be changed";
+
+      expect(screen.queryByText(tipText)).toBeNull(); // tooltip does not start in DOM
+      userEvent.hover(tipIcon);
+      const tooltip = screen.getByText(tipText); // now it's there
```
Dillon-Dumesnil, line 131: Would this error if it wasn't there? Might be my lack of JS knowledge, but I could see this potentially just returning null

docs/decisions/0007-testing.md
```diff
@@ -0,0 +1,41 @@
+# Testing
```
Dillon-Dumesnil, line 1: I think this is good in general. Can we add some more details to it? Maybe I'm being too picky so feel free to tell me no to all of this, but thoughts on:

1. Saying where tests should live? like the redux/API tests live in data/redux.test.js vs component tests live alongside their component. Maybe that's obvious from the code?
2. When factories should be updated and also where they live? Also calling out that we use factories and specifically Rosie for that. Could add a link to that so people can see how to properly do it.
3. Maybe it's just me, but I feel like it would also be helpful to point to some of the different types of tests. i.e. Example of API test is ___ in ___.test.js, example of component test is ___ in ___.test.js, example of snapshot test is ___ in ___.test.js.

src/course-home/data/__factories__/datesTabData.factory.js
mikix: This swap out was part of just updating the mock we had from the (now removed in this PR) fakeData.js file I had for local testing. I wasn't like "2013 sucks" as much as just copying a bunch of dates I had in a different file.

src/course-home/dates-tab/DatesTab.test.jsx
mikix: More the unrealistic thing. I was leaning on my previous work coming up with a bunch of scenarios to manually test in fakeData.js and just encoded the unusual cases here to be done automatically.

Though I guess maybe it could happen if the course team changes the date on when a hw is released after a learner completes it? Seems unlikely.

src/course-home/dates-tab/DatesTab.test.jsx
mikix: One is testing the badge, one is testing the link logic, which felt like two different unit tests to my brain.

src/course-home/dates-tab/DatesTab.test.jsx
mikix: It will error. See https://testing-library.com/docs/react-testing-library/cheatsheet

`queryByText` will return null if it's missing. `getByText` will error out if there's 0 or more than 1.

docs/decisions/0007-testing.md
mikix: Sorry I merged before this feedback. Could be added in a future test addition.

src/course-home/dates-tab/DatesTab.test.jsx
Dillon-Dumesnil: Fair. I suppose I would have offered a more descriptive name for the first test in that case. `handles verified only` felt generic enough to be covering all of the verified only which is why I thought they could be combined. Not a big deal.

src/course-home/dates-tab/DatesTab.test.jsx
Dillon-Dumesnil: Awesome, thank you!

PR #212, opened 2020-09-16
AA-368: Expand course outline section that contains resume block

ciduarte:


src/course-home/outline-tab/Section.jsx
```diff
@@ -27,12 +28,16 @@ function Section({
     },
   } = useModel('outline', courseId);
 
-  const [open, setOpen] = useState(expand);
+  const [open, setOpen] = useState(defaultOpen);
```
Dillon-Dumesnil, line 31: Is the change here going to have an effect on the expandAll button?

Dillon-Dumesnil (approved):
One question about the Expand All functionality with this change, but subject to your answer there, looks good.

src/course-home/outline-tab/Section.jsx
ciduarte: No it won't. The `useEffect` hook a few lines down will watch for changes in `expand` and behave as expected.

PR #213, opened 2020-09-16
Refer to 'Existing experience' as 'Legacy experience'

kdmccormick:
The new experience is the most common case on edx.org, so "existing" is starting to get misleading.

Also, we decided in standup that "Legacy" is a good term to standardize on for these pre-MFE experiences.

### DIFF
--- a/src/instructor-toolbar/InstructorToolbar.jsx
+++ b/src/instructor-toolbar/InstructorToolbar.jsx
@@ -66,7 +66,7 @@ export default function InstructorToolbar(props) {
           )}
           {urlLms && (
             <span className="mx-1 my-1">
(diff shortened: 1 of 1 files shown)


davidjoy (approved):

marcotuts:
One suggested scope change here - or perhaps a separate PR if easier - defaulting instructors to the new experience versus the legacy one is I think a step we still need to take / process. I believe only learners default to the new. is that right?

davidjoy:
That sounds correct - staff are excluded from any redirects (meaning they stay in whatever experience they're in), and the code otherwise 'defaults' to the legacy experience and redirects students to the new one.

kdmccormick:
@marcotuts @davidjoy Makes sense. That would be an edx-platform PR; are you good with this merging now Marco?

PR #222, opened 2020-09-25
AA-363: Tests for Outline Tab and Widgets

ciduarte:


src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -1,107 +1,385 @@
...
-      const staffMessage = 'You are viewing this course as staff, and are not enrolled.';
+      let extraText;
+      let alertMessage;
+      let staffMessage;
+
+      beforeEach(() => {
+        extraText = outlineTabData.enroll_alert.extra_text;
+        alertMessage = `You must be enrolled in the course to see course content. ${extraText}`;
+        staffMessage = 'You are viewing this course as staff, and are not enrolled.';
+      });
```
ciduarte, line 307: This ticket excludes alert tests. There's another ticket to handle these

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -1,107 +1,385 @@
...
+        });
+        axiosMock.onGet(outlineUrl).reply(200, outlineTabDataLongMessage);
+        await executeThunk(thunks.fetchOutlineTab(courseId), store.dispatch);
+
+        render(<OutlineTab />, { store });
+      });
 
-    expect(screen.getByRole('link', { name: 'Resume Course' })).toBeInTheDocument();
+      it('shortens message', async () => {
+        expect(screen.getByTestId('short-welcome-message-iframe')).toBeInTheDocument();
```
mikix, line 148: Maybe also look for the absence of a show more/less button?

src/course-home/data/__factories__/courseBlocks.factory.js
```diff
@@ -64,7 +64,7 @@ export default function buildSimpleCourseBlocks(courseId, title, options = {}) {
     { type: 'chapter', children: sequenceBlock.map(block => block.id), resume_block: false },
     { courseId },
   );
-  const courseBlock = options.courseBlocks || Factory.build(
+  const courseBlock = options.courseBlock || Factory.build(
```
Dillon-Dumesnil, line 73: Nice catch

src/course-home/outline-tab/DateSummary.jsx
```diff
@@ -35,11 +35,10 @@ export default function DateSummary({
           && <div className="font-weight-bold mt-2">{dateBlock.title}</div>}
         </div>
         {dateBlock.description
-          && <div className="date-summary-text m-0 mt-1">{dateBlock.description}</div>}
+          && <div className="date-summary-text mt-1">{dateBlock.description}</div>}
```
Dillon-Dumesnil, line 38: :chefkiss:

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -1,107 +1,385 @@
...
   const outlineUrl = new RegExp(`${getConfig().LMS_BASE_URL}/api/course_home/v1/outline/*`);
   const courseMetadataUrl = new RegExp(`${getConfig().LMS_BASE_URL}/api/course_home/v1/course_metadata/*`);
 
   beforeEach(async () => {
-    store = await initializeTestStore({ excludeFetchCourse: true, excludeFetchSequence: true, courseMetadata });
+    store = initializeStore();
+
+    courseMetadata = Factory.build('courseHomeMetadata');
+    courseId = courseMetadata.courseId;
+    outlineTabData = Factory.build('outlineTabData');
```
Dillon-Dumesnil: Why switch these to the beforeEach vs the describe? Just for my own curiosity.

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -1,107 +1,385 @@
...
-    axiosMock.onGet(courseMetadataUrl).reply(200, courseHomeMetadata);
     logUnhandledRequests(axiosMock);
-    await executeThunk(thunks.fetchOutlineTab(courseMetadata.id), store.dispatch);
+    await executeThunk(thunks.fetchOutlineTab(courseId), store.dispatch);
   });
 
-  it('displays link to start course', () => {
-    render(<OutlineTab />);
-    expect(screen.getByRole('link', { name: 'Start Course' })).toBeInTheDocument();
+  function buildCourseBlocksWithCompletion(isComplete, courseTitle) {
```
Dillon-Dumesnil: This kind of feels like it would live in the factory where you can pass in if you want completion data or not. I would also assume we would always want to supply a default since that is a more natural mock to what we send from the backend.

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -1,107 +1,385 @@
...
+      const outlineTabDataResumeBlock = Factory.build('outlineTabData', {
+        courseId,
+        course_blocks: { blocks: courseBlocks.blocks },
+      });
+      axiosMock.onGet(outlineUrl).reply(200, outlineTabDataResumeBlock);
+      await executeThunk(thunks.fetchOutlineTab(courseId), store.dispatch);
+
+      render(<OutlineTab />, { store });
+      const collapseNode = screen.getByRole('button', { name: /Title of Section/ });
+      expect(collapseNode).toHaveAttribute('aria-expanded', 'true');
```
Dillon-Dumesnil: Nit: `collapseNode` -> `expandedNode`? Something that makes it more clear what its state is.

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -1,107 +1,385 @@
...
+    });
+
+    it('handles expand/collapse all button click', () => {
```
Dillon-Dumesnil, line 79: Great test and very clear documenting. Thanks!

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -1,107 +1,385 @@
...
+        userEvent.click(showLessButton);
+        showLessButton = screen.queryByRole('button', { name: 'Show Less' });
+        expect(showLessButton).not.toBeInTheDocument();
+        showMoreButton = screen.getByRole('button', { name: 'Show More' });
+        expect(showMoreButton).toBeInTheDocument();
+      });
+    });
+
+    it('does not display if no update available', async () => {
+      const outlineTabDataSansUpdate = Factory.build('outlineTabData', {
```
Dillon-Dumesnil, line 172: Sans. Nice

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -1,107 +1,385 @@
...
+    let goalUrl;
+    const goalOptions = [
+      ['certify', 'Earn a certificate'],
+      ['complete', 'Complete the course'],
+      ['explore', 'Explore the course'],
+      ['unsure', 'Not sure yet'],
+    ];
+
+    beforeEach(() => {
+      goalUrl = new RegExp(`${getConfig().LMS_BASE_URL}/api/course_home/v1/save_course_goal`);
```
Dillon-Dumesnil: This is a constant, right? Why redefine before each test?

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -1,107 +1,385 @@
...
+        act(() => {
+          userEvent.click(dropdownButtonNode);
+        });
+
+        // Select a new goal
+        const unsureButtonNode = screen.getByRole('button', { name: 'Not sure yet' });
+        expect(unsureButtonNode).toBeInTheDocument();
+        act(() => {
+          userEvent.click(unsureButtonNode);
+        });
```
Dillon-Dumesnil: Curious. Difference between of `act(() => userEvent.click...` vs `userEvent.click`?

Dillon-Dumesnil (commented):
Couple comments and questions, but overall looking great. Super excited to have all these tests!

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -1,107 +1,385 @@
...
+
+        // Verify the request was made
+        const mockPost = await axiosMock.history.post;
```
mikix: huh ok

mikix (approved):

src/course-home/outline-tab/OutlineTab.test.jsx
ciduarte: Part of my answer is that I believe this might prevent mocked data/behaviors leaking from one test to another... (the other part of my answer is because MT did it this way in the DatesTab tests and I'd like parity)

Dillon-Dumesnil: Would you mind trying to move it back into the describe and see if it has any effects on testing time? I wonder if this has any impacts on the performance of tests. 
Good point about it being good to refresh that if we were seeing leaking. Hopefully we aren't modifying that exact object, but maybe we are.

mikix: I don't recall why I did that tbh.

src/course-home/outline-tab/OutlineTab.test.jsx
ciduarte: Good point

src/course-home/outline-tab/OutlineTab.test.jsx
ciduarte: Yeah you're right I could just pass completion/resume variables into the options parameter `buildSimpleCourseBlocks` function. In my first pass through I wanted to leave that function _simple_ as the name suggests, but including those defaults makes sense.

src/course-home/outline-tab/OutlineTab.test.jsx
ciduarte: So `act(() => {});` is a wrapper for any event that triggers a state change in a component. I had sprinkled these around in an effort to solve a warning about needing to use `act` but it didn't work.

I'm glad you said this because I went back and realized this warning was being triggered by async calls, and the right way to handle this is to wrap the expect calls in an `await waitFor(() => { expect(); });` (documenting for future me who will probably need to hear this again: https://medium.com/@davidwcai/react-testing-library-and-the-not-wrapped-in-act-errors-491a5629193b)

I'm going to remove the `act` calls, and switch the ones that would need them to `fireEvent.click`. `fireEvent` is an alternative to `userEvent` that integrates `act` for you

src/course-home/outline-tab/OutlineTab.test.jsx
ciduarte: @Dillon-Dumesnil you crushed it 😱 
before: 2s 279ms
after: 1s 936 ms

src/course-home/outline-tab/OutlineTab.test.jsx
ciduarte: realized I can remove this line and use `await waitFor(() => {});` around the assertions

PR #228, opened 2020-09-30
Log unexpected course structure as info, not error

stvstnfrd:


davidjoy:
There's one more instance of this in course-home. :) I'd change that too.

davidjoy (approved):

PR #234, opened 2020-10-06
AA-389: Updating Routing for CourseExit Page

Dillon-Dumesnil:
The CourseExit page uses the TabContainer because it seems to be a better solution
than adding it into the paths for the CoursewareContainer, despite the CoursewareContainer
already doing the correct fetch. The reason for this is because within the
CoursewareContainer, it would still be necessary to check for the /course-exit path (due
to the CoursewareContainer trying to greedily match sequenceId (read: it will try and
look up 'course-exit' as a sequence)). That effectively defeats the purpose of using the
routing in the first place so instead, we place it in a TabContainer.

The InstructorToolbar didMount logic became necessary once we had a page (CourseExit) that does a redirect on a quick exit.
As a result, it unmouunts the InstructorToolbar (which will be remounted by the new component),
but the InstructorToolbar's MasqueradeWidget has an outgoing request. Since it is unmounted
during that time, it raises an error about a potential memory leak. By stopping the render
when the InstructorToolbar is unmounted, we avoid the memory leak.

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -52,7 +52,7 @@ subscribe(APP_READY, () => {
             </TabContainer>
           </Route>
           <Route path="/course/:courseId/course-exit">
-            <TabContainer tab="exit" fetch={fetchCourse} isExitPage>
+            <TabContainer tab="courseware" fetch={fetchCourse} slice="courseware">
               <CourseExit />
             </TabContainer>
           </Route>
--- a/src/instructor-toolbar/InstructorToolbar.jsx
+++ b/src/instructor-toolbar/InstructorToolbar.jsx
@@ -36,11 +36,19 @@ function getStudioUrl(courseId, unitId) {
 }
 
 export default function InstructorToolbar(props) {
+  // This didMount logic became necessary once we had a page that does a redirect on a quick exit.
+  // As a result, it unmounts the InstructorToolbar (which will be remounted by the new component),
+  // but the InstructorToolbar's MasqueradeWidget has an outgoing request. Since it is unmounted
+  // during that time, it raises an error about a potential memory leak. By stopping the render
+  // when the InstructorToolbar is unmounted, we avoid the memory leak.
+  // NOTE: This was originally added because of the CourseExit page redirect. Once that page stops
+  //   doing a redirect because a CourseExit experience exists for all learners, this could be removed
   const [didMount, setDidMount] = useState(false);
   useEffect(() => {
     setDidMount(true);
+    // Returning this function here will run setDidMount(false) when this component is unmounted
     return () => setDidMount(false);
-  }, []);
+  });
 
   const {
     courseId,
@@ -58,10 +66,7 @@ export default function InstructorToolbar(props) {
   const urlStudio = getStudioUrl(courseId, unitId);
   const [masqueradeErrorMessage, showMasqueradeError] = useState(null);
 
-  if (!didMount) {
-    return null;
-  }
-  return (
+  return (!didMount ? null : (
     <div>
       <div className="bg-primary text-white">
         <div className="container-fluid py-3 d-md-flex justify-content-end align-items-start">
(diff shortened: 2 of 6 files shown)


src/index.jsx
```diff
@@ -51,8 +51,17 @@ subscribe(APP_READY, () => {
...
           </Route>
+          {/*
+            The CourseExit page uses the TabContainer because it seems to be a better solution
+            than adding it into the paths for the CoursewareContainer, despite the CoursewareContainer
+            already doing the correct fetch. The reason for this is because within the
+            CoursewareContainer, it would still be necessary to check for the /course-exit path (due
+            to the CoursewareContainer trying to greedily match sequenceId (read: it will try and
+            look up 'course-exit' as a sequence)). That effectively defeats the purpose of using the
+            routing in the first place so instead, we place it in a TabContainer.
+          */}
```
Dillon-Dumesnil: Is this too much?

src/index.jsx
```diff
@@ -51,8 +51,17 @@ subscribe(APP_READY, () => {
...
           <Route path="/course/:courseId/course-exit">
-            <TabContainer tab="exit" fetch={fetchCourse} isExitPage>
+            <TabContainer tab="courseware" fetch={fetchCourse} slice="courseware">
```
Dillon-Dumesnil, line 55: Realized tab should be "courseware" so it appropriately underlines the "Course" tab.

src/instructor-toolbar/masquerade-widget/MasqueradeWidget.jsx
```diff
@@ -16,8 +16,6 @@ import {
 import messages from './messages';
 
 class MasqueradeWidget extends Component {
-  isMounted = false;
```
Dillon-Dumesnil, line 19: Doesn't seem like any of the isMounted logic in this file is actually necessary as long as the InstructorToolbar is appropriately not trying to render once its been unmounted

src/instructor-toolbar/InstructorToolbar.jsx
```diff
@@ -36,11 +36,17 @@ function getStudioUrl(courseId, unitId) {
 }
 
 export default function InstructorToolbar(props) {
+  // This didMount logic became necessary once we had a page that does a redirect on a quick exit.
+  // As a result, it unmouunts the InstructorToolbar (which will be remounted by the new component),
+  // but the InstructorToolbar's MasqueradeWidget has an outgoing request. Since it is unmounted
+  // during that time, it raises an error about a potential memory leak. By stopping the render
+  // when the InstructorToolbar is unmounted, we avoid the memory leak.
```
Dillon-Dumesnil: Also this, too much?

src/index.jsx
mikix: This feels like a lot of detail for the main src/index.js file. The routing doesn't look weird or complicated on its face, so it might not deserve such explanation.

Could some of this detail be moved to the CourseExit page where you talk about the ismount stuff?

src/instructor-toolbar/InstructorToolbar.jsx
mikix: This is probably appropriate, just because the pattern is new to folks like us, so it might be confusing to other non-react-pros.

You have a typo in `unmouunts`

src/instructor-toolbar/masquerade-widget/MasqueradeWidget.jsx
mikix: Nice. Way cleaner without this stuff.

mikix (approved):
:heart:

src/index.jsx
Dillon-Dumesnil: The ismount stuff is in InstructorToolbar, not CourseExit.

And yeah, I was thinking that too. Like face value, it does look very normal (and I guess with this PR it is now a pattern we support). I'll remove. It's at least captured in this commit/PR

src/index.jsx
ciduarte: what about slice for the routes above? 🤔

PR #237, opened 2020-10-07
AA-388: CourseExitPage UI improvements

ciduarte:
### Desktop
_Downloadable_
CourseCelebrationUI

_Requesting_
CourseCelebration_Requesting

_Earned but not available_
CourseCelebration_EarnedButNotAvailable

_Unverified_
CourseCelebration_Unverified

### Responsive (mobile size)
I took a stab at making this responsive because we didn't have a mock for it. We can go through changes at the UX review.
CourseCelebrationResponsiveUI


src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -31,7 +31,7 @@ function CourseCelebration({ intl }) {
     certStatus,
     certWebViewUrl,
     downloadUrl,
-  } = certificateData || { downloadUrl: 'google.com', certWebViewUrl: 'duckduckgo.com' };
+  } = certificateData;
```
Dillon-Dumesnil, line 44: Thank you.

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -43,6 +43,7 @@ function CourseCelebration({ intl }) {
       {intl.formatMessage(messages.dashboardLink)}
     </Hyperlink>
   );
+  // todo: remove this hardcoded link to edX support
```
Dillon-Dumesnil, line 56: 👍

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -91,17 +92,31 @@ function CourseCelebration({ intl }) {
...
+            />
+          </div>
+          <div className="mb-2">
+            <FormattedMessage
+              id="courseCelebration.certificateBody.notAvailable.accessCertificate"
+              defaultMessage="You will be able to access your certificate any time from your
+                {dashboardLink} and {profileLink}."
+              values={{ dashboardLink, profileLink }}
+            />
+          </div>
```
Dillon-Dumesnil, line 127: We can find out during UX review, but I think Eugene might be okay with this all being one message (like what we do in the downloadable case). Hilary's wireframes had them split, but I think she was open to change based on the wireframes.

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -114,6 +129,7 @@ function CourseCelebration({ intl }) {
       buttonText = intl.formatMessage(messages.verifyIdentityButton);
       buttonLocation = verifyIdentityUrl;
       title = intl.formatMessage(messages.certificateHeaderUnverified);
+      // todo: check for idVerificationSupportLink null
```
Dillon-Dumesnil, line 141: 👍

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -131,43 +147,70 @@ function CourseCelebration({ intl }) {
...
+            src={celebration}
+            alt={`${intl.formatMessage(messages.congratulationsImage)}`}
+            className="img-fluid"
+            style={{ width: '36rem' }}
+          />
+        </div>
+        <div className="col-12 px-0 px-md-5">
+          <div className="row w-100 m-0 p-4 bg-primary-100">
+            <div className="col order-1 order-md-0 pl-0 pr-0 pr-md-5">
+              <div className="h4">{title}</div>
```
Dillon-Dumesnil, line 187: I think you might have had a comment in the original PR about using an `h4` here? This is what we want?

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -131,43 +147,70 @@ function CourseCelebration({ intl }) {
...
+              <p>{message}</p>
+              {/* The requesting status needs a different button because it does a POST instead of a GET */}
+              {certStatus === 'requesting' && (
+                <Button
+                  variant="outline-primary"
+                  onClick={() => dispatch(requestCert(courseId))}
+                  style={{ backgroundColor: '#FFFFFF' }}
```
Dillon-Dumesnil: These are necessary? Huh. I was expecting those to be something we could have gotten rid of.

src/courseware/course/course-exit/messages.js
```diff
@@ -34,42 +43,41 @@ const messages = defineMessages({
...
   },
   requestCertificateButton: {
     id: 'courseCelebration.requestCertificateButton',
     defaultMessage: 'Request certificate',
-    description: 'Text for request certificate button in certificate banner',
   },
   shareHeader: {
     id: 'courseCelebration.shareHeader',
     defaultMessage: 'You have completed your course. Share your success on social media or email.',
-    description: 'Social media/email share header',
```
Dillon-Dumesnil, line 62: For my future reference, when are descriptions not needed?

src/courseware/course/course-exit/messages.js
```diff
@@ -34,42 +43,41 @@ const messages = defineMessages({
...
     id: 'courseCelebration.verifyIdentityButton',
     defaultMessage: 'Verify now',
-    description: 'Text for verify identity button in certificate banner',
+    description: 'Button to verify the identify of the user',
   },
   viewCertificateButton: {
     id: 'courseCelebration.viewCertificateButton',
     defaultMessage: 'View now',
-    description: 'Text for view certificate button in certificate banner',
+    description: 'Button to view the course certificate',
```
Dillon-Dumesnil, line 79: For example, this was changed to `Button to view the course certificate`, but you didn't change the request one to `Button to request the course certificate`. What's the difference?

Dillon-Dumesnil (approved):
Couple small questions, but nothing blocking. Looks great

src/courseware/course/course-exit/CourseCelebration.jsx
ciduarte: Good point

src/courseware/course/course-exit/CourseCelebration.jsx
ciduarte: Yeah my comment there was about _not_ using an `<h4>` for a11y reasons. The `className="h4"` just overrides the styling, which is what we want. We want to avoid using the heading tag and changing that DOM element when there aren't other headings preceding it

src/courseware/course/course-exit/CourseCelebration.jsx
ciduarte: Do you mean the `backgroundColor`? I can bring this up in the UX review. I went to remove it, but the background didn't match the mock so I left it 🤷‍♀️

src/courseware/course/course-exit/CourseCelebration.jsx
Dillon-Dumesnil: Ah yeah, that makes more sense. Thanks for clearing up!

src/courseware/course/course-exit/CourseCelebration.jsx
Dillon-Dumesnil: Yeah, the backgroundColor. I looked back at it and it seems to be because the button is on the cert banner and so if it has a clear background, it takes on the color of the banner. This just forces it to white. Might be better to at least switch from `#FFFFFF` to `white`? Not sure if there is a `variant` option that can take care of this. The `#FFFFFF` just raised a flag.

src/courseware/course/course-exit/messages.js
ciduarte: So my understanding of the descriptions is that they're used for by the (human) translators to gain context on what the text is so they can give us the most accurate translation.

Descriptions aren't always necessary because with a string like "You have completed your course. Share your success on social media or email.", the description of "social media/email share header" doesn't necessarily communicate any new information about the context to the person translating that string.

I'm realizing, though, it might be best to leave these descriptions as is, as to not trigger a re-translation? I'm not sure how those get triggered... but... my thought was to delete these because I don't think they're good examples of descriptions.

src/courseware/course/course-exit/messages.js
ciduarte: The description on "Request certificate" seemed repetitive and like it didn't add necessary context, so I deleted it. But I think the right move would've been to change it to what you have here. I wasn't comparing the two when I did it, I was hacking away at subpar examples and only replaced them if it felt necessary.

Unrelated, this is flagging to me an a11y concern because "view now" doesn't seem like enough context to communicate to the user what action is being taken w/ that button. So thanks for making me look again

src/courseware/course/course-exit/CourseCelebration.jsx
ciduarte: Oooooh I didn't realize your concern was with the hex value. Yeah I'll switch this to `white`

src/courseware/course/course-exit/messages.js
Dillon-Dumesnil: I think it would be best to just remove them. Hopefully our translating service is smart enough to realize the string didn't change, and I'd prefer our codebase clean.

src/courseware/course/course-exit/messages.js
Dillon-Dumesnil: Oh you know, that's what I'm here for. I'm fine with whatever you choose. Just wanted some more information on when/how to use descriptions here. Basically, if it isn't adding more info than the default message, use it.

src/courseware/course/course-exit/CourseCelebration.jsx
Dillon-Dumesnil: Magic numbers (letters?) Carla! Can't be having something like that.

src/courseware/course/course-exit/messages.js
ciduarte: Also realizing the course exit page isn't even released so time on re-translating isn't a huge concern

PR #241, opened 2020-10-13
TNL-7299 - Hide sequences that are time limited (special exams)

davidjoy:
If a sequence has its isTimeLimited flag set, then show a spinner instead of the sequence content.  The CoursewareContainer, meanwhile, will be attempting to redirect to the legacy experience.

This prevents a situation where we temporarily show proctored/special exam content to users before their exam starts.

### DIFF
--- a/src/courseware/course/sequence/Sequence.jsx
+++ b/src/courseware/course/sequence/Sequence.jsx
@@ -113,6 +113,20 @@ function Sequence({
     );
   }
 
+  /*
+  TODO: When the micro-frontend supports viewing special exams without redirecting to the legacy
+  experience, we can remove this whole conditional. For now, though, we show the spinner here
+  because we expect CoursewareContainer to be performing a redirect to the legacy experience while
+  we're waiting. That redirect may take a few seconds, so we show the spinner in the meantime.
+  */
+  if (sequenceStatus === 'loaded' && sequence.isTimeLimited) {
+    return (
+      <PageLoading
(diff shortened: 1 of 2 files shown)


src/courseware/course/sequence/Sequence.jsx
```diff
@@ -113,6 +113,20 @@ function Sequence({
     );
   }
 
+  /*
+  TODO: When the micro-frontend supports viewing special exams without redirecting to the legacy
```
stvstnfrd, line 117: Do we have a ticket for this? I thought so, but not sure...

stvstnfrd (approved):
Just a TODO-related question, but otherwise, looks good :+1:

src/courseware/course/sequence/Sequence.jsx
davidjoy: We don't... this would presumably get cleaned up as part of implementing special exams in the MFE.  

Unless this conditional is removed, they'll never show up, so we can probably be pretty confident that it'll get noticed. :D

davidjoy: I mean, I think there's probably an epic for it?  But maybe not tasks yet...

marcotuts: v1.4 should have a story... looking for it now...

marcotuts: oh sorry this comment was of course To-DO specific and I missed that. UNtil the epic (v1.5) is done I expect this to-do would remain.

PR #246, opened 2020-10-16
AA-377: Add non-passing course exit screen

mikix:
- Adds a non-passing cert learner course exit screen
- Moves all the logic about what course-exit mode we're in into a utility method in the course-exit folder
- Moves all the logic about how the 'Next' button should read into a utility method in the course-exit folder

Some other more nitty gritty thoughts:
- Moved the bottom dashboard link to a DashboardFooter component to share it. I've seen mocks with other footer text/links, so we might end up adding more of footer components.
- I've tried to keep most string ids starting with `courseExit.` rather than like `courseNonPassing.` or whatever, because they can move around a bit and we shouldn't be changing ids ideally.
- I don't love the kind of complicated row / column nesting we have in CourseCelebration, but I wanted to keep parity for now in CourseNonPassing. Maybe at some point we can refactor out the common bits and make that a little less nested.
- I used an Alert for my color box, but CourseCelebration does it manually. Is one better than another?
- No tests yet. We should probably at least test the utility method that spits out an enumeration, yah? Pages themselves are kind of simple, though rarely seen. I can work on a utility method test.

Wireframe: https://www.figma.com/file/GRsxJFhsMaUr1etLut6dDC/Course-Completion?node-id=142%3A1028

**Screenshots**
Screenshot from 2020-10-19 09-23-14
Screenshot from 2020-10-19 09-22-59
Screenshot from 2020-10-16 13-32-28

### DIFF
--- a/src/courseware/course/course-exit/CourseNonPassing.jsx
+++ b/src/courseware/course/course-exit/CourseNonPassing.jsx
@@ -1,0 +1,51 @@
+import React from 'react';
+
+import {
+  injectIntl, intlShape,
+} from '@edx/frontend-platform/i18n';
+import { Helmet } from 'react-helmet';
+import { useParams } from 'react-router-dom';
+import { Alert, Button } from '@edx/paragon';
+import { getConfig } from '@edx/frontend-platform';
+
+import { useModel } from '../../../generic/model-store';
+
+import DashboardFootnote from './DashboardFootnote';
+import messages from './messages';
+
+function CourseNonPassing({ intl }) {
+  const { courseId } = useParams();
+  const { tabs } = useModel('courses', courseId);
+
+  // Get progress tab link for 'view grades' button
+  const progressTab = tabs.find(tab => tab.slug === 'progress');
+  const progressLink = progressTab && progressTab.url;
+
+  return (
+    <>
+      <Helmet>
+        <title>{`${intl.formatMessage(messages.endOfCourseTitle)} | ${getConfig().SITE_NAME}`}</title>
+      </Helmet>
+      <div className="row w-100 mx-0 mb-4 px-5 py-4 border border-light justify-content-center">
+        <div className="col-12 p-0 h2 text-center">
+          { intl.formatMessage(messages.endOfCourseHeader) }
+        </div>
+        <Alert variant="primary" className="col col-lg-10 mt-4 d-flex align-items-start">
+          <div className="flex-grow-1 mr-5">{ intl.formatMessage(messages.endOfCourseDescription) }</div>
+          {progressLink && (
+            <Button variant="primary" className="flex-shrink-0" href={progressLink}>
+              {intl.formatMessage(messages.viewGradesButton)}
+            </Button>
+          )}
+        </Alert>
+        <DashboardFootnote />
+      </div>
+    </>
+  );
+}
+
+CourseNonPassing.propTypes = {
+  intl: intlShape.isRequired,
+};
+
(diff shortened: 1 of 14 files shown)


src/courseware/course/course-exit/CourseNonPassing.jsx
```diff
@@ -0,0 +1,37 @@
...
+import { getConfig } from '@edx/frontend-platform';
+
+import messages from './messages';
+import DashboardFootnote from './DashboardFootnote';
+
+function CourseNonPassing({ intl }) {
+  return (
+    <>
+      <Helmet>
+        <title>{`End of Course | ${getConfig().SITE_NAME}`}</title>
```
ciduarte: End of Course should be translated

src/courseware/course/course-exit/CourseNonPassing.jsx
```diff
@@ -0,0 +1,37 @@
...
+      </Helmet>
+      <div className="row w-100 mx-0 mb-4 px-5 py-4 border border-light justify-content-center">
+        <div className="col-12 p-0 h2 text-center">
+          { intl.formatMessage(messages.endOfCourseHeader) }
+        </div>
+        <Alert variant="info" className="col col-lg-10 mt-4 d-flex align-items-start">
```
ciduarte: I didn't think of switching this to an Alert in CourseCelebration because it had already been done manually. I think using an Alert is fine, but it adds a border that isn't included in the mock, and the background color should be `primary-100` (i.e. `bg-primary-100`).

[Here's the difference between `info` and `primary` if you're curious.](https://paragon-edx.netlify.app/foundations/colors)

src/courseware/course/course-exit/CourseNonPassing.jsx
```diff
@@ -0,0 +1,37 @@
...
+          <div className="flex-grow-1 mr-5">{ intl.formatMessage(messages.endOfCourseDescription) }</div>
```
ciduarte, line 34: Do you know the difference between these two mocks? I'm not sure why they both exist but one has bold writing and the other doesn't...

https://www.figma.com/file/GRsxJFhsMaUr1etLut6dDC/Course-Completion?node-id=142%3A1028

https://www.figma.com/file/GRsxJFhsMaUr1etLut6dDC/Course-Completion?node-id=153%3A873

src/courseware/course/course-exit/DashboardFootnote.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+function DashboardFootnote({ intl }) {
+  const dashboardLink = (
+    <Hyperlink
+      style={{ textDecoration: 'underline' }}
+      destination={`${getConfig().LMS_BASE_URL}/dashboard`}
+    >
+      {intl.formatMessage(messages.dashboardLink)}
+    </Hyperlink>
+  );
```
ciduarte: These links should be `text-gray-700`. (I don't believe the mocks were updated but this was decided in the UX Review)

ciduarte (approved):
I enjoyed the refactoring, thanks! Some small styling things

src/courseware/course/course-exit/CourseNonPassing.jsx
mikix: We talked about this on slack, but I based mine off one Eugene gave me (the one without bold text). We'll circle back with her on reviewing this.

ciduarte (approved):
Looks great!

PR #249, opened 2020-10-19
AA-385: Add LinkedIn Add to Profile Button to Course Celebrate

Dillon-Dumesnil:


Dillon-Dumesnil:
Mobile View: 
image

Mid size view: 
image

Standard View: 
image

Dillon-Dumesnil:
Experience after clicking button: 
image

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -19,6 +21,8 @@ import { useModel } from '../../../generic/model-store';
 import { requestCert } from '../../../course-home/data/thunks';
 import DashboardFootnote from './DashboardFootnote';
 
+const LINKEDIN_BLUE = '#007fb1';
```
mikix, line 24: Is this something we could pull from react-share? I doubt it, just... feels annoying to have this here in case they ever switch up their color.

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -192,9 +205,20 @@ function CourseCelebration({ intl }) {
                   {buttonText}
                 </Button>
               )}
+              {certStatus === 'downloadable' && linkedinAddToProfileUrl && (
+                <Button
+                  className="mr-2 mb-2 mb-sm-0"
+                  href={linkedinAddToProfileUrl}
+                  onClick={() => logClick('linkedin')}
+                  style={{ backgroundColor: LINKEDIN_BLUE, padding: '0.25rem 1.25rem 0.25rem 0.5rem' }}
```
mikix, line 213: Bootstrap padding wasn't right size here?

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -192,9 +205,20 @@ function CourseCelebration({ intl }) {
                   {buttonText}
                 </Button>
               )}
+              {certStatus === 'downloadable' && linkedinAddToProfileUrl && (
+                <Button
+                  className="mr-2 mb-2 mb-sm-0"
+                  href={linkedinAddToProfileUrl}
+                  onClick={() => logClick('linkedin')}
```
mikix, line 212: That's the same log click string as the social share button -- we probably want to distinguish them? Like 'linkedin-cert' here or something?

mikix (approved):
Looks pretty.

src/courseware/course/course-exit/CourseCelebration.jsx
Dillon-Dumesnil: I actually think we want to keep this as linkedin and update the event name. We had `'edx.ui.lms.celebration.social_share.clicked'` and I can update to something better.

src/courseware/course/course-exit/CourseCelebration.jsx
Dillon-Dumesnil: Unfortunately not. The icon kind of messed things up so we couldn't just use standard bootstrap padding.

src/courseware/course/course-exit/CourseCelebration.jsx
Dillon-Dumesnil: Yeah, I don't think so. We could also use a custom color here too. I just needed a hardcoded value because of the use in the fill of the button and also the icon.

src/courseware/course/course-exit/CourseCelebration.jsx
Dillon-Dumesnil: Updated to `'edx.ui.lms.course_celebration.linkedin_add_to_profile.clicked'`

Dillon-Dumesnil:
Tests will be coming in AA-397

PR #250, opened 2020-10-19
AA-362: Add tests for outline tab alerts

mikix:


src/course-home/data/__factories__/outlineTabData.factory.js
```diff
@@ -17,27 +17,30 @@ Factory.define('outlineTabData')
...
-    userTimezone: 'UTC',
-  })
-  .attr('handouts_html', [], () => '<ul><li>Handout 1</li></ul>')
-  .attr('offer_html', [], () => '<div>Great offer here</div>')
   .attr('resume_course', ['host', 'courseId'], (host, courseId) => ({
     has_visited_course: false,
     url: `${host}/courses/${courseId}/jump_to/block-v1:edX+Test+Block@12345abcde`,
   }))
-  .attr('welcome_message_html', [], () => '<p>Welcome to this course!</p>');
+  .attrs({
```
Dillon-Dumesnil, line 28: 👏

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -21,63 +21,70 @@ jest.mock('@edx/frontend-platform/analytics');
...
   let axiosMock;
 
-  const courseMetadataUrl = new RegExp(`${getConfig().LMS_BASE_URL}/api/course_home/v1/course_metadata/*`);
-  const goalUrl = new RegExp(`${getConfig().LMS_BASE_URL}/api/course_home/v1/save_course_goal`);
-  const outlineUrl = new RegExp(`${getConfig().LMS_BASE_URL}/api/course_home/v1/outline/*`);
+  const courseId = 'course-v1:edX+Test+run';
+  const courseMetadataUrl = `${getConfig().LMS_BASE_URL}/api/course_home/v1/course_metadata/${courseId}`;
+  const enrollmentUrl = `${getConfig().LMS_BASE_URL}/api/enrollment/v1/enrollment`;
+  const goalUrl = `${getConfig().LMS_BASE_URL}/api/course_home/v1/save_course_goal`;
+  const outlineUrl = `${getConfig().LMS_BASE_URL}/api/course_home/v1/outline/${courseId}`;
```
Dillon-Dumesnil, line 28: I approve this message

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -21,63 +21,70 @@ jest.mock('@edx/frontend-platform/analytics');
...
+  }
 
   beforeEach(async () => {
     axiosMock = new MockAdapter(getAuthenticatedHttpClient());
-    axiosMock.onGet(courseMetadataUrl).reply(200, courseMetadata);
-    axiosMock.onGet(outlineUrl).reply(200, outlineTabData);
+
+    // Set defaults for network requests
+    setMetadata();
+    setTabData();
```
Dillon-Dumesnil: I'm not sure how lightweight these are in reality, but I wonder if these need to be in a `beforeEach` if the intention is that tests will be calling them and passing in the attributes and options they care about. Now any test that does specify their own attributes is just throwing away this initial call which feels wasteful.

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -21,63 +21,70 @@ jest.mock('@edx/frontend-platform/analytics');
...
-
-      render(<OutlineTab />, { store });
+      await fetchAndRender();
       const expandedSectionNode = screen.getByRole('button', { name: /Title of Section/ });
       expect(expandedSectionNode).toHaveAttribute('aria-expanded', 'true');
     });
 
-    it('handles expand/collapse all button click', () => {
-      render(<OutlineTab />, { store });
+    it('handles expand/collapse all button click', async () => {
```
Dillon-Dumesnil, line 88: Should all tests be run with this `async` param? I see you making the change, but not sure what it means.

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -96,52 +103,40 @@ describe('Outline Tab', () => {
...
-      render(<OutlineTab />, { store });
+      await fetchAndRender();
       expect(screen.getByTitle('Incomplete section')).toBeInTheDocument();
     });
   });
 
   describe('Welcome Message', () => {
-    it('does not render show more/less button under 200 characters', () => {
-      render(<OutlineTab />, { store });
+    it('does not render show more/less button under 200 characters', async () => {
```
Dillon-Dumesnil: RIP merge conflicts

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -276,67 +256,48 @@ describe('Outline Tab', () => {
...
 
   describe('Alert List', () => {
     describe('Enrollment Alert', () => {
-      let extraText;
       let alertMessage;
       let staffMessage;
 
       beforeEach(() => {
-        extraText = outlineTabData.enroll_alert.extra_text;
+        const defaultTabData = Factory.build('outlineTabData');
```
Dillon-Dumesnil: Should this instead use `setTabData();`?

src/course-home/data/__factories__/outlineTabData.factory.js
```diff
@@ -17,27 +17,30 @@ Factory.define('outlineTabData')
...
+      content_type_gating_enabled: false,
+      missed_gated_content: false,
+      missed_deadlines: false,
+    },
+    enroll_alert: {
+      can_enroll: true,
+      extra_text: 'Contact the administrator.',
+    },
+    handouts_html: '<ul><li>Handout 1</li></ul>',
+    offer_html: null,
```
Dillon-Dumesnil, line 44: Why was this dropped from having a value?

src/course-home/data/__factories__/outlineTabData.factory.js
```diff
@@ -17,27 +17,30 @@ Factory.define('outlineTabData')
...
-  })
-  .attr('handouts_html', [], () => '<ul><li>Handout 1</li></ul>')
-  .attr('offer_html', [], () => '<div>Great offer here</div>')
   .attr('resume_course', ['host', 'courseId'], (host, courseId) => ({
     has_visited_course: false,
     url: `${host}/courses/${courseId}/jump_to/block-v1:edX+Test+Block@12345abcde`,
   }))
-  .attr('welcome_message_html', [], () => '<p>Welcome to this course!</p>');
+  .attrs({
+    course_expired_html: null,
```
Dillon-Dumesnil, line 29: Why was this dropped from having a value?

Dillon-Dumesnil (commented):
Couple questions about some of the implementation. Love the cleanup.

src/course-home/outline-tab/OutlineTab.test.jsx
mikix: I dunno how lightweight they are either. They do two things -- build a data object and set it as the return for axios mock. I had assumed that was cheap, and we didn't care about being slightly wasteful.

I guess the alternative would be moving the "defaults" to the outer layer as constants and set them as axios defaults in `beforeEach`? I do need the default object later anyway in one of the tests...

src/course-home/outline-tab/OutlineTab.test.jsx
mikix: Tests need the `async` tag if they call `await` inside. The `fetchAndRender` call is async (uses await) so the parent tests have to be too. I don't think it hurts to make all tests async anyway, but I believe the compiler/linter catches any missing async tag already.

src/course-home/outline-tab/OutlineTab.test.jsx
mikix: This was just getting the default tab data so that it could extract a string argument from it.

src/course-home/data/__factories__/outlineTabData.factory.js
mikix: This and `course_expired_html` were dropped from having values because (a) when I went to test that the alert for it worked, it felt weird that it was already set (and what, I should test that unsetting it made it go away?) -- these felt like rare events that they would be set, not the "default state" and (b) one of Carla's tests already null'd them out because they were getting in her way with extraneous alerts that made it harder to find the alert she cared about.

Dillon-Dumesnil: Yeah, I remember seeing that. I just noticed the snapshot changing with these changes and wanted to make sure it was intentional. All good to me.

src/course-home/outline-tab/OutlineTab.test.jsx
mikix: Done.

Dillon-Dumesnil (approved):

PR #251, opened 2020-10-19
AA-381: Fix shorten welcome message bug

ciduarte:
- Switched from 200 character count to 100 word limit for parity with old experience

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,8 @@
     "react-share": "4.2.1",
     "redux": "4.0.5",
     "regenerator-runtime": "0.13.7",
-    "reselect": "4.0.0"
+    "reselect": "4.0.0",
+    "truncate-html": "^1.0.3"
   },
   "devDependencies": {
     "@edx/frontend-build": "5.2.7",
--- a/package-lock.json
+++ b/package-lock.json
@@ -2905,6 +2905,14 @@
         "@babel/types": "^7.3.0"
       }
     },
+    "@types/cheerio": {
+      "version": "0.22.22",
+      "resolved": "https://registry.npmjs.org/@types/cheerio/-/cheerio-0.22.22.tgz",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -58,7 +58,8 @@
     "react-share": "4.2.1",
     "redux": "4.0.5",
     "regenerator-runtime": "0.13.7",
-    "reselect": "4.0.0"
+    "reselect": "4.0.0",
+    "truncate-html": "^1.0.3"
```
mikix, line 62: A. I hate how json doesn't use trailing commas.
B. We don't generally use the ^ syntax with renovate (and renovate will drop it next time it look at these I think)

mikix (approved):
Seems simple enough! Good find!

Dillon-Dumesnil (approved):
It all looks good to me.

Should we add in a test to cover the case that brought up this issue in the first place? Like ensure all tags are appropriately closed as part of truncating? Or that we are accurately ignoring certain tags? Not sure if that's just getting into the part of testing that a library does what it should. Just thinking out loud.

package.json
ciduarte: Ooh is it worth manually dropping the ^ to be safe?

mikix: Naw I think it's fine to leave in. Just talking out loud over here

ciduarte:
> It all looks good to me.
> 
> Should we add in a test to cover the case that brought up this issue in the first place? Like ensure all tags are appropriately closed as part of truncating? Or that we are accurately ignoring certain tags? Not sure if that's just getting into the part of testing that a library does what it should. Just thinking out loud.

Nothing actually triggered us noticing this bug, I just realized that it was being truncated incorrectly when writing the test for this. I think that might fall into testing the library too much, and there are some complications with testing the welcome message html itself because it all gets put into an iframe.

PR #257, opened 2020-10-23
AA-397: Course Exit Tests

Dillon-Dumesnil:


src/courseware/course/Course.test.jsx
```diff
@@ -57,21 +57,6 @@ describe('Course', () => {
     jest.spyOn(console, 'warn').mockImplementation(() => {});
     jest.spyOn(console, 'error').mockImplementation(() => {});
 
-    // Mock media queries, because `Celebration` modal uses `react-break` for responsive breakpoints.
```
Dillon-Dumesnil, line 60: Removed because it was added to the setupTest file so it runs by default.

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -194,7 +197,7 @@ function CourseCelebration({ intl }) {
           <Alert variant="primary" className="row w-100 m-0">
             <div className="col order-1 order-md-0 pl-0 pr-0 pr-md-5">
               <div className="h4">{title}</div>
-              <p>{message}</p>
+              {message}
```
Dillon-Dumesnil, line 200: This change (and the above changes) are because I saw an error about a `<p>` having a `<div>` inside of it and that not being liked. So now I just have the message specify the tags it wants.

src/courseware/course/sequence/sequence-navigation/SequenceNavigation.jsx
```diff
@@ -60,7 +60,7 @@ function SequenceNavigation({
   const renderNextButton = () => {
     const exitText = getCourseExitText(courseId, intl);
     const buttonOnClick = isLastUnit ? goToCourseExitPage : nextSequenceHandler;
-    const buttonText = isLastUnit && exitText ? exitText : intl.formatMessage(messages.nextButton);
+    const buttonText = (isLastUnit && exitText) ? exitText : intl.formatMessage(messages.nextButton);
```
Dillon-Dumesnil, line 63: This was a readability thing for me, but if others don't like it, I'm happy to switch.

src/courseware/course/sequence/sequence-navigation/UnitNavigation.test.jsx
```diff
@@ -77,10 +73,46 @@ describe('Unit Navigation', () => {
     expect(screen.getByRole('button', { name: /next/i })).toBeEnabled();
   });
 
-  it('displays end of course message instead of the "Next" button as needed', () => {
+  it('has the "Next" button disabled for the last unit in the sequence if there is no Exit Page', () => {
```
Dillon-Dumesnil, line 76: There just wasn't a test for this which felt odd. /shrug

src/courseware/course/course-exit/CourseExit.test.jsx
```diff
@@ -0,0 +1,148 @@
...
+
+    logUnhandledRequests(axiosMock);
+  });
+
+  describe('Course Exit routing', () => {
+    it('Routes to celebration for a celebration status', async () => {
+      setMetadata({
+        certificate_data: {
+          cert_status: 'downloadable',
+          cert_web_view_url: '/certificates/cooluuidgoeshere',
```
mikix, line 56: :dark_sunglasses: uuids only, please.

src/courseware/course/course-exit/CourseExit.test.jsx
```diff
@@ -0,0 +1,148 @@
...
+        },
+      });
+      await fetchAndRender(<CourseExit />);
+      expect(screen.getByText('Congratulations!')).toBeInTheDocument();
```
mikix, line 60: You've gone with `get` plus `toBe`? It feels "wasteful" to me, but I suppose it's defensive against the `getBy` call changing to something else.

mikix (approved):
Yay! Thanks for all the test design and cleanup.

src/courseware/course/course-exit/CourseExit.test.jsx
Dillon-Dumesnil: Honestly, I debated it. I went with both because I think it helps make it more clear. I was even thinking back to django tests and I never liked it when things were implicit. Like a Model.objects.get(blah) and that ending the test and requiring the implicit knowledge that that would raise an error otherwise. Now anyone coming in doesn't need to know about the get raising an error (although it still will if something is removed) and they know it is checking for it to be in the document.

I'm happy to update our `getBy` calls to not have it, I was just trying to go for a clarity thing. But it might just be personal preference.

PR #259, opened 2020-10-23
AA-390: Add Social Share icons to Course Exit

ciduarte:
**Social share icons within Course Celebration, desktop and responsive**
<div>
Screen Shot 2020-10-26 at 10 03 56 AM
Screen Shot 2020-10-26 at 10 04 01 AM
</div>

**Social share icons within celebration modal, desktop and responsive**
The UI hasn't changed here, but some refactoring around `<SocialIcons />` impacted the celebration modal. Screenshots for sanity.
<div>
Desktop first section milestone modal
Responsive first section milestone modal
</div>

### DIFF
--- a/src/courseware/course/celebration/messages.js
+++ b/src/courseware/course/celebration/messages.js
@@ -13,11 +13,6 @@ const messages = defineMessages({
     id: 'learning.celebration.earned',
     defaultMessage: 'You earned it!',
   },
-  emailBody: {
-    id: 'learning.celebration.emailBody',
-    defaultMessage: 'What are you spending your time learning?',
-    description: 'Body when sharing course progress via email',
-  },
   emailSubject: {
     id: 'learning.celebration.emailSubject',
     defaultMessage: "I'm on my way to completing {title} online with {platform}!",
@@ -32,15 +27,7 @@ const messages = defineMessages({
     id: 'learning.celebration.share',
     defaultMessage: 'Take a moment to celebrate and share your progress.',
   },
-  shareEmail: {
-    id: 'learning.celebration.share.email',
-    defaultMessage: 'Share your progress via email.',
-  },
-  shareService: {
-    id: 'learning.celebration.share.service',
-    defaultMessage: 'Share your progress on {service}.',
-  },
-  social: {
+  socialMessage: {
     id: 'learning.celebration.social',
     defaultMessage: 'I’m on my way to completing {title} online with {platform}. What are you spending your time learning?',
     description: 'Shown when sharing course progress on a social network',
--- a/src/courseware/course/course-exit/CourseCelebration.jsx
+++ b/src/courseware/course/course-exit/CourseCelebration.jsx
@@ -21,6 +21,7 @@ import { useModel } from '../../../generic/model-store';
 import { requestCert } from '../../../course-home/data/thunks';
 import DashboardFootnote from './DashboardFootnote';
 import UpgradeFootnote from './UpgradeFootnote';
+import SocialIcons from '../../social-share/SocialIcons';
 
 const LINKEDIN_BLUE = '#007fb1';
 
@@ -95,11 +96,11 @@ function CourseCelebration({ intl }) {
   let certificateImage = certificate;
   let footnote;
   let message;
-  let title;
+  let certHeader;
   // These cases are taken from the edx-platform `get_cert_data` function found in lms/courseware/views/views.py
   switch (certStatus) {
     case 'downloadable':
-      title = intl.formatMessage(messages.certificateHeaderDownloadable);
+      certHeader = intl.formatMessage(messages.certificateHeaderDownloadable);
       message = (
         <p>
           <FormattedMessage
@@ -123,7 +124,7 @@ function CourseCelebration({ intl }) {
       break;
     case 'earned_but_not_available': {
       const endDate = <FormattedDate value={end} day="numeric" month="long" year="numeric" />;
-      title = intl.formatMessage(messages.certificateHeaderNotAvailable);
+      certHeader = intl.formatMessage(messages.certificateHeaderNotAvailable);
       message = (
         <>
           <p>
@@ -150,14 +151,14 @@ function CourseCelebration({ intl }) {
     }
     case 'requesting':
       buttonText = intl.formatMessage(messages.requestCertificateButton);
-      title = intl.formatMessage(messages.certificateHeaderRequestable);
+      certHeader = intl.formatMessage(messages.certificateHeaderRequestable);
       message = (<p>{intl.formatMessage(messages.requestCertificateBodyText)}</p>);
       footnote = <DashboardFootnote />;
       break;
     case 'unverified':
       buttonText = intl.formatMessage(messages.verifyIdentityButton);
       buttonLocation = verifyIdentityUrl;
-      title = intl.formatMessage(messages.certificateHeaderUnverified);
+      certHeader = intl.formatMessage(messages.certificateHeaderUnverified);
       // todo: check for idVerificationSupportLink null
       message = (
         <p>
@@ -174,7 +175,7 @@ function CourseCelebration({ intl }) {
     case 'audit_passing':
     case 'honor_passing':
       if (verifiedMode) {
-        title = intl.formatMessage(messages.certificateHeaderUpgradable);
+        certHeader = intl.formatMessage(messages.certificateHeaderUpgradable);
         message = (
           <p>
             <FormattedMessage
@@ -225,8 +226,15 @@ function CourseCelebration({ intl }) {
         </div>
         <div className="col-12 p-0 font-weight-normal lead text-center">
           {intl.formatMessage(messages.shareHeader)}
+          <SocialIcons
+            analyticsId="edx.ui.lms.course_exit.social_share.clicked"
+            className="mt-2"
+            courseId={courseId}
+            emailSubject={messages.socialMessage}
+            socialMessage={messages.socialMessage}
+          />
         </div>
-        <div className="col-12 my-4 px-0 px-md-5 text-center">
+        <div className="col-12 mt-3 mb-4 px-0 px-md-5 text-center">
           <OnMobile>
             <img
               src={CelebrationMobile}
@@ -244,10 +252,10 @@ function CourseCelebration({ intl }) {
           </OnAtLeastTablet>
         </div>
         <div className="col-12 px-0 px-md-5">
-          {title && (
+          {certHeader && (
           <Alert variant="primary" className="row w-100 m-0">
             <div className="col order-1 order-md-0 pl-0 pr-0 pr-md-5">
-              <div className="h4">{title}</div>
+              <div className="h4">{certHeader}</div>
               {message}
               {/* The requesting status needs a different button because it does a POST instead of a GET */}
               {certStatus === 'requesting' && (
--- a/src/courseware/course/course-exit/CourseExit.test.jsx
+++ b/src/courseware/course/course-exit/CourseExit.test.jsx
@@ -118,6 +118,24 @@ describe('Course Exit Pages', () => {
       expect(screen.getByRole('button', { name: 'Request certificate' })).toBeInTheDocument();
     });
 
+    it('Displays social share icons', async () => {
+      setMetadata({ certificate_data: { cert_status: 'unverified' }, marketing_url: 'https://edx.org' });
+      await fetchAndRender(<CourseCelebration />);
+      expect(screen.getByRole('button', { name: 'linkedin' })).toBeInTheDocument();
+      expect(screen.getByRole('button', { name: 'facebook' })).toBeInTheDocument();
+      expect(screen.getByRole('button', { name: 'twitter' })).toBeInTheDocument();
+      expect(screen.getByRole('button', { name: 'email' })).toBeInTheDocument();
+    });
+
+    it('Does not display social share icons if no marketing URL', async () => {
+      setMetadata({ certificate_data: { cert_status: 'unverified' } });
+      await fetchAndRender(<CourseCelebration />);
+      expect(screen.queryByRole('button', { name: 'linkedin' })).not.toBeInTheDocument();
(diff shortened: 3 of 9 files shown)


src/index.scss
```diff
@@ -233,7 +233,6 @@ $primary: #1176B2;
   }
 
   .previous-btn, .next-btn {
-    min-width: 4rem;
```
ciduarte, line 236: unrelated fix, adjusting the sequence nav buttons

src/courseware/course/celebration/messages.js
```diff
@@ -32,16 +32,8 @@ const messages = defineMessages({
...
-    defaultMessage: 'Share your progress via email.',
-  },
-  shareService: {
-    id: 'learning.celebration.share.service',
-    defaultMessage: 'Share your progress on {service}.',
-  },
-  social: {
-    id: 'learning.celebration.social',
+  socialMessage: {
+    id: 'learning.celebration.socialMessage',
```
mikix: I wouldn't change the `id`. That wipes out any existing translations for the string in transifex, afaik.

Changing the dict name from `social` to `socialMessage` is great. But I'd say leave the `id` as the old value.

src/courseware/course/course-exit/messages.js
```diff
@@ -45,6 +45,16 @@ const messages = defineMessages({
     defaultMessage: 'Download my certificate',
     description: 'Button to download the course certificate',
   },
+  emailBody: {
+    id: 'courseCelebration.emailBody',
```
mikix: tiny tiny nit: I personally have been trying to have the `id`s be a bit more generic? Like `courseExit.social.emailBody` maybe -- a bigger beginning namespace, just so if we have to shift them around, the `id` doesn't get quite as stale.

src/courseware/course/course-exit/messages.js
```diff
@@ -45,6 +45,16 @@ const messages = defineMessages({
...
+    defaultMessage: 'I just completed {title} with {platform}!',
+    description: 'Body when sharing course progress via email',
+  },
+  emailSubject: {
+    id: 'courseCelebration.emailSubject',
+    defaultMessage: 'What are you spending your time learning?',
+    description: 'Subject when sharing course progress via email',
+  },
```
mikix: This is the same as on the celebration modal, yah?

Maybe it's worth the `SocialIcons` component having defaults for this stuff? Like I also noticed the two instances of it share the same hashtag. If the component itself had defaults and the instances supplied the interesting different bits, maybe that would be less duplication?

But I also totally get if you think it's cleaner if that component stays out of the business of actual content.

src/courseware/social-share/SocialIcons.jsx
```diff
@@ -87,9 +100,25 @@ function SocialIcons({ courseId, intl }) {
...
 SocialIcons.propTypes = {
+  analyticsId: PropTypes.string,
+  className: PropTypes.string,
   courseId: PropTypes.string.isRequired,
+  emailBody: PropTypes.node,
+  emailSubject: PropTypes.node,
+  hashtags: PropTypes.arrayOf(PropTypes.string),
   intl: intlShape.isRequired,
+  // eslint-disable-next-line react/forbid-prop-types
+  socialMessage: PropTypes.object,
```
mikix: Would `PropTypes.shape({})` get rid of the eslint warning?

mikix (approved):

src/courseware/course/course-exit/messages.js
ciduarte: Yeaahh, I kept going back and forth on this. I was thinking it would be cleaner to keep the messaging separated, but I'm not opposed to adding default values to reduce the duplication

ciduarte: Realized I flipped the subject and body in course exit, so glad I'm revisiting

src/courseware/social-share/messages.js
```diff
@@ -0,0 +1,19 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  defaultEmailBody: {
+    id: 'learning.social.share.email.defaultBody',
```
mikix: Uh, same comment on changing `id`s for the copied strings in this file. It'll wipe out any existing translations, afaik.

mikix (approved):

src/courseware/course/course-exit/CourseExit.test.jsx
```diff
@@ -113,6 +113,24 @@ describe('Course Exit Pages', () => {
...
+      expect(screen.getByRole('button', { name: 'email' })).toBeInTheDocument();
+    });
+
+    it('Does not display social share icons if no marketing URL', async () => {
+      setMetadata({ certificate_data: { cert_status: 'unverified' } });
+      await fetchAndRender(<CourseCelebration />);
+      expect(screen.queryByRole('button', { name: 'linkedin' })).not.toBeInTheDocument();
+      expect(screen.queryByRole('button', { name: 'facebook' })).not.toBeInTheDocument();
+      expect(screen.queryByRole('button', { name: 'twitter' })).not.toBeInTheDocument();
+      expect(screen.queryByRole('button', { name: 'email' })).not.toBeInTheDocument();
```
Dillon-Dumesnil, line 136: Just another thought on tests, but you could also do a queryByClass (or something) and just check for the `social-icons` class and do it in one line, yeah?

src/courseware/course/course-exit/messages.js
```diff
@@ -45,6 +45,11 @@ const messages = defineMessages({
     defaultMessage: 'Download my certificate',
     description: 'Button to download the course certificate',
   },
+  emailSubject: {
```
Dillon-Dumesnil: Do we need to have `emailSubject` and `socialMessage` be separate messages in this file? They have the exact same message so no need to have translators do it twice. Maybe both can be the `shareCompletionMessage` or something and you just pass it into the SocialIcons component for both the emailSubject and socialMessage.

src/setupTest.js
```diff
@@ -63,6 +63,7 @@ export function initializeMockApp() {
   mergeConfig({
     INSIGHTS_BASE_URL: process.env.INSIGHTS_BASE_URL || null,
     STUDIO_BASE_URL: process.env.STUDIO_BASE_URL || null,
+    TWITTER_URL: process.env.TWITTER_URL || null,
```
Dillon-Dumesnil, line 66: ? should this maybe just be specified in the .env.test file or whatever it is? I don't have the best understanding of what's happening here.

src/courseware/social-share/SocialIcons.jsx
```diff
@@ -67,17 +82,17 @@ function SocialIcons({ courseId, intl }) {
...
+        quote={intl.formatMessage(socialMessage, { platform: getConfig().SITE_NAME, title })}
         url={`${socialUtmMarketingUrl}&utm_source=facebook`}
       >
         <FacebookIcon round size={32} />
         <span className="sr-only">{intl.formatMessage(messages.shareService, { service: 'Facebook' })}</span>
       </FacebookShareButton>
       <EmailShareButton
         beforeOnClick={() => logClick('email')}
-        body={`${intl.formatMessage(messages.emailBody)}\n\n`}
+        body={`${(emailBody || defaultEmailBody)}\n\n`}
```
Dillon-Dumesnil: Can you get out of doing this by just setting the defaultProps value of emailBody below from `null` to `defaultEmailBody`?

src/courseware/social-share/SocialIcons.jsx
```diff
@@ -44,20 +59,20 @@ function SocialIcons({ courseId, intl }) {
...
       </LinkedinShareButton>
-      { twitterAccount && (
+      {twitterAccount && (
         <TwitterShareButton
           beforeOnClick={() => logClick('twitter')}
           className="ml-2"
-          hashtags={['myedxjourney']}
-          title={intl.formatMessage(messages.social, { platform: `@${twitterAccount}`, title })}
+          hashtags={hashtags}
+          title={intl.formatMessage(socialMessage, { platform: `@${twitterAccount}`, title })}
```
Dillon-Dumesnil: What happens when socialMessage is null? Should this be required to supply? Should we use a default value here?

src/courseware/social-share/SocialIcons.jsx
```diff
@@ -29,12 +38,18 @@ function SocialIcons({ courseId, intl }) {
...
 
+  const defaultEmailBody = intl.formatMessage(messages.defaultEmailBody);
+
   const twitterUrl = getConfig().TWITTER_URL;
   const twitterAccount = twitterUrl && twitterUrl.substring(twitterUrl.lastIndexOf('/') + 1);
 
   const logClick = (service) => {
+    if (!analyticsId) {
+      return;
+    }
```
Dillon-Dumesnil, line 47: :chefkiss:

src/courseware/social-share/messages.js
```diff
@@ -0,0 +1,19 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  defaultEmailBody: {
+    id: 'learning.celebration.emailBody',
+    defaultMessage: 'What are you spending your time learning?',
+    description: 'Body when sharing course progress via email',
+  },
+  shareEmail: {
+    id: 'learning.social.share.email',
```
Dillon-Dumesnil: Nit: `learning.social.shareEmail`

src/courseware/social-share/messages.js
```diff
@@ -0,0 +1,19 @@
...
+    defaultMessage: 'Share your progress via email.',
+  },
+  shareService: {
+    id: 'learning.social.share.service',
```
Dillon-Dumesnil: Nit: `learning.social.shareService`

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -175,8 +177,15 @@ function CourseCelebration({ intl }) {
         </div>
         <div className="col-12 p-0 font-weight-normal lead text-center">
           {intl.formatMessage(messages.shareHeader)}
+          <SocialIcons
+            analyticsId="edx.ui.lms.course_exit.social_share.clicked"
+            className="mt-2"
+            courseId={courseId}
+            emailSubject={intl.formatMessage(messages.emailSubject, { platform: getConfig().SITE_NAME, title })}
+            socialMessage={messages.socialMessage}
```
Dillon-Dumesnil, line 234: See my below message about combining emailSubject and socialMessage.

If it turns out they cannot be combined, then I'm curious why we bother having the emailSubject and socialMessage passed into the SocialIcons component differently? I understand the emailSubject uses the SITE_NAME where the socialMessage changes based on what service it is, but maybe you pass both in the same was as the socialMessage and the SocialIcons component always does the formatting? I like consistency.

Dillon-Dumesnil: That will change the need for both CourseCelebration and CelebrationModal from needing to pull in `title` to support this and it can just be pulled from the SocialIcons every time. Oh and hey, the title is already pulled in there anyway.

Dillon-Dumesnil (commented):
Looks good! Just a couple random comments. Let me know if you want to chat about any of them :)

src/setupTest.js
ciduarte: Needed this here to mock what happens in `src/index.js` so we have access to the Twitter URL for testing. Without this, the Twitter social button won't render

src/courseware/social-share/SocialIcons.jsx
ciduarte: Yeah this was a workaround because I had tried putting in `messages.defaultEmailBody` into the defaultProps, but it was complaining? I just tried it again and it works!! So I'm going to throw that into the defaultProps and do the translation here instead

src/courseware/social-share/SocialIcons.jsx
ciduarte: Ooh yeah this throws an error because intl looks for an id in socialMessage.

I was considering blocking each icon from rendering based on the messages being supplied... But I'm wondering if the component should be more flexible than that? I don't want to jump the gun and ask product for a default value because we only have two use cases, but maybe an empty string would work? What do you think about `{socialMessage && <FacebookShareButton />}` (not rendering the button at all) versus
`quote={socialMessage ? intl.formatMessage(socialMessage, { platform: getConfig().SITE_NAME, title }) : ''}` (supplying an empty string)?

src/courseware/course/course-exit/CourseCelebration.jsx
ciduarte: Yeah this is a good callout, I changed it so that in all cases (emailBody, emailSubject, socialMessage), you'll pass in the message object, and `<SocialIcons />` will take care of the `intl.formatMessage`

src/courseware/social-share/SocialIcons.jsx
Dillon-Dumesnil: I would prefer the empty string I think. Better to let them share and they fill in their own message vs taking away their ability to share at all. And ideally the link still has all the tracking information we care about anyway.

src/courseware/social-share/SocialIcons.jsx
Dillon-Dumesnil: Awesome, thank ya!

src/setupTest.js
Dillon-Dumesnil: Yeah, this stuff is just weird to me. Like I don't get why they need to exist here when we already define them in .env.test. Like we don't call out LMS_BASE_URL, but everything that uses that still works. Why are these three different? Just doesn't sit well with me

src/courseware/course/course-exit/CourseExit.test.jsx
ciduarte: React Testing Library doesn't support `ByClass` 😕 I could add a `data-testid` to the social-icons div and grab the element `ByTestId`... worth the one liner?

Dillon-Dumesnil: Nah, it's fine

PR #263, opened 2020-10-26
AA-197: Handle non-cert learners that can upgrade

mikix:
Tell them about verified certificates and link to ecommerce.

Also fixes AA-376 by handling the no-verified-mode-to-upgrade-to case (which just shows no box and no footnote at bottom of celebration screen).

Screenshot from 2020-10-26 14-59-24


src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -87,6 +90,10 @@ function CourseCelebration({ intl }) {
 
   let buttonLocation;
   let buttonText;
+  let buttonBackground = 'bg-white';
+  let buttonVariant = 'outline-primary';
+  let certificateImage = certificate;
+  let footnote;
```
mikix, line 96: I've made a fair bit more of this page "configurable". I feel like this page deserves to be broken up a bit? But I didn't want to bite that refactoring off right now (other people are working on this page right now and I think it would muddy these changes themselves).

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -37,14 +39,15 @@ function CourseCelebration({ intl }) {
     certificateData,
     end,
     linkedinAddToProfileUrl,
+    verifiedMode,
```
Dillon-Dumesnil, line 42: Nitty-picky: What are your thoughts on renaming this variable to `courseHasVerifiedMode`?

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -159,6 +169,44 @@ function CourseCelebration({ intl }) {
...
+            { /* todo: remove this hardcoded link to edX support */ }
+            {getConfig().SUPPORT_URL && (
+              <Hyperlink
+                className="text-gray-700"
+                style={{ textDecoration: 'underline' }}
+                destination={`${getConfig().SUPPORT_URL}/hc/en-us/articles/206502008-What-is-a-verified-certificate`}
+              >
+                {intl.formatMessage(messages.verifiedCertificateSupportLink)}
+              </Hyperlink>
+            )}
```
Dillon-Dumesnil, line 198: Any reason you defined this inline vs with the other links above?

src/courseware/course/course-exit/CourseCelebration.jsx
Dillon-Dumesnil, line 96: Why doesn't footnote get a default value? I understand that when the learner can upgrade, we change the messaging with their access expiration, but if they have no access expiration, we don't show them the DashboardFootnote? Why not? I know it has messaging about always being able to access the course, but isn't that what they have since there is no access expiration?

src/courseware/course/sequence/sequence-navigation/UnitNavigation.test.jsx
```diff
@@ -73,8 +73,17 @@ describe('Unit Navigation', () => {
     expect(screen.getByRole('button', { name: /next/i })).toBeEnabled();
   });
 
-  it('has the "Next" button disabled for the last unit in the sequence if there is no Exit Page', () => {
-    render(<UnitNavigation {...mockData} unitId={unitBlocks[unitBlocks.length - 1].id} />);
+  it('has the "Next" button changed for the last unit in the sequence if there is no Exit Page', async () => {
```
Dillon-Dumesnil: Nit: I would keep the test name to saying the "Next" button is disabled since that is what happens. Changed feels different. Like we "change" the button when we show CourseExit material. This is disabling it.

src/courseware/course/course-exit/CourseExit.test.jsx
```diff
@@ -123,6 +128,36 @@ describe('Course Exit Pages', () => {
       expect(screen.queryByRole('img', { name: 'Sample certificate' })).not.toBeInTheDocument();
     });
 
+    it('Displays upgrade link', async () => {
+      // Keep this in sync with "audit only" test below
```
Dillon-Dumesnil: Can you add some more details to these comments about what is meant to be kept in sync?

src/courseware/course/course-exit/CourseExit.test.jsx
```diff
@@ -123,6 +128,36 @@ describe('Course Exit Pages', () => {
...
+      setMetadata({
+        certificate_data: { cert_status: 'audit_passing' },
+        verified_mode: {
+          access_expiration_date: '9999-08-06T12:00:00Z',
+          upgrade_url: 'http://localhost:18130/basket/add/?sku=8CF08E5',
+          price: 600,
+          currency_symbol: '€',
```
Dillon-Dumesnil, line 138: Expensive

src/courseware/course/course-exit/CourseExit.test.jsx
```diff
@@ -123,6 +128,36 @@ describe('Course Exit Pages', () => {
...
+      });
+      await fetchAndRender(<CourseCelebration />);
+      expect(screen.getByText('Upgrade to pursue a verified certificate')).toBeInTheDocument();
+      expect(screen.getByText('For €600 you will unlock access', { exact: false })).toBeInTheDocument();
+      expect(screen.getByRole('link', { name: 'Upgrade now' })).toBeInTheDocument();
+      const node = screen.getByText('Access to this course and its materials', { exact: false });
+      expect(node.textContent).toMatch(/until August 6, 9999\./);
+    });
+
+    it('Displays nothing if audit only', async () => {
```
Dillon-Dumesnil, line 151: Nit: `'Displays nothing if course does not have a certificate mode'`

src/courseware/course/course-exit/CourseExit.test.jsx
```diff
@@ -123,6 +128,36 @@ describe('Course Exit Pages', () => {
       expect(screen.queryByRole('img', { name: 'Sample certificate' })).not.toBeInTheDocument();
     });
 
+    it('Displays upgrade link', async () => {
```
Dillon-Dumesnil: Nit: `'Displays upgrade link if course has certificate mode'`

src/courseware/course/course-exit/messages.js
```diff
@@ -93,6 +97,18 @@ const messages = defineMessages({
...
     defaultMessage: 'You have completed your course. Share your success on social media or email.',
   },
+  upgradeButton: {
+    id: 'courseExit.upgradeButton',
+    defaultMessage: 'Upgrade now',
+  },
+  upgradeLink: {
+    id: 'courseExit.upgradeLink',
+    defaultMessage: 'upgrade now',
+  },
```
Dillon-Dumesnil, line 107: Little upsetting these both can't just be `upgradeMessage` or something because of the `U` vs `u`

src/courseware/course/course-exit/utils.js
```diff
@@ -27,19 +29,28 @@ function getCourseExitMode(courseId) {
...
-    certStatus,
-  } = certificateData;
-  const isEligibleForCertificate = NON_CERTIFICATE_STATUSES.indexOf(certStatus) === -1;
+  // Set defaults for our status-calculated variables, used when no certificateData is provided.
+  // This happens when `get_cert_data` in edx-platform returns None, which it does if we are
+  // in a certificate-earning mode, but the certificate is not available (maybe they didn't pass
+  // or course is not set up for certificates or something). Audit users will always have a
+  // certificateData sent over.
+  let isCelebratoryStatus = true;
+  let isEligibleForCertificate = true;
```
Dillon-Dumesnil, line 42: Wonky. Thanks for the comment.

src/courseware/course/course-exit/CourseCelebration.jsx
Dillon-Dumesnil: Ah, saw below that this isn't a boolean, but an object. Ignore me

Dillon-Dumesnil (approved):
Don't think I have anything hard blocking, but a couple questions around some of the code changes. Fair bit of nits, but you know me, I'm a nit picker.

~Also reminder to myself about checking the `verified_mode` variable takes into account other certificate modes and not just verified courses.~

Right, only matters because we are specifically saying to upgrade into verified. All good.

src/courseware/course/course-exit/CourseCelebration.jsx
mikix: I was trying to keep the cases more contained, rather than bits of each spread throughout the file. I think this class has grown too big and needs to be chunked up a bit. Just trying to help that future effort.

src/courseware/course/course-exit/utils.js
mikix: I am still not 100% on the realistic ways this happens. I _think_ it mostly happens in "bad setup" or corner case situations? But yeah, a weird quirk.

mikix: Or non-passing, I think also. Which is fairly common. And we handle that separately.

Dillon-Dumesnil: I think it might also be possible if a course hasn't "activated" their certificate.

mikix:
> Also reminder to myself about checking the verified_mode variable takes into account other certificate modes and not just verified courses.
>
> Right, only matters because we are specifically saying to upgrade into verified. All good.

`verified_mode` is for any upgradable cert mode, now that I've landed the platform PR. Which can include professional, if there is an audit & professional course.

I think our wording around `verified certificates` is still fine _for users_ though, since professional certs still need some sort of verification yeah? From their pov (not knowing that `verified` is an internal keyword as well), both cert types are verified certificates?

src/courseware/course/course-exit/CourseCelebration.jsx
mikix: Hmm, OK. Well you're asking two different questions:

1) Should footnote have a default value? I still think no. If we get a status we don't understand (no certificate data maybe?), I don't want to have language that might not be true. And in the audit-but-past-the-upgrade-deadline case (`verifiedMode` is null in that case), it's definitely not true.

2) Should the upgrade-deadline-but-no-access-expiration case use the normal DashboardFootnote? I can see that argument. I could add it back for that case.

src/courseware/course/sequence/sequence-navigation/UnitNavigation.test.jsx
mikix: Ah yeah that was a refactor I undid, but didn't change the title back.

src/courseware/course/course-exit/CourseCelebration.jsx
Dillon-Dumesnil: Ah right, past the upgrade deadline is null. I forgot about that. Okay, very well about no default value.

Dillon-Dumesnil:
> > Also reminder to myself about checking the verified_mode variable takes into account other certificate modes and not just verified courses.
> > Right, only matters because we are specifically saying to upgrade into verified. All good.
> 
> `verified_mode` is for any upgradable cert mode, now that I've landed the platform PR. Which can include professional, if there is an audit & professional course.
> 
> I think our wording around `verified certificates` is still fine _for users_ though, since professional certs still need some sort of verification yeah? From their pov (not knowing that `verified` is an internal keyword as well), both cert types are verified certificates?

Not necessarily (no-id-professional is for exactly that case), but I agree, I think verified is fine.

mikix:
> Not necessarily (no-id-professional is for exactly that case), but I agree, I think verified is fine.

Ugh, right. But audit & professional is a rare (deprecated) beast, right? Even more so I'd think for audit & no-id-professional?

Dillon-Dumesnil:
> Ugh, right. But audit & professional is a rare (deprecated) beast, right? Even more so I'd think for audit & no-id-professional?

🤞 I sure hope so

PR #265, opened 2020-10-27
AA-402: send segment events for course-exit

mikix:
Specifically, a "I visited this page" event, with information on which variant was seen. And "I clicked this button" events, for our various calls to action.

https://openedx.atlassian.net/browse/AA-402

### DIFF
--- a/src/courseware/course/course-exit/CourseCelebration.jsx
+++ b/src/courseware/course/course-exit/CourseCelebration.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 
 import {
   FormattedDate, FormattedMessage, injectIntl, intlShape,
@@ -9,7 +9,6 @@ import { useDispatch, useSelector } from 'react-redux';
 import { LinkedinIcon } from 'react-share';
 import { Alert, Button, Hyperlink } from '@edx/paragon';
 import { getConfig } from '@edx/frontend-platform';
-import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 import { getAuthenticatedUser } from '@edx/frontend-platform/auth';
 
 import CelebrationMobile from './assets/celebration_456x328.gif';
@@ -22,6 +21,7 @@ import { requestCert } from '../../../course-home/data/thunks';
 import DashboardFootnote from './DashboardFootnote';
 import UpgradeFootnote from './UpgradeFootnote';
 import SocialIcons from '../../social-share/SocialIcons';
+import { logClick, logVisit } from './utils';
 
 const LINKEDIN_BLUE = '#007fb1';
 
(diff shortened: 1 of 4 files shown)


src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -80,22 +80,16 @@ function CourseCelebration({ intl }) {
...
-      is_staff: administrator,
-      service,
-    });
-  };
-
   let buttonLocation;
   let buttonText;
   let buttonBackground = 'bg-white';
   let buttonVariant = 'outline-primary';
+  let buttonEvent = null;
```
Dillon-Dumesnil, line 88: Why use `= null`? What happens if you use `let buttonEvent;` and it is never set? Undefined would be my guess. Is that okay?

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -80,22 +80,16 @@ function CourseCelebration({ intl }) {
...
   let certificateImage = certificate;
   let footnote;
   let message;
   let title;
+  let visitEvent = 'celebration_generic';
```
Dillon-Dumesnil, line 93: Why set a value here if it is given a value in every possible case?

src/courseware/course/course-exit/utils.js
```diff
@@ -68,4 +70,38 @@ function getCourseExitText(courseId, intl) {
   }
 }
 
-export { COURSE_EXIT_MODES, getCourseExitMode, getCourseExitText };
+// Meant to be used as part of a button's onClick handler.
+// For convenience, you can pass a falsy event and it will be ignored.
```
Dillon-Dumesnil, line 74: Curious, what is the use case for this?

src/courseware/course/course-exit/utils.js
```diff
@@ -68,4 +70,38 @@ function getCourseExitText(courseId, intl) {
...
+  }
+
+  sendTrackEvent(`edx.ui.lms.course_exit.${event}.clicked`, {
+    course_id: courseId,
+    is_staff: administrator,
+  });
+};
+
+// Use like the following to call this only once on initial page load:
+// useEffect(() => logVisit(courseId, administrator, variant), [courseId, administrator, variant]);
```
Dillon-Dumesnil, line 87: :chefkiss:

Dillon-Dumesnil (approved):
LGTM. Couple small questions

src/courseware/course/course-exit/CourseCelebration.jsx
mikix: Yeah unset variables are undefined. I could have left it undefined. I guess I was just vibing a clearer "this page left blank intentionally" indicator.

src/courseware/course/course-exit/CourseCelebration.jsx
mikix: It's not given a value in every possible case. Unknown statuses (like if we get an null certData object) or audit-only end up as `generic`.

src/courseware/course/course-exit/utils.js
mikix: If you have a complicated if/else/switch tree like we have and you might not set the event name for onClick. (In our case, we handle it everywhere. Or so I think. But defensively, we should check for an unset event name. And I think it makes more sense to do that inside the function than outside it.)

src/courseware/course/course-exit/CourseCelebration.jsx
Dillon-Dumesnil: But we gate on unknown statuses in CourseExit.jsx, right? Like you only enter CourseCelebration.jsx iff you have one of the statuses that match with the switch statement below.

src/courseware/course/course-exit/utils.js
Dillon-Dumesnil: Yeah, I'm here for it. Was just curious.

src/courseware/course/course-exit/CourseCelebration.jsx
mikix: No, null certData goes to generic celebration (they are cert learner, but we don't know their status).

Plus, audit-only falls through and gets the generic variant.

Dillon-Dumesnil: Oh dang. 👍

PR #268, opened 2020-11-02
fix: make LTI modals full-screen

davidjoy:
Fixes https://openedx.atlassian.net/browse/TNL-7410

This causes LTI modals in courseware to take up the whole screen.  It does this by creating a new “dialogClassName” value that we then use to override the default heights/widths of the Bootstrap modal.

We also remove the title of the iframe, which just takes up space and detracts from the LTI content.

### DIFF
--- a/src/courseware/course/sequence/Unit.jsx
+++ b/src/courseware/course/sequence/Unit.jsx
@@ -147,14 +147,17 @@ function Unit({
                     allow="microphone *; camera *; midi *; geolocation *; encrypted-media *"
                     frameBorder="0"
                     src={modalOptions.url}
+                    style={{
+                      width: '100%',
+                      height: '100%',
+                    }}
                   />
                 )}
             </>
           )}
-          title={modalOptions.title}
           onClose={() => { setModalOptions({ open: false }); }}
           open
-          dialogClassName="modal-lg"
+          dialogClassName="modal-lti"
         />
       )}
       <div className="unit-iframe-wrapper">
--- a/src/index.scss
+++ b/src/index.scss
@@ -319,6 +319,27 @@ $primary: #1176B2;
   }
 }
 
+// This class forces any modals using 'modal-lti' as their dialogClassName to take up the whole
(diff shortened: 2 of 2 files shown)


src/courseware/course/sequence/Unit.jsx
```diff
@@ -147,14 +147,17 @@ function Unit({
                     allow="microphone *; camera *; midi *; geolocation *; encrypted-media *"
                     frameBorder="0"
                     src={modalOptions.url}
+                    style={{
```
davidjoy, line 150: Necessary to force the iframe to take up the whole modal - surprised we didn't have this before.  Without it the iframe is the default size, which is like... 300x200.

src/courseware/course/sequence/Unit.jsx
```diff
@@ -147,14 +147,17 @@ function Unit({
...
+                      width: '100%',
+                      height: '100%',
+                    }}
                   />
                 )}
             </>
           )}
-          title={modalOptions.title}
```
davidjoy, line 154: This title isn't something we show in the old experience - it's just big text at the top of the modal which takes up space and is meaningless for LTI.

kdmccormick (approved):
Haven't tested it out, but the code is good :+1:

src/index.scss
```diff
@@ -319,6 +319,27 @@ $primary: #1176B2;
...
+.modal-lti {
+  height: 100%;
+  max-width: 100%;
+
+  // I don't like this.  We need to set a height of 100% on a div created by react-focus-on, a
+  // package we use in our Modal.  That div has no class name or ID, so instead we're uniquely
+  // identifying it by based on a unique attribute it has which its siblings don't share.
+  > div[data-focus-lock-disabled=false] {
+    height: 100%;
+  }
```
kdmccormick, line 334: Blech! Nice hack, though, and thanks for the explanation.

awaisdar001 (approved):
👍

PR #270, opened 2020-11-06
AA-356: add more config overrides for branding

mikix:
Remove some edX-specific branding / links (like support URLs) in favor of values from configuration.

Images (sample certificates) are still branded for now. We can get them later.

https://github.com/edx/edx-internal/pull/3604
https://github.com/edx/edge-internal/pull/262


.env
```diff
@@ -15,7 +15,12 @@ ORDER_HISTORY_URL=null
 REFRESH_ACCESS_TOKEN_ENDPOINT=null
 SEGMENT_KEY=null
 SITE_NAME=null
-TWITTER_URL=null
+SOCIAL_UTM_MILESTONE_CAMPAIGN=null
 STUDIO_BASE_URL=
 SUPPORT_URL=null
+SUPPORT_URL_CALCULATOR_MATH=null
+SUPPORT_URL_ID_VERIFICATION=null
+SUPPORT_URL_VERIFIED_CERTIFICATE=null
```
mikix, line 23: I chose to add individual URL settings for these because:
* A site might want to only replace some of them, if they're lazy (like the calculator math one is not really site specific)
* A site might in theory of have these help pages in different places? Like a wiki or a help center or a blog post. I dunno. Didn't want to force everyone into the same help center box we use.
* It's not much harder, just slap the same domain in the front of each.

We still use `SUPPORT_URL` in the header for the generic "Help" menu item.

src/courseware/course/course-sock/CourseSock.jsx
```diff
@@ -150,8 +160,11 @@ export default class CourseSock extends Component {
                       &mdash; <FormattedMessage
                         id="coursesock.upsell.learner"
                         description="Name of learner"
-                        defaultMessage="{ name }, edX Learner"
-                        values={{ name: 'Christina Fong' }}
+                        defaultMessage="{name}, {siteName} Learner"
+                        values={{
+                          name: 'Christina Fong',
+                          siteName: getConfig().SITE_NAME,
+                        }}
```
mikix, line 167: So... is Christina Fong a real person such that this becomes disingenuous on a different site? I had kind of assumed no, but maybe they are.

mikix: @davestgermain I think you wrote this code? Do you remember - are these like real people that we shouldn't misrepresent on random open edx installations?

mikix: Oh sorry I guess I want @davidjoy

davidjoy: Oh, huh.  I may have to punt to @marcotuts to see if he has any background on where these peeps came from.  I wasn't aware of them.

mikix: DJoy mentioned maybe @marcotuts would know?

davidjoy: @marcotuts says these are pseudonyms, so the exact name doesn't matter, but that said, it seems odd that we'd leave this here for Open edX.  I don't know that we have a good way of only including this for edx.org right now, though.  He also says maybe the Sock may go away soon.  It doesn't sound like it's worth investing too much time in getting this "right".

mikix: There are little pics of the people too -- are those stock images or real? But yah, maybe this is fine enough for now if we're just redoing the sock.

ciduarte (approved):
I agree with your decision around individual URLs. Weird stuff happening with that Course Sock... but if it's going away 🤷

PR #275, opened 2020-11-17
Submit events to the tracking logs, not just Segment

stvstnfrd:
Due to an unfortunate naming choice, we apparently had _not_ been
sending navigation events to the tracking logs' `/track` backend, but
just to the Segment.io tracking backend.

Note: There are still other portions of this repo _outside_ the learning
sequence (Course Tools, Celebration, etc.) that only send to Segment and
not the tracking logs, though maybe this is by design, those being "new"
events?

### DIFF
--- a/src/courseware/course/sequence/Sequence.jsx
+++ b/src/courseware/course/sequence/Sequence.jsx
@@ -3,7 +3,10 @@ import React, {
   useEffect, useContext, useState,
 } from 'react';
 import PropTypes from 'prop-types';
-import { sendTrackEvent } from '@edx/frontend-platform/analytics';
+import {
+  sendTrackEvent,
+  sendTrackingLogEvent,
+} from '@edx/frontend-platform/analytics';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import { useSelector } from 'react-redux';
 import { history } from '@edx/frontend-platform';
@@ -69,6 +72,7 @@ function Sequence({
       payload.target_tab = targetIndex + 1;
     }
     sendTrackEvent(eventName, payload);
+    sendTrackingLogEvent(eventName, payload);
   };
 
   const { add, remove } = useContext(UserMessagesContext);


src/courseware/course/sequence/Sequence.jsx
```diff
@@ -69,6 +72,7 @@ function Sequence({
       payload.target_tab = targetIndex + 1;
     }
     sendTrackEvent(eventName, payload);
+    sendTrackingLogEvent(eventName, payload);
```
stvstnfrd, line 75: Is this really all we were missing?

src/courseware/course/sequence/Sequence.jsx
```diff
@@ -69,6 +72,7 @@ function Sequence({
       payload.target_tab = targetIndex + 1;
     }
     sendTrackEvent(eventName, payload);
```
stvstnfrd, line 74: Really unfortunate naming clash/confusion :(

kdmccormick: +1

src/courseware/course/sequence/Sequence.jsx
kdmccormick: I believe so :upside_down_face:

kdmccormick:
@stvstnfrd 
> Note: There are still other portions of this repo outside the learning
sequence (Course Tools, Celebration, etc.) that only send to Segment and
not the tracking logs, though maybe this is by design, those being "new"
events?

From ["What are research events"](https://openedx.atlassian.net/wiki/spaces/AN/pages/13959560/Overview+What+are+Research+Events):
>Research events are high-quality, tested and validated events that we are deliver in regular packages to partner universities for research purposes. Unlike product/BI/debug events, research events are a product that edX has committed to providing externally. They are also used internally as input to the analytics pipeline for reporting needs. 

So Research Events are what we want to send to the tracking logs *and* segment, whereas the product/BI/debug events are should only go to segment. So I guess those course tools / celebration events never "graduated" into being proper research events, so they can just stay in segment. However these sequence navigation events follow a long-standing schema that we purposefully replicated in the MFE.

At least, that's my understanding. I agree that the API names are quite unfortunate here.

kdmccormick (approved):

brianhw:
Thanks for jumping on this so quickly!

PR #276, opened 2020-11-17
Conslidate CI test scripts and add Github Workflow

stvstnfrd:


Makefile
```diff
@@ -52,3 +52,18 @@ pull_translations:
...
 	git diff --exit-code package-lock.json
+
+.PHONY: test
+test:
+	make validate-no-uncommitted-package-lock-changes
+	npm run i18n_extract
+	npm run lint -- --max-warnings 0
+	npm run test
+	npm run build
+	npm run is-es5
```
kdmccormick: terminology nit: consider calling this "validate". from the other edX repos I've worked in, it seems like `test` targets are specifically for unit tests, whereas `validate` targets include tests, quality, and anything else. 

otherwise, this is looking good.

Makefile
```diff
@@ -52,3 +52,18 @@ pull_translations:
 validate-no-uncommitted-package-lock-changes:
 	# Checking for package-lock.json changes...
 	git diff --exit-code package-lock.json
+
+.PHONY: test
+test:
```
davidjoy: Personally I think I'd just skip the makefile here and put these steps in the workflow itself... not sure the indirection buys us much.  I don't think this is a thing that humans ever really want to do...

Makefile
stvstnfrd: @kdmccormick Done, `s/test/validate/`.

Makefile
kdmccormick: I do like being able to easily run everythin" before pushing code sometimes.

Also, it's nice to have the authoritative list of "stuff to validate" outside of a specific CI vendor's schema. I feel bad for repos that have a bunch of logic packed into `.travis.yml` that someone will need to unpack and translate to GA.

davidjoy: Many MFEs/frontend libraries repos will put that list of tests/validations in a commit hook, defining it in package.json, fwiw.  ¯\_(ツ)_/¯

davidjoy: Er, to be clear, I don't feel strongly about this. :D

kdmccormick: While we're bikeshedding, the first thing I do after checking out an MFE is disable the pre-commit hooks. Let me run the validation suite on my own terms!

But yeah, I don't feel strongly either. However it ends up, I'll find a way to make the tooling work for me :stuck_out_tongue:

davidjoy (approved):

PR #279, opened 2020-11-20
AA-303: Celebrate a user's first ever discussion post!

Dillon-Dumesnil:


src/courseware/course/Course.jsx
```diff
@@ -10,7 +10,7 @@ import useOfferAlert from '../../alerts/offer-alert';
 
 import Sequence from './sequence';
 
-import { CelebrationModal, shouldCelebrateOnSectionLoad } from './celebration';
+import { FirstSectionCelebrationModal, shouldCelebrateOnSectionLoad } from './celebration';
```
mikix, line 13: Yeah, that was an awfully broad name I was taking with that modal.  :P

src/courseware/course/celebration/messages.js
```diff
@@ -9,6 +9,18 @@ const messages = defineMessages({
     id: 'learning.celebration.congrats',
     defaultMessage: 'Congratulations!',
   },
+  conversation: {
+    id: 'learning.celebration.conversation',
+    defaultMessage: 'Nice job being part of the conversation.',
+  },
+  discussionBodyText1: {
+    id: 'learning.celebration.discussionBodyText1',
+    defaultMessage: "Learners who participate in discussions are 10x more likely to complete their course than those who don't.",
```
mikix: nit: You know how I get about apostrophes. On the next message too.

src/courseware/course/celebration/utils.jsx
```diff
@@ -16,18 +16,34 @@ function handleNextSectionCelebration(sequenceId, nextSequenceId, nextUnitId) {
...
+
+  // Tell our analytics
+  sendCelebrationSegmentEvent(courseId, 'edx.ui.lms.celebration.first_section.opened');
+}
+
+function recordFirstDiscussionCelebration(courseId) {
+  /* postCelebrationComplete should start being used once the Discussion MFE exists
+     and we no longer record the first discussion post in the JS handler in edx-platform
+     See edx-platform/common/static/common/js/discussion/views/new_post_view.js and
+     edx-platform/cms/static/common/js/discussion/views/discussion_thread_view.js */
```
mikix, line 39: I haven't read that code yet, but just bookmarking this because my gut thought is "why not move that recording here anyway, since this is when we know the user saw it" -- but I'm sure that thought will be soothed by further reading.

But this is a helpful comment regardless.

src/courseware/course/celebration/utils.jsx
```diff
@@ -63,4 +79,16 @@ function shouldCelebrateOnSectionLoad(courseId, sequenceId, unitId, celebrateFir
   return shouldCelebrate;
 }
 
-export { handleNextSectionCelebration, recordFirstSectionCelebration, shouldCelebrateOnSectionLoad };
+function shouldCelebrateOnDiscussionPost(firstDiscussion, firstDiscussionUserBucket) {
+  // Bucket 0 === Control group which does not get the discussion celebration.
+  // That check can be removed when we stop using the flag as an experiment.
+  return firstDiscussion && firstDiscussionUserBucket === 0;
```
mikix: My gut reaction was to not love this bucket variable. But it is a good solution.

src/courseware/course/celebration/FirstDiscussionCelebrationModal.jsx
```diff
@@ -0,0 +1,91 @@
...
+      onClose={() => {
+        // Update our local copy of course data from LMS
+        console.log('updating model');
+        dispatch(updateModel({
+          modelType: 'courses',
+          model: {
+            id: courseId,
+            celebrations: {
+              firstDiscussion: false,
+            },
```
mikix, line 68: This feels a bit boundary-leaky to me. Like maybe this should emit a signal and the component that holds this should turn off a state variable?

The other modal doesn't do this, but it also has some weirdly complex local storage solution for this problem, which it needed because different pages wrote and read the value. So probably not a good role model (role modal? eh?) here.

src/courseware/course/celebration/FirstDiscussionCelebrationModal.jsx
```diff
@@ -0,0 +1,91 @@
...
+
+  useEffect(() => {
+    if (open) {
+      recordFirstDiscussionCelebration(courseId);
+    }
+  }, [open]);
+
+  let normativeDataBodyText;
+  // Bucket 2 corresponds to showing normative data in the body text
+  if (firstDiscussionUserBucket === 0) {
```
mikix: Should be `=== 2` yeah?

mikix: (Remind me again why we aren't using different buckets for the two texts? Seems like bad experiment design to not be able to tell which of them is better?)

mikix (approved):
Looks nice. Didn't run it, but just scanned code.

src/courseware/course/celebration/utils.jsx
Dillon-Dumesnil: Reminder for myself to make sure this switches to `!== 0` since it should not show in the 0 case.

src/courseware/course/celebration/FirstDiscussionCelebrationModal.jsx
Dillon-Dumesnil: Yeah, it should. This is one of the things that I have flipped right now for testing purposes.

And I believe because the experiment isn't about _which_ text does better, but rather if the text itself does better. So we have control group (no celly), experiment 1 (celly, no normative), and experiment 2 (celly with normative). Then we just randomize the text.
I will say, it does feel a bit weird to me too because if one of the texts out performs, we would probably want to use that everytime, but that's just not how the experiment is designed. I remember asking about that too, but it was pushed back on /shrug

src/alerts/enrollment-alert/messages.js
```diff
@@ -18,7 +18,7 @@ const messages = defineMessages({
   },
   success: {
     id: 'learning.enrollment.success',
-    defaultMessage: "You've successfully enrolled in this course!",
+    defaultMessage: 'You’ve successfully enrolled in this course!',
```
Dillon-Dumesnil, line 21: After MT's comment about using apostrophe's, I decided to go and update the other uses in this repo. Do y'all know if this will trigger it to be wiped on the translation side of things.

mikix: Naw - you can change `defaultMessage` all day, and as long as you don't change the `id`, existing translations aren't thrown out. Which is also why you have to be careful changing `defaultMessage` to mean something different without changing `id` too.

Dillon-Dumesnil: That's wild. Glad for this scenario though!

openedx-webhooks:
Although this pull request is already closed, I've created [OSPR-6549](https://openedx.atlassian.net/browse/OSPR-6549) so that we can track it in Jira.

PR #281, opened 2020-11-20
AA-131: Landing page for anonymous or un-enrolled users

ciduarte:
**Logged Out, LMS vs MFE**
<div>

_Private Outline_
<div>
AA-131_LMS_LoggedOut
AA-131_LoggedOut_PrivateOutline_wBrand
</div>

_Public Course_ (no alerts, outline links available, widgets available if populated)
<div>
AA-131_LMS_LoggedOut_PublicCourse
AA-131_LoggedOut_PublicCourse_wBrand
</div>
</div>

**Logged in learner who is able to enroll, LMS vs MFE**

_Private Course_
<div>
AA-131_LMS_LoggedIn_CanEnroll_PrivateCourse
AA-131_LoggedIn_CanEnroll_PrivateOutline_wBrand
</div>

_Public Course_
<div>
AA-131_LMS_LoggedIn_CanEnroll_PublicCourse
AA-131_LoggedIn_CanEnroll_PublicCourse_wBrand
</div>

**Logged in learner who is not able to enroll (copy matches LMS, forgot to screenshot w/ Brand.. oops)**
AA-131_LoggedIn_CannotEnroll

**Courseware support for public course**
AA-131_LoggedOut_PublicCourse_Courseware


src/alerts/logistration-alert/LogistrationAlert.jsx
```diff
@@ -2,31 +2,38 @@ import React from 'react';
...
+import { Hyperlink } from '@edx/paragon';
 
 import { Alert } from '../../generic/user-messages';
 import messages from './messages';
 
 function LogistrationAlert({ intl }) {
   const signIn = (
-    <a href={`${getLoginRedirectUrl(global.location.href)}`}>
+    <Hyperlink
+      style={{ textDecoration: 'underline' }}
```
ciduarte, line 13: MT has made a point about condensing these underlined Hyperlinks into a component because we use them everywhere. Some future Paragon release (if this isn't already out as I'm typing this) _should_ have an inline link class name that we can replace all of the redundant styling with.

src/alerts/private-course-alert/PrivateCourseAlert.jsx
```diff
@@ -0,0 +1,104 @@
...
+      className="p-0 border-0 align-top"
+      style={{ textDecoration: 'underline' }}
+      onClick={enrollClickHandler(courseId)}
+    >
+      {intl.formatMessage(enrollmentMessages.enroll)}
+    </Button>
+  );
+
+  // TODO: Pull this registration URL building out into a function, like the login one above.
+  // This is complicated by the fact that we don't have a REGISTER_URL env variable available.
```
ciduarte: Stole this from the existing LogistrationAlert... should I create a REGISTER_URL env variable? Haven't looked into whether this is a quick fix.

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -31,12 +30,6 @@ function OutlineTab({ intl }) {
 
   const {
     title,
-    start,
-    end,
-    enrollmentStart,
-    enrollmentEnd,
-    enrollmentMode,
-    isEnrolled,
```
ciduarte, line 42: These were left here from early days, we weren't using them 👋

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -65,29 +58,18 @@ function OutlineTab({ intl }) {
   const [goalToastHeader, setGoalToastHeader] = useState('');
   const [expandAll, setExpandAll] = useState(false);
 
-  // Above the tab alerts (appearing in the order listed here)
-  const logistrationAlert = useLogistrationAlert();
-  const enrollmentAlert = useEnrollmentAlert(courseId);
-
```
ciduarte, line 77: Moved these into `<LoadedTabPage />` to be above the tabs as in LMS

src/alerts/enrollment-alert/hooks.js
```diff
@@ -30,10 +32,10 @@ export function useEnrollmentAlert(courseId) {
   return { clientEnrollmentAlert: EnrollmentAlert };
 }
 
-export function useEnrollClickHandler(courseId, successText) {
+export function useEnrollClickHandler(courseIdMemo, successText) {
   const [loading, setLoading] = useState(false);
   const { addFlash } = useContext(UserMessagesContext);
-  const enrollClickHandler = useCallback(() => {
+  const enrollClickHandler = useCallback((courseId) => () => {
```
ciduarte: `courseId` was originally out of scope, which is why it was broken in our bug bash

src/generic/messages.js
```diff
@@ -6,6 +6,26 @@ const messages = defineMessages({
     defaultMessage: 'Close',
     description: 'Text used as an aria-label to describe closing or dismissing a component',
   },
+  registerLowercase: {
```
ciduarte, line 9: Not sure if there's a better way to handle/name these...

src/alerts/logistration-alert/messages.js
```diff
@@ -1,16 +0,0 @@
...
-    id: 'learning.logistration.login',
-    defaultMessage: 'sign in',
-    description: 'Text in a link, prompting the user to log in.  Used in "learning.logistration.alert"',
-  },
-  register: {
-    id: 'learning.logistration.register',
-    defaultMessage: 'register',
-    description: 'Text in a link, prompting the user to create an account.  Used in "learning.logistration.alert"',
-  },
-});
```
ciduarte, line 14: Moved these to `generic/messages.js`

src/course-home/data/api.js
```diff
@@ -7,7 +7,7 @@ function normalizeCourseHomeCourseMetadata(metadata) {
   return {
     ...data,
     tabs: data.tabs.map(tab => ({
-      slug: tab.tabId,
+      slug: tab.tabId === 'courseware' ? 'outline' : tab.tabId,
```
ciduarte, line 12: The "Course" tab wasn't active while in course home because our course metadata calls it "courseware"

src/alerts/enrollment-alert/EnrollmentAlert.jsx
```diff
@@ -33,8 +33,8 @@ function EnrollmentAlert({ intl, payload }) {
   }
 
   const button = canEnroll && (
-    <Button disabled={loading} variant="link" className="p-0 border-0 align-top" onClick={enrollClickHandler}>
-      {intl.formatMessage(messages.enroll)}
+    <Button disabled={loading} variant="link" className="p-0 border-0 align-top" style={{ textDecoration: 'underline' }} onClick={enrollClickHandler(courseId)}>
+      {intl.formatMessage(messages.enroll)}.
```
mikix: That should probably go inside the string to be translated. I believe fixed-width languages like Chinese use fixed-width versions of a lot of punctuation.

．vs .

src/alerts/enrollment-alert/hooks.js
```diff
@@ -11,13 +12,14 @@ import { postCourseEnrollment } from './data/api';
 const EnrollmentAlert = React.lazy(() => import('./EnrollmentAlert'));
 
 export function useEnrollmentAlert(courseId) {
+  const { authenticatedUser } = useContext(AppContext);
```
mikix, line 15: Is this a different thing than `import { getAuthenticatedUser } from '@edx/frontend-platform/auth';`?

I see both in the codebase.

src/alerts/enrollment-alert/hooks.js
mikix: Oof my mistake.

src/alerts/private-course-alert/PrivateCourseAlert.jsx
mikix: I think the comment in `LogistrationAlert` was talking about adding a `getRegisterRedirectUrl` to frontend-platform/auth, like they have a `getLoginRedirectUrl` -- but that they might want a REGISTER_URL first, to base that function off of? Adding API to our auth library doesn't sound like a quick fix, if I'm reading that comment right.

src/course-header/Header.jsx
```diff
@@ -84,36 +86,57 @@ function Header({
...
-              <Dropdown.Item href={getConfig().ORDER_HISTORY_URL}>
-                {intl.formatMessage(messages.orderHistory)}
-              </Dropdown.Item>
-            )}
-            <Dropdown.Item href={getConfig().LOGOUT_URL}>
-              {intl.formatMessage(messages.signOut)}
-            </Dropdown.Item>
-          </Dropdown.Menu>
-        </Dropdown>
+        {authenticatedUser && (
```
mikix, line 67: It might be cleaner / easier-to-read-this-component, if these two dropdowns were broken into their own components like `AuthenticatedUserDropdown` and `AnonymousUserDropdown` or something like that. But just an optional nit.

src/index.jsx
```diff
@@ -96,7 +96,7 @@ initialize({
   },
   // TODO: Remove this once the course blocks api supports unauthenticated
   // access and we are prepared to support public courses in this app.
-  requireAuthenticatedUser: true,
+  requireAuthenticatedUser: false,
```
mikix: Probably should drop the comment. And you can also drop the setting at all, since it defaults to `false`.

mikix (approved):
Nice! As mentioned in slack, it would also be good to confirm that `public` and `public_outline` visibility levels work like they should.

(public_outline -- just outline, but they don't have links)
(public -- outline with links and handouts is also visible)

src/course-header/Header.jsx
ciduarte: Yeah that's a good idea

src/alerts/enrollment-alert/hooks.js
ciduarte: That's a good question... Both calls seem to return identical objects

Both AppContext and getAuthenticatedUser are coming from edx/frontend-platform, so I don't know if one has an advantage over the other in terms of efficiency or source of truth-ness.. Might be why we use them interchangeably 🤔

src/alerts/private-course-alert/PrivateCourseAlert.jsx
ciduarte: Ooh I see that, you're right. I skipped past the word "function" and went straight to "we should just add the env variable then"

I will not be doing that

src/alerts/enrollment-alert/EnrollmentAlert.jsx
```diff
@@ -33,8 +33,8 @@ function EnrollmentAlert({ intl, payload }) {
   }
 
   const button = canEnroll && (
-    <Button disabled={loading} variant="link" className="p-0 border-0 align-top" onClick={enrollClickHandler}>
-      {intl.formatMessage(messages.enroll)}
+    <Button disabled={loading} variant="link" className="p-0 border-0 align-top" style={{ textDecoration: 'underline' }} onClick={enrollClickHandler(courseId)}>
```
davidjoy: I think there's a bug here - `enrollClickHandler(courseId)` will call the handler and return `null`, setting the `onClick` handler to `null` but evaluating the `enrollClickHandler` on every re-render.

To make this work as written, you'd need to put a closure around this so that there's a function for `onClick` to evaluate when it's triggered:

```
onClick={() => enrollClickHandler(courseId)}
```

The above is the fix.  Everything below this is about the "right" way to do it:

Note that adding closures like that is generally a bad idea if they're being passed to a child component (like `Button`), as it means every re-render of `EnrollmentAlert` will force a re-render of the `Button`, since you're passing in a new function definition on every re-render.  It's not a big deal here since it won't happen that often, but this is the problem that `useCallback` fixes.

To fix it, you'd have to declare your helpful little closure as its own useCallback hook:

```
const enrollCourseClickHandler = useCallback(
  () => enrollClickHandler(courseId), 
  [enrollClickHandler, courseId]
);

// ... later...

onClick={enrollCourseClickHandler}
```

_Then_ you won't be re-rendering the Button needlessly. :)   

(Functional components are a pain in the butt, IMHO)

src/alerts/enrollment-alert/hooks.js
davidjoy: In this case, the `AppContext` is better because it means that if `authenticatedUser` changes, the function will be re-rendered.  `getAuthenticatedUser` is better outside the react lifecycle.

davidjoy (changes requested):
Er, submitting so these comments will show up.  Still reviewing. ;)

src/alerts/enrollment-alert/hooks.js
```diff
@@ -46,7 +56,7 @@ export function useEnrollClickHandler(courseId, successText) {
       setLoading(false);
       global.location.reload();
     });
-  }, [courseId]);
+  }, [courseIdMemo]);
```
davidjoy: Curious about why this `courseIdMemo` is necessary.  This whole alerting on data-state situation is so complicated. (criticism of myself)

src/alerts/private-course-alert/PrivateCourseAlert.jsx
```diff
@@ -0,0 +1,115 @@
...
+      style={{ textDecoration: 'underline' }}
+      destination={`${getLoginRedirectUrl(global.location.href)}`}
+    >
+      {intl.formatMessage(genericMessages.signInSentenceCase)}
+    </Hyperlink>
+  );
+
+  return (
+    <Alert type="welcome">
+      {anonymousUser && (
```
davidjoy, line 65: I'm assuming codecov is saying there's no test for this clause...

src/course-header/AnonymousUserDropdown.jsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+
+import { getConfig } from '@edx/frontend-platform';
+import { getLoginRedirectUrl } from '@edx/frontend-platform/auth';
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import { Button } from '@edx/paragon';
+
+import genericMessages from '../generic/messages';
+
+function AnonymousUserDropdown({ intl }) {
```
davidjoy: `Dropdown` seems like a misnomer for this thing... Is it maybe just a `Menu`?  It's just two buttons side-by-side, right?

src/course-home/data/api.js
davidjoy: Should we normalize this at the API layer?  The risk is that we end up introducing the same bug because someone used the `tabId` elsewhere and didn't see this.  The "normalize" functions in the api.js files exist as an "anti-corruption layer" that normalizes data at the "edge" of the app and ensures that we can use it with confidence and with a consistent taxonomy within the bulk of the app.  Worth considering, perhaps!

davidjoy: I wrote something here about normalizing this in api.js and didn't realize I was staring at api.js 😀

davidjoy: Maybe worth a comment here to describe why this is necessary, for posterity.

src/course-home/outline-tab/OutlineTab.jsx
davidjoy: Thank goodness.  RIP.  

Turns out I was wrong and you all _didn't_ need all that, huh?  Hah!

davidjoy (approved):
So are there any pages in the learning MFE that still require authentication and should automatically redirect to login if an anonymous user tries to access them?

A few comments, but in general looks great.

src/alerts/enrollment-alert/EnrollmentAlert.jsx
ciduarte: Thanks for this explanation! This actually worked as intended because I used currying in the enrollClickHandler, which was only setting `courseId` and returning the rest of the callback function.

```
const enrollClickHandler = useCallback((courseId) => () => {
    return "something";
}

enrollClickHandler(courseId);
>> () => { return "something" }

enrollClickHandler(courseId)();
>> "something"
```

_But_, I did this because we found that the enrollClickHandler wasn't working in a bug bash, and through local testing it seemed like `courseId` was out of scope (which led to this currying and `courseIdMemo` madness)... Now that I'm sanity checking, it seems like this might have been working in the first place? Aliens??? So my changes here and to enrollClickHandler are seemingly unnecessary... _sigh_

src/alerts/enrollment-alert/hooks.js
ciduarte: Revisiting this, I think courseIdMemo is an unnecessary change. Longer explanation in a previous comment 🙂

src/course-header/AnonymousUserDropdown.jsx
ciduarte: Ooh yeah good catch!

mikix (approved):

PR #283, opened 2020-11-23
Update MFE to use the new “brand-openedx” repository.

davidjoy:
This will allow the MFE to be re-branded by overriding this default implementation.  More detail here:

https://github.com/edx/brand-openedx

To test this branch locally with your own theme (perhaps brand-edx.org):

1. `npm install` so you have the most recent dependencies.
2. As a peer of `frontend-app-learning`:

```git clone git@github.com:edx/brand-edx.org.git```
```git clone git@github.com:edx/paragon.git``` (if you haven't already)

3. In frontend-app-learning, create a `module.config.js` file in the root of the repository with these contents:

```
module.exports = {
  /*
  Modules you want to use from local source code.  Adding a module here means that when this app
  runs its build, it'll resolve the source from peer directories of this app.

  moduleName: the name you use to import code from the module.
  dir: The relative path to the module's source code.
  dist: The sub-directory of the source code where it puts its build artifact.  Often "dist".
  */
  localModules: [
    { moduleName: '@edx/brand', dir: '../brand-edx.org' },
    { moduleName: '@edx/paragon/scss/core', dir: '../paragon', dist: 'scss/core' },
    { moduleName: '@edx/paragon', dir: '../paragon', dist: 'dist' },
  ],
};
```

4. Start your dev server.  `npm start`

You should see the MFE load with the brand-edx.org brand, the most noticeable difference being that the spinner and headings will render with the "elm" color - a dark green.

### CODE REVIEW

**Review by ciduarte — APPROVED** [review]
Looked through Engage's live features:
- Dates Tab (including reset dates banner)
- Reset dates banner within courseware
- Celebration modal

InstructorToolbar light buttons were brought to David's attention. Everything else is functional! 👍

File: .env.development [hunk 1]
```diff
@@ -9,6 +9,10 @@ LANGUAGE_PREFERENCE_COOKIE_NAME='openedx-language-preference'
 LMS_BASE_URL='http://localhost:18000'
 LOGIN_URL='http://localhost:18000/login'
 LOGOUT_URL='http://localhost:18000/logout'
+LOGO_URL=https://edx-cdn.org/v3/default/logo.svg
+LOGO_TRADEMARK_URL=https://edx-cdn.org/v3/default/logo-trademark.svg
+LOGO_WHITE_URL=https://edx-cdn.org/v3/default/logo-white.svg
+FAVICON_URL=https://edx-cdn.org/v3/default/favicon.ico
```
kdmccormick, line 15: [for my own understanding] How does `brand-${XYZ}` override these?

package.json
```diff
@@ -36,7 +36,6 @@
   "dependencies": {
     "@edx/brand": "npm:@edx/brand-openedx@1.1.0",
     "@edx/frontend-component-footer": "10.0.11",
-    "@edx/frontend-component-header": "2.0.6",
```
kdmccormick, line 38: [kudos] As always, thanks for breaking these changes into logical commits!

src/courseware/data/api.js
```diff
@@ -158,6 +158,15 @@ function normalizeSequenceMetadata(sequence) {
...
       title: sequence.display_name,
+      /*
+      Example structure of gated_content when prerequisites exist:
+      {
+        prereq_id: 'id of the prereq section',
+        prereq_url: 'unused by this frontend',
+        prereq_section_name: 'Name of the prerequisite section',
+        gated: true,
+        gated_section_name: 'Name of this gated section',
+      */
```
kdmccormick, line 169: :clap:

.env.test
```diff
@@ -4,10 +4,15 @@ BASE_URL='http://localhost:2000'
 CREDENTIALS_BASE_URL='http://localhost:18150'
 CSRF_TOKEN_API_PATH='/csrf/api/v1/token'
 ECOMMERCE_BASE_URL='http://localhost:18130'
+ENTERPRISE_LEARNER_PORTAL_HOSTNAME='localhost:8734'
 LANGUAGE_PREFERENCE_COOKIE_NAME='openedx-language-preference'
 LMS_BASE_URL='http://localhost:18000'
 LOGIN_URL='http://localhost:18000/login'
 LOGOUT_URL='http://localhost:18000/logout'
+LOGO_URL=https://edx-cdn.org/v3/default/logo.svg
+LOGO_TRADEMARK_URL=https://edx-cdn.org/v3/default/logo-trademark.svg
```
kdmccormick, line 13: [callout] Looks like we've replaced the "Powered by Open edX" logo in the footer with the same overlapping gray circles that are in the header. If that was intentional, please disregard this :)

package.json
```diff
@@ -34,11 +34,11 @@
     "url": "https://github.com/edx/frontend-app-learning/issues"
   },
   "dependencies": {
-    "@edx/frontend-component-footer": "10.0.11",
-    "@edx/frontend-component-header": "2.0.6",
+    "@edx/brand": "npm:@edx/brand-openedx@1.1.0",
```
kdmccormick, line 37: [callout] Looks like we've gone back to the royal blue that the MFE originally had before we switched it to pale blue in order to match the legacy experience. Are we going back to royal blue intentionally? If so, sounds good to me.

[critique, non-blocking] an effect of this change is that the navigation links no longer match the in-XBlock links.

master:
image


this branch:
image

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -24,7 +25,7 @@ import UpgradeFootnote from './UpgradeFootnote';
 import SocialIcons from '../../social-share/SocialIcons';
 import { logClick, logVisit } from './utils';
 
-const LINKEDIN_BLUE = '#007fb1';
+const LINKEDIN_BLUE = '#2867B2';
```
kdmccormick, line 28: [trying to understand] I assume LinkedIn changed its color scheme, and this is just us catching up?

package.json
```diff
@@ -34,11 +34,11 @@
...
+    "@edx/frontend-component-footer": "^10.1.0",
     "@edx/frontend-enterprise": "4.2.3",
-    "@edx/frontend-platform": "1.5.4",
-    "@edx/paragon": "12.0.5",
+    "@edx/frontend-platform": "1.8.0",
+    "@edx/paragon": "^12.3.1",
```
kdmccormick, line 41: image
image

[critique, non-blocking] These drop-downs feel cramped. Some more spacing between the text and th downward arrow may look more appealing.

package.json
kdmccormick, line 37: [critique] The slashes separating the breadcrumbs seem too light on the white background, especially compared to the darkness of the text. This might be an a11y issue.

master:
image

this branch:
image

kdmccormick (approved):
I have a few questions and comments, but the only one I'd consider "blocking" is the slashes in the breadcrumbs. Even then, given the urgency of this, I'm good you merging this as-is at your discretion :+1: 

(this set of comments is for the openedx theme; I reviewed the edx.org theme separately)

package.json
kdmccormick: actually, I'm realizing that all secondary chrome text seems too light:

image

As a user, I might think that Progress/Discussion/Notes/Instructor are disabled until I hovered over them.

kdmccormick (commented):
I tried it out with the edx.org theme. Looks pretty good!

Things that look better in the edx.org theme than openedx the theme:
* The spacing in the drop-downs looks better.
* The links in the breadcrumbs seem *closer* in color the the in-XBlock links, although not identical. You'd probably have to be looking for the difference to spot it.

Things that seem amiss with the edx.org theme:
* It's still using the openedx placeholder logo (although maybe that's just a local configuration issue?)
* The secondary text color still strikes me as being too light against the white background.

image

.env.development
davidjoy: It doesn't - the environment-specific variables being supplied to a build override them.  In edx.org's case, this is based on our edx internal variables.  For an Open edX repository, they'd supply these on the command line manually (along with everything else).

.env.test
davidjoy: Yes, I think so?  That was done in a separate PR in frontend-component-footer.  @adamstankiewicz may be able to shed some light on that.

src/courseware/course/course-exit/CourseCelebration.jsx
davidjoy: @ciduarte - can you shed any light here?

.env.development
kdmccormick: Got it; thanks.

package.json
davidjoy: So the "royal blue" is "Bootstrap blue" - the default Open edX theme now uses Bootstrap defaults.  So that's the why, anyway.  The inconsistent coloring is a good callout... this is the result of edx-platform continuing to have the existing edX theme coloring, while the MFE is now default Bootstrap.  

I think we need to get edx-platform to use a similar Bootstrap-blue based theme to fix this, which may have to wait until the rebrand is through to be fixed.  This isn't a great Open edX experience.  I think I'll make a task about it.

https://openedx.atlassian.net/browse/TNL-7772

PR #286, opened 2020-12-01
AA-476: Send org_key with tracking events

mikix:
Also convert the course tool click event from accidentally being a segment event into a log event, like it is on the LMS.

https://openedx.atlassian.net/wiki/spaces/DE/pages/2097119925/Improved+Event+Data+Fields+-+FY21Q2+Q3

### DIFF
--- a/src/courseware/social-share/SocialIcons.jsx
+++ b/src/courseware/social-share/SocialIcons.jsx
@@ -31,6 +31,7 @@ function SocialIcons({
 }) {
   const {
     marketingUrl,
+    org,
     title,
   } = useModel('courses', courseId);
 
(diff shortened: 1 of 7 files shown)


src/courseware/social-share/SocialIcons.jsx
```diff
@@ -48,7 +49,9 @@ function SocialIcons({
 
     const { administrator } = getAuthenticatedUser();
     sendTrackEvent(analyticsId, {
-      course_id: courseId,
+      org_key: org,
+      courserun_key: courseId,
+      course_id: courseId, // should only be courserun_key, but left as-is for historical reasons
```
ciduarte, line 54: Will we be able to remove this in the future? Or will we be sending two courseId's forever

ciduarte (approved):
One small question, LGTM

src/courseware/social-share/SocialIcons.jsx
mikix: I don't know - that would be DE's call I guess, I don't know how they are merging data, or how long back they need/want to go?
I asked them if I should send both, and they were like, yes please. So :shrug:

doctoryes: Having both fields in the event will help in making an asynchronous shift over to using the new field in various event-processing pipelines. The need for both fields won't be forever - but it's difficult to say how long they'll be needed.

PR #289, opened 2020-12-04
AA-383: add outline sidebar upgrade card

mikix:
This PR adds a sidebar upgrade card.
- I've tried to remain extremely faithful to the LMS experience, since I know this is a high traffic item and even seemingly small changes to upgrade visuals and language can have a noticable effect.
- This is not gated by a waffle flag, although the old LMS one is. The waffle flag is unconditionally on in the LMS (not even course-level opt-outs). So it was probably just a rollout flag and should be removed now? But I didn't want to take that on. But going forward with the MFE, this won't be gated.
- I've also added the course sock to the outline (as it is shown in the LMS). This was needed to power the "Learn More" button, which scrolls down to the course sock.
- Event names and properties were taken from the LMS. One difference is that the LMS uses the `courseRunKey` property and that's not possible with frontend-platform's analytics library. It forces everything to snake case (`course_run_key`). So I just use the new-style property name (`courserun_key`) instead, if it's going to be different anyway.

Related to https://github.com/edx/edx-platform/pull/25765


mikix:
@ekangedx @hilaryg123 - FYI on screenshots above. I've tried to not change anything from LMS, but let me know if I should.

src/course-home/outline-tab/widgets/UpgradeCard.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+import { sendTrackingLogEvent } from '@edx/frontend-platform/analytics';
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import { Button } from '@edx/paragon';
+
+import messages from '../messages';
+import { useModel } from '../../../generic/model-store';
+import VerifiedCert from '../../../generic/assets/edX_verified_certificate.png';
+
+function UpgradeCard({ courseId, intl, onLearnMore }) {
```
Dillon-Dumesnil, line 12: Is there a convention for where to place `intl` in the list of props? Like you supply courseId and onLearnMore so I personally prefer those coming first, but /shrug

src/course-home/outline-tab/widgets/UpgradeCard.jsx
```diff
@@ -0,0 +1,81 @@
...
+  const logClick = () => {
+    sendTrackingLogEvent('edx.bi.course.upgrade.sidebarupsell.clicked', eventProperties);
+    sendTrackingLogEvent('edx.course.enrollment.upgrade.clicked', {
+      location: 'sidebar-message',
+    });
+  };
+
+  return (
+    <section className="mb-4 p-4 outline-sidebar-upgrade-card">
+      <h2 className="h6">{intl.formatMessage(messages.upgradeTitle)}</h2>
```
Dillon-Dumesnil: Should we have className `h6` with the tag `h2`? Feels dirty, but I don't claim much knowledge here.

src/courseware/course/Course.jsx
```diff
@@ -83,7 +83,7 @@ function Course({
           open
         />
       )}
-      {canShowUpgradeSock && verifiedMode && <CourseSock verifiedMode={verifiedMode} />}
+      {canShowUpgradeSock && <CourseSock verifiedMode={verifiedMode} />}
```
Dillon-Dumesnil, line 86: 👏

Dillon-Dumesnil (approved):
All LGTM. Couple small questions regarding style

src/course-home/outline-tab/widgets/UpgradeCard.jsx
mikix: Fair question. I was alphabetically sorting. And it sounds like you were expecting a kind of "importance" sorting where we put the auto-generated-props at the end?

I guess I just love all my props equally, while you play favorites. But seriously, I don't have a strong pref which approach we take.

src/course-home/outline-tab/widgets/UpgradeCard.jsx
mikix: I thought about that too! But this was an approach that I copied from elsewhere, and I suspected that an a11y pass had already been done where like, maybe we shouldn't skip header levels or something? @ciduarte was I merely cargo culting and we should just use `h6` or did this approach maybe come from Jeff Witt?

ciduarte: I believe this is what we want here. The `className="h6"` is merely for styling, while the `<h2>` is for a11y purposes!

Random info: You might see this done differently in other spots because there's a certain className that can't override the element, I forget what the actual scenario is... but it looks like `<div role="heading" aria-level="[some level]">`. I believe this is a less favorable approach because we'd rather use the `<h>` tags, but it was our only option to get the styling we wanted from Paragon.

src/course-home/outline-tab/widgets/UpgradeCard.jsx
ciduarte: I prefer alphabetical FWIW :)

src/course-home/outline-tab/widgets/UpgradeCard.jsx
```diff
@@ -0,0 +1,81 @@
...
+            symbol: verifiedMode.currencySymbol,
+          })}
+        </Button>
+        {onLearnMore && (
+          <Button
+            variant="link"
+            size="sm"
+            onClick={onLearnMore}
+          >
+            {intl.formatMessage(messages.learnMore)}
```
ciduarte, line 64: "Learn more" as the only text is a bit ambiguous. I believe our options are:
1. Add an ID to the `<h2>` upgrade title, and then add an `aria-labelledby=""` to this button
2. Add screen reader only text following the "learn more" text in this button

I think I prefer option 1..

ciduarte (approved):
LGTM, one comment about a11y. Also, how does the responsive/mobile view look? Assuming it's the same, but wanted to double check

src/course-home/outline-tab/widgets/UpgradeCard.jsx
mikix: Fixed by going with option 1! Thanks for suggestion.

mikix:
> Also, how does the responsive/mobile view look? Assuming it's the same, but wanted to double check

Yeah, presumably looks like you'd expect:
Screenshot from 2020-12-08 10-48-48

PR #291, opened 2020-12-07
[REV-1512] Add flyover box and button to toggle it for the REV1512 experiment

MatthewPiatetsky:
This part would be trickier to do in optimizely so adding it in react

testing3

### DIFF
--- a/src/courseware/course/Course.jsx
+++ b/src/courseware/course/Course.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { Helmet } from 'react-helmet';
 import { useDispatch } from 'react-redux';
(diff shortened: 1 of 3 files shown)


src/courseware/course/CourseBreadcrumbs.jsx
```diff
@@ -36,6 +36,8 @@ export default function CourseBreadcrumbs({
   courseId,
   sectionId,
   sequenceId,
+  toggleREV1512Flyover, /* This line should be reverted after the REV1512 experiment */
```
julianajlk, line 39: I don't think this shows as a comment, try `// This line should...`?

src/courseware/course/Course.jsx
```diff
@@ -50,6 +50,33 @@ function Course({
   const celebrateFirstSection = celebrations && celebrations.firstSection;
   const celebrationOpen = shouldCelebrateOnSectionLoad(courseId, sequenceId, unitId, celebrateFirstSection, dispatch);
 
+  // The below block of code should be reverted after the REV1512 experiment
+  const [REV1512FlyoverEnabled, setREV1512FlyoverEnabled] = useState(false);
```
julianajlk, line 54: I haven't used hooks yet, just a question - is this where the state is set for `REV1512FlyoverEnabled`? Meaning it will be the same as what's in `useState()`, in this case false.

src/courseware/course/Course.jsx
```diff
@@ -50,6 +50,33 @@ function Course({
...
+  window.enableREV1512Flyover = () => {
+    setREV1512FlyoverEnabled(true);
+  };
+  const [REV1512FlyoverHidden, setREV1512FlyoverHidden] = useState(true);
+  const isREV1512FlyoverVisible = () => {
+    const getCookie = (name) => {
+      const match = document.cookie.match(`${name}=([^;]*)`);
+      return match ? match[1] : undefined;
+    };
+    return REV1512FlyoverEnabled && (!REV1512FlyoverHidden || getCookie('REV1512FlyoverHidden') === 'false');
```
inventhouse: Constantly switching senses between "visible" and "hidden" nomenclature is inconsistent, it would be better to pick one and stick with it.

src/courseware/course/Course.jsx
MatthewPiatetsky: i think this is just defaulting it to false

src/courseware/course/CourseBreadcrumbs.jsx
MatthewPiatetsky: i think these also work as the block comment format? how would i check

src/courseware/course/Course.jsx
```diff
@@ -50,6 +50,33 @@ function Course({
...
+    const setCookie = (cookieName, value, domain, exdays) => {
+      const cookieDomain = (typeof domain === 'undefined') ? '' : `domain=${domain};`;
+      const exdate = new Date();
+      exdate.setDate(exdate.getDate() + exdays);
+      const cookieValue = escape(value) + ((exdays == null) ? '' : `; expires=${exdate.toUTCString()}`);
+      document.cookie = `${cookieName}=${cookieValue};${cookieDomain}path=/`;
+    };
+    const isVisible = isREV1512FlyoverVisible();
+    setCookie('REV1512FlyoverHidden', isVisible);
+    setREV1512FlyoverHidden(isVisible);
```
inventhouse: Using the opposite senses to toggle the cookie and state like this looks very much like a mistaken use of the wrong sense.

inventhouse (approved):
Some feedback but not blocking for experiment code

src/courseware/course/CourseBreadcrumbs.jsx
inventhouse: I think this would be a syntax error if it were not treated as a comment, but it is strange that it isn't styling like one; that said I _strongly_ prefer line-comments over block-comments unless there's a good reason for something to use the block style

src/courseware/course/Course.jsx
MatthewPiatetsky: good idea, switched

src/courseware/course/Course.jsx
MatthewPiatetsky: switched

MatthewPiatetsky:
barring blocking comments, i will merge this tomorrow morning so it isn't deployed in the late afternoon

PR #293, opened 2020-12-07
[REV-1521] Add new lock paywall component for the Value Prop experiment

jhan217:
## What did we do?
Added a new lock paywall component to be shown as part of the Value Prop experiment.
Added a classname to the existing lock paywall component for easier event tracking in Optimizely.
Updated the Unit to toggle between the old and new lock paywall components based off a `window.` setting from the console (or in our case, from Optimizely)

## Why are we doing this?
Part of the Value Prop experiment is to redesign existing upsell messages to clarify the value of upgrading and persuade more people to purchase a certificate. We'd like to test out a new upsell banner to see if it is more effective than the existing banner. 

Business logic for showing/hiding the new banner is in Optimizely (but you can trigger it by calling `window.toggleValuePropPaywallLock()` from any Chrome console that has that page enabled. However, this should only affect the display logic of the page and not affect anything on the backend. State persists between the React page (e.g. loading the next/previous items in the module) but does not if you leave the page (e.g. to back to Course home).

## Testing
This was tested locally against devstack. Since this is part of an experiment, we intend for this code to be short lived/not persist in the codebase forever, so we are prioritizing speed over code quality/durability. Very basic unit tests were written against the new module (cloned from the existing LockPaywall's tests), and we don't expect any regressions/existing tests to break as this is completely new functionality. 

## Other
Ticket: https://openedx.atlassian.net/browse/REV-1521
Mockups: https://www.figma.com/file/1EaMYJ8FecwzbbSv5YIpEC/Value-Prop-Final-Designs?node-id=0%3A1
More information about the value prop experiment: https://openedx.atlassian.net/browse/REV-1512
Full page view (English): 
Screen Shot 2020-12-08 at 1 35 02 PM
Mobile view (English):
Screen Shot 2020-12-08 at 1 34 29 PM
Screen Shot 2020-12-08 at 1 34 36 PM
Screen Shot 2020-12-08 at 1 34 42 PM
Full page view (Spanish):
Screen Shot 2020-12-11 at 9 53 11 AM
Mobile view (Spanish):
Screen Shot 2020-12-11 at 9 53 21 AM
Screen Shot 2020-12-11 at 9 53 29 AM
With discount (Spanish):
Screen Shot 2020-12-11 at 10 01 05 AM


src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.scss
```diff
@@ -0,0 +1,58 @@
+@import "~@edx/brand/paragon/variables";
```
jhan217: Not super sure why I had to copy this here given that it's already in `indes.scss` but without it I can't use semantic names for colors.

jhan217: Welp looks like the [CI](https://github.com/edx/frontend-app-learning/pull/293/checks?check_run_id=1519293687) didn't like that...
```ERROR in ./src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.scss
Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js):
ModuleBuildError: Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
SassError: Undefined variable.
  ╷
4 │     background-color: $gray-100;
  │                       ^^^^^^^^^```

Happy to use the straight up hex color code unless someone has a better idea

File: src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx [hunk 2]
```diff
@@ -0,0 +1,164 @@
...
+    return null;
+  }
+  const {
+    currencySymbol,
+    price,
+    upgradeUrl,
+  } = verifiedMode;
+
+  let upgradeButtonText;
+  if (document.querySelector('.price.discount') !== null) {
```
MatthewPiatetsky, line 37: i'm not convinced this will work so i would make sure to test this (though if you think it would be a hassle to do locally you could test on production)

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
```diff
@@ -0,0 +1,164 @@
...
+    const discountPrice = document.querySelector('.price.discount').textContent;
+    upgradeButtonText = (
+      <>
```
MatthewPiatetsky, line 68: does this generate a div? or is it the react shorthand for making a wrapper element but doesn't show up on the dom?

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
```diff
@@ -0,0 +1,164 @@
...
+        <span className="font-weight-bold" style={{ paddingRight: '5px' }}>
+          {intl.formatMessage(messages['learn.lockPaywall.upgrade.link.text'], {
```
MatthewPiatetsky: given that you already have an an object of messages, i feel like it could potentially be pretty simple to store the hardcoded spanish versions there too. then you could index into the english or spanish one based on the user setting.

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
```diff
@@ -0,0 +1,164 @@
...
+        href="https://www.edx.org/verified-certificate"
+      >
+        {intl.formatMessage(messages['lock.description.reason1.linkText'])}
+      </a>
+    </b>
+  );
+
+  const userAgent = typeof window.navigator === 'undefined' ? '' : navigator.userAgent;
+  const isMobile = Boolean(
+    userAgent.match(/Android|BlackBerry|iPhone|iPad|iPod|Opera Mini|IEMobile|WPDesktop/i),
```
MatthewPiatetsky, line 108: typically we do the mobile stuff with media queries but it looks like this works fine from your screenshots.  
we can always check with a few phones during the experiment QA.

MatthewPiatetsky (approved):
i'd probably consider either adding REV1512 to the variables or adding comments to the code containing REV1512 like I did in my PR to make sure that it is easy to revert the code when we're done with the experiment (sometimes can be a bit harder if you have a bunch of conflicts when rebasing and you're not sure if you got everything)

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
jhan217: Screen Shot 2020-12-08 at 2 22 28 PM

I was able to set this up locally by:
1. set the `courseware.courseware_mfe` to on for everyone
2. set the `discounts.enable_discounting` to on for everyone
3. create a new account + enroll in the demo course

Looks like the selector works, but the contents contain the currency symbol, so I'll strip that out :)

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
jhan217: @muselesscreator mentioned that tags are needed when you set variables in React like this - if you don't have an html element available to wrap around the code, you can use empty tags instead.

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
jhan217: Not too familiar with how Spanish translations work in the mfes, @muselesscreator suggested that using `intl.formatMessage` was a relatively pain-free way of doing translations. Do you have any thoughts on how to best do the Spanish translations within the code?

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
jhan217: This works alright with the chrome device emulator. This doesn't work super well with resized windows on desktops, but that's not a use case UX was particularly concerned about last time I talked to them (it also seemed like a higher lift to support custom resized windows).

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
MatthewPiatetsky: well you could just add both the english and spanish ones to the object you already defined and then differentiate them the same way you did in optimizely 
but you can also try it this way and just see how that goes so you have a point of comparison

julianajlk: Depending on when we start the experiment and this code is merged, it might take longer to get the strings translated in transifex, just an fyi to keep an eye on them if you decide to go that route, you can use the `<FormattedMessage>` component

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
julianajlk: Have you checked Firefox? Sometimes it has some differences from Chrome. 
Question on the alignment of the text in the mobile screenshots, for the text `"when you upgrade, you:"` is it supposed to align with the checkmarks and not with the certificate image? Looks strange (haven't checked the figma files, I'm assuming it's like that there?)

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
davidjoy: The MFE handles internationalization and localization.  We should not be manually adding Spanish strings to our messages files.  If you want to ensure that the messages are translated into Spanish, you should make sure that happens in Transifex before launching the experiment.  

If you add Spanish messages to the messages file, translators are going to see Spanish messages that they'll be asked to translate from English to whatever language they're working in, which is just a bad scene.

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
```diff
@@ -35,7 +35,7 @@ function LockPaywall({
         <p className="mb-0">
           <span>{intl.formatMessage(messages['learn.lockPaywall.content'])}</span>
           &nbsp;
-          <a href={upgradeUrl}>
+          <a className="lock_paywall_upgrade_link" href={upgradeUrl}>
```
davidjoy, line 38: Nit: we generally use dashes instead of underscores for CSS class names.

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
davidjoy: Not sure I understand why we need to use document.querySelector here... is this a DOM element that's somehow outside react?  How did it get there?  If it's just holding data that we want to use, could we have just attached that data to window so that we can access it?  Passing data through the DOM is kinda a round-about way of sharing it.

muselesscreator: The data is generated in edx-platform and put directly into the referenced element in the upgrade banner.
Is currently not exposed to any kind of API hook we can find.  If this moves beyond an experiment, it will definitely need to be exposed to a separate query.

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
jhan217: After a round of discussions, we came to the conclusion that we'll be hardcoding the English and Spanish strings directly and not go through Transifex at all. Factors into the decision: 
1. This is temporary code (intended to be deleted in a few weeks)
2. The rest of our experiment only does English or Spanish
3. Transifex requires additional work outside of this PR (deploying jobs + updating Transifex) that will also need to be done/potentially undone as part of the experiment + experiment cleanup

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
```diff
@@ -0,0 +1,168 @@
...
+          {intl.formatMessage(messages['learn.lockPaywall.upgrade.link.text'], {
+            currencySymbol,
+            price: discountPrice,
+          })}
+        </span>
+        <span style={{ textDecoration: 'line-through' }}>
+          ({intl.formatMessage(messages['learn.lockPaywall.upgrade.link.strikethrough.price'], {
+          currencySymbol,
+          price,
+        })})
```
muselesscreator: this indentation is really hard to read.

muselesscreator (approved):
As long as it gets sign-off from TnL, this seems reasonable to me as an experimental/temporary approach.

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
jhan217: Looks fine in Firefox; not sure about text alignment but we can fix that after the UX review. The Figma specs were all with the old brand styles/font so some of the alignments are already different in the new brand style.

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
jhan217: This is a DOM element that is provided in from edx-platform / outside of the learning MFE (it's being created by python in the backend at https://github.com/edx/edx-platform/blob/master/openedx/features/discounts/utils.py#L53). Are there any specific concerns about using `document.querySelector` to get this data? 

When @muselesscreator and @MatthewPiatetsky and I investigated, we found:
1. edx-platform 's first discount banner and the learning mfe with the banner we're modifying are both in the same window, not separate iFrames, so there shouldn't be any issues with accessing the `.price.discount` element from production. This has been tested to work locally in devstack.
2. There is a race condition (e.g. the edx first discount banner needs to be loaded before the lock paywall banner) that might happen, but in the worst case scenario we'd only be affecting the text of the button (i.e. not showing the discount)
3. This code will only be run when we trigger it from Optimizely, so it should only break our experiment, if that. 
4. We're prioritizing speed over quality for temporary experiment code, so we're hoping to avoid making too many changes where possible.

cc'ing @emma-green if she has thoughts on this

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
jhan217: fixed

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
jhan217: fixed, I think

muselesscreator (approved):
LGTM

jhan217:
@davidjoy and @ciduarte mentioned they did not have any additional concerns about the PR over Slack.

PR #301, opened 2020-12-08
AA-503: Course Celebration view for users in verification pending state

Dillon-Dumesnil:
Learners were having questions when we would continue showing them the
'Verify Now' button if they had a submitted a verification attempt
already.

image

### DIFF
--- a/src/courseware/data/api.js
+++ b/src/courseware/data/api.js
@@ -139,6 +139,7 @@ function normalizeMetadata(metadata) {
     courseExitPageIsActive: metadata.course_exit_page_is_active,
     certificateData: camelCaseObject(metadata.certificate_data),
     verifyIdentityUrl: metadata.verify_identity_url,
+    verificationStatus: metadata.verification_status,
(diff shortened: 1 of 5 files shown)


src/courseware/data/api.js
```diff
@@ -139,6 +139,7 @@ function normalizeMetadata(metadata) {
     courseExitPageIsActive: metadata.course_exit_page_is_active,
     certificateData: camelCaseObject(metadata.certificate_data),
     verifyIdentityUrl: metadata.verify_identity_url,
+    verificationStatus: metadata.verification_status,
```
mikix, line 142: aw come on, put that one up so it's alphabetical. Though... i guess surrounding items aren't either. ugh.

mikix (approved):

src/courseware/data/api.js
Dillon-Dumesnil: I put it in the same order as I did in the API. Maybe I suck at both? But that whole endpoint needs alphabetizing. I just don't have the heart MT :(

PR #302, opened 2020-12-09
Remove Travis-CI integration

stvstnfrd:
This is admittedly the much abridged version, but...
Due to changes at Travis-CI that lead to prohibitively slow build/wait
times, we (edx) decided to switch CI/validation provider to Github
Actions [1].

While we had originally intended to keep the integration available,
following Travis-CI's drop in available open-source credits, it no
longer appears to be a viable option for the community.

As such, and given our already present vendor lock-in at Github,
we recommend using Github Actions for CI/validation in this repo.

However, for those not hosting in the Github ecosystem, you can
configure your testrunner to run `make validate.ci`, for the same
effect; the Github Action is just a wrapper for this call.

- [0] Fixes: TNL-7784
- [1] #276


kdmccormick (approved):
LGTM, and thanks for the commit message detailing the reasoning 👍

README.rst
```diff
@@ -10,8 +10,6 @@ Introduction
 
 React app for edX learning.
 
-.. |Build Status| image:: https://api.travis-ci.org/edx/frontend-app-learning.svg?branch=master
```
stvstnfrd, line 13: Note: This does mean the badge goes away (instead of being "updated");
does that matter to us, especially given the tight coupling of Github Actions?

kdmccormick: IMO it's a nice-to-have, but since an identical build status is always visible on in the latest commit anyway it's not a big deal to me 
image

kdmccormick: Either way, :heavy_check_mark:

README.rst
```diff
@@ -1,4 +1,4 @@
-|Build Status| |Coveralls| |npm_version| |npm_downloads| |license|
```
stvstnfrd, line 1: I missed this on the first pass.

README.rst
stvstnfrd: Yeah, and the last commit is visible on the top of the main repo/README page, so we still get it there anyway, albeit higher up on the page (near the top toolbar, instead of inside the README).

PR #304, opened 2020-12-10
ignore for now

MatthewPiatetsky:
Add flyover box and button to toggle it for the REV1512 experiment
This part would be trickier to do in optimizely so adding it in react
REV-1512


openedx-webhooks:
Although this pull request is already closed, I've created [OSPR-6502](https://openedx.atlassian.net/browse/OSPR-6502) so that we can track it in Jira.

PR #306, opened 2020-12-11
TNL-7185: Stop using dangerouslySetInnerHTML in alerts

mikix:
Render offer and access-expiration alerts ourselves from newly passed in backend data, rather than from provided HTML blobs.

https://openedx.atlassian.net/browse/TNL-7185

Related to https://github.com/edx/edx-platform/pull/25863

### DIFF
--- a/src/alerts/access-expiration-alert/messages.js
+++ b/src/alerts/access-expiration-alert/messages.js
@@ -1,0 +1,10 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  upgradeNow: {
+    id: 'learning.accessExpiration.upgradeNow',
+    defaultMessage: 'Upgrade now',
+  },
+});
+
+export default messages;
--- a/src/alerts/offer-alert/OfferAlert.jsx
+++ b/src/alerts/offer-alert/OfferAlert.jsx
@@ -1,24 +1,98 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import {
+  FormattedMessage, FormattedDate, injectIntl, intlShape,
+} from '@edx/frontend-platform/i18n';
+import { Hyperlink } from '@edx/paragon';
 
 import { Alert, ALERT_TYPES } from '../../generic/user-messages';
+import messages from './messages';
 
-function OfferAlert({ payload }) {
+function OfferAlert({ intl, payload }) {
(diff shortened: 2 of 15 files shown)


src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -218,6 +218,8 @@ function CourseCelebration({ intl }) {
         } else {
           footnote = <DashboardFootnote variant={visitEvent} />;
         }
+      } else {
+        visitEvent = 'celebration_audit_no_upgrade';
```
mikix, line 222: Unrelated, but thrown in after a conversation with Kaitlin and Tim.

src/alerts/access-expiration-alert/messages.js
```diff
@@ -0,0 +1,10 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  upgradeNow: {
+    id: 'learning.accessExpiration.upgradeNow',
+    defaultMessage: 'Upgrade now',
```
mikix, line 6: This string is duplicated a few places. But... I dunno, this might be an instance where duplication is better/easier than having a central location for this dumb string and creating interdependencies?

Dillon-Dumesnil: Yeah, I'm not sure what the best way to handle this is. I like deduplication, I also like not having interdependencies. I'm okay having this listed multiple times.

src/alerts/offer-alert/OfferAlert.jsx
```diff
@@ -1,24 +1,98 @@
...
 
-function OfferAlert({ payload }) {
+function OfferAlert({ intl, payload }) {
   const {
-    rawHtml,
+    offer,
+    userTimezone,
   } = payload;
-  return rawHtml && (
+  const timezoneFormatArgs = userTimezone ? { timeZone: userTimezone } : {};
```
Dillon-Dumesnil: Nit: Move this below the `if (!offer)` or `if (!accessExpiration)` lines since it only has meaning if it makes it past that line. Could be defined with the deconstruction.

I just happened to miss this line because whenever I see the early exit, I expect not much to happen above it.

Dillon-Dumesnil (approved):
LGTM

PR #309, opened 2020-12-15
[REV-1512] Add back classes to first purchase offer banner that are used in the REV-1512 optimizely test

MatthewPiatetsky:
These classes were being used to set some of the discount info in our optimizely test, looks like they were removed in https://github.com/edx/frontend-app-learning/pull/306

### DIFF
--- a/src/alerts/offer-alert/OfferAlert.jsx
+++ b/src/alerts/offer-alert/OfferAlert.jsx
@@ -34,14 +34,16 @@ function OfferAlert({ intl, payload }) {
         {intl.formatMessage(messages.srPrices, { discountedPrice, originalPrice })}
       </span>
       <span aria-hidden="true">
-        {discountedPrice} <del>{originalPrice}</del>
+        {/* the price discount and price original classes can be removed post REV-1512 experiment */}
+        <span className="price discount">{discountedPrice}</span> <del className="price original">{originalPrice}</del>
       </span>
     </>
   );
 
   return (
     <Alert type={ALERT_TYPES.INFO}>
-      <span className="font-weight-bold">
+      {/* the first-purchase-offer-banner class can be removed post REV-1512 experiment */}
+      <span className="font-weight-bold first-purchase-offer-banner">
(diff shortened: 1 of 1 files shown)


jhan217 (approved):

julianajlk (approved):

src/alerts/offer-alert/OfferAlert.jsx
```diff
@@ -34,13 +34,13 @@ function OfferAlert({ intl, payload }) {
...
-        {discountedPrice} <del>{originalPrice}</del>
+        <span className="price discount">{discountedPrice}</span> <del className="price original">{originalPrice}</del>
       </span>
     </>
   );
 
   return (
-    <Alert type={ALERT_TYPES.INFO}>
+    <Alert type={ALERT_TYPES.INFO} class="first-purchase-offer-banner">
       <span className="font-weight-bold">
```
dianekaplan: Curious why Alert uses class instead of className?

mikix: Yeah I think that should be className

MatthewPiatetsky: should be using className

mikix (approved):
Sorry, didn't mean to interfere with your experiment!

src/alerts/offer-alert/OfferAlert.jsx
MatthewPiatetsky: fixed

mikix:
Landing this quickly should be a priority. But I'd also like to see some comments that these otherwise unused classnames are actually used outside the code base for an experiment. Otherwise we won't know when it's safe to remove them and/or some other eng will remove them too early in a cleanup later.

src/alerts/offer-alert/OfferAlert.jsx
dianekaplan: thanks!

dianekaplan (approved):

PR #311, opened 2020-12-15
AA-449: Show discounted price in upgrade buttons & course exit

mikix:
If an offer is active for the user, show the discounted price (and a struck-out original price) on upgrade buttons in the course sock and outline sidebar.

Also show the discount code and price in the course exit upgrade screen.

https://openedx.atlassian.net/browse/AA-449

Screenshot from 2020-12-15 12-10-56
Screenshot from 2020-12-15 11-36-24
Screenshot from 2020-12-15 11-36-40

### DIFF
--- a/src/alerts/offer-alert/OfferAlert.jsx
+++ b/src/alerts/offer-alert/OfferAlert.jsx
@@ -6,6 +6,7 @@ import {
 import { Hyperlink } from '@edx/paragon';
 
 import { Alert, ALERT_TYPES } from '../../generic/user-messages';
+import { FormattedPricing } from '../../generic/upgrade-button';
 import messages from './messages';
 
 function OfferAlert({ intl, payload }) {
@@ -20,25 +21,11 @@ function OfferAlert({ intl, payload }) {
 
   const {
     code,
-    discountedPrice,
     expirationDate,
-    originalPrice,
     percentage,
     upgradeUrl,
   } = offer;
   const timezoneFormatArgs = userTimezone ? { timeZone: userTimezone } : {};
-
-  const fullPricing = (
-    <>
-      <span className="sr-only">
-        {intl.formatMessage(messages.srPrices, { discountedPrice, originalPrice })}
-      </span>
-      <span aria-hidden="true">
-        {/* the price discount and price original classes can be removed post REV-1512 experiment */}
-        <span className="price discount">{discountedPrice}</span> <del className="price original">{originalPrice}</del>
-      </span>
-    </>
-  );
 
   return (
     <Alert type={ALERT_TYPES.INFO}>
@@ -57,7 +44,7 @@ function OfferAlert({ intl, payload }) {
                 {...timezoneFormatArgs}
               />
             ),
-            fullPricing,
+            fullPricing: <FormattedPricing offer={offer} />,
             percentage,
           }}
         />
--- a/src/generic/course-sock/CourseSock.jsx
+++ b/src/generic/course-sock/CourseSock.jsx
@@ -4,12 +4,12 @@ import { getConfig } from '@edx/frontend-platform';
 import { FormattedMessage } from '@edx/frontend-platform/i18n';
 import LearnerQuote1 from './assets/learner-quote.png';
 import LearnerQuote2 from './assets/learner-quote2.png';
+import { UpgradeButton } from '../upgrade-button';
 import VerifiedCert from '../assets/edX_certificate.png';
 
 export default class CourseSock extends Component {
   constructor(props) {
     super(props);
-    this.verifiedMode = props.verifiedMode;
     this.state = { showUpsell: false };
     this.sockElement = React.createRef();
   }
@@ -31,7 +31,7 @@ export default class CourseSock extends Component {
   }
 
   render() {
-    if (!this.verifiedMode) {
+    if (!this.props.verifiedMode) {
       return null;
     }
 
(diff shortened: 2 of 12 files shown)


src/generic/course-sock/CourseSock.jsx
```diff
@@ -65,22 +65,14 @@ export default class CourseSock extends Component {
...
+                      size="lg"
+                      offer={this.props.offer}
+                      verifiedMode={this.props.verifiedMode}
+                      className="mb-3"
                       data-creative="original_sock"
                       data-position="sock"
-                    >
-                      <FormattedMessage
-                        id="coursesock.upsell.upgrade"
-                        defaultMessage="Upgrade ({symbol}{price} {currency})"
```
mikix, line 76: Note that my unification of this code has removed the `{currency}` bit from the sock button. Before my change it would say something like "Upgrade ($150 USD)". Now it either says "Upgrade ($150)" or "Upgrade ($120 -$150-)"

The use of the USD is not in the LMS (this was an addition by the MFE). And I wasn't sure how that would have played in the strikeout scenario either. The mocks I've been given for the text elsewhere don't use USD either. So I figured it was safe to remove? @davidjoy this was St Germain code originally - do you know origin of the USD bit?

src/alerts/offer-alert/OfferAlert.jsx
```diff
@@ -20,25 +21,12 @@ function OfferAlert({ intl, payload }) {
...
   } = offer;
   const timezoneFormatArgs = userTimezone ? { timeZone: userTimezone } : {};
 
-  const fullPricing = (
-    <>
-      <span className="sr-only">
-        {intl.formatMessage(messages.srPrices, { discountedPrice, originalPrice })}
-      </span>
-      <span aria-hidden="true">
-        {discountedPrice} <del>{originalPrice}</del>
```
MatthewPiatetsky: we'll need the classes i added in my other PR but i'm merging that shortly so you can rebase

src/generic/course-sock/CourseSock.jsx
MatthewPiatetsky: fyi i believe there's two aspects to this 
there's a local currency feature that displays different currencies on marketing site, track selection and checkout, so currency symbol is especially important there
i believe the platform also has a currency feature that we don't use but may or may not be an open source thing

(Hilary might know if we're moving towards displaying currencies on prices or not)

(If you're wondering the difference between symbol and currency in other pages like track selection it can be helpful since many other currencies also use $ as their symbol)

MatthewPiatetsky (commented):
A couple questions
1. The current offer banner includes a deadline for upgrading after which point it disappears, was that removed intentionally? If so that's fine (though our experiment does depend on it, but we can come up with an alternative solution in that case)
2.  Was the goal of the change just to unify the banner across different places including course exit?
The experiment we're currently working on deletes the banner and makes some changes to the value prop messaging. I will assume the relevant product managers have synced on it but just wanted to make sure you were aware

mikix:
> The current offer banner includes a deadline for upgrading after which point it disappears, was that removed intentionally? If so that's fine (though our experiment does depend on it, but we can come up with an alternative solution in that case)

No it was not removed intentionally. To my knowledge, [it is still there](https://github.com/edx/frontend-app-learning/blob/master/src/alerts/offer-alert/OfferAlert.jsx#L49). But this PR also doesn't change anything about that, I believe.

> Was the goal of the change just to unify the banner across different places including course exit?
> The experiment we're currently working on deletes the banner and makes some changes to the value prop messaging. I will assume the relevant product managers have synced on it but just wanted to make sure you were aware

The goal of this change was fixing AA-449, which is about showing the discounted price in all the places we show the normal price (specifically, achieving parity with LMS by adding the discounted price to the outline-sidebar and the course-sock upgrade buttons and also adding the discounted price to the new course exit page).

This PR only touches the offer banner incidentally, in order to share some code. When you say banner, are you also refering to the outline-sidebar card thing?

Hopefully this PR won't affect your experiment at all. It's just adding some strikeout text in a few buttons to match the LMS look. The only public bits it touches are the course sock button and the course exit, neither of which are affected by your experiment, I don't think?

src/generic/course-sock/CourseSock.jsx
mikix: I'm aware that multiple currencies use `$` as a symbol. My concern was matching the current UX. The course sock had an unexpected delta from the UX and I wasn't sure where it came from. This PR brings the MFE back in line with the LMS.

@hilaryg123 your mocks so far have not had any currency codes on them - do you have context for whether the course sock upgrade button should be displaying one?

MatthewPiatetsky: (yup that makes sense, i do think this brings it in line with the lms, just wanted to add some of the context around where we do use local currency in case you weren't familiar)

MatthewPiatetsky:
ahh ok, i saw the screenshot at the top and got confused because I thought both the offer banner and the course exit thing would both look like that, if it is only modified incidentally then my only feedback would be including the "price discount" "price original" and "first-purchase-offer-bannner" classes from the rebase
thanks!

src/generic/course-sock/CourseSock.jsx
hilaryg123: If we don't have to, my preference is not to add the currency code. Looks like we use it quite inconsistently throughout the marketing site and LMS (it is on track selection, but not in the cart, for example). If there's a reason we really need it somewhere, happy to learn more and discuss.

ciduarte (approved):
LGTM

PR #313, opened 2020-12-16
[REV-1517] Add Notification header and X to hide the flyover

MatthewPiatetsky:
Screen Shot 2020-12-15 at 8 52 01 PM

### DIFF
--- a/src/alerts/offer-alert/OfferAlert.jsx
+++ b/src/alerts/offer-alert/OfferAlert.jsx
@@ -34,14 +34,16 @@ function OfferAlert({ intl, payload }) {
         {intl.formatMessage(messages.srPrices, { discountedPrice, originalPrice })}
       </span>
       <span aria-hidden="true">
-        {discountedPrice} <del>{originalPrice}</del>
+        {/* the price discount and price original classes can be removed post REV-1512 experiment */}
+        <span className="price discount">{discountedPrice}</span> <del className="price original">{originalPrice}</del>
       </span>
     </>
   );
 
   return (
     <Alert type={ALERT_TYPES.INFO}>
-      <span className="font-weight-bold">
+      {/* the first-purchase-offer-banner class can be removed post REV-1512 experiment */}
+      <span className="font-weight-bold first-purchase-offer-banner">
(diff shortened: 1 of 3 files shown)


julianajlk (approved):

dianekaplan (commented):
Nothing objectionable stands out to me! happy to approve when build tests are passing

src/alerts/offer-alert/OfferAlert.jsx
```diff
@@ -34,14 +34,16 @@ function OfferAlert({ intl, payload }) {
         {intl.formatMessage(messages.srPrices, { discountedPrice, originalPrice })}
       </span>
       <span aria-hidden="true">
-        {discountedPrice} <del>{originalPrice}</del>
+        {/* the price discount and price original classes can be removed post REV-1512 experiment */}
+        <span className="price discount">{discountedPrice}</span> <del className="price original">{originalPrice}</del>
```
julianajlk, line 38: These are 2 separate classNames? `price` and `discount`, not `price-discount`

inventhouse (approved):

src/alerts/offer-alert/OfferAlert.jsx
MatthewPiatetsky: yup

src/courseware/course/sequence/Sequence.jsx
```diff
@@ -196,9 +197,46 @@ function Sequence({
                 width: '330px',
                 verticalAlign: 'top',
                 marginLeft: '20px',
-                padding: '20px',
+                padding: '0 20px 20px 20px',
```
davidjoy, line 200: Where possible, we should use utility css classes rather than hard coded individual styles: https://paragon-edx.netlify.app/foundations/css-utilities

Something to clean up if the experiment succeeds.

MatthewPiatetsky: thanks! good to know.  
if the experiment does succeed i think we can count on all the code being reverted and written from scratch so we can hopefully take full advantage of that there

PR #315, opened 2020-12-16
[REV-1567] Mobile version of flyover for REV-1512 experiment

inventhouse:


MatthewPiatetsky (approved):
i have some further changes to make so i'm branching off of this branch to avoid having to rebase after

PR #316, opened 2020-12-17
Update business logic for the flyover per UX feedback

MatthewPiatetsky:
Per UX feedback:
1. Default flyover to open
2. Make cookie that saves the open/close state course specific
3. Expire cookie after 4 hours

### DIFF
--- a/src/courseware/course/Course.jsx
+++ b/src/courseware/course/Course.jsx
@@ -60,18 +60,22 @@ function Course({
     const match = document.cookie.match(`${name}=([^;]*)`);
     return match ? match[1] : undefined;
   };
-  const [REV1512FlyoverVisible, setREV1512FlyoverVisible] = useState(getCookie('REV1512FlyoverVisible') === 'true');
+  const userAgent = typeof window.navigator === 'undefined' ? '' : navigator.userAgent;
+  const isMobile = Boolean(
+    userAgent.match(/Android|BlackBerry|iPhone|iPad|iPod|Opera Mini|IEMobile|WPDesktop/i),
+  );
+  const [REV1512FlyoverVisible, setREV1512FlyoverVisible] = useState(isMobile ? false : !(getCookie(`REV1512FlyoverVisible${courseId}`) === 'false'));
   const isREV1512FlyoverVisible = () => REV1512FlyoverEnabled && (REV1512FlyoverVisible || getCookie('REV1512FlyoverVisible') === 'true');
   const toggleREV1512Flyover = () => {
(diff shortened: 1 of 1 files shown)


src/courseware/course/Course.jsx
```diff
@@ -60,18 +60,18 @@ function Course({
...
       const exdate = new Date();
-      exdate.setDate(exdate.getDate() + exdays);
-      const cookieValue = escape(value) + ((exdays == null) ? '' : `; expires=${exdate.toUTCString()}`);
+      exdate.setHours(exdate.getHours() + 4);
+      const cookieValue = escape(value) + ((4 == null) ? '' : `; expires=${exdate.toUTCString()}`);
       document.cookie = `${cookieName}=${cookieValue};${cookieDomain}path=/`;
     };
     const isVisible = isREV1512FlyoverVisible();
-    setCookie('REV1512FlyoverVisible', !isVisible);
+    setCookie(`REV1512FlyoverVisible${courseId}`, !isVisible);
```
inventhouse, line 78: How many cookies should we be creating?  Seems like the cookie-space could get crowded.

inventhouse (approved):

src/courseware/course/Course.jsx
inventhouse: (On the other hand, with a 4-hour expiration, maybe it won't be a problem)

src/courseware/course/Course.jsx
```diff
@@ -60,18 +60,18 @@ function Course({
     const match = document.cookie.match(`${name}=([^;]*)`);
     return match ? match[1] : undefined;
   };
-  const [REV1512FlyoverVisible, setREV1512FlyoverVisible] = useState(getCookie('REV1512FlyoverVisible') === 'true');
+  const [REV1512FlyoverVisible, setREV1512FlyoverVisible] = useState(!(getCookie(`REV1512FlyoverVisible${courseId}`) === 'false'));
```
inventhouse: Defaulting to open would be terrible on Mobile, since that takes over the entire screen

inventhouse (changes requested):

src/courseware/course/Course.jsx
MatthewPiatetsky: i think it should be fine with a 4 hour expiration for the experiment

src/courseware/course/Course.jsx
MatthewPiatetsky: yeah i agree, updated

inventhouse (approved):

PR #329, opened 2020-12-21
Update REV-1512 experiment gated content 

julianajlk:
Updating locked gated content for [REV-1512](https://openedx.atlassian.net/browse/REV-1512) Value Prop Experiment - 
1) Hyperlink color 
2) Alignment of list items and certificate image 
3) Mobile view for screens < 576px

Please note I am not using Paragon variables for breakpoints and colors on purpose to avoid any errors (since we're going on a code freeze today, want to avoid any room for edits). 

`isMobile` could also be replaced with just having one breakpoint at 576px, but would like to avoid breaking mobile if it's working now. Since this is temporary code, will go with being faster/hacky than dry code.

### DIFF
--- a/src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
+++ b/src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
@@ -138,7 +138,7 @@ function LockPaywall({
             />
           </div>
           <div style={{ float: 'left', paddingLeft: '18px', paddingBottom: '24px' }}>
-            <div style={{ paddingBottom: '10px' }}>
+            <div style={{ paddingBottom: '5px' }}>
               {
                 isSpanish
                   ? 'Cuando te cambias a la opción verificada, tú:'
--- a/src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
+++ b/src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
@@ -96,8 +96,9 @@ function LockPaywall({
       <a
         className="value-prop-verified-certificate-link"
         style={{ textDecoration: 'underline', color: '#00688D' }}
+        rel="noopener noreferrer"
+        target="_blank"
         href="https://www.edx.org/verified-certificate"
-        target="blank"
       >
         { (isSpanish) ? 'certificado verificado' : 'verified certificate' }
       </a>
@@ -114,7 +115,7 @@ function LockPaywall({
       <div className={classNames({ 'is-mobile': isMobile })}>
         <div className="font-weight-bold top-banner-text-header">
           <FontAwesomeIcon icon={faLock} className="text-black mr-2 ml-1 lock-icon" style={{ fontSize: '1rem' }} />
-          <span>
+          <span className="top-banner-title">
             {
               isSpanish
                 ? 'Las tareas calificadas están bloqueadas'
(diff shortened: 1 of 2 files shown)


src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
```diff
@@ -94,8 +94,8 @@ function LockPaywall({
   const verifiedCertificateLink = (
     <b>
       <a
-        className="text-gray-700 value-prop-verified-certificate-link"
-        style={{ textDecoration: 'underline' }}
+        className="value-prop-verified-certificate-link"
+        style={{ textDecoration: 'underline', color: '#00688D' }}
```
julianajlk, line 98: The reason the link was not picking up the color for <a> is because it was being overridden. I am adding the hex color here just as a precaution

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
```diff
@@ -129,16 +129,17 @@ function LockPaywall({
...
           <div className="certificate-image-banner-container">
             <img
               alt="Example Certificate"
               src={VerifiedCert}
               className="border-0 certificate-image-banner"
             />
           </div>
-          <div style={{ float: 'left', paddingLeft: '18px', paddingBottom: '24px' }}>
-            <div style={{ paddingBottom: '10px' }}>
+          <div className="list-container" style={{ float: 'left', paddingLeft: '18px', paddingBottom: '24px' }}>
```
julianajlk, line 143: `list-container` is not being currently used but adding it if we need to change CSS in optimizely

jhan217 (approved):

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
```diff
@@ -94,8 +94,8 @@ function LockPaywall({
   const verifiedCertificateLink = (
     <b>
       <a
-        className="text-gray-700 value-prop-verified-certificate-link"
-        style={{ textDecoration: 'underline' }}
+        className="value-prop-verified-certificate-link"
```
jhan217, line 97: Should this link open in a new tab?

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.jsx
```diff
@@ -94,9 +94,10 @@ function LockPaywall({
...
+        style={{ textDecoration: 'underline', color: '#00688D' }}
         href="https://www.edx.org/verified-certificate"
+        target="blank"
```
jhan217: should this be `target="_blank"` ?

src/courseware/course/sequence/lock-paywall-value-prop/LockPaywall.scss
```diff
@@ -65,5 +67,27 @@
             padding-left: 0px;
         }
     }
-}
 
+    @media only screen and (max-width: 576px) {
```
jhan217, line 71: Any particular reason this can't be in the `.is-mobile` class styles?

julianajlk: Because .is-mobile is activated when isMobile is true, I'm not sure at which breakpoint that is exactly but it's smaller than 576px

jhan217: Could you change the definition of `is-mobile` to be `ipad|iphone|etc && screen-size < 576px`, as they should both be showing the same mobile view?

julianajlk: It's not that simple to get the screensize from react (if there is a resize), it requires saving the width to state and updating if the screen changes size, for the purpose of being fast/hacky, I'll leave it as is

PR #334, opened 2021-01-04
AA-545: Add course in progress CourseExit variant

ciduarte:
<div>AA-545_CourseInProgress
AA-545_CourseInProgress_mobile
</div>

### DIFF
--- a/src/courseware/course/course-exit/CourseInProgress.jsx
+++ b/src/courseware/course/course-exit/CourseInProgress.jsx
@@ -14,16 +14,16 @@ import DashboardFootnote from './DashboardFootnote';
 import messages from './messages';
 import { logClick, logVisit } from './utils';
 
-function CourseNonPassing({ intl }) {
+function CourseInProgress({ intl }) {
   const { courseId } = useSelector(state => state.courseware);
   const { org, tabs } = useModel('courses', courseId);
   const { administrator } = getAuthenticatedUser();
 
-  // Get progress tab link for 'view grades' button
-  const progressTab = tabs.find(tab => tab.slug === 'progress');
-  const progressLink = progressTab && progressTab.url;
+  // Get dates tab link for 'view course schedule' button
+  const datesTab = tabs.find(tab => tab.slug === 'dates');
+  const datesTabLink = datesTab && datesTab.url;
 
-  useEffect(() => logVisit(org, courseId, administrator, 'nonpassing'), [org, courseId, administrator]);
+  useEffect(() => logVisit(org, courseId, administrator, 'in_progress'), [org, courseId, administrator]);
 
   return (
     <>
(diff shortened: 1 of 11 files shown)


src/courseware/course/course-exit/utils.js
```diff
@@ -53,12 +56,15 @@ function getCourseExitMode(courseId) {
...
+  if (isEligibleForCertificate && !userHasPassingGrade && !hasScheduledContent) {
     return COURSE_EXIT_MODES.nonPassing;
   }
-  if (isCelebratoryStatus) {
+  if (isCelebratoryStatus && userHasPassingGrade) {
     return COURSE_EXIT_MODES.celebration;
   }
+  if (hasScheduledContent && !userHasPassingGrade) {
+    return COURSE_EXIT_MODES.inProgress;
+  }
```
ciduarte: Can someone sanity check this?

We only want to show the "Non-passing" state if there is no future scheduled content, the user does not have a passing grade, and they are eligible for a certificate (audit-passing or honor-passing).

We only want to show the "Celebration" state if the user has a celebration status in the list above ('audit_passing', 'downloadable', 'earned_but_not_available', 'honor_passing', 'requesting', 'unverified'). Do we need to explicitly include `userHasPassingGrade`?

We only want to show the "In Progress" state if the user does not have a passing grade and still has scheduled content. (And we do not care about their certificate eligibility?)

And we don't support any other cases, so the CourseExit feature would be disabled for anything else.

src/courseware/course/course-exit/utils.js
```diff
@@ -67,6 +73,8 @@ function getCourseExitText(courseId, intl) {
   switch (getCourseExitMode(courseId)) {
     case COURSE_EXIT_MODES.celebration:
       return intl.formatMessage(messages.nextButtonComplete);
+    case COURSE_EXIT_MODES.inProgress:
+      return intl.formatMessage(sequenceNavigationMessages.nextButton);
```
mikix: If you just don't add these lines, you can skip importing another module's messages.js. Returning `null` here will result in the "default" message getting used by the caller, which is what you want here I believe.

src/courseware/course/course-exit/utils.js
mikix: I don't think you want to add `userHasPassingGrade` to the celebration check. Think of audit users, who we show it to anyway. Or possibly folks who are whitelisted for a cert, regardless of their grade (I don't know what `userHasPassingGrade` is set to in that case). I think if the cert status is good, we just let them revel in it.

Could you put the new in-progress check above the others? Maybe that lets you not change the other conditions at all?

src/courseware/course/course-exit/utils.js
mikix: Though maybe adding a comment here or even a fallthrough `case` line right above `default` so that other folks know you didn't forget this case, and we intentionally want to return null.

ciduarte: I realized we rely on `getCourseExitText` returning null to determine whether the navigation button should be active or not.. i.e. if we let this return null, the default message will be "Next", but the button will be disabled.

I don't think it's great to be relying on `getCourseExitText`? I'm going to replace that check w/ `getCourseExitMode` and check if it is disabled instead. That'll let us return the default case here.

ciduarte: Ooooooh I think I take this back I didn't realize getCourseExitMode is being called within getCourseExitText.... feels gross to call it twice... I will think a little bit more about what's happening here

src/courseware/course/course-exit/CourseInProgress.jsx
```diff
@@ -0,0 +1,63 @@
...
+                onClick={() => logClick(org, courseId, administrator, 'view_dates_tab')}
+              >
+                {intl.formatMessage(messages.viewCourseScheduleButton)}
+              </Button>
+            )}
+          </div>
+        </Alert>
+        <DashboardFootnote variant="in_progress" />
+        <CatalogSuggestion variant="in_progress" />
+      </div>
```
Dillon-Dumesnil, line 54: We use snakecase to specify variants? That feels weird.

src/courseware/course/course-exit/utils.js
```diff
@@ -53,6 +55,9 @@ function getCourseExitMode(courseId) {
     isEligibleForCertificate = NON_CERTIFICATE_STATUSES.indexOf(certStatus) === -1;
   }
 
+  if (hasScheduledContent && !userHasPassingGrade) {
```
Dillon-Dumesnil, line 58: Was this a product/UX decision to restrict this to user's without passing grades? If you have a passing grade, but there is more content to be released, we are fine showing them the exit and not letting them know there is more?

src/courseware/course/course-exit/utils.js
```diff
@@ -62,16 +67,23 @@ function getCourseExitMode(courseId) {
   return COURSE_EXIT_MODES.disabled;
 }
 
-// Returns null if course exit is either not active or not handling the current case
-function getCourseExitText(courseId, intl) {
-  switch (getCourseExitMode(courseId)) {
+// Returns null in order to render the default navigation text
+function getCourseExitNavigation(courseId, intl) {
```
Dillon-Dumesnil, line 71: Don't love this being called CourseExit anymore :/ Not sure of a better name, but doesn't just have to do with exiting the course anymore. Oh well.

Dillon-Dumesnil (approved):

src/courseware/course/course-exit/utils.js
ciduarte: Yes I believe we decided in grooming that we would rather show the celebration screen over the "more content is coming" screen if they have a passing grade for now. I'll double check w/ product/UX

PR #335, opened 2021-01-05
AA-501: Send query param to recheck access to xblocks for render

Dillon-Dumesnil:


mikix (approved):
We can talk about this approach in the platform PR, but assuming this is the path forward, this looks fine.

PR #343, opened 2021-01-11
AA-598: canEnroll loop bug fix

ciduarte:


MatthewPiatetsky (approved):
i could try it out on prod where it gave me issues

PR #344, opened 2021-01-19
AA-361: shift dates banner tests

ciduarte:


src/course-home/dates-tab/DatesTab.test.jsx
```diff
@@ -128,4 +127,77 @@ describe('DatesTab', () => {
...
+
+    it('renders resetDatesBanner', async () => {
+      datesTabData.datesBannerInfo = {
+        contentTypeGatingEnabled: true,
+        missedDeadlines: true,
+        missedGatedContent: false,
+        verifiedUpgradeLink: 'http://localhost:18130/basket/add/?sku=8CF08E5',
+      };
+
+      axiosMock.onGet(`${getConfig().LMS_BASE_URL}/api/course_home/v1/dates/${courseId}`).reply(200, datesTabData);
```
mikix, line 195: How easy is it to test the failure case, with toast and all that?

mikix (approved):
Yay tests!

src/course-home/dates-tab/DatesTab.test.jsx
ciduarte: Ooh if we actually click the button? Doesn't seem too bad at first glance I'll attempt it

mikix: Oh right - yeah. The line I commented on wasn't what I meant to comment on. I was just seeing all these 200s and thought "what if resetting failed" - didn't realize this line was just normal dates info.

ciduarte: Ooooh I see. I added in a test about clicking the shift dates button anyways (it feels a little redundant but probably fine to leave it)

PR #345, opened 2021-01-20
Added component for proctoring panel

alangsto:
## [MST-579](https://openedx.atlassian.net/browse/MST-579)

Added proctoring info panel to course outline tab. 

Examples of panel: 

Screen Shot 2021-01-20 at 12 58 51 PM

Screen Shot 2021-01-20 at 12 59 24 PM

See original panel on LMS here: https://github.com/edx/edx-proctoring/pull/727

### DIFF
--- a/src/course-home/data/api.js
+++ b/src/course-home/data/api.js
@@ -130,6 +130,20 @@ export async function getProgressTabData(courseId) {
   }
 }
 
+export async function getProctoringInfoData(courseId) {
+  const url = `${getConfig().LMS_BASE_URL}/api/edx_proctoring/v1/user_onboarding/status?course_id=${courseId}`;
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(url);
+    return data;
+  } catch (error) {
+    const { httpErrorStatus } = error && error.customAttributes;
+    if (httpErrorStatus === 404) {
+      return {};
+    }
+    throw error;
+  }
+}
+
 export async function getOutlineTabData(courseId) {
   const url = `${getConfig().LMS_BASE_URL}/api/course_home/v1/outline/${courseId}`;
   let { tabData } = {};
--- a/src/course-home/outline-tab/OutlineTab.jsx
+++ b/src/course-home/outline-tab/OutlineTab.jsx
@@ -25,6 +25,7 @@ import useOfferAlert from '../../alerts/offer-alert';
 import usePrivateCourseAlert from '../../alerts/private-course-alert';
 import { useModel } from '../../generic/model-store';
 import WelcomeMessage from './widgets/WelcomeMessage';
+import ProctoringInfoPanel from './widgets/ProctoringInfoPanel';
 
 function OutlineTab({ intl }) {
   const {
@@ -162,6 +163,9 @@ function OutlineTab({ intl }) {
         </div>
         {rootCourseId && (
           <div className="col col-12 col-md-4">
+            <ProctoringInfoPanel
+              courseId={courseId}
+            />
             {courseGoalToDisplay && goalOptions.length > 0 && (
               <UpdateGoalSelector
                 courseId={courseId}
--- a/src/course-home/outline-tab/OutlineTab.test.jsx
+++ b/src/course-home/outline-tab/OutlineTab.test.jsx
@@ -7,7 +7,7 @@ import userEvent from '@testing-library/user-event';
 
 import buildSimpleCourseBlocks from '../data/__factories__/courseBlocks.factory';
 import {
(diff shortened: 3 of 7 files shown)


src/course-home/data/api.js
```diff
@@ -130,6 +130,20 @@ export async function getProgressTabData(courseId) {
   }
 }
 
+export async function getProctoringInfoData(courseId) {
```
alangsto, line 133: I chose to not make this function a thunk or add it to the getOutlineTabData because it is only being used in this component. I also did not want to add this data to the outline model, as the data is specific to a course and a learner.

src/course-home/outline-tab/widgets/ProctoringInfoPanel.jsx
```diff
@@ -0,0 +1,96 @@
...
+              </>
+            )}
+            {(readableStatus !== 'verified') && (
+              <>
+                <p>{intl.formatMessage(messages.proctoringPanelGeneralInfo)}</p>
+                <p>{intl.formatMessage(messages.proctoringPanelGeneralTime)}</p>
+              </>
+            )}
+            {showExamLink(status) && (
+              <a className="btn btn-primary btn-block" href={link}>
```
alangsto: This button is currently red on the old course homepage. I'm keeping it blue here for consistency with the rest of the outline but can update based on feedback.

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -473,4 +475,30 @@ describe('Outline Tab', () => {
       });
     });
   });
+
+  describe('Proctoring Info Panel', () => {
```
schenedx, line 479: Do you have a test case where the API returns 404 or basically the learner does not need an onboarding status?

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -162,6 +163,9 @@ function OutlineTab({ intl }) {
         </div>
         {rootCourseId && (
           <div className="col col-12 col-md-4">
+            <ProctoringInfoPanel
```
schenedx, line 166: Should the display of this panel be conditional? If the course does not have proctoring turned on? If the learner is only audit track and so on?

schenedx (commented):
I got two inline questions.

src/course-home/outline-tab/OutlineTab.test.jsx
alangsto: Just added

schenedx (approved):
👍

src/course-home/outline-tab/widgets/ProctoringInfoPanel.scss
```diff
@@ -0,0 +1,10 @@
+.outline-sidebar-proctoring-panel {
+  border: 1px solid $dark-500;
+  border-top: 5px solid #b20610;
```
ciduarte: I would switch this value to `$brand-600`

src/course-home/outline-tab/widgets/ProctoringInfoPanel.scss
```diff
@@ -0,0 +1,10 @@
...
+}
+.proctoring-onboarding-success {
+  border-top: 5px solid #008100;
```
ciduarte: And this value to `$primary-500`

(I would say to switch it to one of the $success variants, but we've been replacing the success colors w/ primary when replicating things from the legacy experience)

src/course-home/outline-tab/widgets/ProctoringInfoPanel.jsx
```diff
@@ -0,0 +1,96 @@
...
+                {intl.formatMessage(messages.proctoringOnboardingButton)}
+              </a>
+            )}
+            <a className="btn btn-outline-primary btn-block" href="https://support.edx.org/hc/en-us/articles/207249428-How-do-proctored-exams-work">
+              {intl.formatMessage(messages.proctoringReviewRequirementsButton)}
+            </a>
```
ciduarte: We talked about how it seems like the convention here is to use anchor tags, but I'm realizing this repo needs to work on moving away from this, and switch to using the Button or HyperLink components...

I would say that we should switch these anchor tags to `<Button href="..." variant="outline-primary" block>{}</>`

src/course-home/data/api.js
ciduarte: In looking at our other thunks, it seems like some of them are just wrappers that don't actually update redux state/models... I can't think of a reason to mimic that here other than "because it's been done that way in the past"... So your approach of not adding a thunk feels safe to me..

ciduarte (approved):
LGTM :) Just a few comments about spots where we could use brand/Paragon

PR #346, opened 2021-01-21
Updated urls for proctoring info panel

alangsto:
Corrected API url and also corrected the link to the onboarding exam. Also added test to ensure that panel will still appear if there is no status (meaning the user is eligible but hasn't started onboarding).


schenedx (approved):
👍

PR #347, opened 2021-01-21
AA-512: iframe styling

ciduarte:


src/course-home/outline-tab/LmsHtmlFragment.css
```diff
@@ -0,0 +1,11 @@
+body a {
+    color: #00688D;
+}
+
+.inline-link a {
+    text-decoration: underline;
+}
+
+.small {
```
mikix: body.small? And body.inline-link above, yeah?

src/course-home/outline-tab/LmsHtmlFragment.css
```diff
@@ -0,0 +1,11 @@
+body a {
+    color: #00688D;
```
mikix, line 2: Not styled per theme... But per your other comments that would be difficult.

mikix (approved):

PR #348, opened 2021-01-22
Separate courses redux model into courseHomeMeta and coursewareMeta

mikix:
- `courses` model becomes either `courseHomeMeta` or `coursewareMeta`
- `PrivateCourseAlert` gets moved to `src/course-home` since it's only used there
- `LoadedTabPage` gets a `metadataModel` prop, defaulting to `courseHomeMeta`

This fixes at least one known issue: Going from a course-home page like the Outline tab to a courseware page and then pressing the "Back" button would result in a failed page load. This is because the two pages have different expected tab names.

It also might stop similar issues from appearing in the future. And let us use `Link` components between the two sides of the MFE.


src/alerts/enrollment-alert/hooks.js
```diff
@@ -13,7 +13,7 @@ const EnrollmentAlert = React.lazy(() => import('./EnrollmentAlert'));
 
 export function useEnrollmentAlert(courseId) {
   const { authenticatedUser } = useContext(AppContext);
-  const course = useModel('courses', courseId);
+  const course = useModel('courseHomeMeta', courseId); // MIKE FIXME
```
ciduarte: 👀

ciduarte (approved):
Nice! Just one lingering "Mike" comment, but this refactor looks clean and makes sense to me 👍

src/alerts/enrollment-alert/hooks.js
mikix: Yikes. Good catch. Was already fixed, just a commenting problem.

PR #351, opened 2021-01-25
AA-491: Update MFE with analytics event

Dillon-Dumesnil:
These already exist in the legacy view. This is just porting them
over into the MFEs

### DIFF
--- a/src/alerts/enrollment-alert/EnrollmentAlert.jsx
+++ b/src/alerts/enrollment-alert/EnrollmentAlert.jsx
@@ -5,6 +5,7 @@ import { Button } from '@edx/paragon';
 
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faSpinner } from '@fortawesome/free-solid-svg-icons';
+import { useModel } from '../../generic/model-store';
 import { Alert, ALERT_TYPES } from '../../generic/user-messages';
 
 import messages from './messages';
@@ -18,8 +19,13 @@ function EnrollmentAlert({ intl, payload }) {
     isStaff,
   } = payload;
 
+  const {
+    org,
+  } = useModel('courseHomeMeta', courseId);
+
   const { enrollClickHandler, loading } = useEnrollClickHandler(
     courseId,
+    org,
     intl.formatMessage(messages.success),
   );
 
--- a/src/course-home/outline-tab/alerts/private-course-alert/PrivateCourseAlert.jsx
+++ b/src/course-home/outline-tab/alerts/private-course-alert/PrivateCourseAlert.jsx
@@ -22,11 +22,13 @@ function PrivateCourseAlert({ intl, payload }) {
   } = payload;
 
   const {
+    org,
     title,
   } = useModel('courseHomeMeta', courseId);
 
   const { enrollClickHandler, loading } = useEnrollClickHandler(
     courseId,
+    org,
     intl.formatMessage(enrollmentMessages.success),
   );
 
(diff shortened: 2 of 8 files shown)


Dillon-Dumesnil:
Events being ported over:
Research Events -> `sendTrackingLogEvent`

- edx.course.home.resume_course.clicked
- edx.course.enrollment.upgrade.clicked (this technically already exists inside of the learning MFE, but only the variant where location: 'sidebar-message' and not the variant for location: 'sock')

Segment Events -> `sendTrackEvent`

- Promotion Viewed (this needs to be implemented for both sidebar upsell and sock upsell)
- Promotion Clicked (this needs to be implemented for both sidebar upsell and sock upsell)
- edx.bi.user.course-home.enrollment
- edx.bi.course.sock.toggle_opened
- edx.bi.course.sock.toggle_closed

Dillon-Dumesnil:
See https://openedx.atlassian.net/wiki/spaces/DE/pages/2097119925/Improved+Event+Data+Fields+-+FY21Q2+Q3 for why I added org_key and courserun_key to the new and old events I saw.

src/alerts/enrollment-alert/EnrollmentAlert.jsx
```diff
@@ -18,8 +19,13 @@ function EnrollmentAlert({ intl, payload }) {
     isStaff,
   } = payload;
 
+  const {
+    org,
+  } = useModel('courseHomeMeta', courseId);
```
mikix, line 24: The pattern we've been using for alerts and model data is to put them into the payload object in hooks.js and read them in the alert. But... I don't know *why*? @davidjoy - do you remember why?

src/course-home/outline-tab/alerts/private-course-alert/PrivateCourseAlert.jsx
```diff
@@ -22,11 +22,13 @@ function PrivateCourseAlert({ intl, payload }) {
   } = payload;
 
   const {
+    org,
```
mikix, line 25: Same hooks.js comment.

mikix (approved):
Sure. Looks reasonable. Just had a question about the hook / model use.

src/alerts/enrollment-alert/EnrollmentAlert.jsx
Dillon-Dumesnil: Ah yeah, I did this because I saw the PrivateCourseAlert using it so I thought it might be an accepted format. Let me know and I'll see what I can do about refactoring

Dillon-Dumesnil: Hey @davidjoy , sorry for the double ping. Just wanted to bump on this to get some clarity on how these hooks files should be. Thanks! And welcome back!

davidjoy: Is this about keeping the alert "dumb" as to where the data's coming from, perhaps?

davidjoy: (I think that'd be the main reason that springs to mind)

davidjoy: This whole system is so bonkers, though, that I really don't feel that strongly/can't remember if there's a stronger reason.

Dillon-Dumesnil: Okay, in that case, I'm going to leave as is and if we decide to refactor at some point, it shouldn't be a hard lift to fix in both spots. Thanks David!

PR #354, opened 2021-01-28
[AA-304] Create three day streak celebration

MatthewPiatetsky:
Show learners a celebratory modal if they visit the learning mfe for 3 days in a row. Call edx-platform API to determine if they should see the celebration.

**[edx-platform PR](https://github.com/edx/edx-platform/pull/26409/files#diff-e1221a569c1c5f896d2be047b50d17c09c967785c90a7fbc2a68f0c8d9543129R562)
[Paragon lightbulb icon PR](https://github.com/edx/paragon/pull/612)
[Paragon close icon PR](https://github.com/edx/paragon/pull/631)
[Mocks](https://www.figma.com/file/B6AF4htYRdesPBNR08Fzvp/Celebratory-Milestones?node-id=312%3A22)
[Jira Ticket](https://openedx.atlassian.net/browse/AA-304)**

**Mobile View:**
Screen Shot 2021-02-15 at 12 05 30 AM

**Desktop View:**
Screen Shot 2021-02-15 at 12 05 01 AM


package-lock.json
```diff
@@ -30,18 +30,18 @@
       }
     },
     "@babel/code-frame": {
```
MatthewPiatetsky, line 32: there are a lot of changes in this file, not sure if this is to be expected if I did npm install?  
i did remove node_modules at one point when i was trying to get the new paragon working

src/shared/streak-celebration/utils.jsx
```diff
@@ -0,0 +1,28 @@
...
+  });
+}
+
+function recordModalClosing(metadataModel, celebrations, org, courseId, dispatch) {
+  // Ensure we only celebrate each streak once
+  dispatch(updateModel({
+    modelType: metadataModel,
+    model: {
+      id: courseId,
+      celebrations: { ...celebrations, shouldCelebrateStreak: false },
```
MatthewPiatetsky, line 22: I was trying only to update the relevant property of the celebrations dict here and in the first section celebration 
but i'm not currently handling the case of two celebrations occurring at the same time.  
I don't think this would happen with the streak and first section celebrations (since the first section celebration would happen after clicking next while the streak would have already happened when the page loaded)
however, two celebrations could occur at the same time, so this may be a case that I should handle here defensively
thoughts?

package.json
```diff
@@ -38,8 +38,8 @@
     "@edx/frontend-component-footer": "10.1.4",
     "@edx/frontend-enterprise": "4.2.3",
     "@edx/frontend-platform": "1.8.4",
-    "@edx/paragon": "12.3.1",
     "@fortawesome/fontawesome-svg-core": "1.2.34",
+    "@edx/paragon": "13.9.0",
```
mikix, line 41: You un-alphabetized this.  :(

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -0,0 +1,115 @@
...
+import { useModel } from '../../generic/model-store';
+import StreakMobileImage from './assets/Streak_mobile.png';
+import StreakDesktopImage from './assets/Streak_desktop.png';
+import messages from './messages';
+import { recordModalClosing, recordStreakCelebration } from './utils';
+
+function getRandomFactoid(intl) {
+  const boldedSectionA = intl.formatMessage(messages.StreakFactoidABoldedSection);
+  const boldedSectionB = intl.formatMessage(messages.StreakFactoidBBoldedSection);
+  const factoids = [
```
mikix, line 18: Factoid is so judgemental. These are simply **facts** - get on board, learners.

(Not a serious suggestion to change the name.)

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -0,0 +1,115 @@
...
+    />),
+    (<FormattedMessage
+      id="learning.streakcelebration.factoidb"
+      defaultMessage="Users who learn 3 days in a row {bolded_section} vs. those who don't."
+      values={{
+        bolded_section: (<b>{boldedSectionB}</b>),
+      }}
+    />),
+  ];
+  return factoids[Math.floor(Math.random() * (factoids.length))];
```
mikix, line 36: I know we've talked about this before, but it seems kind of wild to me that we aren't measuring if one factoid is better than the other.

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -0,0 +1,115 @@
...
+import messages from './messages';
+import { recordModalClosing, recordStreakCelebration } from './utils';
+
+function getRandomFactoid(intl) {
+  const boldedSectionA = intl.formatMessage(messages.StreakFactoidABoldedSection);
+  const boldedSectionB = intl.formatMessage(messages.StreakFactoidBBoldedSection);
+  const factoids = [
+    (<FormattedMessage
+      id="learning.streakcelebration.factoida"
+      defaultMessage="Users who learn 3 days in a row {bolded_section} than those who don't."
```
mikix: The only difference in this outer sentence is `than` vs `vs.`? :-/ Whatever UI wants.

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -0,0 +1,115 @@
...
+  return factoids[Math.floor(Math.random() * (factoids.length))];
+}
+
+function StreakModal({
+  courseId, metadataModel, intl, open, ...rest
+}) {
+  const { org, celebrations } = useModel(metadataModel, courseId);
+  const factoid = getRandomFactoid(intl);
+  // eslint-disable-next-line no-unused-vars
+  const [randomFactoid, setRandomFactoid] = useState(factoid);
```
mikix: Using this state like this, avoids changing the factoid on re-render? Cute. Might be worth a comment, but no biggie.

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+  const OnMobile = layout.is('mobile');
+  const OnDesktop = layout.isAtLeast('desktop');
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    if (open) {
+      recordStreakCelebration(org, courseId);
+    }
+  }, [open]);
```
mikix: Why not add org and courseId to the deps list?

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+  function CloseText() {
+    return (
+      <span>
+        {intl.formatMessage(messages.StreakButton)}
+        <span className="sr-only">{intl.formatMessage(messages.StreakButtonSrOnly)}</span>
```
mikix: Should the normal text be marked `aria-hidden`? The way it is now, the screen reader would say `Keep it up Close modal and continue`.

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -0,0 +1,115 @@
...
+      </span>
+    );
+  }
+
+  return (
+    <div>
+      <Modal
+        dialogClassName="StreakModal modal-dialog-centered"
```
mikix: Capitalized classnames is not common.

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -0,0 +1,115 @@
...
+        body={(
+          <>
+            <p>{intl.formatMessage(messages.StreakBody)}</p>
```
mikix: Likewise, messages properties like this usually follow normal javascript case (`messages.streakBody`).

src/shared/streak-celebration/StreakCelebrationModal.scss
```diff
@@ -0,0 +1,67 @@
...
+      right: 1rem;
+    }
+
+    button::after {
+      content: none;
+    }
+  }
+
+  .modal-title {
+    padding-top: 20px;
```
mikix: I dislike any hardcoded pixels in css, here and elsewhere in this file. Can you make them `rem` values instead? Or even better, use bootstrap's spacing values.

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -0,0 +1,115 @@
...
+    (<FormattedMessage
+      id="learning.streakcelebration.factoida"
+      defaultMessage="Users who learn 3 days in a row {bolded_section} than those who don't."
+      values={{
+        bolded_section: (<b>{boldedSectionA}</b>),
+      }}
+    />),
+    (<FormattedMessage
+      id="learning.streakcelebration.factoidb"
+      defaultMessage="Users who learn 3 days in a row {bolded_section} vs. those who don't."
```
mikix: In both these sentences, we should use proper unicode apostrophes, not ascii ones: ’ not '.

src/shared/streak-celebration/utils.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+import { updateModel } from '../../generic/model-store';
+
+function recordStreakCelebration(org, courseId) {
+  // Tell our analytics
+  const { administrator } = getAuthenticatedUser();
+  sendTrackEvent('edx.ui.lms.celebration.streak.opened', {
+    org_key: org,
+    courserun_key: courseId,
+    course_id: courseId, // should only be courserun_key, but left as-is for historical reasons
```
mikix: In new code, you can leave this one out.

src/shared/streak-celebration/utils.jsx
mikix: But if they both happen at same time, they'd each update the models with their own slices yeah, and it would be fine? Or are you worried about stale `celebrations` values from before the other modal updated the models? You could re-get the `celebrations` data from the redux model bofore updating the model here, maybe to work around that? Seems corner case either way, as you say.

Like, what is the UX in that case? Does the user see two modals right after each other? Or two models on top of each other? I'd guess Hilary would have opinions on what we should do in this weird case.

It's not an insane situation though. Like, if I close my laptop on the last piece of a section for the evening, because I don't want to start a new section right before bed... then in the morning I open it and start again with the same browser tab, I might reasonably see both triggers for the same page request.

src/shared/streak-celebration/messages.js
```diff
@@ -0,0 +1,33 @@
...
+  StreakButton: {
+    id: 'learning.streakcelebration.button',
+    defaultMessage: 'Keep it up',
+  },
+  StreakButtonSrOnly: {
+    id: 'learning.streakcelebration.buttonsronly',
+    defaultMessage: 'Close modal and continue',
+    description: 'Screenreader label for StreakButton text',
+  },
+  StreakFactoidABoldedSection: {
```
ciduarte: Can we make the first letter of all of these lowercase? `streakFactoidABoldedSection`

src/shared/streak-celebration/messages.js
```diff
@@ -0,0 +1,33 @@
...
+    id: 'learning.streakcelebration.factoidaboldedsection',
```
ciduarte: Can we camel case these ID's? `learning.streakCelebration.factoidA.boldedSection`?

src/shared/streak-celebration/StreakCelebrationModal.scss
ciduarte: +1, please use rem values

src/shared/streak-celebration/StreakCelebrationModal.scss
```diff
@@ -0,0 +1,67 @@
...
+      width: 50%;
+    }
+  }
+
+  .calendar {
+    margin-top: 30px;
+    margin-bottom: 30px;
+  }
+
+  .factoidWrapper {
```
ciduarte: `factoid-wrapper`

src/shared/streak-celebration/StreakCelebrationModal.scss
```diff
@@ -0,0 +1,67 @@
...
+      @extend .btn-primary;
+      font-size: 1.2rem;
+      width: 50%;
+    }
+  }
+
+  .calendar {
+    margin-top: 30px;
+    margin-bottom: 30px;
+  }
```
ciduarte: Calendar? Is this meant to be the modal-image? Can we change it to that, if so?

src/shared/streak-celebration/StreakCelebrationModal.scss
```diff
@@ -0,0 +1,67 @@
...
+  .factoidWrapper {
+    font-size: 14px;
+    text-align: left;
+    max-width: 85%;
+  }
+}
+
+
+@media screen and (min-width: 570px) {
+  .StreakModal {
```
ciduarte: `streak-modal`

src/shared/streak-celebration/StreakCelebrationModal.scss
```diff
@@ -0,0 +1,67 @@
+.StreakModal {
```
ciduarte: `streak-modal`

src/shared/streak-celebration/StreakCelebrationModal.scss
```diff
@@ -0,0 +1,67 @@
...
+  .StreakModal {
+    width: 532px;
+    max-width: 532px;
+
+    .calendar {
+      margin-top: 40px;
+      margin-bottom: 40px;
+    }
+
+    .factoidWrapper {
```
ciduarte: factoid-wrapper

docs/decisions/0008-streak-celebration.md
```diff
@@ -0,0 +1,32 @@
...
+
+Relevant Ticket:
+https://openedx.atlassian.net/browse/AA-304
+
+Decisions
+--------
+
+1. **Retrieving streak celebration data**
+We use the course home and courseware metadata APIs to retrieve the data determining whether we should celebrate a streak.
+Some MFEs implement progressive loading and use separate requests to retrieve data that is only necessary for a given component. However, the learning MFE has gone the direction of retrieving all the necessary data upfront.
```
ciduarte: Can we change this first sentence's wording to "Some MFEs implement progressive loading, where individual requests only retrieve the data necessary for a given component."

docs/decisions/0008-streak-celebration.md
```diff
@@ -0,0 +1,32 @@
...
+
+2. **No custom caching for streak celebration data**
+Because we retrieve the streak celebration data as part of the metadata apis we will be retrieving the streak data for every page load even though it cannot change more than once per day. We considered adding custom caching to just this element of the metadata API. We chose to avoid doing this to not overcomplicate the metadata API, as it would make more sense to have holistic rather than piecemeal caching strategies.
```
ciduarte: Editing for clarity and concision: "Although the streak celebration data cannot change more than once per day, it will be retrieved on every page load. This is because the streak celebration data comes through getCourseHomeCourseMetadata and getCourseMetadata. We considered adding custom caching for the streak celebration data, but decided against it in order to avoid overcomplicating the metadata APIs."

docs/decisions/0008-streak-celebration.md
```diff
@@ -0,0 +1,32 @@
+Streak Celebration
```
ciduarte: I went back and forth on the addition of this to our decision docs. These are meant to be architectural decisions, and it feels like most of this is just an example of how a new feature we've written is using the existing architecture that has already been described in previous decision files, along with feature decisions that feel like they belong in the backend.

The only thing I'd consider an architectural decision is our decision to not use custom caching (point #2). But to get to point #2, do we need all of the additional context? Part of me feels like this doc is unnecessary or should live in the backend. 🤷

ciduarte (commented):
Sorry for duplicate comments, hadn't seen what Mike Terry said.

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -0,0 +1,115 @@
...
+            <p className="calendar">
+              <OnMobile>
+                <img src={StreakMobileImage} alt="" className="img-fluid" />
+              </OnMobile>
+              <OnDesktop>
+                <img src={StreakDesktopImage} alt="" className="img-fluid" />
+              </OnDesktop>
+            </p>
+            <div className="factoidRow row mt-3 mx-3 py-3 bg-light-300">
+              <Icon className="col-small ml-3 lightbulb" src={Lightbulb} />
```
ciduarte: Why lightbulb in the className?

docs/decisions/0008-streak-celebration.md
MatthewPiatetsky: changed

docs/decisions/0008-streak-celebration.md
MatthewPiatetsky: changed to "Although the streak celebration data cannot change more than once per day, it will be retrieved on every page load. This is because the streak celebration data comes through the course home and courseware metadata APIs. We considered adding custom caching for the streak celebration data, but decided against it in order to avoid overcomplicating the metadata APIs."

docs/decisions/0008-streak-celebration.md
MatthewPiatetsky: I'm flexible on how and where this team wants to document stuff, i recently worked on a sdn feature where the revenue team wanted me to write up some decisions in an arch doc like this one so I was approaching it similarly
https://github.com/edx/ecommerce/blob/master/docs/decisions/0007-sdn-fallback.rst

i do think it would be nice to have some context around the feature in one place though if future engineers are wondering about certain decisions, rather than two documents, one for frontend and one for backend

package.json
MatthewPiatetsky: fixed

src/shared/streak-celebration/StreakCelebrationModal.jsx
MatthewPiatetsky: can you explain why?

src/shared/streak-celebration/StreakCelebrationModal.jsx
MatthewPiatetsky: i don't remember talking about this?  
is this a question I should ask product on whether we want to run an experiment?

src/shared/streak-celebration/StreakCelebrationModal.jsx
MatthewPiatetsky: changed

src/shared/streak-celebration/StreakCelebrationModal.jsx
MatthewPiatetsky: i think that's what product wanted it to say from this thread
https://edx-internal.slack.com/archives/G019J2RMNEA/p1613086140063800

src/shared/streak-celebration/StreakCelebrationModal.jsx
MatthewPiatetsky: changed

src/shared/streak-celebration/StreakCelebrationModal.jsx
MatthewPiatetsky: changed

src/shared/streak-celebration/StreakCelebrationModal.jsx
MatthewPiatetsky: ah i had css for it but then I removed that so removing the class

src/shared/streak-celebration/StreakCelebrationModal.scss
MatthewPiatetsky: changed

src/shared/streak-celebration/StreakCelebrationModal.scss
MatthewPiatetsky: changed

src/shared/streak-celebration/StreakCelebrationModal.scss
MatthewPiatetsky: changed

src/shared/streak-celebration/StreakCelebrationModal.scss
MatthewPiatetsky: changed

src/shared/streak-celebration/messages.js
MatthewPiatetsky: changed

src/shared/streak-celebration/messages.js
MatthewPiatetsky: changed (i didn't add the period though in the name since i'm not sure if the two periods here is. aconvention that downstream code could be depending on?)

src/shared/streak-celebration/utils.jsx
MatthewPiatetsky: removed

src/shared/streak-celebration/utils.jsx
MatthewPiatetsky: for that particular example i think it would still not happen because you'd see the streak modal when you open your computer back up but you would see the first section modal once you click next 
though there could certainly be overlap with some hypothetical future celebration (i did see what two overlapping modals looks like and its bad lol) 

right now I am getting the celebrations data before updating and only updating the key for the given celebration.  
so if two celebrations come back as true on the same page it would show two modals. 

i could have each celebration just override the celebration dict but that seems a bit weird
perhaps something like a lock where I set a modalIsOpen state variable and have any celebration that opens a modal check the variable before celebrating? then when you close one modal it could popup the second modal

src/shared/streak-celebration/StreakCelebrationModal.jsx
MatthewPiatetsky: added comment

src/shared/streak-celebration/StreakCelebrationModal.scss
MatthewPiatetsky: changed

src/shared/streak-celebration/StreakCelebrationModal.scss
MatthewPiatetsky: changed to rem values

src/shared/streak-celebration/messages.js
```diff
@@ -0,0 +1,34 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  streakHeader: {
+    id: 'learning.streakcelebration.header',
```
ciduarte: nit: `streakCelebration` in all of these IDs :)

src/shared/streak-celebration/StreakCelebrationModal.jsx
ciduarte: I think we talked about this when we first groomed the ticket before you joined the team... but I also can't remember why we decided not to experiment with these... Might be worth asking Kaitlin?

docs/decisions/0008-streak-celebration.md
ciduarte: I haven't looked at your backend PR, but it might be worth keeping this context near the backend code?

Or we could have a Confluence page in our team's space to document the full behavior of this feature. I spoke briefly with Jason about making our Confluence documentation a little bit more robust so other teams have access to our feature decisions/behavior (what behavior to expect, where things live, what flags they might need to access a feature, etc.), and this (along with the other milestones) feel like a good candidate to add there.

Either way, I think what you've documented here is valuable and I don't want to lose it, but it feels like we could find a better spot for it...

ciduarte (approved):
Thanks for fixing those class names. Just one more nit and a suggestion about where to put the doc, but looks good!

(I'm approving based on my comments, but I see on-going convo and an open question to MT so I would also wait for his approval to merge)

src/shared/streak-celebration/utils.jsx
mikix: > you'd see the streak modal when you open your computer back up

Why? Your browser wouldn't auto-refresh the page to make a new request, right?

src/shared/streak-celebration/StreakCelebrationModal.jsx
mikix: My comment was not meant to start a whole discussion. We talked about it in grooming. I don't remember the reason, but we decided not to.

src/shared/streak-celebration/StreakCelebrationModal.jsx
mikix: Ah, yeah I don't mind `Keep it up. Close modal and continue` -- but the code right now is missing the period, which might be hard for a screen reader to read aloud properly.

src/shared/streak-celebration/StreakCelebrationModal.scss
mikix: OK, but for stuff like this:
` @media screen and (min-width: 35.625rem) {`

You can just use pixels. Working with screen sizes is a pixel thing, not related to how zoomed in you are on that screen.

src/shared/streak-celebration/StreakCelebrationModal.jsx
mikix: It looks nicer.

Open any book and they use curved apostrophes, not straight ones. The straight one is just a programmer/typewriter thing. Normies expect curved apostrophes. Same with double quotes.

Also, you already use a proper apostrophe in the `learning.streakCelebration.body` message. And the rest of the MFE uses proper apostrophes. So at least for consistency's sake.

src/shared/streak-celebration/messages.js
MatthewPiatetsky: changed

src/shared/streak-celebration/utils.jsx
MatthewPiatetsky: i guess depends on whether your computer shuts off or sleeps. 
but per our conversation in slack, clarified with UX and Hilary said that we should only show one celebration on a given page and that in the weird edge case to show the three day streak over the first section celebration so I tried to implement that in the courseware/data/api

src/shared/streak-celebration/StreakCelebrationModal.jsx
MatthewPiatetsky: updated

src/shared/streak-celebration/StreakCelebrationModal.scss
MatthewPiatetsky: updated

src/shared/streak-celebration/StreakCelebrationModal.jsx
MatthewPiatetsky: added a period

package-lock.json
mikix: npm install does do a lot - deps of deps get updated when the top one updates, etc. It's probably fine.

src/courseware/course/bookmark/data/redux.test.js
```diff
@@ -5,7 +5,7 @@ import { getConfig } from '@edx/frontend-platform';
 
 import * as thunks from './thunks';
 
-import executeThunk from '../../../../utils';
+import { executeThunk } from '../../../../utils';
```
mikix, line 8: Tsk someone didn't plan ahead and just start with non-default exports.

src/courseware/data/api.js
```diff
@@ -115,6 +116,18 @@ function normalizeTabUrls(id, tabs) {
   });
 }
 
+function oneCelebrationAtATime(celebrations) {
+  // We don't want learners to have too much fun
+  // and see two celebration modals on a single page
+  // If the streak celebration and first section celebrations both come back as true
+  // we will only show the streak celebration
+  const celebrationsObject = celebrations;
```
mikix: Is that meant to be a copy? `{ ...celebrations}` might do instead, if so.

src/courseware/data/api.js
```diff
@@ -115,6 +116,18 @@ function normalizeTabUrls(id, tabs) {
...
+  if (celebrations && celebrations.firstSection && celebrations.streakLengthToCelebrate) {
+    celebrationsObject.firstSection = false;
```
mikix: This is going to have a weird effect. If we hit this edge case, we won't celebrate first section. Fine. But we aren't telling backend to stop setting `firstSection`, so you'll see it at the beginning of your second section, but the UI will act like it's your first section.

Might want to report back to backend to turn off first section celebration in this case?

MatthewPiatetsky: good point, I added an if statement to shouldCelebrateOnSectionLoad for the first section celebration instead.  
not a very extensible option, but might be good enough for the time being?

mikix: Yeah I think the current solution is fine for now. But we may well want a more centralized celebration clearing house system if we grow many more.

mikix (approved):
Love it, let's try this out!  (I didn't run this locally, but assume it looks beatiful and works fine.)

PR #355, opened 2021-01-29
AA-492: Add event data for consumption in the backend

Dillon-Dumesnil:
Related to https://github.com/edx/edx-platform/pull/26264

Important, needs to be merged AFTER the above otherwise the endpoint will start erroring due to the new parameter.

### DIFF
--- a/src/course-home/data/thunks.js
+++ b/src/course-home/data/thunks.js
@@ -90,9 +90,9 @@ export function requestCert(courseId) {
   return async () => postRequestCert(courseId);
 }
 
-export function resetDeadlines(courseId, getTabData) {
+export function resetDeadlines(courseId, model, getTabData) {
   return async (dispatch) => {
-    postCourseDeadlines(courseId).then(response => {
+    postCourseDeadlines(courseId, model).then(response => {
       const { data } = response;
       const {
         header,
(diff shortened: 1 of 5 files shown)


src/course-home/data/thunks.js
```diff
@@ -111,9 +111,10 @@ export async function saveCourseGoal(courseId, goalKey) {
 
 export function processEvent(eventData, getTabData) {
   return async (dispatch) => {
+    const { research_event_data: researchEventData } = eventData;
```
Dillon-Dumesnil, line 116: Pulling this out early so the data doesn't get camelCased and is easier to use when it's passed to the backend.

mikix: Maybe that comment is worth being in-code? But I don't feel strongly - a dev would probably see that.

mikix (approved):

src/course-home/data/thunks.js
Dillon-Dumesnil: Yeah, that's what I'm hoping.

Dillon-Dumesnil: I'll just update it. Need to change edx-platform anyway

PR #357, opened 2021-02-08
AA-614: Show effort estimation if the backend provides it

mikix:
https://openedx.atlassian.net/browse/AA-614

Some screenshots:
Screenshot from 2021-02-08 13-07-04
Screenshot from 2021-02-08 13-06-09
Screenshot from 2021-02-08 13-05-59

### DIFF
--- a/src/course-home/outline-tab/SequenceLink.jsx
+++ b/src/course-home/outline-tab/SequenceLink.jsx
@@ -12,6 +12,7 @@ import { faCheckCircle as fasCheckCircle } from '@fortawesome/free-solid-svg-ico
 import { faCheckCircle as farCheckCircle } from '@fortawesome/free-regular-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
+import EffortEstimate from '../../shared/effort-estimate';
 import { useModel } from '../../generic/model-store';
 import messages from './messages';
 
@@ -62,10 +63,13 @@ function SequenceLink({
               />
             )}
           </div>
-          <div className="col-10 p-0 ml-3 text-break">{displayTitle}</div>
-          <span className="sr-only">
-            , {intl.formatMessage(complete ? messages.completedAssignment : messages.incompleteAssignment)}
-          </span>
+          <div className="col-10 p-0 ml-3 text-break">
+            <span className="align-middle">{displayTitle}</span>
+            <span className="sr-only">
+              , {intl.formatMessage(complete ? messages.completedAssignment : messages.incompleteAssignment)}
+            </span>
+            <EffortEstimate className="ml-3 align-middle" block={sequence} />
+          </div>
         </div>
         {due && (
           <div className="row w-100 m-0 ml-3 pl-3">
--- a/src/courseware/course/sequence/sequence-navigation/UnitNavigationEffortEstimate.jsx
+++ b/src/courseware/course/sequence/sequence-navigation/UnitNavigationEffortEstimate.jsx
@@ -1,0 +1,66 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { useSelector } from 'react-redux';
+
+import EffortEstimate from '../../../../shared/effort-estimate';
+import { sequenceIdsSelector } from '../../../data';
+import { useModel } from '../../../../generic/model-store';
+
+// This component exists to peek ahead at the next subsection or section and grab its estimated effort.
+// If we should be showing the next block's effort, we display the title and effort instead of "Next".
+// This code currently tries to handle both section and subsection estimates. But once AA-659 happens, it can be
+// simplified to one or the other code path.
+
+function UnitNavigationEffortEstimate({ children, sequenceId, unitId }) {
+  const sequenceIds = useSelector(sequenceIdsSelector);
+  const sequenceIndex = sequenceIds.indexOf(sequenceId);
+  const nextSequenceId = sequenceIndex < sequenceIds.length - 1 ? sequenceIds[sequenceIndex + 1] : null;
+  const sequence = useModel('sequences', sequenceId);
+  const nextSequence = useModel('sequences', nextSequenceId);
+  const nextSection = useModel('sections', nextSequence ? nextSequence.sectionId : null);
+
+  if (!sequence || !nextSequence) {
+    return children;
+  }
+
+  const isLastUnitInSequence = sequence.unitIds.indexOf(unitId) === sequence.unitIds.length - 1;
+  if (!isLastUnitInSequence) {
+    return children;
+  }
+
+  let blockToShow = nextSequence;
+  // The experimentation code currently only sets effort on either sequences, sections, or nothing. If we don't have
+  // sequence info, we are either doing sections or nothing. Let's look into it.
+  if (!nextSequence.effortActivities && !nextSequence.effortTime) {
+    if (!nextSection.effortActivities && !nextSection.effortTime) {
+      return children; // control group - no effort estimates at all
+    }
+
+    // Are we at a section border? If so, let's show the next section's effort estimates
+    if (sequence.sectionId !== nextSequence.sectionId) {
+      blockToShow = nextSection;
+    }
+  }
+
+  // Note: we don't use `children` here - we replace it with the next section name.
+  // AA-659: remember to add a translation for Next Up
+  return (
+    <div className="d-inline-block text-wrap">
+      Next Up: {blockToShow.title}
+      <EffortEstimate className="d-block mt-1" block={blockToShow} />
+    </div>
+  );
+}
+
+UnitNavigationEffortEstimate.propTypes = {
+  children: PropTypes.node,
+  sequenceId: PropTypes.string.isRequired,
+  unitId: PropTypes.string,
+};
+
+UnitNavigationEffortEstimate.defaultProps = {
+  children: null,
+  unitId: null,
+};
+
+export default UnitNavigationEffortEstimate;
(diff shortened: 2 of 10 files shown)


src/courseware/course/sequence/sequence-navigation/UnitNavigationEffortEstimate.jsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
```
MatthewPiatetsky, line 1: probably ok for an experiment, but otherwise i think we'd want some documentation on what this feature is. 
and what the specific business logic that this code should be implementing is

src/shared/effort-estimate/EffortEstimate.jsx
```diff
@@ -0,0 +1,52 @@
...
+function EffortEstimate(props) {
+  const {
+    block: {
+      effortActivities,
+      effortTime,
+    },
+    className,
+  } = props;
+
+  // FIXME: This is not properly internationalized. This is just an experiment right now, so I chose to not mark
```
MatthewPiatetsky, line 17: could we maybe make a ticket for "things to fix if effort estimation is implemented" and then mark things in both PRs with that ticket number?  
Seeing the 6 year old TODOs in the edx-platform code makes me sad lol

MatthewPiatetsky (commented):
1. Is the grey color bright enough for a11y contrast rules?
2. Do the estimates show up ok with long section titles?
3. Probably a question for UX but I would have expected that once you get to a high number of minutes it would display something like X hours Y minutes rather than large amount of minutes?

src/shared/effort-estimate/EffortEstimate.jsx
mikix: > Seeing the 6 year old TODOs in the edx-platform code makes me sad lol

Making a ticket and referencing it in code does not fix that problem, as plenty of comments in our code show. :)

But yeah, I can make a ticket. Our team shares your distaste of leftover code, no question.

src/shared/effort-estimate/index.js
```diff
@@ -0,0 +1,3 @@
+import EffortEstimate from './EffortEstimate';
```
ciduarte, line 1: About this `shared` folder - I had been using our `generic` folder to store things that are shared between course home and courseware. Should this live there?

ciduarte (approved):
Not much to add, just wondering about that shared folder.

src/shared/effort-estimate/EffortEstimate.jsx
MatthewPiatetsky: certainly true though i think in the past we've remembered things like this more often when the tickets were in our backlog vs just code comments

src/shared/effort-estimate/index.js
mikix: Oh right - good eye. That's something that Matthew noticed. See https://github.com/edx/frontend-app-learning/blob/master/src/generic/README.md

It looks like TNL meant for that folder to hold stuff that like... could theoretically live in paragon? And I think we took it as a "shared code between two sides of MFE" folder. So Matthew was going to move some stuff into a new `shared` folder for that kind of stuff, and I was trying to use the new location too.

ciduarte: Oh okay great good to know

src/shared/effort-estimate/EffortEstimate.jsx
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  return (
+    <span
+      className={classNames('text-gray-500 text-monospace align-middle', className)}
+      style={{ fontSize: '0.8em' }}
+    >
+      {content}
+    </span>
+  );
```
ciduarte: I'm back with my a11y hat. Curious how we want this to read over a screen reader? I think the questions and activities are sufficient, but for time we might want to add a more context? ex. ", 10 minutes estimated time to complete"

I'll ask this in the UX review, but jotting it down here for my own memory

src/shared/effort-estimate/EffortEstimate.jsx
mikix: You already know this from other sources, but just commenting here for posterity: https://openedx.atlassian.net/browse/AA-659 holds the cleanups we'll need to do.

mikix:
Answering some of Matthew's top-level comments:

> Is the grey color bright enough for a11y contrast rules?

According to my a11y tool, yes.

> Do the estimates show up ok with long section titles?

Yes to my eyes. They wrap and make the box taller.

> Probably a question for UX but I would have expected that once you get to a high number of minutes it would display something like X hours Y minutes rather than large amount of minutes?

We talked about this in UX review, and looks like we'll just stick with minutes for now.

src/shared/effort-estimate/EffortEstimate.jsx
mikix: These are the notes from UX review:

> Min > “Minutes” for screen reader
> “11 minutes next” < add label to section level Aria tag

First one is clear. I'm interpreting the second as having the screen reader see "Next Up: My Cool Title 11 minutes next" -- where the addition is the last word `next`? Is that right? Feels a little redundant with Next Up, but easy enough to add.

mikix: OK after discussion I have not done the "11 minutes next" bit - we think that was a duplicate comment for min->minutes in the Next button.

mikix:
This is ready for a re-review. I believe all comments are addressed. Only particularly interesting change is for a11y: min -> minutes for screen readers.

I'm not doing any tests because (A) it's just an experiment now and (B) this stuff is obvious enough and non-interactive enough that I think it falls under "devs will see it during normal course" and doesn't need a js test.

src/course-home/outline-tab/SequenceLink.jsx
```diff
@@ -62,10 +63,13 @@ function SequenceLink({
...
             )}
           </div>
-          <div className="col-10 p-0 ml-3 text-break">{displayTitle}</div>
-          <span className="sr-only">
-            , {intl.formatMessage(complete ? messages.completedAssignment : messages.incompleteAssignment)}
-          </span>
+          <div className="col-10 p-0 ml-3 text-break">
+            <span className="align-middle">{displayTitle}</span>
+            <span className="sr-only">
+              , {intl.formatMessage(complete ? messages.completedAssignment : messages.incompleteAssignment)}
```
Dillon-Dumesnil, line 69: What's the comma for in sr-only text? So it takes a break between the title and this text?

MatthewPiatetsky (approved):
1. have you checked how it looks on mobile web?
2. how is it switching between the buckets, is it just that the effort estimate will be populated/not populated on certain blocks based on the waffle flag/block transformer and the code doesn't care if its because it doesn't have a time or is in a different bucket?
3. do we need any events for experiment purposes or are we able to measure everything we need by having the ExperimentWaffleFlag?

src/courseware/course/sequence/sequence-navigation/UnitNavigationEffortEstimate.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+function UnitNavigationEffortEstimate({ children, sequenceId, unitId }) {
+  const sequenceIds = useSelector(sequenceIdsSelector);
+  const sequenceIndex = sequenceIds.indexOf(sequenceId);
+  const nextSequenceId = sequenceIndex < sequenceIds.length - 1 ? sequenceIds[sequenceIndex + 1] : null;
+  const sequence = useModel('sequences', sequenceId);
+  const nextSequence = useModel('sequences', nextSequenceId);
+  const nextSection = useModel('sections', nextSequence ? nextSequence.sectionId : null);
+
+  if (!sequence || !nextSequence) {
```
Dillon-Dumesnil, line 22: Is the not sequence just being safe? I see that sequenceId is required so unclear when sequence would be null.

Dillon-Dumesnil (approved):
Small questions, otherwise lgtm

Also echo Matthew's third point in particular in terms of how we are tracking this.

src/courseware/course/sequence/sequence-navigation/UnitNavigationEffortEstimate.jsx
mikix: Yeah safety. I assume some other bit of code would have already exploded if sequenceId wasn't actually in our sequence list.

src/course-home/outline-tab/SequenceLink.jsx
mikix: I assume so? That the voice reading it knows to pause and not read "Ethics of Computing completed" as one thing.

mikix:
> have you checked how it looks on mobile web?

Yeah, any long text wraps cleanly.

> how is it switching between the buckets, is it just that the effort estimate will be populated/not populated on certain blocks based on the waffle flag/block transformer and the code doesn't care if its because it doesn't have a time or is in a different bucket?

Yeah I wrote the MFE to be bucket-dumb in the sense that it will display any estimates it's given at section and subsection levels. The backend only gives it what the current user's bucket decides. Post experiment, all blocks will have the estimate data, and we'll just take out the effort estimation code at the wrong layer in the MFE.

> do we need any events for experiment purposes or are we able to measure everything we need by having the ExperimentWaffleFlag?

I don't believe so, since the bucketing will signal which users / courses are in which bucket. Then we do normal engagement tracking for those cases. There's no real user action to signal on anyway.

mikix:
> Yeah, any long text wraps cleanly.

Whoops, found one case where that wasn't true and fixed it. Thanks for the reminder, Matthew.

MatthewPiatetsky (approved):

PR #363, opened 2021-02-09
Hackathon XXV: Initial implementation of frontend plugins

davidjoy:
Uses webpack 5 module federation to dynamically load code based on a JS file URL.


src/plugin-test/PluginTest.jsx
```diff
@@ -0,0 +1,157 @@
...
+  if (failed) {
+    return <h2>Failed to load plugin script: {plugin.url}</h2>;
+  }
+
+  const PluginComponent = React.lazy(
+    loadPluginComponent(plugin.scope, plugin.module),
+  );
+
+  return (
+    <React.Suspense fallback="Loading Plugin">
```
adamstankiewicz: [consideration] This may already be on your roadmap for this implementation. if so, please ignore; I'm just excited by the potential of Module Federation 😄

I might suggest letting Webpack 5 module federation take care of the dynamic loading of scripts natively as configured in the Webpack config files via `remotes`.

An [example Webpack config](https://github.com/adamstankiewicz/frontend-enterprise-app/blob/main/packages/container/config/webpack.dev.js#L18-L22
) for consuming remoteEntry.js files (note: the remoteEntry.js file is a bit different than a normal Webpack bundle): 

```jsx
remotes: {
  pluginOne: 'pluginOne@//localhost:2001/remoteEntry.js',
  pluginTwo: 'pluginTwo@//localhost:2002/remoteEntry.js',
},
```

A relatively solid pattern that is recommended by this [Udemy course](https://www.udemy.com/course/microfrontend-course/) on micro-frontends and `single-spa` suggests letting your plugins export a generic API that the consuming application (here, frontend-app-learning) can use as necessary to "orchestrate" where and when the plugin should appear.

As an example, this ["Dashboard" plugin](https://github.com/adamstankiewicz/frontend-enterprise-app/blob/main/packages/dashboard/src/bootstrap.js#L7) has a `bootstrap.js` file that exports a generic interface, in this case a `mount` function, which is then [consumed by this host application](https://github.com/adamstankiewicz/frontend-enterprise-app/blob/main/packages/container/src/components/DashboardApp.js#L4). With this approach, the plugin can use any framework (not just React) or even just vanilla JS because the plugin "owns" how it gets mounted into the DOM. The host/container app only knows that a `mount` function exists and calls it when it needs to render that plugin/micro-frontend (but doesn't need to know _how_ the component mounts to the DOM, e.g. React vs. vanilla JS).

The interesting bit is with the routing; a plugin might have its own "nested" routing set up so you need a way to share the same browser history object to communicate route changes.

davidjoy: Hmm, if I'm understanding correctly, this app has a slightly different use case than the one you describe.  It needs "dynamic remotes" - i.e., remotes that are resolved only at runtime. 

It's based off of this sample app here: https://github.com/module-federation/module-federation-examples/tree/master/advanced-api/dynamic-remotes

I've tweaked their taxonomy a bit to make it fit in more with ours, but the idea is that you don't know what remotes exist at build-time, so it can't go in the webpack config.

As for what the plugins expose, yes, I imagine it'd be standardized and generic so that the apps would know how to bootstrap them.

adamstankiewicz: Cool, that makes sense! Slightly different use cases :)

xitij2000:
@davidjoy Since I've been looking at theming recently, I tried combining it with module federation and I have a system running where you can dynamically get a list of themes and apply them at run time. 

I'm hosting a demo here: https://modular.theming.prototypes.kshitij.dev/ 
The source code is here: https://gitlab.com/xitij2000/federated-theme-proto/

I plan to write up an ADR about this, and would love your feedback!

davidjoy:
@xitij2000 Cool!  How does it handle stylesheets?  Is webpack just dynamically adding the CSS as a <style> tag when it loads the JS that imports the scss?

xitij2000:
@davidjoy  My understanding is that there is some level of control here. I haven't experimented with this aspect as much, but the current mechanism I'm using (which webpack calls `lazyStyleTag`) wraps everything in a style tag, and provides an API to apply/unapply a stylesheet using `styleseet.use()` and `stylesheet.unuse()`.

So in this case the theme is just a module that can be directly loaded by the frontend. You can call `use()` on it to put it in use. 

The main aim is to just have the theme loaded dynamically, so it can be replaced easily.

davidjoy:
This is great @xitij2000 - I'm just now starting to pick this work up again and would love to dig into what you have here.  I think it has applicability beyond even theming.  One of my concerns had been how CSS can get dynamically attached to the page, and this feels like it solves it.  Having plugins that can modify the styling of the page could be really valuable. 

More soon!

xitij2000:
@davidjoy Looking forward to seeing what you're working on. 

I definitely see applications beyond theming, for instance, it can be used to embed views from one MFE into another without  needing iframes. For instance, a sidebar could be added by dynamically adding a sidebar plugin that's returned via an API.

davidjoy:
Exactly.  I'm nearly as excited about the non-plugin uses as the plugin ones.

davidjoy:
Closing this PR so it's not hanging around, but will keep the branch.

PR #364, opened 2021-02-11
[MM-P2P] optimizely experiment overrides for course_home

muselesscreator:
Design Doc: https://www.figma.com/file/cTjgRZdeQlJT2A0aEV9Rln/MM-Rev-Shift?node-id=8%3A6
JIRA: https://openedx.atlassian.net/browse/EDUCATOR-5542
Demo:

https://user-images.githubusercontent.com/5533134/107658930-a4a10200-6c54-11eb-8e80-4c1733df415d.mov

### DIFF
--- a/src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
+++ b/src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import {
   FormattedMessage, FormattedDate, injectIntl, intlShape,
@@ -7,8 +7,18 @@ import { Hyperlink } from '@edx/paragon';
 
 import { Alert, ALERT_TYPES } from '../../generic/user-messages';
 import messages from './messages';
+import AccessExpirationAlertMMP2P from './AccessExpirationAlertMMP2P';
 
 function AccessExpirationAlert({ intl, payload }) {
+  /** [MM-P2P] Experiment */
+  const [showMMP2P, setShowMMP2P] = useState(!!window.experiment__home_alert_bShowMMP2P);
+  if (window.experiment__home_alert_showMMP2P === undefined) {
+    window.experiment__home_alert_showMMP2P = (val) => {
+      window.experiment__home_alert_bShowMMP2P = !!val;
+      setShowMMP2P(!!val);
+    };
+  }
+
   const {
     accessExpiration,
     userTimezone,
@@ -46,6 +56,13 @@ function AccessExpirationAlert({ intl, payload }) {
           }}
         />
       </Alert>
+    );
+  }
+
+  /** [MM-P2P] Experiment */
+  if (showMMP2P) {
+    return (
+      <AccessExpirationAlertMMP2P intl={intl} payload={payload} />
     );
   }
 
--- a/src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
+++ b/src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
@@ -62,7 +62,7 @@ function AccessExpirationAlert({ intl, payload }) {
   /** [MM-P2P] Experiment */
   if (showMMP2P) {
     return (
-      <AccessExpirationAlertMMP2P intl={intl} payload={payload} />
+      <AccessExpirationAlertMMP2P payload={payload} />
     );
   }
 
--- a/src/course-home/outline-tab/DateSummary.jsx
+++ b/src/course-home/outline-tab/DateSummary.jsx
@@ -15,10 +15,12 @@ export default function DateSummary({
 
   /** [MM-P2P experiment] */
   const [showMMP2P, setShowMMP2P] = useState(window.experiment__home_dates_bShowMMP2P);
-  window.experiment__home_dates_showMMP2P = (value) => {
-    setShowMMP2P(!!value);
-    window.experiment__home_dates_bShowMMP2P = !!value;
-  };
+  if (window.experiment__home_dates_showMMP2P !== undefined) {
+    window.experiment__home_dates_showMMP2P = (value) => {
+      setShowMMP2P(!!value);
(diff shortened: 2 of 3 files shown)


src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
```diff
@@ -9,6 +9,14 @@ import { Alert, ALERT_TYPES } from '../../generic/user-messages';
 import messages from './messages';
 
 function AccessExpirationAlert({ intl, payload }) {
+  const [showMMP2P, setShowMMP2P] = useState(!!window.experiment__home_alert_bShowMMP2P);
+  if (window.experiment__home_alert_showMMP2P === undefined) {
+    window.experiment__home_alert_showMMP2P = (val) => {
+      window.experiment__home_alert_bShowMMP2P = !!val;
+      setShowMMP2P(!!val);
+    };
+  }
```
mikix, line 20: I'm not used to optimizely myself, so I assume this is normal boilerplate. But woof.

src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
```diff
@@ -53,35 +61,97 @@ function AccessExpirationAlert({ intl, payload }) {
   if (upgradeDeadline && upgradeUrl) {
     deadlineMessage = (
       <>
+        <br />
```
mikix: Why a newline at the top? That's not normally how we add spacing, plus it's a change to the existing experience.

src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
```diff
@@ -53,35 +61,97 @@ function AccessExpirationAlert({ intl, payload }) {
...
+        {showMMP2P
```
mikix: It might have been cleaner to fork the logic by making a replacement alert component as a sibling to this normal one, and in `hooks.js` decide which of the two components you are using.

Less churn for the normal code, in terms of PR conflicts and git blame, etc.

But just a nit / thought for next time.

src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
```diff
@@ -53,35 +61,97 @@ function AccessExpirationAlert({ intl, payload }) {
...
+          ? (
+            <FormattedMessage
+              id="learning.accessExpiration.deadline.mmp2p"
```
mikix: Do we want this translated? Your strings in DateSummary aren't, and it might annoy translators to add potentially throwaway strings that they don't know are throwaway. Same comment for other places in this file where experiment strings exist.

src/course-home/outline-tab/DateSummary.jsx
```diff
@@ -12,6 +12,15 @@ export default function DateSummary({
 }) {
   const linkedTitle = dateBlock.link && isLearnerAssignment(dateBlock);
   const timezoneFormatArgs = userTimezone ? { timeZone: userTimezone } : {};
+
+  /** [MM-P2P experiment] */
+  const [showMMP2P, setShowMMP2P] = useState(window.experiment__home_dates_bShowMMP2P);
+  window.experiment__home_dates_showMMP2P = (value) => {
```
mikix: You want a `undefined` check here, yah, like the other block above?

src/course-home/outline-tab/DateSummary.jsx
```diff
@@ -27,18 +36,34 @@ export default function DateSummary({
           />
         </div>
       </div>
-      <div className="row ml-4 pr-2">
-        <div className="date-summary-text">
-          {linkedTitle
-          && <div className="font-weight-bold mt-2"><a href={dateBlock.link}>{dateBlock.title}</a></div>}
-          {!linkedTitle
-          && <div className="font-weight-bold mt-2">{dateBlock.title}</div>}
+      {activateMMP2P ? (
```
mikix, line 41: Feels like this block could have been less code-disruptive by replacing some variables used like `title` and `description`, rather than copying the block wholesale. But maybe you prefer it this way, to avoid changing the experiment, even if the normal experience changes? Regardless, just a nit.

src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
muselesscreator: Totally fine with removing the translations!

src/course-home/outline-tab/DateSummary.jsx
muselesscreator: yeah, I will update.

src/course-home/outline-tab/DateSummary.jsx
muselesscreator: I tried, but dateBlock did not like being modified, and the alternative was to point the native code back to a new object, which felt more disruptive.

src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
```diff
@@ -49,6 +59,13 @@ function AccessExpirationAlert({ intl, payload }) {
     );
   }
 
+  /** [MM-P2P] Experiment */
+  if (showMMP2P) {
+    return (
+      <AccessExpirationAlertMMP2P intl={intl} payload={payload} />
```
mikix: I had been envisioning that you'd make this fork in `hooks.js` and just give the alerts system your alert component to run, instead of `AccessExpirationAlert`. This works too though. :shrug: 

Note that it's not typically done to pass `intl` in like that -- see how this file calls `injectIntl` in the export line at the bottom? Your component can/should do that too.

src/course-home/outline-tab/DateSummary.jsx
mikix: Or make variables for `title` and `description` that normally just get set to `dateBlock.title` at the top or whatever, and then overridden by your code. That doesn't seem so bad. But whatevs. This is fine too.

mikix (approved):

PR #369, opened 2021-02-22
draft: fix: handle problemsets and videosequences just like sequentials

kdmccormick:
problemset and videoset are both aliases to
sequential (e.g. Subsection). They are purely semantic;
they aren't supposed to affect functionality at all
currently.

https://openedx.atlassian.net/browse/TNL-7978


kdmccormick:
Closed; we'll be deprecating these block types instead (see https://openedx.atlassian.net/browse/TNL-7978)

PR #371, opened 2021-02-25
AA-684: allow sequence link to direct to legacy courseware

ciduarte:


Dillon-Dumesnil (approved):
Same thing if there should be a little test about this stuff, but otherwise, love it. Thanks for doing this Carla!

PR #372, opened 2021-02-26
AA-651: fix dates tab undefined property

ciduarte:


src/course-home/data/api.js
```diff
@@ -117,8 +117,7 @@ export async function getDatesTabData(courseId) {
       return {};
     }
     if (httpErrorStatus === 401) {
-      global.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/course`);
-      return {};
+      global.location.replace(`${getConfig().BASE_URL}/course/${courseId}/home`);
```
ciduarte, line 120: I changed this to point to the MFE because it feels right? :)

Let me know if I should leave it as LMS

src/course-home/data/api.js
```diff
@@ -117,8 +117,7 @@ export async function getDatesTabData(courseId) {
       return {};
     }
     if (httpErrorStatus === 401) {
-      global.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/course`);
-      return {};
```
ciduarte, line 121: The fix was removing this return statement

Dillon-Dumesnil (approved):
Works for me

PR #373, opened 2021-02-26
[MST-666] Add otherCourseApproved and expiringSoon statuses to proctoring panel

bseverino:
Related PR: https://github.com/edx/edx-proctoring/pull/794

* Added two new statuses to the proctoring info panel:
    - `otherCourseApproved`: The user has no onboarding attempts in the current course, but they have at least one verified attempt in a different course.
    - `expiringSoon`: Same as above, but the attempt is expiring within 28 days.
* Converted readable status strings in the proctoring info panel to constants.

---
Screen Shot 2021-02-26 at 2 01 52 PM
---
Screen Shot 2021-02-26 at 2 02 58 PM
---

### DIFF
--- a/src/course-home/outline-tab/OutlineTab.test.jsx
+++ b/src/course-home/outline-tab/OutlineTab.test.jsx
@@ -55,7 +55,11 @@ describe('Outline Tab', () => {
     axiosMock.onPost(enrollmentUrl).reply(200, {});
     axiosMock.onPost(goalUrl).reply(200, { header: 'Success' });
     axiosMock.onGet(outlineUrl).reply(200, defaultTabData);
-    axiosMock.onGet(proctoringInfoUrl).reply(200, { onboarding_status: 'created', onboarding_link: 'test' });
+    axiosMock.onGet(proctoringInfoUrl).reply(200, {
+      onboarding_status: 'created',
+      onboarding_link: 'test',
+      expiration_date: null,
+    });
 
     logUnhandledRequests(axiosMock);
   });
@@ -513,7 +517,11 @@ describe('Outline Tab', () => {
     });
 
     it('appears for verified', async () => {
-      axiosMock.onGet(proctoringInfoUrl).reply(200, { onboarding_status: 'verified', onboarding_link: 'test' });
+      axiosMock.onGet(proctoringInfoUrl).reply(200, {
+        onboarding_status: 'verified',
+        onboarding_link: 'test',
+        expiration_date: null,
+      });
       await fetchAndRender();
       await screen.findByText('This course contains proctored exams');
       expect(screen.queryByRole('link', { name: 'Complete Onboarding' })).not.toBeInTheDocument();
@@ -523,7 +531,11 @@ describe('Outline Tab', () => {
     });
 
     it('appears for rejected', async () => {
-      axiosMock.onGet(proctoringInfoUrl).reply(200, { onboarding_status: 'rejected', onboarding_link: 'test' });
+      axiosMock.onGet(proctoringInfoUrl).reply(200, {
+        onboarding_status: 'rejected',
+        onboarding_link: 'test',
+        expiration_date: null,
+      });
       await fetchAndRender();
       await screen.findByText('This course contains proctored exams');
       expect(screen.queryByRole('link', { name: 'Complete Onboarding' })).toBeInTheDocument();
@@ -533,7 +545,11 @@ describe('Outline Tab', () => {
     });
 
     it('appears for submitted', async () => {
-      axiosMock.onGet(proctoringInfoUrl).reply(200, { onboarding_status: 'submitted', onboarding_link: 'test' });
+      axiosMock.onGet(proctoringInfoUrl).reply(200, {
+        onboarding_status: 'submitted',
+        onboarding_link: 'test',
+        expiration_date: null,
+      });
       await fetchAndRender();
       await screen.findByText('This course contains proctored exams');
       expect(screen.queryByText('Your submitted profile is in review.')).toBeInTheDocument();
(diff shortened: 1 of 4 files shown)


src/course-home/outline-tab/widgets/ProctoringInfoPanel.jsx
```diff
@@ -32,24 +45,37 @@ function ProctoringInfoPanel({ courseId, intl }) {
     return !NO_SHOW_STATES.includes(examStatus);
   }
 
-  function getBorderClass(examStatus) {
+  function getBorderClass() {
     let borderClass = '';
-    if (['submitted', 'second_review_required'].includes(examStatus)) {
+    if (readableStatus === readableStatuses.submitted) {
```
alangsto, line 52: I think we still need to include `second_review_required`, or did we change that in the old experience as well?

src/course-home/outline-tab/widgets/ProctoringInfoPanel.scss
```diff
@@ -1,9 +1,9 @@
 .outline-sidebar-proctoring-panel {
   border: 1px solid $dark-500;
-  border-top: 5px solid $brand-600;
+  border-top: 5px solid $danger-600;
 }
 .proctoring-onboarding-success {
-  border-top: 5px solid $primary-500;
+  border-top: 5px solid $success-500;
```
alangsto: I think Carla had recommended using primary here, @ciduarte can you confirm?

src/course-home/outline-tab/widgets/ProctoringInfoPanel.jsx
bseverino: `second_review_required` also resolves to `readableStatuses.submitted`, so the functionality should be unchanged

src/course-home/outline-tab/widgets/ProctoringInfoPanel.scss
bseverino: Ah, looking at comments on your original PR for this component, you're right. I can change it back to primary.

src/course-home/outline-tab/widgets/ProctoringInfoPanel.jsx
alangsto: Got it, thank you!

alangsto: Got it, thank you!

schenedx (approved):
👍

src/course-home/outline-tab/messages.js
```diff
@@ -168,6 +176,14 @@ const messages = defineMessages({
     id: 'learning.proctoringPanel.message.error',
     defaultMessage: 'An error has occurred during your onboarding exam. Please retry onboarding.',
   },
+  otherCourseApprovedProctoringMessage: {
+    id: 'learning.proctoringPanel.message.otherCourseApproved',
+    defaultMessage: 'Your onboarding profile has been approved in another course, so you are eligible to take proctored exams in this course. However, it is highly recommended that you complete this course\'s onboarding exam in order to ensure that your device still meets the requirements for proctoring.',
```
Dillon-Dumesnil: Mike Terry would want me to point out using the Unicode apostrophe here as a preferable solution. Then it doesn't need to be escaped.

For reference, `'`

src/course-home/outline-tab/widgets/ProctoringInfoPanel.jsx
```diff
@@ -13,16 +13,29 @@ function ProctoringInfoPanel({ courseId, intl }) {
...
+    started: 'started',
+    submitted: 'submitted',
+    otherCourseApproved: 'otherCourseApproved',
+    verified: 'verified',
+    rejected: 'rejected',
+    error: 'error',
+    expiringSoon: 'expiringSoon',
+  };
+
   function getReadableStatusClass(examStatus) {
```
Dillon-Dumesnil, line 28: No need to refactor anything, but this function could also be a good use case for `switch` statements. Just in case you haven't heard of them before. But truly, I have no preference, just a possible teaching moment to let you know about another cool thing.

src/course-home/outline-tab/widgets/ProctoringInfoPanel.jsx
```diff
@@ -13,16 +13,29 @@ function ProctoringInfoPanel({ courseId, intl }) {
...
     } else if (['started', 'ready_to_submit'].includes(examStatus)) {
-      readableClass = 'started';
+      readableClass = readableStatuses.started;
     } else if (['second_review_required', 'submitted'].includes(examStatus)) {
-      readableClass = 'submitted';
-    } else if (['verified', 'rejected', 'error'].includes(examStatus)) {
-      readableClass = examStatus;
+      readableClass = readableStatuses.submitted;
+    } else if (examStatus === 'other_course_approved') {
+      readableClass = readableStatuses.otherCourseApproved;
```
Dillon-Dumesnil: Nit: This could technically already be covered in the catch remaining case below. I understand if y'all want to be explicit about it though.

src/course-home/outline-tab/widgets/ProctoringInfoPanel.jsx
```diff
@@ -13,16 +13,29 @@ function ProctoringInfoPanel({ courseId, intl }) {
...
+    } else if (examStatus in readableStatuses) {
```
Dillon-Dumesnil: Is this line of code going to work? It looks like `examStatus` is returning snake_case, but the keys in `readableStatuses` are camelCase. Might be worth casting `examStatus` to camelCase for this line. This could also allow you to cover the other_course_approved case (see above comment)

I believe it _will_ work, but only for single word statuses, but by adding in a cast to camelCase, it will catch more cases.

src/course-home/outline-tab/widgets/ProctoringInfoPanel.jsx
```diff
@@ -13,16 +13,29 @@ function ProctoringInfoPanel({ courseId, intl }) {
   const [link, setLink] = useState('');
   const [readableStatus, setReadableStatus] = useState('');
 
+  const readableStatuses = {
```
Dillon-Dumesnil, line 17: 👏 Love seeing defining strings in one spot

src/course-home/outline-tab/widgets/ProctoringInfoPanel.scss
```diff
@@ -1,6 +1,6 @@
 .outline-sidebar-proctoring-panel {
   border: 1px solid $dark-500;
-  border-top: 5px solid $brand-600;
+  border-top: 5px solid $danger-600;
```
Dillon-Dumesnil: _danger_ 😬

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -533,23 +545,65 @@ describe('Outline Tab', () => {
...
+      });
       await fetchAndRender();
       await screen.findByText('This course contains proctored exams');
       expect(screen.queryByText('Your submitted profile is in review.')).toBeInTheDocument();
       expect(screen.queryByText('Onboarding profile review, including identity verification, can take 2+ business days.')).toBeInTheDocument();
     });
 
+    it('appears for other_course_approved', async () => {
+      const expirationDate = new Date();
+      // Set expiration date at least one month in the future
```
Dillon-Dumesnil: Small nit on these two tests to more explicitly say in this comment that the month difference is what triggers the expiration text or not.

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -533,23 +545,65 @@ describe('Outline Tab', () => {
...
+      expirationDate.setDate(expirationDate.getDate() + 31);
```
Dillon-Dumesnil: Also a very minor nit: why play with 31? I know it should be > 1 month, but dates sometimes do weird things. Why not just make it like 50 days and be done with it?

Similarly below. Rather than 27 days (to try and account for February), just make it like 10 and then it's never a worry.

src/course-home/outline-tab/widgets/ProctoringInfoPanel.scss
```diff
@@ -1,6 +1,6 @@
 .outline-sidebar-proctoring-panel {
```
Dillon-Dumesnil: So I don't claim to be a CSS expert and am personally cool with this scss file, but I believe we do try to use bootstrap classes inline as much as possible via  `className` in our components. Just wanted to mention in case you want to try switching, but I'm not gonna reject this PR over that so no worries.

Dillon-Dumesnil (approved):
No major changes that I can see need to happen, mostly some nits and small requests, but otherwise fantastic work!

src/course-home/outline-tab/messages.js
bseverino: Unless I'm misunderstanding, wouldn't the Unicode apostrophe need to be escaped anyway?

src/course-home/outline-tab/widgets/ProctoringInfoPanel.scss
bseverino: ~~I was going to leave this as is, but I think it makes sense to switch to inline the more I think about it. I'll refactor this.~~ Changed my mind again as I switched it back to using the `brand` variable, which I don't believe I can use inline. I'll leave this as a consideration for another day.

src/course-home/outline-tab/messages.js
Dillon-Dumesnil: No, it does not need to be escaped because it is just another unicode point. Where the standard apostrophe does need to be because it things you're trying to terminate your string.

bseverino: Oh, I see what you mean now. Thank you!

PR #376, opened 2021-03-04
WS-1740 add course recommendations to celebration page (experiment)

rgraber:
Screen Shot 2021-03-04 at 3 45 44 PM


package.json
```diff
@@ -56,6 +56,7 @@
     "react-router": "5.2.0",
     "react-router-dom": "5.2.0",
     "react-share": "4.2.1",
+    "react-truncate": "2.4.0",
```
mikix: We already use `truncate-html` (see WelcomeMessage.jsx) -- can that just be used instead?  (Or if react-truncate is amazing, swap out the other use of truncate-html?)

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -59,6 +60,10 @@ function CourseCelebration({ intl }) {
     downloadUrl,
   } = certificateData || {};
 
+  /** [WS-1681 experiment] */
+  const [showWS1681, setShowWS1681] = useState(window.experiment__courseware_celebration_bShowWS1681);
+  useEffect(() => { setShowWS1681(window.experiment__courseware_celebration_bShowWS1681); });
```
mikix, line 65: So every render of this component will sync `showWS1681` from the window value? Seems right, just trying to grok.

src/courseware/course/course-exit/CourseRecommendationsExp/data/thunks.exp.js
```diff
@@ -0,0 +1,29 @@
...
+import getCourseRecommendations from './api.exp';
+import { updateModel } from '../../../../../generic/model-store';
+
+export default function fetchCourseRecommendations(courseKey, courseId) {
+  return async (dispatch) => {
+    dispatch(fetchCourseRecommendationsRequest({ courseId }));
+    try {
+      const recommendations = await getCourseRecommendations(courseKey);
+      dispatch(updateModel({
+        modelType: 'coursewareMeta',
```
mikix, line 17: I guess it doesn't really matter, especially for an experiment, but I would have naively expected this stuff to go into its own model, not the coursewareMeta one.

src/courseware/course/course-exit/CourseRecommendationsExp/data/api.exp.js
```diff
@@ -0,0 +1,32 @@
...
+      courseId,
+    },
+  }) => courseId);
+
+  return recommendations.filter(({ uuid: recUuid, courseRunKeys }) => (
+    recUuid !== uuid && courseRunKeys.every((key) => !enrollmentRunIds.includes(key))
+  ));
+}
+
+export default async function getCourseRecommendations(courseKey) {
```
mikix, line 26: Probably deserves an early exit if the disco url isn't set.

src/courseware/course/course-exit/CourseRecommendationsExp/data/api.exp.js
```diff
@@ -0,0 +1,32 @@
...
+  const recommendationsUrl = new URL(`${getConfig().DISCOVERY_API_BASE_URL}/api/v1/course_recommendations/${courseKey}?exclude_utm=true`);
+  const enrollmentsUrl = new URL(`${getConfig().LMS_BASE_URL}/api/enrollment/v1/enrollment`);
+  const recommendationsResponse = await getAuthenticatedHttpClient().get(recommendationsUrl);
+  const enrollmentsResponse = await getAuthenticatedHttpClient().get(enrollmentsUrl);
```
mikix: There is a way to joint wait on two promises, rather than serialize them like this. I don't remember syntax, but searching online should get it.

mikix (commented):
I mostly glossed over the actual new recommendation code, I assume it's lovely. Just a couple of nits, but one real question about the truncate dep.

src/courseware/course/course-exit/CourseRecommendationsExp/course_recommendations.exp.scss
```diff
@@ -0,0 +1,111 @@
+$default-border-color: $info-300;
+.course-recommendations {
```
mikix, line 2: This file is a lot of custom css which we normally try hard to avoid by using bootstrap utility classes and such. And a lot of hardcoded pixels. Since this is just an experiment, fine. But just heads up if this becomes a real person.

package.json
rgraber: do you know if there's a way to make truncate-html truncate by line instead of by word? if not by word is probably fine for the experiment

mikix: Naw, looking at its docs, seems like just words or letters. If react-truncate is fancier, I'm fine with dropping truncate-html for it. Just seemed like a waste to have both.

src/courseware/course/course-exit/CourseCelebration.jsx
rgraber: yes

src/courseware/course/course-exit/CourseRecommendationsExp/course_recommendations.exp.scss
rgraber: With any luck by the time this gets productized we'll have a Card component in paragon and won't need this nonsense.

mikix (approved):

PR #377, opened 2021-03-05
static expiration box

emma-green:
This should look like this: https://openedx.atlassian.net/wiki/spaces/RS/pages/2527068482/Expiration+Box

### DIFF
--- a/src/course-home/data/__snapshots__/redux.test.js.snap
+++ b/src/course-home/data/__snapshots__/redux.test.js.snap
@@ -432,6 +432,7 @@ Object {
           "hasVisitedCourse": false,
           "url": "http://localhost:18000/courses/course-v1:edX+DemoX+Demo_Course/jump_to/block-v1:edX+Test+Block@12345abcde",
         },
+        "timeOffsetMillis": 0,
         "verifiedMode": Object {
           "accessExpirationDate": "2050-01-01T12:00:00",
           "currency": "USD",
--- a/src/course-home/data/api.js
+++ b/src/course-home/data/api.js
@@ -198,6 +198,7 @@ export async function getOutlineTabData(courseId) {
     data,
     headers,
   } = tabData;
+
   const accessExpiration = camelCaseObject(data.access_expiration);
   const canShowUpgradeSock = data.can_show_upgrade_sock;
   const courseBlocks = data.course_blocks ? normalizeOutlineBlocks(courseId, data.course_blocks.blocks) : {};
@@ -210,7 +211,7 @@ export async function getOutlineTabData(courseId) {
   const hasEnded = data.has_ended;
   const offer = camelCaseObject(data.offer);
   const resumeCourse = camelCaseObject(data.resume_course);
-  const timeOffsetMillis = getTimeOffsetMillis(headers.date, requestTime, responseTime);
+  const timeOffsetMillis = getTimeOffsetMillis(headers && headers.date, requestTime, responseTime);
   const verifiedMode = camelCaseObject(data.verified_mode);
   const welcomeMessageHtml = data.welcome_message_html;
 
--- a/src/course-home/data/api.js
+++ b/src/course-home/data/api.js
@@ -208,6 +208,7 @@ export async function getOutlineTabData(courseId) {
     resumeCourse,
     verifiedMode,
     welcomeMessageHtml,
+    timeOffsetMillis: 0,
   };
 }
 
--- a/src/course-home/data/api.js
+++ b/src/course-home/data/api.js
@@ -196,6 +196,7 @@ export async function getOutlineTabData(courseId) {
 
   const {
     data,
+    headers,
   } = tabData;
   const accessExpiration = camelCaseObject(data.access_expiration);
   const canShowUpgradeSock = data.can_show_upgrade_sock;
@@ -209,7 +210,7 @@ export async function getOutlineTabData(courseId) {
   const hasEnded = data.has_ended;
   const offer = camelCaseObject(data.offer);
   const resumeCourse = camelCaseObject(data.resume_course);
-  const timeOffsetMillis = getTimeOffsetMillis(tabData.headers.date, requestTime, responseTime);
+  const timeOffsetMillis = getTimeOffsetMillis(headers.date, requestTime, responseTime);
   const verifiedMode = camelCaseObject(data.verified_mode);
   const welcomeMessageHtml = data.welcome_message_html;
 
--- a/src/course-home/data/api.js
+++ b/src/course-home/data/api.js
@@ -161,11 +161,30 @@ export async function getProctoringInfoData(courseId, username) {
   }
 }
 
+export function getTimeOffsetMillis(headerDate, requestTime, responseTime) {
+  // Time offset computation should move down into the HttpClient wrapper to maintain a global time correction reference
+  // Requires 'Access-Control-Expose-Headers: Date' on the server response per https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#access-control-expose-headers
+
+  let timeOffsetMillis = 0;
+  if (headerDate !== undefined) {
+    const headerTime = Date.parse(headerDate);
+    const roundTripMillis = requestTime - responseTime;
+    const localTime = responseTime - (roundTripMillis / 2); // Roughly compensate for transit time
+    timeOffsetMillis = headerTime - localTime;
+  }
+
+  return timeOffsetMillis;
+}
+
 export async function getOutlineTabData(courseId) {
   const url = `${getConfig().LMS_BASE_URL}/api/course_home/v1/outline/${courseId}`;
   let { tabData } = {};
+  let requestTime = Date.now();
+  let responseTime = requestTime;
   try {
+    requestTime = Date.now();
     tabData = await getAuthenticatedHttpClient().get(url);
+    responseTime = Date.now();
   } catch (error) {
     const { httpErrorStatus } = error && error.customAttributes;
     if (httpErrorStatus === 404) {
@@ -190,6 +209,7 @@ export async function getOutlineTabData(courseId) {
   const hasEnded = data.has_ended;
   const offer = camelCaseObject(data.offer);
   const resumeCourse = camelCaseObject(data.resume_course);
+  const timeOffsetMillis = getTimeOffsetMillis(tabData.headers.date, requestTime, responseTime);
   const verifiedMode = camelCaseObject(data.verified_mode);
   const welcomeMessageHtml = data.welcome_message_html;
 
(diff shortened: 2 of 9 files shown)


src/course-home/outline-tab/widgets/UpgradeCard.jsx
```diff
@@ -1,21 +1,232 @@
...
 
-function UpgradeCard({ courseId, intl, onLearnMore }) {
-  const { org } = useModel('courseHomeMeta', courseId);
-  const {
-    offer,
-    verifiedMode,
-  } = useModel('outline', courseId);
+function UpsellNoFBECardContent() {
+  return (
+    <ul className="fa-ul upgradeCard-ul pt-0">
```
julianajlk: I would name my class as `upgrade-card-ul`, React doesn't have a restriction that I know of (dashed, camel-case, etc) but would be good for consistency with the other classNames already in this file

src/course-home/outline-tab/widgets/UpgradeCard.jsx
```diff
@@ -1,21 +1,232 @@
...
+    );
+  }
+  return (<div className="upsell-warning">{expirationText}</div>);
+}
+
+ExpirationCountdown.propTypes = {
+  hoursToExpiration: PropTypes.number.isRequired,
+};
+
+function AccessExpirationDateBanner({ accessExpirationDate, timezoneFormatArgs }) {
```
julianajlk, line 185: What's the difference between a "card" and a "banner"? I think of them as the same, so to me when I read AccessExpirationDateBanner I think it's a separate banner/card than the UpgradeCard, but in this case it's a piece of it?

julianajlk: Also wondering if it would make sense to pull out some of this stuff to a separate file, and/or a messages.js file. It's hard to follow through all of the different functions, but if that would make it complicated to access the data we need here, nevermind

src/course-home/outline-tab/widgets/UpgradeCard.jsx
```diff
@@ -1,21 +1,232 @@
 import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
+import { faCheck } from '@fortawesome/free-solid-svg-icons';
```
jhan217, line 3: Could you use the `check` icon from https://paragon-edx.netlify.app/foundations/icons?

src/course-home/outline-tab/widgets/UpgradeCard.jsx
```diff
@@ -1,21 +1,232 @@
...
+  return (
+    <ul className="fa-ul upgradeCard-ul pt-0">
+      <li>
+        <span className="fa-li upgradeCard-li"><FontAwesomeIcon icon={faCheck} /></span>
+        <FormattedMessage
+          id="learning.outline.alert.upgradecard.verifiedCertLink"
+          defaultMessage="Earn a {verifiedCertLink} of completion to showcase on your resume"
+          values={{
+            verifiedCertLink: (
+              <a className="inline-link-underline font-weight-bold" rel="noopener noreferrer" target="_blank" href={`${getConfig().MARKETING_SITE_BASE_URL}/verified-certificate`}>verified certificate</a>
```
jhan217, line 22: Could you use a [Hyperlink](https://paragon-edx.netlify.app/components/hyperlink/) component here?

src/course-home/outline-tab/widgets/UpgradeCard.scss
```diff
@@ -1,4 +1,57 @@
...
+  border-radius: 0 !important;
 }
+
+.upgrade-card-header{
+  margin: 1.25rem;
+
+}
+
+.upsell-warning{
+  background-color: #FCF1F4;
```
jhan217: `$danger-100` from https://paragon-edx.netlify.app/foundations/colors?

src/course-home/outline-tab/widgets/UpgradeCard.scss
```diff
@@ -1,4 +1,57 @@
...
+}
+
+.upsell-warning-light{
+  background-color: #FFFAED;
```
jhan217: Should this be `$warning-100`?

src/course-home/outline-tab/widgets/UpgradeCard.scss
```diff
@@ -1,4 +1,57 @@
...
+}
+
+.upsell-warning, .upsell-warning-light{
+  padding-left: 1.25rem;
```
jhan217, line 19: Would it be possible to use any of the [spacing utility classes](https://paragon-edx.netlify.app/foundations/spacing) instead of hardcoding all the values?

src/course-home/outline-tab/widgets/UpgradeCard.scss
```diff
@@ -1,4 +1,57 @@
...
+.discount-info {
+  border-top: 1px solid rgba(0, 0, 0, 0.125);
+}
+
+.inline-link-underline {
+  text-decoration: underline;
+}
+
+.upgrade-card .upgrade-card-message a{
+  color: #00262B;
```
jhan217: `$primary-500`?

jhan217 (commented):
Couple of general comments about potentially using Paragon variables/components

src/course-home/outline-tab/widgets/UpgradeCard.scss
emma-green: Good call out, but I don't think so, because I am not using full rems.

src/course-home/data/api.js
```diff
@@ -206,6 +206,7 @@ export async function getOutlineTabData(courseId) {
     resumeCourse,
     verifiedMode,
     welcomeMessageHtml,
+    timeOffsetMillis: 0,
```
inventhouse: I don't see the code where this is actually computed from the `Date` headers (see these two diffs: [src/courseware/data/api.js](https://github.com/edx/frontend-app-learning/pull/385/files#diff-9a08a1a98535680aabde6f7b44130d363b29ea26469f026235f1f0a19e31a02f), [src/course-home/data/api.js](https://github.com/edx/frontend-app-learning/pull/385/files#diff-4d3a6d853ff9fd1354a56024f06c6718c856c281e295de92c5b09ac1e86774bf))

inventhouse (changes requested):

src/course-home/data/api.js
inventhouse: It would be good to include in the QA as well, that the countdown is correct no matter what your local computer time is set to.

emma-green: how about I write a test for it?

src/course-home/outline-tab/widgets/UpgradeCard.jsx
emma-green: nope. It needs to be fontawesome in order to use it in a list

src/course-home/outline-tab/widgets/UpgradeCard.jsx
emma-green: What would be the advantage of that?

jhan217: If the behavior/attributes of Hyperlinks changed, those changes would automatically apply to this link (if it were a Hyperlink component) as well. Not a big deal, as it looks like both are used in this repo.

src/course-home/outline-tab/widgets/UpgradeCard.jsx
emma-green: There's a banner inside the card that counts down to expiration.

src/course-home/outline-tab/widgets/UpgradeCard.scss
emma-green: yes! Good catch! Thank you!!!

src/course-home/outline-tab/widgets/UpgradeCard.scss
emma-green: Hmmm....That's not the same color as requested. I'll message Eugene -- maybe FFFAED was an error

src/course-home/outline-tab/widgets/UpgradeCard.scss
emma-green: yep! Thanks!

src/course-home/outline-tab/widgets/UpgradeCard.scss
emma-green: She said to do $warning-100. Thanks for the callout!

src/course-home/data/api.js
emma-green: I did a first pass. Please let me know your thoughts :)

inventhouse (approved):

src/course-home/outline-tab/widgets/UpgradeCard.scss
```diff
@@ -1,4 +1,59 @@
-.outline-sidebar-upgrade-card {
-  border: 1px solid $dark-500;
-  border-top: 5px solid $dark-500;
+.upgrade-card {
```
mikix, line 1: This file has a lot of custom css (which also means using a lot of `rem` instead of the bootstrap spacing system).

We prefer to do this sort of thing as bootstrap utility classnames like `m-2` or similar. Keeps sizes the same and reduces all the little bits of css lying around.

Even the stuff you can't do as bootstrap class (like `text-decoration`) we will often do inline to the jsx with a `style={{ textDecoration: 'underline' }}` bit, to keep the styling close to the component.

src/course-home/outline-tab/widgets/UpgradeCard.jsx
mikix: Is that a bug with our paragon implementation? Seems like something we should put on a backlog or try to solve. We are trying to get away from direct uses of FA.

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -224,10 +225,14 @@ function OutlineTab({ intl }) {
...
+                  offer={offer}
+                  verifiedMode={verifiedMode}
+                  accessExpiration={accessExpiration}
+                  contentTypeGatingEnabled={datesBannerInfo.contentTypeGatingEnabled}
+                  userTimezone={userTimezone}
+                  timeOffsetMillis={timeOffsetMillis}
                   courseId={courseId}
-                  onLearnMore={
-                    canShowUpgradeSock ? () => { courseSock.current.showToUser(); } : null
-                  }
```
mikix, line 230: Dropping this removes the need for the `ref` on `<CourseSock>`, so you can drop that bit and variable it saved the ref to.

src/course-home/data/__snapshots__/redux.test.js.snap
```diff
@@ -432,6 +432,7 @@ Object {
           "hasVisitedCourse": false,
           "url": "http://localhost:18000/courses/course-v1:edX+DemoX+Demo_Course/jump_to/block-v1:edX+Test+Block@12345abcde",
         },
+        "timeOffsetMillis": 0,
```
mikix, line 435: I don't see this being set in the upgrade card tests, and it's zero here. Is it worth a test where this is set and we see the change in the UI? (Or maybe there is one and I missed it)

mikix (commented):
Some minor comments, nothing crazy / blocking

PR #379, opened 2021-03-08
Cleaned up all references to REV-1512's value prop experiment

inventhouse:


jhan217 (approved):
Looks fine to me, but I'd defer to @dianekaplan and @muselesscreator in case either of them still need the code.

src/courseware/course/sequence/sequence-navigation/SequenceNavigation.jsx
```diff
@@ -91,22 +81,7 @@ function SequenceNavigation({
       </Button>
       {renderUnitButtons()}
       {renderNextButton()}
-      {/** [MM-P2P] Experiment (additional conditional) */}
-      {!mmp2p.state.isEnabled && REV1512FlyoverEnabled
```
dianekaplan, line 95: I believe the new MM-P2P bit is from @muselesscreator, so let's check with him what he needs in this block. (Everything else is fine from my perspective)

inventhouse: I believe the "additional conditional" changes were meant to have the mmp2p flyover supersede the REV1512 one, and can be safely removed with REV experiment code.

muselesscreator: ++

src/courseware/course/sequence/Sequence.jsx
```diff
@@ -367,16 +195,10 @@ function Sequence({
...
-          {!mmp2p.state.isEnabled && REV1512FlyoverEnabled && isREV1512FlyoverVisible() && (
-            isMobile
-              ? <REV1512FlyoverMobile toggleREV1512Flyover={toggleREV1512Flyover} />
-              : <REV1512Flyover toggleREV1512Flyover={toggleREV1512Flyover} />
-          )}
+
           {/** [MM-P2P] Experiment */}
           {(mmp2p.state.isEnabled && mmp2p.flyover.isVisible) && (
-            isMobile
+            isMobile()
```
muselesscreator, line 201: Nice, this is one of the places I was worried about.

muselesscreator (approved):

PR #380, opened 2021-03-09
MM-P2P G2 Activation

muselesscreator:
# What is this?
This PR is focused on enabling the activation of view modifications for the MM-P2P experiment.
This includes:
* removing upsell banners from the outline tab, dates tab, and courseware
* Adding a upsell sidecard for audit users to outline tab and courseware (that will pre-empt the upcoming Value Prop one for our pages)
* Update language on verified lock paywall based on whether the segment is blocked because it is graded
  * Move lock paywall into its own component to support upcoming rewrite of that component
* Block content with overlay modal for audit users after hard-coded upgrade deadline.

# Resources
Examples doc (screenshots): https://docs.google.com/document/d/1_WoBvvZkm5tQDXXLn4eatHPhuC40SttrWOCM3TH6VZ0/edit#heading=h.8ppru1okalyc

Figma Design: https://www.figma.com/file/cTjgRZdeQlJT2A0aEV9Rln/MM-Rev-Shift?node-id=15%3A139

JIRA link: https://openedx.atlassian.net/browse/EDUCATOR-5598

## Note:
This contains a test object (externalConfig) that is expressly for testing of changes pre-PR in devstack against demo course, and will be deactivated before merge.


MatthewPiatetsky (commented):
If this is an experiment I imagine the main thing we are concerned about is whether it works rather than code quality, hard to review that just based on the code
maybe you could add screenshots or a gif of yourself showing the experiences the experiment is supposed to create?

muselesscreator:
The example doc is full of screenshots

kdmccormick (commented):
@muselesscreator 
* I skimmed through the changes under src/courseware and certain parts of src/experiements. Nothing jumps out at me as needing to change :+1: I figure that someone on your team will give you a full code review.  
* Thanks for clearly annotating all your experimental code!
* The one thing this is missing for me is some sort of brief README or wiki page link where I can find info about the experiment. As someone on-call for this code but generally ignorant of the experiment, I just need a few key pieces of information -- where can I find the timeline of the experiment, how do I know who it's enabled for, who is my MMP2P PoC if I am paged at 2am and think I need to turn something off, etc... you get the point.

If you add that info, then this is good from my perspective. Best of luck with the experiment!

MatthewPiatetsky:
ah wasn't sure if those were from this code
well if you know this works I'm not sure if a review from our team is necessary if this code will be deleted anyways?  
assuming if something breaks the experiment can be turned off in optimizely? (and that you've verified that learning mfe still works with the experiment disabled)
in that case agree with kyle that your team can probably review the specifics of the experiment

src/experiments/mm-p2p/FlyoverTriggerIcon.jsx
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable no-use-before-define */
+import React from 'react';
+
+const FlyoverTriggerIcon = () => (
```
nsprenkle, line 4: Ew. Is this really the way we have to do SVGs? We don't import from an asset file?

muselesscreator: We could, if we were optimizing, or if we had a bunch of them.  but this is generally cleaner for react usage, as they can be modified/treated like other react/jsx objects

nsprenkle: TIL 🤔

jnlapierre (approved):
LGTM, I appreciate the consistent application of React best practices, in particular, the use of triggers and lazy loading

nsprenkle (approved):
@muselesscreator gave us a sync walkthrough.
Sanity-check looks good 👍

PR #385, opened 2021-03-11
Compute offset to correct browser time for accurate expiration countdown

inventhouse:
Uses the `Date` header to compute an offset to correct browser time to offer an accurate countdown based on the server time.  Requires [this edx-platform PR](https://github.com/edx/edx-platform/pull/26979), otherwise falling back to uncorrected browser time.


inventhouse:
Closing old PR

PR #388, opened 2021-03-12
Set flyover visibility default to true

mattcarter:
**What is this?**
This PR is to change the default visibility for the sidecard in the MM-P2P experiment

JIRA link: https://openedx.atlassian.net/browse/EDUCATOR-5630


jnlapierre (approved):
LGTM

PR #390, opened 2021-03-15
Add Sidebar and SidebarNotificationButton components

julianajlk:
**_EDIT: build is failing for unknown reasons, I made a duplicate of this PR and divided the commit that was failing at unrelated tests, and all of them passed. Duplicate PR here https://github.com/edx/frontend-app-learning/pull/414_**

[REV-2125](https://openedx.atlassian.net/browse/REV-2125).

For Value Prop implementation, we will implement a sidebar that is triggered on click of a button with upgrade messaging and CTA. Pls note this ticket does not cover adding the upgrade messaging and CTA, it only covers adding the button trigger and sidebar component. 

Large screen size (> 992px):
Screen Shot 2021-04-08 at 12 49 43 PM
Screen Shot 2021-04-08 at 12 49 51 PM

Responsive at medium screen sizes (> 576px and < 992px): it has the button like large and above screen size, but the sidebar behaves in full width mode like for small screens. 
Screen Shot 2021-04-08 at 12 57 28 PM
Screen Shot 2021-04-08 at 12 57 37 PM

Responsive at small screen sizes (< 576px):
Screen Shot 2021-04-08 at 1 03 12 PM
Screen Shot 2021-04-08 at 1 02 26 PM

### DIFF
--- a/src/courseware/course/Sidebar.jsx
+++ b/src/courseware/course/Sidebar.jsx
@@ -12,7 +12,7 @@ function Sidebar({
 }) {
   const shouldDisplayFullScreen = useWindowSize().width < 992;
   return (
-    <div className="sidebar-container ml-0 ml-lg-4">
+    <section className="sidebar-container ml-0 ml-lg-4" aria-label={intl.formatMessage(messages.sidebarNotification)}>
       {shouldDisplayFullScreen ? (
         <div className="mobile-close-container" onClick={() => { toggleSidebar(); }} onKeyDown={() => { toggleSidebar(); }} role="button" tabIndex="0" alt={intl.formatMessage(messages.responsiveCloseSidebar)}>
           <Icon src={ArrowBackIos} />
@@ -20,17 +20,17 @@ function Sidebar({
         </div>
       ) : null}
       <div className="sidebar-header px-3">
-        <span>{intl.formatMessage(messages.notification)}</span>
+        <span>{intl.formatMessage(messages.notificationTitle)}</span>
         {shouldDisplayFullScreen
           ? null
-          : <Icon src={Close} className="close-btn" onClick={() => { toggleSidebar(); }} onKeyDown={() => { toggleSidebar(); }} role="button" tabIndex="0" alt={intl.formatMessage(messages.closeButton)} />}
+          : <Icon src={Close} className="close-btn" onClick={() => { toggleSidebar(); }} onKeyDown={() => { toggleSidebar(); }} role="button" tabIndex="0" alt={intl.formatMessage(messages.closeSidebarButton)} />}
       </div>
       <div className="sidebar-divider" />
       <div className="sidebar-content">
         {/* REV-2130 TODO: add conditional here to display expiration box or display below message */}
-        <p>You have no new notifications at this time.</p>
+        <p>{intl.formatMessage(messages.noNotificationsMessage)}</p>
       </div>
-    </div>
+    </section>
   );
 }
 
--- a/src/courseware/course/Sidebar.jsx
+++ b/src/courseware/course/Sidebar.jsx
@@ -5,49 +5,43 @@ import { Icon } from '@edx/paragon';
 import { ArrowBackIos, Close } from '@edx/paragon/icons';
 import './Sidebar.scss';
 import messages from './messages';
+import useWindowSize from '../../generic/tabs/useWindowSize';
 
 function Sidebar({
-  intl, sidebarVisible, toggleSidebar, isTabletWidth,
+  intl, toggleSidebar,
 }) {
+  const shouldDisplayFullScreen = useWindowSize().width < 992;
   return (
-    sidebarVisible ? (
-      <div className="sidebar-container ml-0 ml-lg-4">
-        {isTabletWidth
-          ? (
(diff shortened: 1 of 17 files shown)


src/courseware/course/Sidebar.jsx
```diff
@@ -0,0 +1,46 @@
...
+import './Sidebar.scss';
+import messages from './messages';
+import useWindowSize from '../../generic/tabs/useWindowSize';
+
+function Sidebar({
+  intl, toggleSidebar,
+}) {
+  const shouldDisplayFullScreen = useWindowSize().width < 992;
+  return (
+    <section className="sidebar-container ml-0 ml-lg-4">
```
wittjeff: Could you please add **aria-label="sidebar notifications"** here?

src/courseware/course/messages.js
```diff
@@ -0,0 +1,26 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  notificationButton: {
+    id: 'sidebar.notificationButton',
+    defaultMessage: 'Sidebar notification button',
```
wittjeff: Could you please make this a verb or verb phrase like "Show sidebar notifications"?

src/courseware/course/messages.js
```diff
@@ -0,0 +1,26 @@
...
+    description: 'Button to open the sidebar',
+  },
+  closeButton: {
+    id: 'sidebar.closeButton',
+    defaultMessage: 'Close',
```
wittjeff: "Close sidebar notifications"?

src/courseware/course/Sidebar.scss
```diff
@@ -0,0 +1,69 @@
...
+        position: fixed; 
+        top: 0; 
+        bottom: 0; 
+        left: 0; 
+        right: 0;
+        width: 100%;
+        height: 100%;
+        background-color: white;
+        margin: 0;
+        border: none;
```
wittjeff, line 22: If you use **border: transparent** instead of **border: none**, the border will show up as a white box on Windows High Contrast Mode. Sometimes this is useful. I'm not sure if it would help here.  It would also change your box size calculation for the width of the borders.

wittjeff (changes requested):

src/courseware/course/Sidebar.scss
julianajlk: @wittjeff Would it be useful to have the "transparent" border if it's on a full width screen (the sidebar takes up the entire screen)?

src/courseware/course/Sidebar.jsx
```diff
@@ -0,0 +1,46 @@
...
+      <div className="sidebar-header px-3">
+        <span>{intl.formatMessage(messages.notificationTitle)}</span>
+        {shouldDisplayFullScreen
+          ? null
+          : <Icon src={Close} className="close-btn" onClick={() => { toggleSidebar(); }} onKeyDown={() => { toggleSidebar(); }} role="button" tabIndex="0" alt={intl.formatMessage(messages.closeSidebarButton)} />}
+      </div>
+      <div className="sidebar-divider" />
+      <div className="sidebar-content">
+        {/* REV-2130 TODO: add conditional here to display expiration box or display below message */}
+        <p>You have no new notifications at this time.</p>
```
dianekaplan: Will this 'no new notifications' line be translatable (via transifex) this way? Or should it have an entry in messages.js or use FormatMessage?

julianajlk: Good catch! Formatted message will be added in the next commit

src/courseware/course/Sidebar.jsx
```diff
@@ -0,0 +1,46 @@
...
+import { Icon } from '@edx/paragon';
+import { ArrowBackIos, Close } from '@edx/paragon/icons';
+import './Sidebar.scss';
+import messages from './messages';
+import useWindowSize from '../../generic/tabs/useWindowSize';
+
+function Sidebar({
+  intl, toggleSidebar,
+}) {
+  const shouldDisplayFullScreen = useWindowSize().width < 992;
```
jhan217, line 13: Is there a reusable variable that can be used here (and other places where the sizes are hardcoded) so you don't have to hardcode the pixel size? Something like `media-breakpoint-down(xs)` in edx-platform?

src/courseware/course/Sidebar.scss
```diff
@@ -0,0 +1,69 @@
...
+        display: inline-block;
+    }
+    svg {
+        top: 0.4rem;
+        left: 0.8rem;
+    }
+}
+
+.mobile-close {
+    font-weight: 500;
```
jhan217, line 67: Could you use the `font-weight-bold` class from https://paragon-edx.netlify.app/foundations/typography/?

julianajlk: Because I have a separate SCSS file, I'm generally choosing not to add utility classes inline but add all CSS to the file to keep it contained. Except for responsive padding/margin, which are simpler to do than adding media queries.

src/courseware/course/Sidebar.jsx
julianajlk: Good question, there is a way to add to a SCSS file (ie. `$grid-breakpoints`), but not sure about here, will take a look how to do it in JSX.

PR #392, opened 2021-03-15
AA-711: implement upsell click events

ciduarte:
Adding `course_home_green` and `in_course_upgrade` upsell events to course home and courseware.

Descriptions of these events:
https://openedx.atlassian.net/wiki/spaces/RS/pages/1675100377/How+to+find+upsell+link+click+events

### DIFF
--- a/src/course-home/outline-tab/widgets/UpgradeCard.jsx
+++ b/src/course-home/outline-tab/widgets/UpgradeCard.jsx
@@ -37,7 +37,7 @@ function UpgradeCard({ courseId, intl, onLearnMore }) {
   useEffect(() => {
     sendTrackingLogEvent('edx.bi.course.upgrade.sidebarupsell.displayed', eventProperties);
     sendTrackEvent('Promotion Viewed', promotionEventProperties);
-  });
+  }, []);
 
   const logClick = () => {
     sendTrackingLogEvent('edx.bi.course.upgrade.sidebarupsell.clicked', eventProperties);
@@ -46,6 +46,13 @@ function UpgradeCard({ courseId, intl, onLearnMore }) {
       location: 'sidebar-message',
     });
     sendTrackEvent('Promotion Clicked', promotionEventProperties);
+    sendTrackEvent('edx.bi.ecommerce.upsell_links_clicked', {
+      ...eventProperties,
+      linkCategory: 'green_upgrade',
+      linkName: 'course_home_green',
+      linkType: 'button',
+      pageName: 'course_home',
+    });
   };
 
   return (
--- a/src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
+++ b/src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faLock } from '@fortawesome/free-solid-svg-icons';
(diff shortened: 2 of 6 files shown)


src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
```diff
@@ -25,6 +27,20 @@ function LockPaywall({
...
+
+  const eventProperties = {
+    org_key: org,
+    courserun_key: courseId,
+  };
+
+  const logClick = () => {
+    sendTrackingLogEvent('edx.bi.ecommerce.upsell_links_clicked', {
+      ...eventProperties,
+      linkCategory: '(none)',
```
ciduarte, line 39: So, I know this looks weird, but I'm under the impression that `'(none)'` is what they want here... I'll double check w/ rev

See:
https://github.com/edx/edx-platform/pull/24526#discussion_r459447956

https://openedx.atlassian.net/wiki/spaces/RS/pages/1675100377/How+to+find+upsell+link+click+events

src/course-home/outline-tab/widgets/UpgradeCard.jsx
```diff
@@ -46,6 +46,12 @@ function UpgradeCard({ courseId, intl, onLearnMore }) {
       location: 'sidebar-message',
     });
     sendTrackEvent('Promotion Clicked', promotionEventProperties);
+    sendTrackingLogEvent('edx.bi.ecommerce.upsell_links_clicked', {
```
MatthewPiatetsky: what is the difference between sendTrackEvent and sendTrackingLogEvent, first goes to segment and second to splunk?

MatthewPiatetsky (approved):
a review from someone on rev squad familiar with these events might be a good idea

mikix (approved):

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
dianekaplan: Yes sorry- this awkward choice was my fault, you're right that this is the way to keep it consistent

dianekaplan (changes requested):
Excited to see this started so soon!

One general flag: I don't see the actual link names passed along to sendTrackEvent or sendTrackingLogEvent in the code (course_home_green and in_course_upgrade)- not sure which one sends to Segment? 

Our trackUpsellClick function in edx-platform took the link name along with the set of properties (pageName, linkType, linkCategory), so segment events have `event:properties:linkName` that we search on, also used by our New Relic graph to see the relative counts of the different upsell links. In this code I don't see the link names being passed along- is this happening in a subtle way that I'm missing, or still to be done?

ciduarte:
> One general flag: I don't see the actual link names passed along to sendTrackEvent or sendTrackingLogEvent in the code (course_home_green and in_course_upgrade)- not sure which one sends to Segment?
> 
> Our trackUpsellClick function in edx-platform took the link name along with the set of properties (pageName, linkType, linkCategory), so segment events have `event:properties:linkName` that we search on, also used by our New Relic graph to see the relative counts of the different upsell links. In this code I don't see the link names being passed along- is this happening in a subtle way that I'm missing, or still to be done?

The `linkName` is definitely something I missed! Glad you caught that.

ciduarte:
>  not sure which one sends to Segment?

`sendTrackEvent` goes to Segment, so that's what we actually want. Updated it to use that rather than `sendTrackingLogEvent`.

dianekaplan (approved):
Thanks for adding the linkName! No other feedback for these changes, so once you've got the new sendTrackEvent lines under test this will be good to go. Thanks!

PR #394, opened 2021-03-17
[REV-2127] feat: update gated content lock screen to Value Prop designs

dianekaplan:
Ticket [REV-2127](https://openedx.atlassian.net/browse/REV-2127): for the gated content lock page in frontend-app-learning, implement value prop designs [here](https://www.figma.com/file/1EaMYJ8FecwzbbSv5YIpEC/Value-Prop-Final-Designs?node-id=1067%3A89) using paragon Alert component

This includes: change of text, updated use of paragon components, icons, and styling. UPDATE: broke the text changes out into an earlier update [here](https://github.com/edx/frontend-app-learning/commit/6abf8531bb64d3f4581eef3f01a98795a9870e23)

Notes: 
- Change not called out in the figma designs: we'll use the middle design for LG and XL, and the smallest design for XS, S, and MD.  (The designs were made from the old/wider courseware implementation on edx-platform but the new learning MFE has a narrower nav, giving less space for the contents inside, so Eugene modified the request accordingly)
- the Alert component from Paragon doesn't yet support swapping an icon for a bullet for list items (logged in [PAR-429](https://openedx.atlassian.net/browse/PAR-429)). In the meantime, our temporary solution is for the lineitems to use a FontAwesomeIcon together with the message text inside, with special styling in LockPaywall.scss to fit it to the design. 
- there's an [open PR ](https://github.com/edx/paragon/pull/663)pending for some updates to the Alert component (for example background color and elevation 1)

- [x] UX review with Eugene
- [x] accessibility fixes with Axe chrome plugin (automatic and guided)
- [x] accessibility review with Jeff Witt
- [x] sanity check with Spanish translations

larger
smaller
openedx

### DIFF
--- a/src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
+++ b/src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
@@ -1,13 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { faLock } from '@fortawesome/free-solid-svg-icons';
+import { faCheck } from '@fortawesome/free-solid-svg-icons';
 import { sendTrackEvent } from '@edx/frontend-platform/analytics';
-import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
-
+import { FormattedMessage, injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import {
+  Alert, Button, Icon,
+} from '@edx/paragon';
+import { Locked } from '@edx/paragon/icons';
 import messages from './messages';
-import VerifiedCert from '../../../../generic/assets/edX_certificate.png';
+import certificateLocked from '../../../../generic/assets/edX_locked_certificate.png';
 import { useModel } from '../../../../generic/model-store';
+import './LockPaywall.scss';
 
 function LockPaywall({
   intl,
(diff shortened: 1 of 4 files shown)


src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
```diff
@@ -42,33 +46,138 @@ function LockPaywall({
       pageName: 'in_course',
     });
   };
+
+  const lockIcon = (
+    <Icon
+      className="float-left"
+      src={Locked}
+      aria-hidden="true"
+      alt=""
```
julianajlk: a11y feedback: add `role="img"` and an `aria-label` attribute for SVGs

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
```diff
@@ -42,33 +46,138 @@ function LockPaywall({
...
-          <FontAwesomeIcon icon={faLock} className="text-black mr-2 ml-1" style={{ fontSize: '2rem' }} />
-          <span>{intl.formatMessage(messages['learn.lockPaywall.title'])}</span>
-        </h4>
-        <p className="mb-0">
-          <span>{intl.formatMessage(messages['learn.lockPaywall.content'])}</span>
-          &nbsp;
-          <a className="lock_paywall_upgrade_link" href={upgradeUrl} onClick={logClick}>
+    <Alert variant="light">
+      <div className="row">
+        {/* Thin left-hand column for lock */}
```
julianajlk: Not sure if you meant to leave these comments?

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
```diff
@@ -42,33 +46,138 @@ function LockPaywall({
...
+        <div className="col-auto px-0">
+          {lockIcon}
+        </div>
+
+        {/* Wide column for all the middle contents */}
+        <div className="col">
+          <div>
+            <div>
```
julianajlk: Curious to know why the multiple `div`s that don't carry a particular className, they might not be necessary (you can also use `<Rect.Fragment>` if you don't want them to show in the DOM)

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
dianekaplan: As I made updates, having these guideposts made it so much easier to immediately see where I was (so it would help anyone else making updates in the future)- but if that's bad practice I'm happy to remove them. What do you think? Better to leave it out and have the code look cleaner?

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
```diff
@@ -42,33 +46,138 @@ function LockPaywall({
...
+                  </li>
+                </ul>
+              </div>
+            </div>
+          </div>
+        </div>
+
+        {/* Column for upgrade button */}
+        <div
+          className="col-md-auto p-md-0 d-md-flex align-items-md-center mr-md-3"
```
julianajlk, line 154: When the util classes get to a number that is too many, I would use classNames so they can be multi-line, you'll have to import it FYI

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
```diff
@@ -42,33 +46,138 @@ function LockPaywall({
...
+          style={{ textAlign: 'right' }}
```
julianajlk, line 155: Could this be added as a bootstrap class along with the other ones?

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
dianekaplan: thanks! Does that also apply to images that you don't want described? Jeff had mentioned that we don't want the screen reader to describe the lock since it doesn't give additional meaning than the sentence next to it, so I have it be ignored with the alt="".

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
dianekaplan: Good call! I changed things around multiple times and may not need all the divs- will have a fresh look and see if some can be removed

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
dianekaplan: I tried but it wasn't taking effect (hence the brute force) but I'll re-check...

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
dianekaplan: okay I'll try it out!

src/courseware/course/sequence/lock-paywall/messages.js
```diff
@@ -3,24 +3,49 @@ import { defineMessages } from '@edx/frontend-platform/i18n';
...
+  },
+  'learn.lockPaywall.list.bullet1.linktext': {
+    id: 'learn.lockPaywall.list.bullet1.linktext',
+    defaultMessage: 'verified certificate',
+    description: 'Link text for verified certificate info page.',
+  },
+  'learn.lockPaywall.list.bullet2.boldtext': {
+    id: 'learn.lockPaywall.list.bullet2.boldtext',
+    defaultMessage: 'graded assignments',
+    description: 'bolded text for graded assignments',
```
julianajlk: Uppercase `bolded` and period, small detail for consistency, on the other 2 messages too

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
julianajlk: True! Nevermind my comment

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
julianajlk: I knew there was a reason 😅

julianajlk (commented):
Great work Diane! I'm done with all my comments, I would also check a11y with `axe` or another tool if you haven't already

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
julianajlk: `thin left-hand column for lock` doesn't mean much if you're not familiar with the wireframe and how it looks tbh. It's also very specific to each section of the gated content banner (ie. most likely won't be worked on this specifically anytime soon). I would not keep comments like this, but you could ask maybe the owners, engage?

julianajlk: Oh wait, we own that part so we can make that call?

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
dianekaplan: Interesting! Switching the className statement into classNames actually broke the textAlign: right: 
`          classNames={('col-md-auto',
          'p-md-0',
          'd-md-flex',
          'align-items-md-center',
          'mr-md-3')}`
          
So instead I'll break this into multiple lines by using className with a template literal

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
dianekaplan: confirmed! It'll only actually take effect when I have it as a separate style, grumble...

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
dianekaplan: You were right! I had two that weren't needed anymore, so I've removed them now- thanks!

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
dianekaplan: Yeah- I'm convinced- if anyone *does* come back on it they can add whatever temporary help they need, but it's not customary for us to have that in the code already. Removed!

src/courseware/course/sequence/lock-paywall/messages.js
dianekaplan: done!

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
dianekaplan: Update: making that change actually stopped the last class from taking effect (mr-md-3) so I've put it back the original way

julianajlk (approved):

PR #395, opened 2021-03-18
feat: handle courseware paths more liberally

kdmccormick:
```
Valid courseware URLs currently include:
* /course/:courseId
* /course/:courseId/:sequenceId
* /course/:courseId/:sequenceId/:unitId

In this commit we add support for:
* /course/:courseId/:sectionId
* /course/:courseId/:sectionId/:unitId
* /course/:courseId/:unitId

All URL forms still redirect to:
  /course/:courseId/:sequenceId/:unitId

See ADR #8 for more context.
```

I recommend reviewing commit-by-commit. I did my best to break up the feature change from all the refactorings.

Blocks https://github.com/edx/edx-platform/pull/26816
Part of https://openedx.atlassian.net/browse/TNL-7796

### DIFF
--- a/docs/decisions/0008-liberal-courseware-path-handling.md
+++ b/docs/decisions/0008-liberal-courseware-path-handling.md
@@ -1,0 +1,90 @@
+# Liberal courseware path handling
+
+## Status
+
+Accepted
+
+_This updates some of the content in [ADR #2: Courseware page decisions](./0002-courseware-page-decisions.md)._
+
+## Context
+
+The courseware container currently accepts three path forms:
+
+1. `/course/:courseId`
+2. `/course/:courseId/:sequenceId`
+3. `/course/:courseId/:sequenceId/:unitId`
+
+Forms #1 and #2 are always redirected to Form #3 via simple set of rules:
+
+* If the sequenceId is not specified, choose the first sequence in the course.
+* If the unitId is not specified, choose the active unit in the sequence,
+  or the first unit if none are active.
+
+Thus, Form #3 is effectively the canonoical path;
+all Learning MFE units should be served from it.
+We acknowledge that the best user experience is to link directly to the canonoical
+path when possible, since it skips the redirection steps.
+Still, there are times when it is necessary or prudent to link just to a course or
+a sequence.
+
+Through recent work in the LMS, we are realizing that there are _also_ times where it
+would be simpler or more performant to link a user to an
+_entire section without specifying a squence_ or to a
+_unit without including the sequence_.
+Specifically, this capability would let as avoid further modulestore or
+block transformer queries in order to discern the course structure when trying to
+direct a learner to a section or unit.
+Futhermore, we hypothesize that being able to build a Learning MFE courseware link
+with just a unit ID or a section ID will be a nice simplifying quality for future
+development or debugging.
+
+## Decision
+
+The courseware container will accept five total path forms:
+
+1. `/course/:courseId`
+2. `/course/:courseId/:sectionId`
+3. `/course/:courseId/:sectionId/:unitId`
+4. `/course/:courseId/:sequenceId`
+5. `/course/:courseId/:unitId`
+6. `/course/:courseId/:sequenceId/:unitId`
+
+The redirection rules are as follows:
+
+* Forms #1 redirects to Form #4 by selecting the first sequence in the course.
+* Form #2 redirects to Form #4 by selecting to the first sequence in the section.
+* Form #3 redirects to Form #5 by dropping the section ID.
+* Form #4 redirects to Form #6 by choosing the active unit in the sequence
+  (or the first unit, if none are active).
+* Form #5 redirects to Form #6 by filling in the ID of the sequence that the
+  specified unit belongs to (in the edge case where the unit belongs to multiple
+  sequences, the first sequence is selected).
+
+As before, Form #5 is the canonocial courseware path, which is always redirected to
+by any of the other courseware path forms.
+
+## Consequences
+
+The above decision is implemented.
+
+## Further work
+
+At some point, we may decide to further extend the URL scheme to be
+more human-readable.
+
+We can't make UsageKeys themselves more readable because they're tied to student state,
+but we could introduce a new optional `slug` field on Sequences,
+which would be captured and propagated to the learning_sequences API.
+We could eventually do something similar to Units, since those slugs only have to be sequence-local.
+
+So eventually, URLs could look less like:
+
+```
+
+https://learning.edx.org/course/course-v1:edX+DemoX.1+2T2019/block-v1:edX+DemoX.1+2T2019+type@sequential+block@e0a61b3d5a2046949e76d12cac5df493/block-v1:edX+DemoX.1+2T2019+type@vertical+block@52dbad5a28e140f291a476f92ce11996
+```
+
+And more like:
+```
+https://learning.edx.org/course/course-v1:edX+DemoX.1+2T2019/Being_Social/Teams
+```
--- a/src/courseware/CoursewareContainer.jsx
+++ b/src/courseware/CoursewareContainer.jsx
@@ -41,6 +41,32 @@ const checkResumeRedirect = memoize((courseStatus, courseId, sequenceId, firstSe
   }
 });
 
+const checkSectionUnitToUnitRedirect = memoize((courseStatus, courseId, sequenceStatus, section, unitId) => {
+  if (courseStatus === 'loaded' && sequenceStatus === 'failed' && section && unitId) {
+    history.replace(`/course/${courseId}/${unitId}`);
+  }
+});
+
+const checkSectionToSequenceRedirect = memoize((courseStatus, courseId, sequenceStatus, section, unitId) => {
+  if (courseStatus === 'loaded' && sequenceStatus === 'failed' && section && !unitId) {
+    // If the section is non-empty, redirect to its first sequence.
+    if (section.sequenceIds && section.sequenceIds[0]) {
+      history.replace(`/course/${courseId}/${section.sequenceIds[0]}`);
+    // Otherwise, just go to the course root, letting the resume redirect take care of things.
+    } else {
+      history.replace(`/course/${courseId}`);
+    }
+  }
+});
+
+const checkUnitToSequenceUnitRedirect = memoize((courseStatus, courseId, sequenceStatus, unit) => {
+  if (courseStatus === 'loaded' && sequenceStatus === 'failed' && unit) {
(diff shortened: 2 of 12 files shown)


docs/decisions/0008-liberal-courseware-path-handling.md
```diff
@@ -0,0 +1,71 @@
+# Liberal coursesware path handling
```
kdmccormick: coursesware -> courseware

davidjoy (commented):
ADR looks great!

ormsbee:
Doesn't get in the way of this work, but thought I'd like to put in here is that we can put slugs on sequences and units. UsageKeys can't change because they're tied to student state, but a slug could be a new optional field that we capture and propagate to the `learning_sequences` API, and we could eventually do something similar to Units, since those slugs only have to be sequence-local. So eventually, URLs could look less like:

```
https://learning.edx.org/course/course-v1:edX+DemoX.1+2T2019/block-v1:edX+DemoX.1+2T2019+type@sequential+block@e0a61b3d5a2046949e76d12cac5df493/block-v1:edX+DemoX.1+2T2019+type@vertical+block@52dbad5a28e140f291a476f92ce11996
```

And more like:

```https://learning.edx.org/course/course-v1:edX+DemoX.1+2T2019/Being_Social/Teams```

File: docs/decisions/0008-liberal-courseware-path-handling.md [hunk 2]
```diff
@@ -0,0 +1,71 @@
...
+
+1. `/course/:courseId`
+2. `/course/:courseId/:courseId`
+3. `/course/:courseId/:sectionId`
+4. `/course/:courseId/:sequenceId`
+5. `/course/:courseId/:unitId`
+6. `/course/:courseId/:sequenceId/:unitId`
+
+Form #2 may seem silly, but exists so that the URL can be generally formulated as:
+`/course/:courseId/:navigationalBlockId`
```
ormsbee: I'm not sure I understand why we'd formulate (2) as a target URL.

> Form 2 may seem silly, but exists so that the URL can be generally formulated as:
`/course/:courseId/:navigationalBlockId`

If it makes it easier to always generate `/course/:courseId/:navigationalBlockId`, then can we generate it the root course as `/course/:courseId/` (trailing slash, so `navigationalBlockId = ""`)? I worry that things will get confusing with the `CourseKey` vs. the `UsageKey` of the root `CourseBlock`.

kdmccormick: > I worry that things will get confusing with the CourseKey vs. the UsageKey of the root CourseBlock.

@ormsbee Right. I always have to remind my self that a course's root block usage key != its course key, even though they carry the same information. I'll remove the `course/:courseId/:courseId` target path.

kdmccormick:
> And more like: `https://learning.edx.org/course/course-v1:edX+DemoX.1+2T2019/Being_Social/Teams`

@ormsbee Nice! I will add a note to the ADR that this is a potential future direction we may want to go.

src/courseware/data/__factories__/courseBlocks.factory.js
```diff
@@ -84,14 +84,67 @@ export default function buildSimpleCourseBlocks(courseId, title, options = {}) {
...
+ *   Seq---+---Seq      Seq---+---Seq
+ *    |         |        |         |
+ * U--+--U   U--+--U  U--+--U   U--+--U
+ *                          ^
+ *
+ * Each left branch is indexed 0, and each right branch is indexed 1.
+ * So, the caret in the diagram above is pointing to `unitTree[1][0][1]`,
+ * whose parent is `sequenceTree[1][0]`, whose parent is `sectionTree[1]`.
+ */
+export function buildBinaryCourseBlocksTree(courseId, title) {
```
kdmccormick: The whole purpose of this commit is so I can introduce this factory function. I need a non-linear course structure to properly test the URL handling changes I'm making.

src/courseware/data/__factories__/courseBlocks.factory.js
```diff
@@ -84,14 +84,67 @@ export default function buildSimpleCourseBlocks(courseId, title, options = {}) {
...
         course: courseBlock,
       },
     ),
     unitBlocks,
-    sequenceBlock,
-    sectionBlock,
+    sequenceBlocks,
+    sectionBlocks,
+    sequenceBlock: sequenceBlocks,
+    sectionBlock: sectionBlocks[0],
```
kdmccormick: I put these here out of curiosity to see if exposing the old names would fix some tests. It did not.

src/courseware/CoursewareContainer.test.jsx
```diff
@@ -141,7 +141,7 @@ describe('CoursewareContainer', () => {
       courseBlocks = result.courseBlocks;
       unitBlocks = result.unitBlocks;
       // eslint-disable-next-line prefer-destructuring
-      sequenceBlock = result.sequenceBlock[0];
+      sequenceBlock = result.sequenceBlocks[0];
```
kdmccormick: Lots of this diff is just fixing pluralization so that I could think straight.

src/courseware/data/__factories__/courseBlocks.factory.js
```diff
@@ -23,28 +23,28 @@ Factory.define('courseBlocks')
       { courseId },
     ),
   ]))
-  .option('sequence', ['courseId', 'units'], (courseId, child) => Factory.build(
```
kdmccormick: Right, so here's an actual change. The courseBlocks factory could only be build with one section and one sequential block. I want to make it accept any number of sections and sequences.

src/courseware/course/sequence/Sequence.test.jsx
```diff
@@ -281,7 +281,7 @@ describe('Sequence', () => {
     });
 
     it('handles the navigation buttons for empty sequence', async () => {
-      const testSequenceBlock = [Factory.build(
+      const testsequenceBlocks = [Factory.build(
```
kdmccormick: second `s` should be capitalized, but it's a consistent typo, so it's not the cause of the test failures.

davidjoy: `testsequenceBlocks` -> `testSequenceBlocks`

src/courseware/CoursewareContainer.jsx
```diff
@@ -110,15 +138,49 @@ class CoursewareContainer extends Component {
     this.checkFetchCourse(routeCourseId);
     this.checkFetchSequence(routeSequenceId);
 
-    // Redirect to the legacy experience for exams.
-    checkExamRedirect(sequenceStatus, sequence);
-
-    // Determine if we need to redirect because our URL is incomplete.
-    checkContentRedirect(courseId, sequenceStatus, sequenceId, sequence, routeUnitId);
+    // All courseware URLs should normalize to the format /course/:courseId/:sequenceId/:unitId
```
davidjoy, line 141: Shouldn't this PR have a change to index.jsx to update the PageRoute paths that direct to CoursewareContainer?

```
<PageRoute
  path={[
    // More paths here?
    '/course/:courseId/:sequenceId/:unitId',
    '/course/:courseId/:sequenceId',
    '/course/:courseId',
  ]}
  component={CoursewareContainer}
/>
```

davidjoy: Huh, or is it that the `/course/:courseId/:sequenceId` accounts for all "two variable" options?  Okay, maybe I understand after all.

kdmccormick: Exactly. The part parts remain named in terms of their intended function.

If we were to name the path parts in terms of what we accept, it would more accurately be:
```
/course/:courseId/:sequenceOrSectionId/:unitId
/course/:courseId/:blockId
/course/:courseId
```

@davidjoy shall I refactor, or leave it as-is? If I leave it, I could still add a clarifying comment to the `<PageRoute>`.

davidjoy (commented):
I think this is excellently written given the complexity of how this thing loads.  

So this may not be a very welcome thought, and maybe we don't do it, but I'm thinking about whether we can move a lot of this logic into the _thunks_ in order to simplify it, make it more performant, and allow us to do it in a more imperative, less declarative way.  

Right now we have all these "check" functions that take a given state and transition us to the next step along the way. 

What if there was one "check" that said "The URL is not in canonical form. Fix it."

It would call a thunk which would:

- Look at the 1-3 IDs in the URL (which could be passed into the thunk)
- Use course blocks data (already in redux and available from getState() passed into the thunk) to figure out what IDs _should_ be in the URL based on what's there.
- Call any necessary APIs based on the state (resume, sequence metadata)
  - "Oh, I have a section, let me grab the right sequence and unit and call sequence metadata."
  - "Just a course!  Let me call the resume API and find out where I should be and call sequence metadata once I find out."
- Fix the URL all at once when we have the data we need.

I just wonder if this would let us be more direct about the whole thing, though I'm well aware a lot of work has already gone into this. 

It's possible there are a lot of corner cases that make this difficult, but I feel like this logic would be easier to write if you were just in a thunk dealing with the data and didn't have to worry about all of React's state changes and memoization.

kdmccormick:
@davidjoy Thanks! But yeah... as soon as I started writing things like "`sectionViaSequenceIdSelector`" I knew there had to be a better way, but I just wasn't sure what it was :smile: 

I think I understand the gist of your suggestion--push this logic down into a lower layer and do it all at once, instead of relying on a series partially-render-then-redirect steps. That makes sense for a lot of reasons. And I am sensitive to the desire to keep this high-traffic frontend code from becoming tangled and contrived.

But I don't think I have the React/Redux chops to makes that change quickly, and this PR already feels like quite tangent off of the original issue (making the MFE the educator default), which I'm anxious to get shipped. It also seems like the MFE has a big data-handling refactor coming soon, anyway: switching it over to the new Learning Sequences API.

Given all that, I think our options are:
1) Refactor before merging.
2) Merge, but then I follow up with a refactor before calling the educator default ticket "done".
3) Merge, and then you follow up with a refactor (you seem to be clear on what would the refactor would look like; would you have the cycles to do that?).
4) Merge, and then make a ticket to refactor after the LS API migration.

What do you think?

davidjoy:
I'm totally with you.  I think the fourth option (merge and make a ticket) is the right move; especially given the other data-handling refactors.  I don't have the cycles for it either... wish I did!

davidjoy (approved):

kdmccormick:
Sounds good! I made a [follow up ticket](https://openedx.atlassian.net/browse/TNL-8182) (it's not filled in yet) and linked it in a comment in CoursewareContainer. I'll merge + deploy tomorrow in the AM.

PR #397, opened 2021-03-22
[AA-713] fix: remove extra x in first section celebration modal

MatthewPiatetsky:
x icon that was added to paragon as part of the streak celebration modal is redundant with the x added to only the first section celebration modal

fixing this bug:
Screen Shot 2021-03-22 at 8 39 25 AM

### DIFF
--- a/src/courseware/course/celebration/CelebrationModal.scss
+++ b/src/courseware/course/celebration/CelebrationModal.scss
@@ -11,10 +11,6 @@
       position: absolute;
       right: 1rem;
     }
-
-    button::after {
-      content: "⨉";
-    }
   }
 
   .modal-body {


src/courseware/course/celebration/CelebrationModal.scss
```diff
@@ -11,10 +11,6 @@
       position: absolute;
       right: 1rem;
     }
-
-    button::after {
-      content: "⨉";
-    }
```
mikix, line 17: That's my code I think? I can only hope that it was actually useful when I wrote it - I don't _recall_ there being two x's...

mikix (approved):

src/courseware/course/celebration/CelebrationModal.scss
MatthewPiatetsky: per the pr description i think I probably broke this when I added an X icon to paragon for the streak celebration modal but the first section celebration modal already had an x

mikix: Well upstreaming it to paragon is the better thing anyway - I probably should have done that originally myself. But nice catch on this.

PR #401, opened 2021-03-30
mmp2p: handle missing verified-mode info after upgrade deadline

muselesscreator:
It looks like the verified mode data is stripped from the data model after upgrade deadline, and thus our experiment cannot activate anymore, because it is not appropriately reading users as audit.


mattcarter (approved):
approved

PR #404, opened 2021-04-02
feat: use legacy_web_url to redirect to legacy courseware 

kdmccormick:
## Description

As part of making the new courseware experience the
default for staff, the LMS /jump_to/ links that are
exposed by the Course Blocks API via the `lms_web_url`
field will soon direct users to whichever experience
is active to them (instead of always directing to
the legacy experience & relying on the learner
redirect).

Because of this, the MFE can no longer rely on
`lms_web_url` to land a staff user to the legacy
experience. However, the aformentioned change
will also introduce a `legacy_web_url` field
to the API, which we *can* use for this purpose.

## Test instructions

1. Spin up LMS (on edx-platform `master`) and frontend-app-learning (on this branch).
2. Make sure you have a course run that is configured to use the new courseware experience (either by globally enabling `courseware.courseware_mfe`, or by specifically enabling it for that course).
3. Log in as a global staff user or as an instructor for the course.
4. From your course dashboard, navigate to the course, and click "Resume" in the course outline.
5. You should find yourself in the **Legacy Experience**. 
6. Click "View In New Experience". You should be directed to the MFE.
7. Click "Legacy Experience". You should be directed back to legacy frontend.
8. Now, check out [this PR's branch](https://github.com/edx/edx-platform/pull/26816) and restart LMS.
9. Starting again from your course dashboard, navigate to the course, and click "Resume" in the course outline.
10. You should find yourself in the **New Experience**.
11. Click "Legacy Experience". You should be directed to legacy frontend.
12. Click "View In New Experience". You should be directed back to the MFE.

## Merge deadline

Today (Friday 4/2) would be good, so that we're all prepped for the MFE-educator-default release on Monday.

## Supporting information

Part of https://openedx.atlassian.net/browse/TNL-7796

This PR is a prerequisite for the corresponding edx-platform PR, which changes the behavior of lms_web_url and adds legacy_web_url: https://github.com/edx/edx-platform/pull/26816

### DIFF
--- a/src/course-home/data/api.js
+++ b/src/course-home/data/api.js
@@ -56,8 +56,13 @@ export function normalizeOutlineBlocks(courseId, blocks) {
           effortTime: block.effort_time,
           icon: block.icon,
           id: block.id,
-          lmsWebUrl: block.lms_web_url,
-          showLink: !!block.lms_web_url, // we reconstruct the url ourselves as an MFE-internal <Link>
+          // Fall back to `lms_web_url` until `legacy_web_url` is added to API (TNL-7796).
+          legacyWebUrl: block.legacy_web_url || block.lms_web_url,
+          // The presence of an LMS URL for the sequence indicates that we want this
+          // sequence to be a clickable link in the outline (even though, if the new
+          // courseware experience is active, we will ignore `legacyWebUrl` and build a
+          // link to the MFE ourselves).
+          showLink: !!(block.legacy_web_url || block.lms_web_url),
           title: block.display_name,
         };
         break;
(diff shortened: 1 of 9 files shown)


src/course-home/data/api.js
```diff
@@ -56,7 +56,8 @@ export function normalizeOutlineBlocks(courseId, blocks) {
           effortTime: block.effort_time,
           icon: block.icon,
           id: block.id,
-          lmsWebUrl: block.lms_web_url,
+          // Fall back to `lms_web_url` until `legacy_web_url` is added to API (TNL-7796).
+          legacyWebUrl: block.legacy_web_url || block.lms_web_url,
           showLink: !!block.lms_web_url, // we reconstruct the url ourselves as an MFE-internal <Link>
```
doctoryes: Should this `lms_web_url` use the same logic as the line above?

doctoryes (commented):
I won't approve here just to emphasize that my review isn't in-depth - but had one question.

src/course-home/data/api.js
kdmccormick: Good call. I made that change, and added an explanatory comment.

(Even though `!!block.lms_web_url` will be logically equivalent to `!!block.legacy_web_url` in practice, it's best to be consistent)

doctoryes (approved):
Grain of 🧂 and all for my approval - but 👍🏼

PR #415, opened 2021-04-15
[AA-719] Create certificate status component 

MatthewPiatetsky:
Jira Ticket: https://openedx.atlassian.net/browse/AA-719
[PR to add necessary data that was missing](https://github.com/edx/edx-platform/pull/27323) 

The certificate status component appears on the right top part of the progress page
Screen Shot 2021-04-23 at 11 23 43 AM

It shows one of the below ten cases

1. Certificate is not available because learner is not passing the course
Screen Shot 2021-04-15 at 1 00 00 PM

2. Certificate is not available because learner is not passing the course, but the course still has scheduled content in the future
Screen Shot 2021-04-15 at 12 40 26 PM

3. Certificate is available to be requested
Screen Shot 2021-04-15 at 1 01 04 PM

4. Certificate is unavailable because learner has not completed ID verification
Screen Shot 2021-04-15 at 1 01 30 PM

5. Certificate is unavailable because ID verification is pending
Screen Shot 2021-04-23 at 11 20 55 AM

6. Certificate is available to be downloaded
Screen Shot 2021-04-23 at 11 19 54 AM

7. Certificate is available to be viewed
Screen Shot 2021-04-23 at 11 16 27 AM

8. Certificate has been earned but is not available yet because the course is not complete
Screen Shot 2021-04-23 at 11 23 35 AM

9. Certificate is unavailable because learner needs to upgrade
Screen Shot 2021-04-15 at 4 55 59 PM

10. When none of the status cases apply, the component does not appear
Screen Shot 2021-04-23 at 11 22 29 AM

### DIFF
--- a/src/course-home/progress-tab/ProgressTab.test.jsx
+++ b/src/course-home/progress-tab/ProgressTab.test.jsx
@@ -27,6 +27,11 @@ describe('Progress Tab', () => {
   const defaultMetadata = Factory.build('courseHomeMetadata', { id: courseId });
   const defaultTabData = Factory.build('progressTabData');
 
+  function setMetadata(attributes, options) {
+    const courseMetadata = Factory.build('courseHomeMetadata', { id: courseId, ...attributes }, options);
+    axiosMock.onGet(courseMetadataUrl).reply(200, courseMetadata);
+  }
+
   function setTabData(attributes, options) {
     const progressTabData = Factory.build('progressTabData', attributes, options);
     axiosMock.onGet(progressUrl).reply(200, progressTabData);
@@ -191,6 +196,155 @@ describe('Progress Tab', () => {
       await fetchAndRender();
       expect(screen.getByText('Detailed grades')).toBeInTheDocument();
       expect(screen.getByText('You currently have no graded problem scores.')).toBeInTheDocument();
+    });
+  });
+
+  describe('Certificate Status', () => {
+    Object.defineProperty(window, 'matchMedia', {
+      writable: true,
+      value: jest.fn().mockImplementation(query => {
+        const matches = !!(query === 'screen and (min-width: 768px)' || query === 'screen and (min-width: 992px)');
+        return {
+          matches,
+          media: query,
+          onchange: null,
+          addListener: jest.fn(), // deprecated
+          removeListener: jest.fn(), // deprecated
+          addEventListener: jest.fn(),
+          removeEventListener: jest.fn(),
+          dispatchEvent: jest.fn(),
+        };
+      }),
+    });
+
+    describe('enrolled user', () => {
+      beforeEach(async () => {
+        setMetadata({ is_enrolled: true });
+      });
+
+      it('Displays text for nonPassing case when learner does not have a passing grade', async () => {
+        setTabData({
+          user_has_passing_grade: false,
+        });
+        await fetchAndRender();
+        expect(screen.getByText('In order to qualify for a certificate, you must have a passing grade.')).toBeInTheDocument();
+      });
+
+      it('Displays text for inProgress case when more content is scheduled and the learner does not have a passing grade', async () => {
+        setTabData({
+          user_has_passing_grade: false,
+          has_scheduled_content: true,
+        });
+        await fetchAndRender();
+        expect(screen.getByText('It looks like there is more content in this course that will be released in the future. Look out for email updates or check back on your course for when this content will be available.')).toBeInTheDocument();
+      });
+
+      it('Displays request certificate link', async () => {
+        setTabData({
+          certificate_data: { cert_status: 'requesting' },
+          user_has_passing_grade: true,
+        });
+        await fetchAndRender();
+        expect(screen.getByRole('button', { name: 'Request certificate' })).toBeInTheDocument();
+      });
+
+      it('Displays verify identity link', async () => {
+        setTabData({
+          certificate_data: { cert_status: 'unverified' },
+          user_has_passing_grade: true,
+          verification_data: { link: 'test' },
+        });
+        await fetchAndRender();
+        expect(screen.getByRole('link', { name: 'Verify ID' })).toBeInTheDocument();
+      });
+
+      it('Displays verification pending message', async () => {
+        setTabData({
+          certificate_data: { cert_status: 'unverified' },
+          verification_data: { status: 'pending' },
+          user_has_passing_grade: true,
+        });
+        await fetchAndRender();
+        expect(screen.getByText('Your ID verification is pending and your certificate will be available once approved.')).toBeInTheDocument();
+        expect(screen.queryByRole('link', { name: 'Verify ID' })).not.toBeInTheDocument();
+      });
+
+      it('Displays download link', async () => {
+        setTabData({
+          certificate_data: {
+            cert_status: 'downloadable',
+            download_url: 'fake.download.url',
+          },
+          user_has_passing_grade: true,
+        });
+        await fetchAndRender();
+        expect(screen.getByRole('link', { name: 'Download my certificate' })).toBeInTheDocument();
+      });
+
+      it('Displays webview link', async () => {
+        setTabData({
+          certificate_data: {
+            cert_status: 'downloadable',
+            cert_web_view_url: '/certificates/cooluuidgoeshere',
+          },
+          user_has_passing_grade: true,
+        });
+        await fetchAndRender();
+        expect(screen.getByRole('link', { name: 'View my certificate' })).toBeInTheDocument();
+      });
+
+      it('Displays certificate is earned but unavailable message', async () => {
+        setTabData({
+          certificate_data: { cert_status: 'earned_but_not_available' },
+          user_has_passing_grade: true,
+        });
+        await fetchAndRender();
+        expect(screen.queryByText('Certificate status')).toBeInTheDocument();
+      });
+
+      it('Displays upgrade link when available', async () => {
+        setTabData({
+          certificate_data: { cert_status: 'audit_passing' },
+          verified_mode: {
+            upgrade_url: 'http://localhost:18130/basket/add/?sku=8CF08E5',
+          },
+        });
+        await fetchAndRender();
+        // Keep these text checks in sync with "audit only" test below, so it doesn't end up checking for text that is
+        // never actually there, when/if the text changes.
+        expect(screen.getByText('You are in an audit track and do not qualify for a certificate. In order to work towards a certificate, upgrade your course today.')).toBeInTheDocument();
+        expect(screen.getByRole('link', { name: 'Upgrade now' })).toBeInTheDocument();
+      });
+
+      it('Displays nothing if audit only', async () => {
+        setTabData({
+          certificate_data: { cert_status: 'audit_passing' },
+          verified_mode: null,
+        });
+        await fetchAndRender();
+        // Keep these queries in sync with "upgrade link" test above, so we don't end up checking for text that is
+        // never actually there, when/if the text changes.
+        expect(screen.queryByText('You are in an audit track and do not qualify for a certificate. In order to work towards a certificate, upgrade your course today.')).not.toBeInTheDocument();
+        expect(screen.queryByRole('link', { name: 'Upgrade now' })).not.toBeInTheDocument();
+      });
+
+      it('Does not display the certificate component if it does not match any statuses', async () => {
+        setTabData({
+          certificate_data: {
+            cert_status: 'bogus_status',
+          },
+          user_has_passing_grade: true,
+        });
+        setMetadata({ is_enrolled: true });
+        await fetchAndRender();
+        expect(screen.queryByTestId('certificate-status-component')).not.toBeInTheDocument();
+      });
+    });
+
+    it('Does not display the certificate component if the user is not enrolled', async () => {
+      setMetadata({ is_enrolled: false });
+      await fetchAndRender();
+      expect(screen.queryByTestId('certificate-status-component')).not.toBeInTheDocument();
     });
   });
 });
--- a/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
+++ b/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
@@ -1,12 +1,173 @@
 import React from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import {
+  FormattedDate, FormattedMessage, injectIntl, intlShape,
+} from '@edx/frontend-platform/i18n';
 
-function CertificateStatus() {
+import { Button, Card } from '@edx/paragon';
+import { getConfig } from '@edx/frontend-platform';
+import { useModel } from '../../../generic/model-store';
+import { COURSE_EXIT_MODES, getCourseExitMode } from '../../../courseware/course/course-exit/utils';
+import { DashboardLink, IdVerificationSupportLink, ProfileLink } from '../../../shared/links';
+import { requestCert } from '../../data/thunks';
+import messages from './messages';
+
+function CertificateStatus({ intl }) {
+  const {
+    courseId,
+  } = useSelector(state => state.courseHome);
+
+  const {
+    isEnrolled,
+  } = useModel('courseHomeMeta', courseId);
+
+  const {
+    certificateData,
+    hasScheduledContent,
+    userHasPassingGrade,
+  } = useModel('progress', courseId);
+
+  const mode = getCourseExitMode(
+    certificateData,
+    hasScheduledContent,
+    isEnrolled,
+    userHasPassingGrade,
+  );
+  const dispatch = useDispatch();
+
+  const {
+    end,
+    verificationData,
+    certificateData: {
+      certStatus,
+      certWebViewUrl,
+      downloadUrl,
+    },
+    verifiedMode,
+  } = useModel('progress', courseId);
+
+  let certCase;
+  let body;
+  let buttonAction;
+  let buttonLocation;
+  let buttonText;
+  let endDate;
+
+  const dashboardLink = <DashboardLink />;
+  const idVerificationSupportLink = <IdVerificationSupportLink />;
+  const profileLink = <ProfileLink />;
+
+  if (mode === COURSE_EXIT_MODES.nonPassing) {
+    certCase = 'notPassing';
+    body = intl.formatMessage(messages[`${certCase}Body`]);
+  } else if (mode === COURSE_EXIT_MODES.inProgress) {
+    certCase = 'inProgress';
+    body = intl.formatMessage(messages[`${certCase}Body`]);
+  } else if (mode === COURSE_EXIT_MODES.celebration) {
+    switch (certStatus) {
+      case 'requesting':
+        // Requestable
+        certCase = 'requestable';
+        buttonAction = () => { dispatch(requestCert(courseId)); };
+        body = intl.formatMessage(messages[`${certCase}Body`]);
+        buttonText = intl.formatMessage(messages[`${certCase}Button`]);
+        break;
+
+      case 'unverified':
+        certCase = 'unverified';
+        if (verificationData.status === 'pending') {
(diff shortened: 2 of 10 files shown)


src/course-home/progress-tab/ProgressTab.test.jsx
```diff
@@ -44,6 +53,10 @@ describe('Progress Tab', () => {
     axiosMock.onGet(courseMetadataUrl).reply(200, defaultMetadata);
     axiosMock.onGet(progressUrl).reply(200, defaultTabData);
 
+    // Set necessary defaults for Certificate Status component
+    setMetadata({ is_enrolled: true });
+    setTabData({ certificate_data: {} });
```
MatthewPiatetsky: I'm setting a default here because otherwise the code errors when destructuring certificateData
let me know if I should do this a different way

src/course-home/progress-tab/certificate-status/utils.js
```diff
@@ -0,0 +1,66 @@
...
+  'honor_passing',
+  'requesting',
+  'unverified',
+];
+const NON_CERTIFICATE_STATUSES = [
+  'audit_passing',
+  'honor_passing',
+];
+
+function getCertificateStatusMode(courseId) {
```
mikix: This is basically a copy/paste from course exit yeah? We should consolidate that code rather than copy it. It's finicky and if we want these two pieces of UI to mirror each other (i.e. we change it in one and we want the other to have same new logic).

mikix (commented):
Looks pretty!

src/course-home/progress-tab/ProgressTab.test.jsx
mikix: I think there's also a way to specify a default when destructuring? But if we start doing that, probably better to do it at the api/thunk level rather than every consuming component. Doing this in tests is fine though, I think? Carla might have a better opinion here.

src/course-home/progress-tab/certificate-status/utils.js
MatthewPiatetsky: these are the same from a business logic perspective, but different from an implementation perspective since they get data from different models.  
but i guess I can pull out getting the data into the main methods and then pass all the data in as arguments

mikix: Yeah protecting the business logic from drifting seems most important.

If pulling the data out makes some surrounding code gross, you could keep these two methods as convenience wrappers around a third business logic utility method that takes the data as a direct argument.

src/course-home/progress-tab/ProgressTab.test.jsx
```diff
@@ -27,8 +27,17 @@ describe('Progress Tab', () => {
...
+    const courseMetadata = Factory.build('courseHomeMetadata', { id: courseId, ...attributes }, options);
+    axiosMock.onGet(courseMetadataUrl).reply(200, courseMetadata);
+  }
+
   function setTabData(attributes, options) {
-    const progressTabData = Factory.build('progressTabData', attributes, options);
+    const dataAttributes = attributes;
+    if (!attributes.certificate_data) {
+      dataAttributes.certificate_data = {};
+    }
```
ciduarte: We can just switch `certificate_data` in progressTabData.factory.js from `null` to `{}`, removing the need for `setTabData` on line 58.

src/course-home/progress-tab/ProgressTab.test.jsx
```diff
@@ -44,6 +53,10 @@ describe('Progress Tab', () => {
     axiosMock.onGet(courseMetadataUrl).reply(200, defaultMetadata);
     axiosMock.onGet(progressUrl).reply(200, defaultTabData);
 
+    // Set necessary defaults for Certificate Status component
+    setMetadata({ is_enrolled: true });
```
ciduarte: Can we move this line into a `beforeEach(() => {})` within the `describe('Certificate Status')`?

src/course-home/progress-tab/ProgressTab.test.jsx
```diff
@@ -82,4 +95,132 @@ describe('Progress Tab', () => {
       expect(screen.getByText('You currently have no graded problem scores.')).toBeInTheDocument();
     });
   });
+
+  describe('Certificate Status', () => {
+    it('Displays text for nonPassing case when learner does not have a passing grade', async () => {
+      setTabData({
+        certificate_data: {},
```
ciduarte: If we default this value in the factory, we won't need this line

src/course-home/progress-tab/ProgressTab.test.jsx
```diff
@@ -82,4 +95,132 @@ describe('Progress Tab', () => {
...
+        user_has_passing_grade: false,
+      });
+      await fetchAndRender();
+      expect(screen.getByText('In order to qualify for a certificate, you must have a passing grade.')).toBeInTheDocument();
+    });
+
+    it('Displays text for inProgress case when more content is scheduled and the learner does not have a passing grade', async () => {
+      setTabData({
+        certificate_data: {},
```
ciduarte: Same here

src/course-home/progress-tab/ProgressTab.test.jsx
ciduarte: I see you're testing a not-enrolled case towards the end. I think you can wrap all the tests that require enrolled to be true in another `describe('enrolled user')` and put the before each in there..

src/course-home/progress-tab/certificate-status/messages.js
```diff
@@ -0,0 +1,91 @@
...
+  // Certificate available, download
+  downloadableHeader: {
+    id: 'downloadableHeader',
+    defaultMessage: 'Your certificate is available!',
+  },
+  downloadableBody: {
+    id: 'downloadableBody',
+    defaultMessage: 'Showcase your accomplishment on LinkedIn or your resume today. You can download your certificate now and access it any time from your Dashboard and Profile.',
+  },
+  // (include links to Dashboard and Profile)
```
ciduarte: Is this a leftover todo?

src/course-home/progress-tab/certificate-status/messages.js
```diff
@@ -0,0 +1,91 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  // Not passing
```
ciduarte: I don't think the comments in this file are necessary

src/course-home/progress-tab/certificate-status/utils.js
```diff
@@ -0,0 +1,66 @@
...
+    certificateData,
+    hasScheduledContent,
+    userHasPassingGrade,
+  } = useModel('progress', courseId);
+
+  const authenticatedUser = getAuthenticatedUser();
+
+  if (!authenticatedUser || !isEnrolled) {
+    return CERTIFICATE_STATUS_MODES.disabled;
+  }
```
ciduarte: So we check if the user is enrolled in course exit in order to redirect, but I'm curious if we want the same behavior for the progress tab? What if a user has completed a course, but unenrolled? Would they still be able to see their grades and certificate status here? Probably a question for product

src/course-home/progress-tab/certificate-status/utils.js
ciduarte: I agree we should consolidate where we can.

ciduarte (commented):
Looks good so far! I think you're going to address the course exit overlaps, so I can re-review when that's pushed up. 👍

src/course-home/progress-tab/certificate-status/messages.js
```diff
@@ -0,0 +1,91 @@
...
+    defaultMessage: 'View my certificate',
+  },
+  // Not available yet
+  notAvailableHeader: {
+    id: 'notAvailableHeader',
+    defaultMessage: 'Your certificate will be available soon!',
+  },
+  notAvailableBody: {
+    id: 'notAvailableBody',
+    defaultMessage: 'Your certificate will be available soon! After this course officially ends on {end_date}, you will receive an email notification with your certificate.',
```
ciduarte, line 66: Did we want "Your certificate will be available soon!" again in the body?

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
```diff
@@ -1,12 +1,196 @@
...
+    >
+      {intl.formatMessage(courseExitMessages.profileLink)}
+    </Hyperlink>
+  );
+
+  let buttonLocation;
+  let buttonPrefix;
+  let certCase;
+  let body;
+  let button;
```
ciduarte: Should this be `buttonText`?

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
```diff
@@ -1,12 +1,196 @@
...
+      className="text-gray-700"
+      style={{ textDecoration: 'underline' }}
+      destination={`${getConfig().LMS_BASE_URL}/u/${username}`}
+    >
+      {intl.formatMessage(courseExitMessages.profileLink)}
+    </Hyperlink>
+  );
+
+  let buttonLocation;
+  let buttonPrefix;
```
ciduarte: I think this would be cleaner if we defined a `buttonAction` and `buttonHref` and built the button in the return, rather than carrying along a possibly null `buttonPrefix`.

Then in the return we could do `<Button variant="outline-brand" onClick={buttonAction} href={buttonLocation}>{buttonText}</Button>`

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
```diff
@@ -1,12 +1,196 @@
...
+      <Card>
+        <Card.Body>
+          <Card.Title><h3 className="h4">{header}</h3></Card.Title>
+          <Card.Text>
+            {body}
+          </Card.Text>
+          {buttonPrefix}
+          {button && buttonLocation && (
+            <Button variant="outline-brand" href={buttonLocation}>{button}</Button>
+          )}
```
ciduarte: I suggested the switch above because it feels more clean to consolidate these two buttons into one child:

`<Button variant="outline-brand" onClick={buttonAction} href={buttonLocation}>{buttonText}</Button>`

Unless there's a case I'm unaware of where two buttons may appear here?

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
MatthewPiatetsky: sure i can rename the variable

src/course-home/progress-tab/ProgressTab.test.jsx
MatthewPiatetsky: good idea, thanks!

src/course-home/progress-tab/ProgressTab.test.jsx
MatthewPiatetsky: sure, updated

src/course-home/progress-tab/ProgressTab.test.jsx
MatthewPiatetsky: addressed with Carla's suggestion

src/course-home/progress-tab/certificate-status/messages.js
MatthewPiatetsky: fixed, good catch, thanks!

src/course-home/progress-tab/certificate-status/utils.js
MatthewPiatetsky: that's a good question, though I don't think we would handle any tab-wide cases like that at the certificate status component level

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
MatthewPiatetsky: good call, i think this is one of the cases where the business logic differs from course exit, so this code can be simplified given that we don't have a two-button case

src/courseware/course/course-exit/links.jsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import { getAuthenticatedUser } from '@edx/frontend-platform/auth';
+import { getConfig } from '@edx/frontend-platform';
+import { Hyperlink } from '@edx/paragon';
+
+import messages from './messages';
+
+function IntlDashboardLink({ intl }) {
```
ciduarte: Nice!

I wonder if we could put this links.jsx file in a more general location? I know we use the dashboard and profile links a few times around this repo and it'd be nice to re-use these components! Maybe the `generic` folder?

src/courseware/course/course-exit/utils.js
```diff
@@ -26,18 +25,16 @@ const NON_CERTIFICATE_STATUSES = [ // no certificate will be given, though a val
...
-function getCourseExitMode(courseId) {
-  const {
-    certificateData,
-    courseExitPageIsActive,
-    hasScheduledContent,
-    isEnrolled,
-    userHasPassingGrade,
-  } = useModel('coursewareMeta', courseId);
-
+function getCourseExitMode(
```
ciduarte, line 28: I feel like this function should be pulled into the `generic` folder and be renamed getCourseCertificateStatus

src/courseware/course/course-exit/links.jsx
MatthewPiatetsky: good idea!
i moved it to the shared folder that I created as part of the streak celebration work, since it turned out the generic folder was created with the intention of housing code not specific to the learning mfe while the shared code is for learning mfe specific code shared between different folders

src/courseware/course/course-exit/utils.js
MatthewPiatetsky: i'm unsure whether pulling this function out would make things clearer since getCourseExitNavigation would get more confusing if getCourseExitMode and COURSE_EXIT_MODES were in a different file and its all rather heavy on the implementation details/specific arguments vs the more easily reusable links stuff. 
happy to move this anyway even if i disagree if you have a strong opinion about it though

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
```diff
@@ -1,12 +1,173 @@
...
-      <h3 className="h4">Certificate status</h3>
+    <section data-testid="certificate-status-component" className="text-dark-700 mb-4 p-4">
+      <Card>
+        <Card.Body>
+          <Card.Title><h3 className="h4">{header}</h3></Card.Title>
+          <Card.Text>
+            {body}
+          </Card.Text>
+          {buttonText && (buttonLocation || buttonAction) && (
+            <Button variant="outline-brand" onClick={buttonAction} href={buttonLocation}>{buttonText}</Button>
```
ciduarte: didn't notice this earlier, the mocks show this button as a `block` button that spans the whole width

example: `<Button variant="outline-brand" block>`

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
```diff
@@ -1,12 +1,173 @@
...
+
+  const header = intl.formatMessage(messages[`${certCase}Header`]);
 
-function CertificateStatus() {
   return (
-    <section className="text-dark-700 rounded shadow-sm mb-4 p-4">
-      {/* TODO: AA-719 */}
-      <h3 className="h4">Certificate status</h3>
+    <section data-testid="certificate-status-component" className="text-dark-700 mb-4 p-4">
+      <Card>
```
ciduarte: this card should have a background of light 200 (ex. `bg-light-200`)

src/course-home/progress-tab/certificate-status/messages.js
```diff
@@ -0,0 +1,82 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  notPassingHeader: {
+    id: 'notPassingHeader',
+    defaultMessage: 'Certificate Status',
```
ciduarte: `Status` should be lowercase

src/course-home/progress-tab/certificate-status/messages.js
```diff
@@ -0,0 +1,82 @@
...
+    id: 'downloadableBody',
+    defaultMessage: 'Showcase your accomplishment on LinkedIn or your resume today. You can download your certificate now and access it any time from your Dashboard and Profile.',
+  },
+  downloadableButton: {
+    id: 'downloadableButton',
+    defaultMessage: 'Download my certificate',
+  },
+  viewableButton: {
+    id: 'viewableButton',
+    defaultMessage: 'View my certificate',
```
ciduarte, line 58: "my" shouldn't be in both of these buttons

src/course-home/progress-tab/certificate-status/messages.js
```diff
@@ -0,0 +1,82 @@
...
+    id: 'inProgressHeader',
+    defaultMessage: 'More content is coming soon!',
+  },
+  inProgressBody: {
+    id: 'inProgressBody',
+    defaultMessage: 'It looks like there is more content in this course that will be released in the future. Look out for email updates or check back on your course for when this content will be available.',
+  },
+  requestableHeader: {
+    id: 'requestableHeader',
+    defaultMessage: 'Certificate Status',
```
ciduarte: Status should be lowercase

src/course-home/progress-tab/certificate-status/messages.js
```diff
@@ -0,0 +1,82 @@
...
+    id: 'requestableBody',
+    defaultMessage: 'Congratulations, you qualified for a certificate! In order to access your certificate, request it below.',
+  },
+  requestableButton: {
+    id: 'requestableButton',
+    defaultMessage: 'Request certificate',
+  },
+  unverifiedHeader: {
+    id: 'unverifiedHeader',
+    defaultMessage: 'Certificate Status',
```
ciduarte: Status should be lowercase

ciduarte (commented):
Just a few styling comments - sorry I didn't get to these before!

Looking at the mocks, it seems like we want the certificate status component on mobile view to be under completion and above grades, but to move to the side panel on desktop.

Screen Shot 2021-04-28 at 11 04 18 AM

We have a nifty `layoutGenerator` that we use in this repo where you could set a breakpoint for where we add the side panel (>= 992px). So in Progress Tab you could plop them in the right spots by doing something like this:
```
const layout = layoutGenerator({
  mobile: 0,
  desktop: 992,
});

const OnMobile = layout.is('mobile');
const OnAtLeastDesktop = layout.isAtLeast('desktop');

<OnMobile>
  <CertificateStatus />
</OnMobile>
<OnAtLeastDesktop>
  <CertificateStatus />
</OnAtLeastDesktop>
```

src/course-home/progress-tab/certificate-status/messages.js
ciduarte: sorry I capitalized these in the initial ticket write up!

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
ciduarte: Oh and this card needs a shadow, I think we're using `shadow-sm`

src/course-home/progress-tab/certificate-status/messages.js
MatthewPiatetsky: https://edx-internal.slack.com/archives/G019J2RMNEA/p1619188106007500 based on this thread it sounds like we want the my

MatthewPiatetsky:
i guess the mocks changed at some point relative to the screenshot in the ticket?  
updated screenshots
Screen Shot 2021-04-28 at 11 50 08 AM

Screen Shot 2021-04-28 at 11 50 23 AM

src/course-home/progress-tab/certificate-status/messages.js
ciduarte: Makes sense!

ciduarte (approved):

PR #416, opened 2021-04-15
Create 0009-discussions-integration.md

davidjoy:


docs/decisions/0009-discussions-integration.md
```diff
@@ -0,0 +1,16 @@
...
+
+Furthermore, we will need to revisit this if it ever becomes the case that an LTI-based discussions provider implements in-context discussions.  Our approach should ideally be roughly compatible with LTI concepts and techniques, though I'd strongly suggest we _don't_ build anything LTI-specific for the foreseeable future.
+
+The suggested implementation involves:
+
+- Updating the sequence metadata API to include information about whether a particular unit should load in-context discussions.  For now, this could take the form of a `discussions_url` added to the unit objects inside the `items` array.  Longer term we'll want a way to signal to the frontend-app-learning MFE the plugins that should be available on the page, but that likely will come from some other API.  I don't think we should worry about that broader plugin configuration as part of this.
```
xitij2000, line 13: Instead of directly providing a URL, perhaps the sequence metadata can just signal that a particular unit has in-context discussions, and the discussions code in the learning MFE could derive the URL from the unit's usage key?  If eventually the sidebar is based on a frontend plugin, each plugin could handle checking whether it's enabled and what to show based on the course key and the usage key.

docs/decisions/0009-discussions-integration.md
```diff
@@ -0,0 +1,16 @@
...
+- Adding a new `InContextSidebar` React component to frontend-app-learning to house the discussions integration.  The medium to long-term goal is that discussions will not be the only experience plugin loaded in this sidebar, so it would be appropriate to add a thin layer of abstraction between the components that render the discussions iframe and the components that handle the sidebar's behavior.
+- Have the discussions iframe load the page at `discussions_url`
+- If necessary, use a postMessage-based API to communicate into the iframe and pass additional data into it.  I don't know of a use-case for this for the discussions project, but wanted to mention it.  The schema of these messages would ideally be LTI-friendly, though the LTI specification doesn't have a lot to say about a postMessage schema.  This document suggests a reasonable message schema: https://github.com/bracken/lti_messaging
```
xitij2000, line 16: I think this would be useful for the course tab embed at least for resizing dynamically.

docs/decisions/0009-discussions-integration.md
davidjoy: If there are multiple discussions providers that can do in-context discussion, I think we want the MFE pretty agnostic to how the iframe URL is constructed or where it points. 🤔

davidjoy: (And if we eventually have plugins in here, the MFE won't be able to build URLs because there's unlikely to be a consistent URL schema for those...)

It's early and I'm half awake of course, so I may be missing something, admittedly. :)

xitij2000: My thinking here was that, since we're using LTI currently, the next step for in-context discussions will be based on LTI, so they will need more context as an iframe URL won't suffice. We will either need to add more information to the sequence API, or have that information in a new API dedicated for this. 

i.e. the sequence metadata signals that a unit has an associated discussion. The learning MFE then calls the provider-specific code for what/how to display content in the sidebar. To start with this can just support an iframe based on cs_comments. For LTI providers there can be another code path that uses LTI embeds here, and in the future it can support frontend plugins that allow other kinds of custom providers.

Does that make sense? This way the backend might not need any plugin, just generic APIs that the frontend can use to derive how to embed each provider.

davidjoy: I think so - I just don't think the frontend should be responsible for building a URL if at all possible.  In the future it can't know what LTI provider has been configured in the backend, and shouldn't know provider-specific URL schemas.  If we need to pass additional information into the provider's iframe, then we could either add query string parameters or use postMessage (I'd lean toward the latter).

davidjoy:
Going to close this, as I don't think there's value in it/it may not reflect reality.  @xitij2000 feel free to chime in if you think we should merge this instead.

xitij2000:
> Going to close this, as I don't think there's value in it/it may not reflect reality. @xitij2000 feel free to chime in if you think we should merge this instead.

I think there are bits in it that need to be updated, but roughly it is the plan that's being followed. I don't see how we can return the discussion information in the present form of the learning sequences API since it doesn't concern itself with Units, however I have a WIP PR (https://github.com/edx/edx-platform/pull/29300) that adds this data to the course blocks API and the learning MFE can use the data from there to render in-context discussions.

PR #418, opened 2021-04-21
fix: AA-738: Switch our use of FormattedTime to use hourCycle

Dillon-Dumesnil:
We had a bug reported where learners were seeing a due date like
March 24, 24:59 instead of March 25, 00:59. This is a bug that only
shows up in Chrome. The hour12 flag overrides the hourCycle flag so
we are just going to swap the two. h23 means a 24 hour format ranging
from 0-23 (there also exists a h24 option which goes from 1-24).

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat
for any additional details on the options.

### DIFF
--- a/src/course-home/outline-tab/SequenceLink.jsx
+++ b/src/course-home/outline-tab/SequenceLink.jsx
@@ -96,7 +96,7 @@ function SequenceLink({
                       day="numeric"
                       month="short"
                       year="numeric"
-                      hour12={false}
+                      hourCycle="h23"
                       timeZoneName="short"
                       value={due}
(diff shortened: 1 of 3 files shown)


MatthewPiatetsky (approved):
Not sure if we frequently use FormattedTime in a similar way but could we have or risk introducing this bug elsewhere?

Dillon-Dumesnil:
Good catch. Two other spots we use this (in start and end date alerts). I have updated those as well.

PR #421, opened 2021-04-23
fix: AA-663: Update header text for CourseCompletion

Dillon-Dumesnil:
If the marketing url is not set, we shouldn't have a message
about sharing.

https://openedx.atlassian.net/browse/AA-663

### DIFF
--- a/src/courseware/course/course-exit/CourseCelebration.jsx
+++ b/src/courseware/course/course-exit/CourseCelebration.jsx
@@ -45,6 +45,7 @@ function CourseCelebration({ intl }) {
     certificateData,
     end,
     linkedinAddToProfileUrl,
+    marketingUrl,
     offer,
     org,
     relatedPrograms,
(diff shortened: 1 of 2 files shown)


src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -287,7 +288,8 @@ function CourseCelebration({ intl }) {
           {intl.formatMessage(messages.congratulationsHeader)}
         </div>
         <div className="col-12 p-0 font-weight-normal lead text-center">
-          {intl.formatMessage(messages.shareHeader)}
+          {intl.formatMessage(messages.completedCourseHeader)}
+          {marketingUrl && ' ' + intl.formatMessage(messages.shareMessage)}
```
Dillon-Dumesnil: Wasn't sure if this or 
```
{marketingUrl ?  intl.formatMessage(messages.shareMessage) : intl.formatMessage(messages.completedCourseHeader)}
```
would be preferred. Let me know if you have any thoughts!

MatthewPiatetsky (approved):

src/courseware/course/course-exit/CourseCelebration.jsx
mikix: If you did the terniary clause, you'd need both strings on the left side, yeah, not just the one? (Unless you're saying the translation would have both?)

I think the way you did it is good.

mikix (approved):

src/courseware/course/course-exit/CourseCelebration.jsx
Dillon-Dumesnil: Ah whoops, yeah, the ternary clause would use the old version of the shareMessage message which included both strings

PR #422, opened 2021-04-23
fix: A couple of fixes for lilac

Dillon-Dumesnil:
I was originally only going to do my most recent commit to fix the header in course completion, but realized most of the other recent commits also were fixes that would likely be good for lilac to have.

Please see the individual commits for the relevant changes.


MatthewPiatetsky (commented):
This seems like a lot of changes, do any need anything in the release notes?

Dillon-Dumesnil:
I just scanned the changes and don't think that's necessary. It's mostly minor bug fixes and the biggest change is the still unreleased progress tab, but we can document that when it is fully released, likely in the next release.

PR #423, opened 2021-04-26
fix: use new xblock handler URLs for sequence blocks

mikix:
They've changed to proper new-style handlers, so the URL also changed. This will let us get the fix for green checkmarks showing up as audit users pass FBE units, even though those units aren't actually complete.

https://openedx.atlassian.net/browse/AA-409
https://github.com/edx/edx-platform/pull/27422

### DIFF
--- a/src/courseware/data/api.js
+++ b/src/courseware/data/api.js
@@ -204,48 +204,22 @@ export async function getSequenceMetadata(sequenceId) {
   return normalizeSequenceMetadata(data);
 }
 
-const getSequenceXModuleHandlerUrl = (courseId, sequenceId) => `${getConfig().LMS_BASE_URL}/courses/${courseId}/xblock/${sequenceId}/handler/xmodule_handler`;
+const getSequenceHandlerUrl = (courseId, sequenceId) => `${getConfig().LMS_BASE_URL}/courses/${courseId}/xblock/${sequenceId}/handler`;
 
 export async function getBlockCompletion(courseId, sequenceId, usageKey) {
-  // Post data sent to this endpoint must be url encoded
-  // TODO: Remove the need for this to be the case.
-  // TODO: Ensure this usage of URLSearchParams is working in Internet Explorer
-  const urlEncoded = new URLSearchParams();
-  urlEncoded.append('usage_key', usageKey);
-  const requestConfig = {
-    headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
-  };
-
   const { data } = await getAuthenticatedHttpClient().post(
-    `${getSequenceXModuleHandlerUrl(courseId, sequenceId)}/get_completion`,
-    urlEncoded.toString(),
-    requestConfig,
+    `${getSequenceHandlerUrl(courseId, sequenceId)}/get_completion`,
+    { usage_key: usageKey },
   );
-
-  if (data.complete) {
-    return true;
-  }
-
-  return false;
+  return data.complete === true;
 }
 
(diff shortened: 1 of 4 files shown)


src/courseware/data/api.js
```diff
@@ -204,48 +204,22 @@ export async function getSequenceMetadata(sequenceId) {
...
+    `${getSequenceHandlerUrl(courseId, sequenceId)}/get_completion`,
+    { usage_key: usageKey },
   );
-
-  if (data.complete) {
-    return true;
-  }
-
-  return false;
+  return !!data.complete;
```
Dillon-Dumesnil: 👏

src/courseware/data/api.js
```diff
@@ -204,48 +204,22 @@ export async function getSequenceMetadata(sequenceId) {
   return normalizeSequenceMetadata(data);
 }
 
-const getSequenceXModuleHandlerUrl = (courseId, sequenceId) => `${getConfig().LMS_BASE_URL}/courses/${courseId}/xblock/${sequenceId}/handler/xmodule_handler`;
+const getSequenceHandlerUrl = (courseId, sequenceId) => `${getConfig().LMS_BASE_URL}/courses/${courseId}/xblock/${sequenceId}/handler`;
 
 export async function getBlockCompletion(courseId, sequenceId, usageKey) {
-  // Post data sent to this endpoint must be url encoded
-  // TODO: Remove the need for this to be the case.
-  // TODO: Ensure this usage of URLSearchParams is working in Internet Explorer
```
Dillon-Dumesnil, line 212: ... Wonder whatever happened here.

src/courseware/data/api.js
```diff
@@ -204,48 +204,22 @@ export async function getSequenceMetadata(sequenceId) {
...
-  const requestConfig = {
-    headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
-  };
-
   const { data } = await getAuthenticatedHttpClient().post(
-    `${getSequenceXModuleHandlerUrl(courseId, sequenceId)}/goto_position`,
-    urlEncoded.toString(),
-    requestConfig,
+    `${getSequenceHandlerUrl(courseId, sequenceId)}/goto_position`,
+    // Position is 1-indexed on the server and 0-indexed in this app. Adjust here.
```
Dillon-Dumesnil, line 220: :chef_kiss:

Dillon-Dumesnil (approved):
Subject to approval of the backend, but all these changes look good.

src/courseware/data/api.js
MatthewPiatetsky: wouldn't  ` return data.complete === true `be more correct?
(like checked here https://github.com/edx/edx-platform/pull/27422/files#diff-a647370de737de1375504ef9c1cc9857e68deae43de99f116edb87740e217212R422)
not that we should expect data.complete to have a value of 126 or "bob", but if it did have those values we wouldn't necessarily want to mark it as complete
of course this should work fine too so just a nitpicky thought

MatthewPiatetsky (approved):
were you able to test this locally?

mikix:
> were you able to test this locally?

Yeah with the new backend, both master and this PR worked as expected (master had the bug but was able to call old handlers fine, this PR called new fixed handlers). But this needs the new backend code to be in prod first.

src/courseware/data/api.js
mikix: Yeah fair. Might be more defensive to check for true especially.

PR #425, opened 2021-04-28
Account activation pop up added:

adeelehsan:
After the registration show account activation pop up to learner if next url is dashboard or course-view. Its a one time reminder only.
VAN-435
Screen Shot 2021-05-21 at 11 20 52 AM


davidjoy (changes requested):
Apologies, but there are a number of things I think we're going to need to fix.

1. Please use the AlertModal from Paragon - it should remove the need to have custom CSS for this.  `Modal` has a poor API and is soon to be deprecated.  Documentation is here: https://paragon-edx.netlify.app/components/modal/alert-modal/ 
2. We can't hard code "edX" into our MFEs... these are used by the Open edX community.  If we want to have a button that says "Continue to edX" the "edX" part of that needs to be set to the SITE_NAME configuration variable and parameterized in the i18n message.  
3. Please use more unambiguous `id` attributes for your FormattedMessages and more descriptive descriptions - "cta" isn't really specific to this functionality, and translators may need more details in the descriptions to make these understandable.
4. I see several warnings on the PR about packages that aren't in package.json, such as universal-cookie.  These need to be fixed.
5. There are other warnings about using anchor tags as buttons - these are accessibility issues.  You may need to add a `role` to the anchor tag if it appears as a link and has an onClick handler.

adeelehsan:
@davidjoy please review it again. I have addressed the requested changes.

davidjoy:
Hi @adeelehsan - this looks so good!  Thanks for making those changes.

The one last thing I didn't think of at the time would be to add a few component tests for `AccountActivationAlert`, focusing on:

- Proper hiding/showing of the component depending on cookie state (presumably you'll have to mock js-cookie...)
- Proper calling of the activation email endpoint when a user clicks the button.

Basically we want to make sure it shows up only when it's supposed to, and actually calls the API it's supposed to call!

davidjoy (changes requested):
Thank you for indulging me so far, @adeelehsan.  I hate to say it, but a test that shows that the alert _doesn't_ show up unless the cookie is true seems important.  Otherwise we don't know whether this is being shown to all users all the time.

src/alerts/logistration-alert/AccountActivationAlert.jsx
```diff
@@ -0,0 +1,81 @@
...
+          {sendEmailTag}."
+          description="Message for account activation alert which is shown after the registration"
+          values={{
+            boldEmail: <b>{getAuthenticatedUser().email}</b>,
+            sendEmailTag: (
+            // eslint-disable-next-line jsx-a11y/anchor-is-valid
+              <a href="#" role="button" onClick={handleOnClick}>
+                <FormattedMessage
+                  id="account-activation.resend.link"
+                  defaultMessage="resend the email"
```
davidjoy: Should this message start with a capital?  I.e., "Resend the email"

adeelehsan: According to the design it should be "resend the email"

src/alerts/logistration-alert/AccountActivationAlert.jsx
```diff
@@ -0,0 +1,85 @@
...
+function AccountActivationAlert() {
+  const [showModal, setShowModal] = useState(false);
+  const handleOnClick = () => {
+    sendActivationEmail();
+    setShowModal(false);
+  };
+
+  const showAccountActivationAlert = Cookies.get('show-account-activation-popup');
+  if (showAccountActivationAlert === 'true') {
+    Cookies.remove('show-account-activation-popup', { path: '/', domain: process.env.SESSION_COOKIE_DOMAIN });
```
davidjoy, line 30: Need this domain be specified?  Does it default to the current domain?

Also, if necessary, custom environment variables need to be added to edx-internal so they show up during the build, and should also be added to the object returned by `getConfig()`, rather than being accessed directly from `process.env`.

davidjoy (approved):
Small reminder re: the new environment variable.  Otherwise looks good.

src/alerts/logistration-alert/AccountActivationAlert.jsx
adeelehsan: domain is `.edx.org`. Already created the PR for internal. thanks for the reminder.

src/alerts/logistration-alert/AccountActivationAlert.jsx
```diff
@@ -0,0 +1,85 @@
...
+      />
+    </Button>
+  );
+
+  return (
+    <AlertModal
+      isOpen={showModal}
+      title={title}
+      footerNode={button}
+      onClose={() => ({})}
```
mikix, line 128: This feels odd - is this a required prop?

mikix (approved):
I'm a little confused about this case - from the screenshot looks like the user is already logged in, but the modal is talking about logging back in. But regardless, I'm sure you and UX have it all sorted.

src/alerts/logistration-alert/AccountActivationAlert.jsx
adeelehsan: Yes, that's why passing empty implementation.

mikix (approved):

PR #426, opened 2021-04-29
AA-723: progress tab locked content experience

ciduarte:
Mobile and desktop view of locked content experience
<div>
Screen Shot 2021-04-29 at 11 10 36 AM
Screen Shot 2021-04-29 at 11 10 18 AM
</div>

### DIFF
--- a/src/course-home/progress-tab/ProgressTab.test.jsx
+++ b/src/course-home/progress-tab/ProgressTab.test.jsx
@@ -161,6 +161,22 @@ describe('Progress Tab', () => {
       expect(screen.getByText('B: 80%-90%'));
       expect(screen.getByText('F: <80%'));
     });
+
+    it('renders locked feature preview when user has locked content', async () => {
+      setTabData({
+        completion_summary: {
+          complete_count: 1,
+          incomplete_count: 1,
+          locked_count: 1,
+        },
+      });
+      await fetchAndRender();
+      expect(screen.getByText('locked feature')).toBeInTheDocument();
+    });
+    it('does not render locked feature preview when user does not have locked content', async () => {
+      await fetchAndRender();
+      expect(screen.queryByText('locked feature')).not.toBeInTheDocument();
+    });
   });
 
   describe('Grade Summary', () => {
@@ -200,21 +216,23 @@ describe('Progress Tab', () => {
   });
 
   describe('Certificate Status', () => {
-    Object.defineProperty(window, 'matchMedia', {
-      writable: true,
-      value: jest.fn().mockImplementation(query => {
-        const matches = !!(query === 'screen and (min-width: 768px)' || query === 'screen and (min-width: 992px)');
-        return {
-          matches,
-          media: query,
-          onchange: null,
-          addListener: jest.fn(), // deprecated
-          removeListener: jest.fn(), // deprecated
-          addEventListener: jest.fn(),
-          removeEventListener: jest.fn(),
-          dispatchEvent: jest.fn(),
-        };
-      }),
+    beforeAll(() => {
+      Object.defineProperty(window, 'matchMedia', {
+        writable: true,
+        value: jest.fn().mockImplementation(query => {
+          const matches = !!(query === 'screen and (min-width: 992px)');
+          return {
+            matches,
+            media: query,
+            onchange: null,
+            addListener: jest.fn(), // deprecated
+            removeListener: jest.fn(), // deprecated
+            addEventListener: jest.fn(),
+            removeEventListener: jest.fn(),
+            dispatchEvent: jest.fn(),
+          };
+        }),
+      });
     });
 
     describe('enrolled user', () => {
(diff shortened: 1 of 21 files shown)


src/course-home/progress-tab/ProgressTab.test.jsx
```diff
@@ -200,21 +216,23 @@ describe('Progress Tab', () => {
...
-          addEventListener: jest.fn(),
-          removeEventListener: jest.fn(),
-          dispatchEvent: jest.fn(),
-        };
-      }),
+    beforeAll(() => {
+      Object.defineProperty(window, 'matchMedia', {
+        writable: true,
+        value: jest.fn().mockImplementation(query => {
+          const matches = !!(query === 'screen and (min-width: 992px)');
```
ciduarte, line 223: The first query was only necessary because this block wasn't wrapped in a `beforeAll` so I updated & removed it so the query would more accurately represent the CertificateStatus component's scenario

src/course-home/progress-tab/grades/messages.js
```diff
@@ -29,6 +29,22 @@ const messages = defineMessages({
...
     defaultMessage: 'You’re currently passing this course with a grade of {letterGrade} ({minGrade}-{maxGrade}%)',
   },
+  courseGradePreviewHeader: {
+    id: 'progress.courseGrade.preview.header',
+    defaultMessage: 'locked feature',
+  },
+  courseGradePreviewHeaderAriaHidden: {
+    id: 'progress.courseGrade.preview.header',
+    defaultMessage: 'Preview of a ',
+  },
```
ciduarte: These are broken up per Jeff Witt's suggestion to make the preview aria-hidden to avoid screen reader confusion.

src/course-home/progress-tab/grades/course-grade/GradeRangeTooltip.jsx
```diff
@@ -27,51 +27,55 @@ function GradeRangeTooltip({ intl, passingGrade }) {
   const gradeRangeEntries = Object.entries(gradeRange);
 
   return (
-    <div className="col-auto m-0 pl-2 pr-0">
```
ciduarte, line 30: GitHub doesn't make this easy to see but I just removed this wrapping `<div>` and added `iconButtonClassname` to the `<IconButton />`

src/course-home/progress-tab/ProgressTab.test.jsx
MatthewPiatetsky: thanks!

src/course-home/progress-tab/certificate-status/messages.js
```diff
@@ -2,79 +2,79 @@ import { defineMessages } from '@edx/frontend-platform/i18n';
 
 const messages = defineMessages({
   notPassingHeader: {
-    id: 'notPassingHeader',
+    id: 'progress.certificateStatus.notPassingHeader',
```
MatthewPiatetsky, line 5: thanks!

src/course-home/progress-tab/grades/course-grade/CourseGrade.jsx
```diff
@@ -29,18 +33,24 @@ function CourseGrade({ intl }) {
 
   passingGrade = Number(passingGrade.toFixed(0));
 
+  const isLocked = lockedCount > 0;
+  const applyLockedOverlay = isLocked ? 'locked-overlay' : '';
```
MatthewPiatetsky, line 37: couldn't you just do 
```const lockedOverlay = lockedCount ? 'locked-overlay' : '';```

**Review by MatthewPiatetsky — APPROVED** [review]
1. do we think there is potential for people to get confused and think the 0% in the screenshot is their current grade?
2. What's the screenreader behavior here?
3. wasn't there something with a lock icon inside the donut or is that separate from this?

**Comment by ciduarte** [comment]
@MatthewPiatetsky responses to your questions:
> 1. do we think there is potential for people to get confused and think the 0% in the screenshot is their current grade?

I'm not sure what you mean by this. This will reflect their current grade, and if it is 0%, then that's what it will show.

> 2. What's the screenreader behavior here?

We've hidden the locked overlay section using `aria-hidden`, so all of the locked-overlay portion will be invisible to a screen reader. We've also hidden the text "Preview of a", so the screen reader will only read "Locked feature. Unlock to view grades and work towards a certificate."

> 3. wasn't there something with a lock icon inside the donut or is that separate from this?

That was covered in an earlier ticket!

File: src/courseware/course/sequence/sequence-navigation/SequenceNavigation.test.jsx [hunk 6]
```diff
@@ -70,7 +70,7 @@ describe('Sequence Navigation', () => {
     expect(testData.onNavigate).not.toHaveBeenCalled();
     // TODO: Not sure if this is working as expected, because the `contentType="lock"` will be overridden by the value
     //  from Redux. To make this provide a `fa-icon` lock we could introduce something like `overriddenContentType`.
-    expect(unitButton.firstChild).toHaveClass('fa-edit');
+    expect(unitButton.firstChild.firstChild).toHaveClass('fa-edit');
```
ciduarte, line 73: I believe a Paragon update nested this element further.

PR #427, opened 2021-04-29
feat: Deny course staff visibility of legacy courseware.

doctoryes:
https://openedx.atlassian.net/browse/TNL-8203

Currently, course staff can always view their courses
in the Legacy courseware experience.

With this change, course staff will *not* be able
to view their courses if the New (MFE) courseware
experience has been enabled for them.

This does not affect global staff, and it does not
affect courses that are still running in the Legacy
experience.

Adds a new parameter returned by the course metadata API
used by the courseware MFE to determine if the button to
show legacy experience should be displayed or not.

Companion PR: https://github.com/edx/edx-platform/pull/27469

### DIFF
--- a/src/instructor-toolbar/InstructorToolbar.jsx
+++ b/src/instructor-toolbar/InstructorToolbar.jsx
@@ -53,9 +53,14 @@ export default function InstructorToolbar(props) {
   const {
     courseId,
     unitId,
+    canViewLegacyCourseware,
   } = props;
   const urlInsights = getInsightsUrl(courseId);
   const urlLegacy = useSelector((state) => {
+    if (!canViewLegacyCourseware) {
+      return undefined;
+    }
+
     if (!unitId) {
       return undefined;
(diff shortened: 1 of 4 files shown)


kdmccormick (approved):

src/instructor-toolbar/InstructorToolbar.jsx
```diff
@@ -53,9 +53,14 @@ export default function InstructorToolbar(props) {
   const {
     courseId,
     unitId,
+    canViewLegacyCourseware,
   } = props;
   const urlInsights = getInsightsUrl(courseId);
   const urlLegacy = useSelector((state) => {
+    if (!canViewLegacyCourseware) {
```
awaisdar001, line 60: Just curious to know if `canViewLegacyCourseware` is false - none of the following options/redirect URLs will be available?
* Legacy experience
* Studio
* Insights

PR #432, opened 2021-05-03
fix: update sequence navigation buttons to use iconBefore and iconAfter

ciduarte:
A recent Paragon update wraps the children of `<Button />` in a `<span />`

ex:
```
<Button>
  <span className="btn-icons-container">
    {children}
  </span>
</Button>
```

This clashed with the custom CSS for the sequence navigation buttons, making the content of the buttons disappear on mobile.
Screen Shot 2021-05-03 at 5 25 56 PM

I updated the sequence navigation buttons to use the `iconBefore` and `iconAfter` utilities provided by Paragon, and this fixed subsequent styling clashes that came from the custom CSS applied to these buttons.

**Screenshots**
Screen Shot 2021-05-03 at 6 52 50 PM

Before and after:
<div>
Screen Shot 2021-05-03 at 6 54 34 PM
Screen Shot 2021-05-03 at 6 52 42 PM
</div>

### DIFF
--- a/src/index.scss
+++ b/src/index.scss
@@ -255,10 +255,6 @@
     @media (max-width: -1 + map-get($grid-breakpoints, 'sm')) {
       padding-top: 1rem;
       padding-bottom: 1rem;
-
-      span {
-        @include sr-only();
-      }
     }
 
     @media (min-width: map-get($grid-breakpoints, 'sm')) {
--- a/src/courseware/course/sequence/sequence-navigation/SequenceNavigation.jsx
+++ b/src/courseware/course/sequence/sequence-navigation/SequenceNavigation.jsx
@@ -1,9 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Button } from '@edx/paragon';
+import { ChevronLeft, ChevronRight } from '@edx/paragon/icons';
 import classNames from 'classnames';
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { faChevronLeft, faChevronRight } from '@fortawesome/free-solid-svg-icons';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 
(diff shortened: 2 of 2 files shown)


src/index.scss
```diff
@@ -255,10 +255,6 @@
     @media (max-width: -1 + map-get($grid-breakpoints, 'sm')) {
       padding-top: 1rem;
       padding-bottom: 1rem;
-
-      span {
-        @include sr-only();
-      }
```
ciduarte, line 261: This mixin was applying the following css to that new Paragon span (essentially hiding its content):
```
@media (max-width: 575px)
.sequence-navigation .next-btn span, .sequence-navigation .previous-btn span {
    position: absolute;
    width: 1px;
    height: 1px;
    overflow: hidden;
    clip: rect(0,0,0,0);
    white-space: nowrap;
}
```

This mixin is seemingly unnecessary, as an `sr-only` utility class already exists through Paragon which can be appended to the element in question.

ciduarte, line 261: This mixin was applying the following css to that new Paragon span:
```
@media (max-width: 575px)
.sequence-navigation .next-btn span, .sequence-navigation .previous-btn span {
    position: absolute;
    width: 1px;
    height: 1px;
    overflow: hidden;
    clip: rect(0,0,0,0);
    white-space: nowrap;
}
```

This mixin is seemingly unnecessary, as an `sr-only` utility class already exists through Paragon which can be appended to the element in question.

davidjoy:
Hmm, feels like there's not enough padding on the right side of Previous and the left side of Next. Did it always look that way?

ciduarte:
> Hmm, feels like there's not enough padding on the right side of Previous and the left side of Next. Did it always look that way?

@davidjoy I included before and after screenshots (the two mobile views) and it looks like it was that way before!

davidjoy (approved):

PR #433, opened 2021-05-04
fix: accurately represent pass/fail grade range

ciduarte:


src/course-home/progress-tab/grades/course-grade/CourseGradeFooter.jsx
```diff
@@ -35,7 +35,7 @@ function CourseGradeFooter({ intl, passingGrade }) {
   const OnMobile = layout.is('mobile');
   const OnAtLeastTablet = layout.isAtLeast('tablet');
 
-  const hasLetterGrades = !gradeRange.pass;
```
ciduarte, line 38: this key wasn't accurately representing the pass/fail scenario, as "pass" can be changed to any value in Studio

Dillon-Dumesnil (approved):

PR #434, opened 2021-05-05
Fix height issue in safari

awaisdar001:
This PR fixes LTI modal height for the safari browser. `height:100%` doesn't seem to work there. 

This fix is specific to the Safari browser. 
In the new learning experience, the LTI modal contained inside `iframe` does not respect  `height:100%`. To fix that I have changed it to have a 100 percent view height using `height: 100vh`. 

[TNL-7410](https://openedx.atlassian.net/browse/TNL-7410)

Before fix:
Screen Shot 2021-05-05 at 10 53 58 PM

After Fix:
Screen Shot 2021-05-05 at 10 53 45 PM

### DIFF
--- a/src/courseware/course/sequence/Unit.jsx
+++ b/src/courseware/course/sequence/Unit.jsx
@@ -174,7 +174,7 @@ function Unit({
                     src={modalOptions.url}
                     style={{
                       width: '100%',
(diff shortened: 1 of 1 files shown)


ciduarte (approved):
Looks good to me!

PR #437, opened 2021-05-06
fix: Remove confusing language on proctoring info panel

alangsto:
## [MST-700 ](https://openedx.atlassian.net/browse/MST-700)

Remove language that says learners with a verified onboarding can take exams. This created confusion for learners who did not realize they must also complete IDV before taking a proctored exam.


schenedx (approved):
👍

PR #440, opened 2021-05-07
fix checkmark bullets to vertically align with text

dianekaplan:
The top of the checkmarks was not quite at the height of the text. This fixes that bug.


julianajlk (approved):
I'm assuming no need for `!important` like in Emma's case?

dianekaplan:
> I'm assuming no need for `!important` like in Emma's case?

Sadly I'll have to verify in stage: my local is already showing the correct alignment

PR #442, opened 2021-05-10
fix: i18n translations for UpgradeCard

julianajlk:
[REV-2224](https://openedx.atlassian.net/browse/REV-2224).

Translations with missing formatted strings for https://github.com/edx/frontend-app-learning/pull/377

**Pls note:**
`<Hyperlink>` component from Paragon is not used here because it looks like it has a bug, it adds an extra space at the end of the link. So keeping it as `<a>` tag with styling instead. 
Screen Shot 2021-05-11 at 12 16 38 PM

### DIFF
--- a/src/course-home/outline-tab/widgets/UpgradeCard.jsx
+++ b/src/course-home/outline-tab/widgets/UpgradeCard.jsx
@@ -11,6 +11,20 @@ import { Hyperlink } from '@edx/paragon';
 import { UpgradeButton } from '../../../generic/upgrade-button';
 
 function UpsellNoFBECardContent() {
+  const verifiedCertLink = (
+    <Hyperlink
+      destination={`${getConfig().MARKETING_SITE_BASE_URL}/verified-certificate`}
+      className="font-weight-bold"
+      target="_blank"
+      rel="noopener noreferrer"
+    >
+      <FormattedMessage
+        id="learning.outline.widgets.upgradeCard.verifiedCertLink"
+        defaultMessage="verified certificate"
+      />
+    </Hyperlink>
+  );
+
   return (
     <ul className="fa-ul upgrade-card-ul pt-0">
       <li>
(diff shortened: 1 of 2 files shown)


src/course-home/outline-tab/widgets/UpgradeCard.jsx
```diff
@@ -6,28 +6,39 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { sendTrackEvent, sendTrackingLogEvent } from '@edx/frontend-platform/analytics';
 import { FormattedDate, FormattedMessage, injectIntl } from '@edx/frontend-platform/i18n';
 import { getConfig } from '@edx/frontend-platform';
+import { Alert } from '@edx/paragon';
 
 import { UpgradeButton } from '../../../generic/upgrade-button';
 
 function UpsellNoFBECardContent() {
+  const verifiedCertLink = (
+    <Alert.Link
```
julianajlk, line 15: This should be <Hyperlink>, change coming soon in the next commit

src/course-home/outline-tab/widgets/UpgradeCard.jsx
```diff
@@ -10,24 +10,29 @@ import { getConfig } from '@edx/frontend-platform';
...
+  );
+
   return (
     <ul className="fa-ul upgrade-card-ul pt-0">
       <li>
         <span className="fa-li upgrade-card-li"><FontAwesomeIcon icon={faCheck} /></span>
         <FormattedMessage
-          id="learning.outline.alert.upgradecard.verifiedCertLink"
+          id="learning.outline.widgets.upgradeCard.verifiedCertMessage"
           defaultMessage="Earn a {verifiedCertLink} of completion to showcase on your resume"
```
dianekaplan, line 28: I'm curious about the alert -> widgets change for the ids: when are they each typically used?

julianajlk: Emma modeled it from another file in the "alert" folder, so I changed it to "widgets" because `UpgradeCard` is in that folder. For Transifex purposes, I believe it does not matter how you name your ids, it will not throw an error if it's not technically the correct "path" since it's not really a path

emma-green (approved):

PR #443, opened 2021-05-10
AA-792: Progress tab UI fixes

ciduarte:
- Responsive page break @ `md` rather than `lg`
- Increase padding @ top of outline & progress titles
- Contain 100% grade tooltip within card
- Vertical centering for currently passing/non-passing message
- Fix grade range tooltip wrapping by itself

Screen Shot 2021-05-10 at 2 53 32 PM
Screen Shot 2021-05-10 at 2 53 18 PM

### DIFF
--- a/src/course-home/progress-tab/grades/course-grade/CourseGradeFooter.jsx
+++ b/src/course-home/progress-tab/grades/course-grade/CourseGradeFooter.jsx
@@ -55,30 +55,34 @@ function CourseGradeFooter({ intl, passingGrade }) {
     }
   }
 
+  const icon = isPassing ? <Icon src={CheckCircle} className="text-success-300 d-inline-flex align-bottom" />
+    : <Icon src={WarningFilled} className="d-inline-flex align-bottom" />;
+
   return (
-    <div className={`row w-100 m-0 px-4 py-3 py-md-4 align-items-baseline rounded-bottom ${isPassing ? 'bg-success-100' : 'bg-warning-100'}`}>
-      <div className="col-auto p-0 align-self-md-center">
-        {isPassing && (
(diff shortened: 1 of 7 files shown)


src/course-home/progress-tab/grades/course-grade/CourseGradeFooter.jsx
```diff
@@ -55,30 +55,34 @@ function CourseGradeFooter({ intl, passingGrade }) {
...
           </span>
         </OnMobile>
         <OnAtLeastTablet>
-          <span className="h4 m-0">
+          <span className="h4 m-0 align-bottom">
             {footerText}
             {hasLetterGrades && (
-              <GradeRangeTooltip iconButtonClassName="h3 ml-1" passingGrade={passingGrade} />
+              <span style={{ whiteSpace: 'nowrap' }}>
+                &nbsp;
```
Dillon-Dumesnil, line 83: What is this doing?

MatthewPiatetsky (approved):
a bit hard to visualize the changes without before screenshots but code looks good to me
(i think the Alert component could come in handy next time for some of the icon switching?)

also looks like codecov is unhappy

Dillon-Dumesnil (approved):
Similar to Matthew, sure!

src/course-home/progress-tab/grades/course-grade/CourseGradeFooter.jsx
ciduarte: a #hack, it's making it so the tooltip doesn't wrap away from the text. the `&nbsp;` doesn't work on its own because the tooltip doesn't behave like text, so the `whiteSpace: nowrap` on the span is joining the`&nbsp;` with the tooltip and restricting wrapping, and the `&nbsp;` is holding onto the preceding `footerText`

PR #444, opened 2021-05-11
[AA-759] fix: don't throw an error when there is no verified mode

MatthewPiatetsky:


src/tab-page/LoadedTabPage.jsx
```diff
@@ -60,15 +60,17 @@ function LoadedTabPage({
...
-      <StreakModal
-        courseId={courseId}
-        metadataModel={metadataModel}
-        streakLengthToCelebrate={streakLengthToCelebrate}
-        isStreakCelebrationOpen={isStreakCelebrationOpen}
-        closeStreakCelebration={closeStreakCelebration}
-        AA759ExperimentEnabled={AA759ExperimentEnabled}
-        verifiedMode={verifiedMode}
-      />
+      {streakLengthToCelebrate && (
```
MatthewPiatetsky: add back condition to running the StreakModal component - there isn't a case where we need to run the streak modal code when this value is false (even if the modal is closed)

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -74,18 +74,24 @@ function StreakModal({
...
+  let upgradeUrl;
+  let mode;
+  let offer;
 
-  const offer = {
-    discountedPrice: (mode.price * 0.85).toFixed(2).toString(),
-    originalPrice: mode.price.toString(),
-    upgradeUrl: mode.upgradeUrl,
-  };
+  if (verifiedMode) {
```
MatthewPiatetsky, line 84: all these changes are just coding defensively for when verifiedMode isn't there

mikix (approved):
Sure. `offer` can be left undefined like that no problem?

src/tab-page/LoadedTabPage.jsx
```diff
@@ -37,12 +37,12 @@ function LoadedTabPage({
 
   const activeTab = tabs.filter(tab => tab.slug === activeTabSlug)[0];
 
+  const { verifiedMode } = useModel(metadataModel, courseId);
```
ciduarte: can we add this field to the useModel above?

src/tab-page/LoadedTabPage.jsx
ciduarte: nit: I'd rather us early exit within the StreakModal component

MatthewPiatetsky:
offer is set if you've set verified mode
but the stuff using offer only appears if AA759ExperimentEnabled is true and i've added an extra safety check to only set that one to true if verified mode is true

src/tab-page/LoadedTabPage.jsx
MatthewPiatetsky: tbh the existing way seems a tiny bit clearer to me, but sure i've removed this and added an early exit

ciduarte (approved):

PR #445, opened 2021-05-12
AA-790: progress tab handle unenrolled/unauthenticated users

ciduarte:
**Unauthenticated learner visiting progress tab**
https://user-images.githubusercontent.com/25124041/118039849-c76bb100-b33e-11eb-9a0a-06575b80892d.mov

**Unenrolled learner visiting progress tab**
The speed of this one is questionable but I partially blame devstack
https://user-images.githubusercontent.com/25124041/118039778-b4f17780-b33e-11eb-8817-ef4b88c01387.mov

### DIFF
--- a/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
+++ b/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
@@ -23,8 +23,11 @@ function CertificateStatus({ intl }) {
 
   const {
     certificateData,
+    end,
     hasScheduledContent,
     userHasPassingGrade,
+    verificationData,
+    verifiedMode,
   } = useModel('progress', courseId);
 
   const mode = getCourseExitMode(
@@ -35,16 +38,15 @@ function CertificateStatus({ intl }) {
   );
   const dispatch = useDispatch();
 
-  const {
-    end,
-    verificationData,
-    certificateData: {
-      certStatus,
-      certWebViewUrl,
-      downloadUrl,
-    },
-    verifiedMode,
-  } = useModel('progress', courseId);
+  let certStatus;
+  let certWebViewUrl;
+  let downloadUrl;
+
+  if (certificateData) {
+    certStatus = certificateData.certStatus;
+    certWebViewUrl = certificateData.certWebViewUrl;
+    downloadUrl = certificateData.downloadUrl;
+  }
 
   let certCase;
   let body;
@@ -66,7 +68,6 @@ function CertificateStatus({ intl }) {
   } else if (mode === COURSE_EXIT_MODES.celebration) {
     switch (certStatus) {
       case 'requesting':
-        // Requestable
         certCase = 'requestable';
         buttonAction = () => { dispatch(requestCert(courseId)); };
         body = intl.formatMessage(messages[`${certCase}Body`]);
--- a/src/course-home/progress-tab/grades/grade-summary/AssignmentTypeCell.jsx
+++ b/src/course-home/progress-tab/grades/grade-summary/AssignmentTypeCell.jsx
@@ -24,7 +24,7 @@ AssignmentTypeCell.propTypes = {
 
 AssignmentTypeCell.defaultProps = {
   footnoteId: '',
-  footnoteMarker: '',
+  footnoteMarker: null,
 };
 
 export default AssignmentTypeCell;
--- a/src/shared/streak-celebration/StreakCelebrationModal.jsx
+++ b/src/shared/streak-celebration/StreakCelebrationModal.jsx
@@ -188,22 +188,24 @@ function StreakModal({
 
 StreakModal.defaultProps = {
   isStreakCelebrationOpen: false,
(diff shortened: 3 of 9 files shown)


MatthewPiatetsky (commented):
could you include a gif or screenshot of the behavior?
also, it seems like this would be redirected once the page has fully loaded
I remember we ran some optimizely tests where we did redirects on the frontend and those had a pretty bad effect on conversions because we were essentially doubling the load time for the page the user was trying to see

normally I'd suggest just redirecting on the backend, but not sure if this can be avoided with MFEs? assuming cloudflare would be too complex for this? i might ask about this in the dev channel
edit: asked in dev channel, i guess there isn't really a simple way to avoid this cost, so redirecting here is fine

src/alerts/enrollment-alert/hooks.js
```diff
@@ -17,6 +18,9 @@ export function useEnrollmentAlert(courseId) {
   const course = useModel('courseHomeMeta', courseId);
   const outline = useModel('outline', courseId);
   const enrolledUser = course && course.isEnrolled !== undefined && course.isEnrolled;
+  if (!enrolledUser && !outline) {
+    global.location.replace(`${getConfig().BASE_URL}/course/${courseId}/home`);
+  }
```
Dillon-Dumesnil: Do you run a risk of an infinite loop here or does going to `/home` result in `outline` always existing? i.e. this is basically saying if you're on any other tab and not enrolled, go to outline? Maybe add a comment here to make that more clear.

Dillon-Dumesnil (approved):
LGTM

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
```diff
@@ -35,16 +38,15 @@ function CertificateStatus({ intl }) {
...
-  const {
-    end,
-    verificationData,
-    certificateData: {
-      certStatus,
-      certWebViewUrl,
-      downloadUrl,
-    },
-    verifiedMode,
-  } = useModel('progress', courseId);
```
Dillon-Dumesnil, line 47: 🤦  Were we just grabbing from this same model twice? Good catch and clean up

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
```diff
@@ -35,16 +38,15 @@ function CertificateStatus({ intl }) {
...
+  let certStatus;
+  let certWebViewUrl;
+  let downloadUrl;
```
Dillon-Dumesnil, line 43: What's the difference between this and what was here before? Wouldn't it have been undefined before if there wasn't certificateData and same now?

src/course-home/progress-tab/grades/grade-summary/AssignmentTypeCell.jsx
```diff
@@ -24,7 +24,7 @@ AssignmentTypeCell.propTypes = {
 
 AssignmentTypeCell.defaultProps = {
   footnoteId: '',
-  footnoteMarker: '',
+  footnoteMarker: null,
```
Dillon-Dumesnil, line 27: Sure 🤷‍♂️

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -188,22 +188,24 @@ function StreakModal({
 
 StreakModal.defaultProps = {
   isStreakCelebrationOpen: false,
+  streakLengthToCelebrate: 0,
```
Dillon-Dumesnil: Is there a chance this could bite us? Like first time visiting and the value is 0 or something? Maybe set it to something impossible like -1?

src/tab-page/LoadedTabPage.jsx
```diff
@@ -45,7 +45,7 @@ function LoadedTabPage({
   return (
     <>
       <Helmet>
-        <title>{`${activeTab.title} | ${title} | ${getConfig().SITE_NAME}`}</title>
+        <title>{`${activeTab ? `${activeTab.title} | ` : ''}${title} | ${getConfig().SITE_NAME}`}</title>
```
Dillon-Dumesnil, line 48: Nice

Dillon-Dumesnil (commented):
Couple of tiny questions

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
ciduarte: certificateData wasn't populated when masquerading as our generic users, so this was throwing an error. It didn't like trying to retrieve certStatus, etc. from a null certificateData

src/course-home/progress-tab/grades/grade-summary/AssignmentTypeCell.jsx
ciduarte: just cleaning up console warnings, the props expect a number and I had it as a string 🤦

Dillon-Dumesnil (approved):

PR #447, opened 2021-05-12
feat: Support classifying certain errors as ignored errors in an MFE.

doctoryes:
Ignored errors are sent to New Relic as page actions instead of JS errors,
allowing those errors to still be tracked as occurring but without causing
unnecessary alerts.

Ignored errors are configured per-MFE, *not* globally.

This added IGNORED_ERROR_REGEXES parameter will be overridden by edx-internal's value.

TNL-7924


.env
```diff
@@ -6,6 +6,7 @@ CSRF_TOKEN_API_PATH=null
 DISCOVERY_API_BASE_URL=null
 ENTERPRISE_LEARNER_PORTAL_HOSTNAME=null
 ECOMMERCE_BASE_URL=null
+IGNORED_ERROR_REGEXES=null
```
robrap: I'm unclear. Will all MFEs require this change, or is this change not actually needed until you want to set this?

doctoryes: I've written the code which reads the value quite defensively - so the absence of the process environment variable shouldn't affect MFEs which don't define it:
https://github.com/edx/frontend-platform/pull/169/files#diff-d74140d8aaffeb831691e7b7c82f403433680e5d35735dc594afeda7a1effbb8R98

robrap: Cool. So you are just adding this for fun?

kdmccormick: > for fun?

Most variables here are in that category; I think it has merit as lite-documentation.

kdmccormick (commented):
.env changes look good. Once your frontend-platform change merges, I think you will need to bump the version pin in package-lock.json here too, right?

doctoryes:
That's a step I'd forgotten, @kdmccormick - thanks. Yes - this PR will merge after the frontend-platform PR because of that version bump.

.env
doctoryes: It was also unclear to me if the variable had to exist in the MFE to be properly overridden by the value in `prod_config.yml` - do either of you know?

doctoryes: I found the answer here:
https://github.com/edx/tubular/blob/33180e8a8859b044bb1e2e761e95baf97e9a7c22/tubular/scripts/frontend_utils.py#L89-L95
The `.env*` files aren't involved at all with the production deploy. But especially since this config variable is per-MFE, it makes sense to add it in this PR for development purposes.

doctoryes:
@kdmccormick @robrap Please re-review - thanks.

robrap (approved):

PR #448, opened 2021-05-14
fix: remove currency symbol to avoid double currency symbol with some upgrade buttons

MatthewPiatetsky:
i was adding a currency symbol but discounted_price and original_price both come with the currency symbol already


julianajlk (approved):
Thank you!

PR #449, opened 2021-05-14
fix: grade range bug

ciduarte:


src/course-home/progress-tab/grades/course-grade/CourseGrade.jsx
```diff
@@ -24,14 +24,7 @@ function CourseGrade({ intl }) {
...
 
-  let passingGrade;
-  if (gradeRange.pass) {
-    passingGrade = gradeRange.pass * 100;
-  } else {
-    passingGrade = Object.entries(gradeRange).pop()[1] * 100;
-  }
-
-  passingGrade = Number(passingGrade.toFixed(0));
+  const passingGrade = Number((Math.min(...Object.values(gradeRange)) * 100).toFixed(0));
```
mikix, line 27: There's really never two grades below `passing`? Like no one does "Incomplete: 10%, F: 50%, D: 60%" or whatever? I wouldn't think so either and presumably this code is perfect, but assumptions like this make me nervous.

src/course-home/progress-tab/grades/course-grade/GradeRangeTooltip.jsx
```diff
@@ -24,7 +24,18 @@ function GradeRangeTooltip({ intl, iconButtonClassName, passingGrade }) {
...
+
+  const orderedGradeRange = [];
+  while (gradeRangeKeys.length > 0) {
+    const maxGradeCutoff = Math.max(...gradeRangeValues);
+    const gradeIndex = gradeRangeValues.indexOf(maxGradeCutoff);
+    const maxLetterGrade = gradeRangeKeys[gradeIndex];
+    orderedGradeRange.push([maxLetterGrade, maxGradeCutoff]);
+    gradeRangeKeys.splice(gradeIndex, 1);
+    gradeRangeValues.splice(gradeIndex, 1);
+  }
```
mikix: The above logic maybe should be split into a method to isolate it and make it a little more obvious what's happening here. 

Answers here had nice and short versions too: https://stackoverflow.com/questions/13694883/converting-an-object-literal-to-a-sorted-array

mikix (approved):

src/course-home/progress-tab/grades/course-grade/CourseGrade.jsx
ciduarte: It doesn't seem like in Studio we allow users to change which letter grade is considered the "passing" grade, it seems like it's always the last value (i.e. anything below the last grade range surfaced)... but I'd also like to get this right, so I can dig a little and push this assumption to see what else I can find

PR #450, opened 2021-05-14
[AA-746] fix: add hour/minute to assignment due dates and update dates timeline with branding styles

MatthewPiatetsky:
Left: Mocks____________________________________________________________Right: local dates tab
Screen Shot 2021-05-17 at 10 48 28 AM Screen Shot 2021-05-17 at 10 53 27 AM

[Figma Link](https://www.figma.com/file/hEwShVPQozMJQw2OB0dM5T/Dates?node-id=2%3A2390)

### DIFF
--- a/src/course-home/dates-tab/Badge.jsx
+++ b/src/course-home/dates-tab/Badge.jsx
@@ -5,7 +5,7 @@ import classNames from 'classnames';
 export default function Badge({ children, className }) {
   return (
     <span
-      className={classNames('dates-badge small ml-2', className)}
+      className={classNames('dates-badge x-small ml-2', className)}
       data-testid="dates-badge"
     >
       {children}
--- a/src/course-home/dates-tab/Badge.jsx
+++ b/src/course-home/dates-tab/Badge.jsx
@@ -2,11 +2,12 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-export default function Badge({ children, className }) {
+export default function Badge({ children, className, ...rest }) {
   return (
     <span
-      className={classNames('dates-badge x-small ml-2', className)}
+      className={classNames('dates-badge x-small ml-2 position-absolute', className)}
       data-testid="dates-badge"
+      {...rest}
     >
       {children}
     </span>
--- a/src/course-home/dates-tab/Day.jsx
+++ b/src/course-home/dates-tab/Day.jsx
@@ -2,7 +2,12 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { useSelector } from 'react-redux';
-import { FormattedDate, injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import {
+  FormattedDate,
+  FormattedTime,
+  injectIntl,
+  intlShape,
+} from '@edx/frontend-platform/i18n';
 import { Tooltip, OverlayTrigger } from '@edx/paragon';
 import { faInfoCircle } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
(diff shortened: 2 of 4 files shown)


src/course-home/dates-tab/Day.jsx
```diff
@@ -70,16 +76,26 @@ function Day({
...
+            <div key={item.title + item.date} className={classNames(textColor, 'small')} data-testid="dates-item">
               <div>
-                <span className="font-weight-bold small mt-1">
-                  {item.assignmentType && `${item.assignmentType}: `}{title}
+                <span className="small">
+                  <span className="font-weight-bold">{item.assignmentType && `${item.assignmentType}: `}{title}</span>
+                  {showDueDateTime && (
+                    <span>
+                      <span className="mx-1">due</span>
+                      <FormattedTime
```
mikix, line 92: You'll want to pass in the user timezone too - search in the MFE for `timezoneFormatArgs` to see other places we do this.

src/course-home/dates-tab/Timeline.jsx
```diff
@@ -66,7 +66,7 @@ export default function Timeline({ mmp2p }) {
   return (
     <ul className="list-unstyled m-0">
       {groupedDates.map((groupedDate) => (
-        <Day key={groupedDate.date} {...groupedDate} mmp2p={mmp2p} />
+        <Day key={groupedDate.date} {...groupedDate} showTime mmp2p={mmp2p} />
```
mikix: Why make it a toggle? I think this is the only place we use `<Day>`. You're thinking we might conditionalize that flag in the future in some cases?

MatthewPiatetsky: yeah i guess we don't need an argument, removed
(will this also show on the dates widget then on course home? not sure which flags i need to enable that or if what i have is already the new version)

src/course-home/dates-tab/Day.jsx
MatthewPiatetsky: thanks, added!

mikix: ah that was easy already there  :P

src/course-home/dates-tab/Timeline.jsx
mikix: I don't believe this code is used on course home. Is this change suppose to show there or not?

MatthewPiatetsky: ah ok, ticket was only referencing the dates tab so should be fine then

mikix (approved):
I'd say drop the `showTime` flag, but I'll approve now to unblock ya.

src/course-home/dates-tab/Badge.jsx
```diff
@@ -2,10 +2,11 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-export default function Badge({ children, className }) {
+export default function Badge({ children, className, style }) {
   return (
     <span
-      className={classNames('dates-badge small ml-2', className)}
+      style={style}
```
mikix: This was fine, but another way to do this would be to pass on the the remaining props as passthroughs to this span, since you don't really care about the `style` value here, you're just acting as go-between. Search for `...rest` or `...passthroughProps` or maybe some other variation for examples.

src/course-home/dates-tab/badgelist.jsx
```diff
@@ -77,6 +77,7 @@ function getBadgeListAndColor(date, intl, item, items) {
     },
   ];
   let color = null; // first color of any badge
+  const marginTopStyle = item ? { marginTop: 0 } : { marginTop: '2px' };
```
mikix, line 80: `mt-1` or `mt-2` weren't suitable here?

mikix (approved):

src/course-home/dates-tab/badgelist.jsx
MatthewPiatetsky: i believe i initially tried mt-1 and it was 4px

src/course-home/dates-tab/Badge.jsx
MatthewPiatetsky: ah good point, will change that

MatthewPiatetsky: switched the style prop to ...rest

PR #455, opened 2021-05-18
AA-814: course grade footer fix

ciduarte:


src/course-home/progress-tab/grades/course-grade/CurrentGradeTooltip.jsx
```diff
@@ -21,7 +21,7 @@ function CurrentGradeTooltip({ intl, tooltipClassName }) {
     },
   } = useModel('progress', courseId);
 
-  const currentGrade = percent * 100;
+  const currentGrade = Number((percent * 100).toFixed(0));
```
Dillon-Dumesnil, line 24: Why the casting to Number? I would have thought the toFixed would be ensuring that?

src/course-home/progress-tab/grades/course-grade/CourseGradeFooter.jsx
```diff
@@ -40,10 +40,9 @@ function CourseGradeFooter({ intl, passingGrade }) {
 
   if (isPassing) {
     if (hasLetterGrades) {
-      const letterGrades = Object.keys(gradeRange);
-      const gradeIndex = letterGrades.indexOf(letterGrade);
       const minGrade = gradeRange[letterGrade] * 100;
-      const maxGrade = gradeIndex > 0 ? gradeRange[letterGrades[gradeIndex - 1]] * 100 : 100;
+      const gradeRangeValues = [...Object.values(gradeRange).filter((grade) => (grade * 100 > minGrade))];
```
Dillon-Dumesnil: Is this filter necessary? If I'm understanding it correctly, it's saying to only give the values if they are > minGrade, but is that ever not the case? I thought they define minGrade by being everything above it.

src/course-home/progress-tab/grades/course-grade/CourseGradeFooter.jsx
```diff
@@ -40,10 +40,9 @@ function CourseGradeFooter({ intl, passingGrade }) {
...
+      const maxGrade = gradeRangeValues.length ? Math.min(...gradeRangeValues) * 100 : 100;
```
Dillon-Dumesnil: Math.min, not Math.max?

Dillon-Dumesnil (commented):
Couple questions, but I think it could all just be my misunderstanding of this PR

src/course-home/progress-tab/grades/course-grade/CourseGradeFooter.jsx
ciduarte: Yeah this is misleading, I can choose better variable names. `minGrade` here isn't the minimum grade needed to pass the course, but the minimum cutoff for the letter grade. So what I'm trying to do here is find the next highest grade, which would be the cutoff for the next letter grade.

For example, an object might look like this:
```
{
  B: .8,
  C: .7,
  A: .9,
}
```

and if a student has a B, we need to display `80-90%`

ciduarte: So in your next comment where you ask if I want Math.min or Math.max, I want Math.min. My intention is to strip anything equal to or less than the learner's letter grade cutoff, and then find the next highest grade cutoff to find the top of the range.

src/course-home/progress-tab/grades/course-grade/CurrentGradeTooltip.jsx
ciduarte: I needed to add the `toFixed`, but `toFixed` casts this to a String. So I cast it back to a Number because it's used in calculations for pixel/em positioning in the SVG.

Dillon-Dumesnil: Of course it does. Sounds good to me. I'll complain separately about how rounding a number converts to a string.

src/course-home/progress-tab/grades/course-grade/CourseGradeFooter.jsx
Dillon-Dumesnil: Ah okay! To check my understanding:
1. we know the learner has a letter grade `C`
2. To get the minimum threshold for that grade, we just grab it from the object `gradeRange['C'] = .7 => 70%`
3. Then to get the max grade, we first filter out all grades below our letter grade C (in this case, leaving B and A).
4. But we only want the one that comes next so we take the minimum out of the remaining values, getting us B with .8 => 80%

That all track?

ciduarte: Yes!

src/course-home/outline-tab/messages.js
```diff
@@ -22,7 +22,7 @@ const messages = defineMessages({
   },
   dates: {
     id: 'learning.outline.dates',
-    defaultMessage: 'Upcoming Dates',
+    defaultMessage: 'Important dates',
```
Dillon-Dumesnil, line 25: TY

src/course-home/progress-tab/grades/course-grade/CourseGradeFooter.jsx
```diff
@@ -40,15 +40,14 @@ function CourseGradeFooter({ intl, passingGrade }) {
 
   if (isPassing) {
     if (hasLetterGrades) {
-      const letterGrades = Object.keys(gradeRange);
-      const gradeIndex = letterGrades.indexOf(letterGrade);
-      const minGrade = gradeRange[letterGrade] * 100;
-      const maxGrade = gradeIndex > 0 ? gradeRange[letterGrades[gradeIndex - 1]] * 100 : 100;
+      const minGradeRangeCutoff = gradeRange[letterGrade] * 100;
+      const gradeRangeValues = [...Object.values(gradeRange).filter((grade) => (grade * 100 > minGradeRangeCutoff))];
```
Dillon-Dumesnil: Nit: `gradeRangeValues` -> `possibleMaxGradeRangeValues`
Thinking it _might_ make it more clear of what you're actually doing with those. and then when you pull out the range cutoff, it is more clear that you're taking the minimum of those possible maxes.
I don't really know though. Whatever you think is best.

Dillon-Dumesnil (approved):
Small nit idea, but otherwise LGTM

PR #461, opened 2021-05-25
AA-813: fix progress tab related links

ciduarte:
We can't use the react-router `<Link />` to switch between tabs because this is a client side navigation. It doesn't actually refresh the page and send a request to the server.

### DIFF
--- a/src/course-home/progress-tab/grades/detailed-grades/DetailedGrades.jsx
+++ b/src/course-home/progress-tab/grades/detailed-grades/DetailedGrades.jsx
@@ -1,8 +1,9 @@
 import React from 'react';
 import { useSelector } from 'react-redux';
-import { Link } from 'react-router-dom';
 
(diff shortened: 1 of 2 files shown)


mikix:
> We can't use the react-router <Link /> to switch between tabs because this is a client side navigation. It doesn't actually refresh the page and send a request to the server.

I don't think that needs to be true? Otherwise one page apps would be in a heap of trouble. I believe redux is designed to be able to refill its data as needed, if driven correctly. That's why we like, look up the course metadata by course key - we were preparing for a world where a single client side navigation session went as far as hitting two courses.

I don't know what we're doing wrong that would not make it work though.

mikix (approved):
As we discussed on slack, this looks right and forget my previous comment. Might be worth a one line comment above these though? Like “we wish these were Links, but waffles…”

PR #463, opened 2021-05-25
AA-806: upgrade deadline passed on progress tab

ciduarte:
- Updated copy on locked preview
- Hid certificate status component
<div>Screen Shot 2021-05-25 at 4 36 54 PM
Screen Shot 2021-05-25 at 4 36 46 PM
</div>

### DIFF
--- a/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
+++ b/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
@@ -134,6 +134,8 @@ function CertificateStatus({ intl }) {
           body = intl.formatMessage(messages[`${certCase}Body`]);
           buttonLocation = verifiedMode.upgradeUrl;
           buttonText = intl.formatMessage(messages[`${certCase}Button`]);
+        } else {
+          certCase = null; // Do not render the certificate component if the upgrade deadline has passed
         }
         break;
 
--- a/src/course-home/progress-tab/grades/messages.js
+++ b/src/course-home/progress-tab/grades/messages.js
@@ -37,9 +37,13 @@ const messages = defineMessages({
     id: 'progress.courseGrade.preview.header.ariaHidden',
     defaultMessage: 'Preview of a ',
   },
-  courseGradePreviewBody: {
-    id: 'progress.courseGrade.preview.body',
-    defaultMessage: 'Unlock to view grades and work towards a certificate',
+  courseGradePreviewUnlockCertificateBody: {
+    id: 'progress.courseGrade.preview.body.unlockCertificate',
+    defaultMessage: 'Unlock to view grades and work towards a certificate.',
+  },
+  courseGradePreviewUpgradeDeadlinePassedBody: {
+    id: 'progress.courseGrade.preview.body.upgradeDeadlinePassed',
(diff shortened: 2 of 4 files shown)


src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
```diff
@@ -134,6 +134,8 @@ function CertificateStatus({ intl }) {
           body = intl.formatMessage(messages[`${certCase}Body`]);
           buttonLocation = verifiedMode.upgradeUrl;
           buttonText = intl.formatMessage(messages[`${certCase}Button`]);
+        } else {
+          certCase = null; // Do not render the certificate component if the upgrade deadline has passed
```
ciduarte, line 138: Thought it was appropriate to determine whether the upgrade deadline has passed based on our `VerifiedModeSerializerMixin`'s response being null: https://github.com/edx/edx-platform/blob/9c5992521abfb30f5286570592d31c5e23e3e4f7/openedx/core/djangoapps/courseware_api/utils.py#L50-L51

https://github.com/edx/edx-platform/blob/9c5992521abfb30f5286570592d31c5e23e3e4f7/lms/djangoapps/courseware/utils.py#L68

let me know if you think this isn't sufficient

MatthewPiatetsky: it looks like the default certCase below is already null so i think this is redundant if this is an else?  
or am I misunderstanding somethin

Dillon-Dumesnil: Yeah, slight subtlety here with how switches work. The default case is hit if none of the other cases match. This `else` is covering when the learner first enters either the `audit_passing` or `honor_passing` case and then we see that the verifiedMode does not exist.

So I think it might be a small difference between certCase = `undefined` vs `null`. Is that correct Carla?

src/course-home/progress-tab/grades/messages.js
```diff
@@ -37,9 +37,13 @@ const messages = defineMessages({
     id: 'progress.courseGrade.preview.header.ariaHidden',
     defaultMessage: 'Preview of a ',
   },
-  courseGradePreviewBody: {
+  courseGradePreviewUnlockBody: {
     id: 'progress.courseGrade.preview.body',
```
Dillon-Dumesnil: Nit: Update this body to be more descriptive like you did with the upgradeDeadlinePassed message. String is changing anyway so might as well update the id too

Dillon-Dumesnil (approved):
Small clarification and nit, but otherwise good.

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
MatthewPiatetsky: ah true i guess the break statement ends the switch rather than going to the default case so yeah it would be undefined vs null
though i believe the end result would still be the same in either case since both are falsey and the component would not show up in either case
however, this is probably fine since marking this explicit case + comment will make the behavior  a little clearer
alternatively we might also be able to omit this section and instantiate certCase to null, then it would be null in both cases rather than undefined vs null

PR #464, opened 2021-05-26
[BD-26] feat: [EDUCATOR-5799] add temporary flag for enabling/disabling proctored exams

ihor-romaniuk:
**Description:** add temporary waffle flag for enabling disabling proctored exams

**Jira:** https://openedx.atlassian.net/browse/EDUCATOR-5799

### DIFF
--- a/src/courseware/course/sequence/Sequence.jsx
+++ b/src/courseware/course/sequence/Sequence.jsx
@@ -48,6 +48,7 @@ function Sequence({
   const unit = useModel('units', unitId);
   const sequenceStatus = useSelector(state => state.courseware.sequenceStatus);
   const specialExamsEnabledWaffleFlag = useSelector(state => state.courseware.specialExamsEnabledWaffleFlag);
+  const proctoredExamsEnabledWaffleFlag = useSelector(state => state.courseware.proctoredExamsEnabledWaffleFlag);
   const shouldDisplaySidebarButton = useWindowSize().width < responsiveBreakpoints.small.minWidth;
 
   const handleNext = () => {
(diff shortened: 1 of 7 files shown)


openedx-webhooks:
Thanks for the pull request, @ihor-romaniuk! I've created [BLENDED-853](https://openedx.atlassian.net/browse/BLENDED-853) to keep track of it in Jira.   More details are on the [BD-26](https://openedx.atlassian.net/wiki/spaces/COMM/pages/2523627980/BD-26+Proctoring+MFE+and+configuration) project page.


src/courseware/course/sequence/Sequence.jsx
```diff
@@ -140,12 +141,18 @@ function Sequence({
...
   we're waiting. That redirect may take a few seconds, so we show the spinner in the meantime.
   */
-  if (sequenceStatus === 'loaded' && sequence.isTimeLimited && !specialExamsEnabledWaffleFlag) {
-    return (
-      <PageLoading
-        srMessage={intl.formatMessage(messages['learn.loading.learning.sequence'])}
-      />
-    );
+  if (sequenceStatus === 'loaded') {
+    const isDisabledSpecialExams = sequence.isTimeLimited && !specialExamsEnabledWaffleFlag;
```
zacharis278: nit: can we name this something like `shouldRedirectSpecialExams` since this boolean is really about if we should redirect this specific exam instead of "is the feature disabled". Just a bit easier to read

UvgenGen: Renamed

zacharis278 (approved):
we'll merge right after the edx-platform PR. Looks like that one got hung up on a frozen CI job. I just kicked that off so it runs again

zacharis278 (approved):

PR #465, opened 2021-05-26
[MST-783] Add honor code component

bseverino:
[MST-783](https://openedx.atlassian.net/browse/MST-783)

This component blocks access to graded units when the user is required to sign the integrity agreement for the course (currently behind a waffle flag via https://github.com/edx/edx-platform/pull/27669). Once signed, it will not appear for the course again.

Screen Shot 2021-05-25 at 3 45 56 PM


schenedx (approved):
👍 LGTM

PR #466, opened 2021-05-27
fix: gated content banner discount display

julianajlk:
[REV-2232](https://openedx.atlassian.net/browse/REV-2232).

Bug fix for enabling the display of a discount in the upgrade button in the gated content banner. 
This is part of the Value Prop work.

Screen Shot 2021-05-28 at 12 35 26 PM
Thanks @dianekaplan for checking out this locally and for the screenshot :)

**Note on Codecov:**
The -0.01% decrease in Codecov looks like it's a fraction issue (covered lines ÷ total lines) rather than a "no coverage" for the added lines issue. I've discussed it with Purchase Squad, and will go ahead and ignore the failing Codecov report.

### DIFF
--- a/src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
+++ b/src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
@@ -155,7 +155,6 @@ function LockPaywall({
             offer={offer}
             onClick={logClick}
             verifiedMode={verifiedMode}
-            className="lock_paywall_upgrade_link"
           />
         </div>
       </div>
--- a/src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
+++ b/src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
@@ -5,12 +5,13 @@ import { faCheck } from '@fortawesome/free-solid-svg-icons';
 import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 import { FormattedMessage, injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import {
-  Alert, Button, Icon,
+  Alert, Icon,
 } from '@edx/paragon';
 import { Locked } from '@edx/paragon/icons';
 import messages from './messages';
 import certificateLocked from '../../../../generic/assets/edX_locked_certificate.png';
 import { useModel } from '../../../../generic/model-store';
+import { UpgradeButton } from '../../../../generic/upgrade-button';
 import './LockPaywall.scss';
 
 function LockPaywall({
@@ -19,6 +20,7 @@ function LockPaywall({
 }) {
   const course = useModel('coursewareMeta', courseId);
   const {
+    offer,
     org,
     verifiedMode,
   } = course;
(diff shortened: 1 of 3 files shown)


src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
```diff
@@ -154,17 +151,11 @@ function LockPaywall({
           className="col-md-auto p-md-0 d-md-flex align-items-md-center mr-md-3"
           style={{ textAlign: 'right' }}
         >
-          <Button
-            className="lock_paywall_upgrade_link"
-            href={upgradeUrl}
+          <UpgradeButton
```
dianekaplan, line 154: ooh nice- I didn't know this UpgradeButton existed! I only knew about the paragon Button.  Before using a paragon item, should we be in the habit to check if they've been extended into components? 

For this ticket the natural starting point is to see where it's working- (UpgradeCard, which uses UpgradeButton)- but trying to think about how to find it the first time around...

src/courseware/course/sequence/lock-paywall/messages.js
```diff
@@ -11,11 +11,6 @@ const messages = defineMessages({
     defaultMessage: 'Upgrade to gain access to locked features like this one and get the most out of your course.',
     description: 'Message shown to indicate that a piece of content is unavailable to audit track users.',
   },
-  'learn.lockPaywall.upgrade.link': {
-    id: 'learn.lockPaywall.upgrade.link',
-    defaultMessage: 'Upgrade for {currencySymbol}{price}',
-    description: 'A link users can click that navigates their browser to the upgrade payment page.',
-  },
   'learn.lockPaywall.example.alt': {
```
dianekaplan, line 14: Looks cleaner without this- is the translation now covered in UpgradeButton?

dianekaplan (approved):
Looks good to me- just a couple curiosity questions!

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
julianajlk: That's a good question - I found it because I had seen it in the `UpgradeCard` component code, so that doesn't help your question... In this case  you're correct, `<UpgradeButton>` is using paragon's `Button` but it's a component within learning MFE and not a paragon one. So I guess searching for `<Button` could give you some clues.

src/courseware/course/sequence/lock-paywall/messages.js
julianajlk: Correct 👍 By doing using one component, it also removes the need to do translations in all the different places that it uses this exact language, so that's handy.

src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
dianekaplan: Cool- it's a good thing you did! It looks like UpgradeCard and LockPaywall had a different button all along and we just modified the contents around them, but now they can be consistent!

PR #471, opened 2021-06-02
[AA-759] feat: update streak discount coupon expiration date

MatthewPiatetsky:
This was a placeholder date that is now more accurately updated relative to when we think the experiment will run


Dillon-Dumesnil (approved):
Assuming this was cleared via someone, but yeah, works for me.

PR #477, opened 2021-06-07
feature: Improve the Certificate Alerts in the outline to support new statuses

staubina:
feature: Improve the Certificate Alerts in the outline to support new statuses

[MICROBA-678]

These changes refactor the CertificateAvailableAlert and add new
features to it to support more status alerts for certificates. It
attempts to do so in an iterative manner so that new/updated alerts can
be included over time.

**Notes**
* This PR focuses on the refactor and addition of View My Certificate status for web certificates only right now. In a follow on PR we will add/update the other states.
* I included screenshots of the Cert not available status to demonstrate that it still works.

Screen Shot 2021-06-08 at 12 07 31 PM
Screen Shot 2021-06-08 at 12 07 57 PM
Screen Shot 2021-06-01 at 7 40 05 AM


src/course-home/outline-tab/alerts/certificate-available-alert/CertificateStatusAlert.jsx
```diff
@@ -0,0 +1,94 @@
+import React from 'react';
```
staubina: Note: I renamed the time and as a result, it looks like it is all new. Much of this code is inherited from the original alert.

src/course-home/outline-tab/alerts/certificate-available-alert/CertificateStatusAlert.jsx
```diff
@@ -0,0 +1,94 @@
...
+    )
+  } else if (certStatusType === CERT_STATUS_TYPE.DOWNLOADABLE) {
+    header = intl.formatMessage(certMessages.certStatusDownloadableHeader);
+    buttonMessage = intl.formatMessage(certStatusMessages.viewableButton);
+    buttonVisible = true;
+  }
+
+
+  return (
+    <Alert variant={ variant }>
```
staubina: Note: I switched to using a Paragon alert here.

Tj-Tracy (approved):

PR #478, opened 2021-06-07
fix: Scroll page when html anchor tag clicked

KristinAoki:
Jira issue: [TNL-8312](https://openedx.atlassian.net/browse/TNL-8312)

This PR changes the default function of HTML anchor tags that focus on another element on the page. Previously the anchor tags that were set to scroll on the page to another element would open the link outside of the iframe and redirect the parent page. As a result, users would have to have to click the back arrow to navigate back to the course and continue the unit. Now when the anchor tag is used to focus it and send the location of the focus element to the parent page to scroll to the correct location. This change will impact the Learner.

This change is connected to a change in the edx-platform repo, [PR 27872](https://github.com/edx/edx-platform/pull/27872)


src/courseware/course/sequence/Unit.jsx
```diff
@@ -118,6 +118,8 @@ function Unit({
       } else if (type === 'plugin.modal') {
         payload.open = true;
         setModalOptions(payload);
+      } else if (event.data.offset) {
```
kdmccormick, line 121: Please add a comment explaining this block a bit. If I weren't aware of the corresponding change you made in LMS, I might have a very hard time tracking down why this event is being handed at all.

kdmccormick (changes requested):
Code looks great and works great 👍🏻 Just one request for an additional comment.

kdmccormick (approved):
Looks great! I'll send you some info on how to deploy this. Unlike edx-platform, it's a more manual process.

PR #479, opened 2021-06-09
fix: pin special exams library

zacharis278:
this library is in development, we should not blindly update versions. Removing the carrot so we don't accidentally pull in code with a bug 😬


schenedx (approved):
👍

schenedx:
There is a plan to move the carrot back sometime later, right?

zacharis278:
actually, I think with the renovate bot handling updates we shouldn't ever need it

PR #480, opened 2021-06-09
fix: updating progress tab to better respect showGrades field

ciduarte:
A bit of a refactor to make it so the learner's current grade respects the `showGrades` field (aka the `show_completeness` Studio setting which controls whether the learner sees their grades).

### DIFF
--- a/src/course-home/data/__snapshots__/redux.test.js.snap
+++ b/src/course-home/data/__snapshots__/redux.test.js.snap
@@ -542,6 +542,7 @@ Object {
           "isPassing": true,
           "letterGrade": "pass",
           "percent": 1,
+          "visiblePercent": 1,
         },
         "courseId": "course-v1:edX+DemoX+Demo_Course_1",
         "end": "3027-03-31T00:00:00Z",
@@ -549,11 +550,12 @@ Object {
         "gradingPolicy": Object {
           "assignmentPolicies": Array [
             Object {
+              "averageGrade": 1,
               "numDroppable": 1,
-              "numTotal": 2,
               "shortLabel": "HW",
               "type": "Homework",
               "weight": 1,
+              "weightedGrade": 1,
             },
           ],
           "gradeRange": Object {
--- a/src/course-home/data/api.js
+++ b/src/course-home/data/api.js
@@ -3,6 +3,81 @@ import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 import { logInfo } from '@edx/frontend-platform/logging';
 import { appendBrowserTimezoneToUrl } from '../../utils';
 
+const calculateAssignmentTypeGrades = (points, assignmentWeight, numDroppable) => {
+  let dropCount = numDroppable;
+  // Drop the lowest grades
+  while (dropCount && points.length >= dropCount) {
+    const lowestScore = Math.min(...points);
+    const lowestScoreIndex = points.indexOf(lowestScore);
+    points.splice(lowestScoreIndex, 1);
+    dropCount--;
+  }
+  let averageGrade = 0;
+  let weightedGrade = 0;
+  if (points.length) {
+    averageGrade = points.reduce((a, b) => a + b, 0) / points.length;
+    weightedGrade = averageGrade * assignmentWeight;
+  }
+  return { averageGrade, weightedGrade };
+};
+
+function normalizeAssignmentPolicies(assignmentPolicies, sectionScores) {
+  const gradeByAssignmentType = {};
+  assignmentPolicies.forEach(assignment => {
+    // Create an array with the number of total assignments and set the scores to 0
+    // as placeholders for assignments that have not yet been released
+    gradeByAssignmentType[assignment.type] = {
+      grades: Array(assignment.numTotal).fill(0),
+      numAssignmentsCreated: 0,
+      numTotalExpectedAssignments: assignment.numTotal,
+    };
+  });
+
+  sectionScores.forEach((chapter) => {
+    chapter.subsections.forEach((subsection) => {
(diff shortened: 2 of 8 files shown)


src/course-home/data/api.js
```diff
@@ -3,6 +3,80 @@ import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
...
+      gradeByAssignmentType[assignmentType].numAssignmentsCreated = numAssignmentsCreated;
+    });
+  });
+
+  return assignmentPolicies.map((assignment) => {
+    const { averageGrade, weightedGrade } = calculateAssignmentTypeGrades(
+      gradeByAssignmentType[assignment.type].grades,
+      assignment.weight,
+      assignment.numDroppable,
+    );
```
ciduarte, line 68: This logic hasn't changed (except for line 38 where I've added `&& subsection.showGrades`), it's just been moved from `<GradeSummary />` and `<GradeSummaryTable />`

MatthewPiatetsky (approved):
not sure if you think a test of this working is necessary?

src/course-home/data/api.js
Dillon-Dumesnil: Good call out. Thank you

src/course-home/data/__snapshots__/redux.test.js.snap
```diff
@@ -542,18 +542,19 @@ Object {
...
         "courseId": "course-v1:edX+DemoX+Demo_Course_1",
         "end": "3027-03-31T00:00:00Z",
         "enrollmentMode": "audit",
         "gradingPolicy": Object {
           "assignmentPolicies": Array [
             Object {
+              "averageGrade": 1,
               "numDroppable": 1,
-              "numTotal": 2,
-              "shortLabel": "HW",
```
Dillon-Dumesnil: ...? (read: why did these two things disappear? Do we not need them anymore? Did we never?)

src/course-home/data/api.js
```diff
@@ -132,7 +206,24 @@ export async function getProgressTabData(courseId) {
   try {
     const { data } = await getAuthenticatedHttpClient().get(url);
     const camelCasedData = camelCaseObject(data);
+
+    const assignmentPolicies = normalizeAssignmentPolicies(
+      camelCasedData.gradingPolicy.assignmentPolicies,
+      camelCasedData.sectionScores,
+    );
+    camelCasedData.gradingPolicy.assignmentPolicies = assignmentPolicies;
```
Dillon-Dumesnil: Do you need to set it separately? Why not just have `camelCasedData.gradingPolicy.assignmentPolicies = normalizeAssignmentPolicies(`?

src/course-home/data/api.js
```diff
@@ -132,7 +206,24 @@ export async function getProgressTabData(courseId) {
...
+
+    // Accumulate the weighted grades by assignment type to calculate the learner facing grade. The grades within
+    // assignmentPolicies have been filtered by what's visible to the learner.
+    camelCasedData.courseGrade.visiblePercent = assignmentPolicies.reduce(
+      (accumulator, assignment) => accumulator + assignment.weightedGrade, 0,
+    );
+
+    // We replace gradingPolicy.gradeRange with the original data to preserve the intended casing for the grade.
+    // For example, if a grade range key is "A", we do not want it to be camel cased (i.e. "A" would become "a")
+    // in order to preserve a course team's desired grade formatting.
```
Dillon-Dumesnil, line 228: Great comment

src/course-home/progress-tab/grades/grade-summary/GradeSummaryTable.jsx
```diff
@@ -27,24 +24,6 @@ function GradeSummaryTable({
 
   const footnotes = [];
 
-  const calculateAssignmentTypeGrades = (points, assignmentWeight, numDroppable) => {
```
Dillon-Dumesnil, line 30: This is the stuff that's all handled in the api.js file now, right? Just to get my mental ducks in a row.

Dillon-Dumesnil (approved):
Couple small questions, mostly clarifying, but looks great. Digging the comments in this PR/code.

Oh and +1 to Matthew about there likely being some value in a test that shows how certain assignments won't render based on this change as opposed to normal.

src/course-home/progress-tab/grades/grade-summary/GradeSummaryTable.jsx
ciduarte: Yes it is

src/course-home/data/__snapshots__/redux.test.js.snap
ciduarte: Oops we need the short label, nice catch.

We needed `numTotal` to do all of that grade normalizing in `api.js`, but the rest of our components don't need to access that field, so I didn't include it in that `normalizeAssignmentPolicies` return (which is why it disappears from this snapshot).

src/course-home/data/api.js
ciduarte: I set `assignmentPolicies` here because I use it again below on line 218

ciduarte (commented):
Yeah I agree w/ adding a test. I can mock a staff vs a normal learner with `showGrades` as `false` to show how these components are meant to change.

edit: mocking the staff version of this doesn't actually capture anything new because `showGrades` would just be `true`

src/course-home/data/api.js
Dillon-Dumesnil: Could you not use the `camelCasedData.gradingPolicy.assignmentPolicies` in place?

This is all nit in case that wasn't clear.

ciduarte: Yeah we could, just felt lengthy. I'm fine with switching it 👍

PR #483, opened 2021-06-11
fix: update assignment policy logic

ciduarte:
[AA-836](https://openedx.atlassian.net/browse/AA-836)

### DIFF
--- a/src/course-home/data/api.js
+++ b/src/course-home/data/api.js
@@ -35,7 +35,7 @@ function normalizeAssignmentPolicies(assignmentPolicies, sectionScores) {
 
   sectionScores.forEach((chapter) => {
     chapter.subsections.forEach((subsection) => {
-      if (!(subsection.hasGradedAssignment && subsection.showGrades)) {
+      if (!(subsection.hasGradedAssignment && subsection.showGrades && subsection.numPointsPossible)) {
         return;
       }
       const {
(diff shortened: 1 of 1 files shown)


src/course-home/data/api.js
```diff
@@ -35,22 +35,27 @@ function normalizeAssignmentPolicies(assignmentPolicies, sectionScores) {
...
       const {
         assignmentType,
         numPointsEarned,
         numPointsPossible,
       } = subsection;
-      let {
-        numAssignmentsCreated,
-      } = gradeByAssignmentType[assignmentType];
 
-      if (!numPointsPossible) {
```
ciduarte, line 50: moved this as a nit of mine

src/course-home/data/api.js
```diff
@@ -35,22 +35,27 @@ function normalizeAssignmentPolicies(assignmentPolicies, sectionScores) {
...
+      // If a subsection's assignment type does not match an assignment policy in Studio,
+      // we won't be able to include it in this accumulation of grades by assignment type.
+      // This may happen if a course author has removed/renamed an assignment policy in Studio and
+      // neglected to update the subsection's of that assignment type
```
ciduarte, line 50: Not sure if this comment is overkill. This behavior wasn't obvious to me so it felt worth documenting

src/course-home/data/api.js
```diff
@@ -35,22 +35,27 @@ function normalizeAssignmentPolicies(assignmentPolicies, sectionScores) {
...
+      if (!gradeByAssignmentType[assignmentType]) {
```
MatthewPiatetsky, line 51: couldn't you do the below statement and then check if (!numAssignmentsCreated) to avoid using the same code twice?

MatthewPiatetsky (approved):

src/course-home/data/api.js
ciduarte: The bug that was happening in prod was due to trying to retrieve `numAssignmentsCreated` from null, so we need to check if `gradeByAssignmentType[assignmentType]` exists first.

PR #486, opened 2021-06-14
feat: make course recommendations a part of the course celebration

rgraber:
mobile:
localhost_2000_course_course-v1_edX+DemoX+Demo_Course_course-end

desktop:
Screen Shot 2021-06-17 at 3 50 18 PM


src/courseware/course/course-exit/CatalogSuggestion.jsx
```diff
@@ -33,7 +33,7 @@ function CatalogSuggestion({ intl, variant }) {
   );
 
   return (
-    <div className="row w-100 mx-0 my-2 justify-content-center">
+    <div className="row w-100 mx-0 my-2 justify-content-center" data-testid="catalog-suggestion">
```
Vyraal1, line 36: Why the custom HTML5 attribute? Is it something specific to how this repo handles passing props?

src/courseware/data/__factories__/index.js
```diff
@@ -1,2 +1,3 @@
 import './courseMetadata.factory';
 import './sequenceMetadata.factory';
+import './courseRecommendations.factory';
```
Vyraal1, line 3: Where do the factories get consumed/used after being defined and imported? Was curious to see what happens to them

src/courseware/course/course-exit/CatalogSuggestion.jsx
rgraber: it's specifically for testing with rosie.js. It's a pattern I took from other places in the code

src/courseware/data/__factories__/index.js
rgraber: Defining a factory "registers" it with the rosie.js framework. Then we can use it to make test objects for unit tests.

Vyraal1 (approved):

srwang (approved):

PR #487, opened 2021-06-14
fix: Autoscroll on page when using jump_to_id

KristinAoki:
This PR changes the default function of HTML anchor tags that focus on another element on a different page in the course. Previously the anchor tags that were set to scroll on a different page would open up the page but remain at the top of the page. As a result, users would have to manually scroll to the correct part of the page if that information is known or they will read the entire page. Now when the anchor tag is used to focus on a different page, the link of the new page is checked for a hash and sends the location of the hash to the parent page to scroll to the correct location. This change will impact the Learner in the New Experience view.

This PR is connected to [PR-27952](https://github.com/edx/edx-platform/pull/27952) in the edx-platform repo.

### DIFF
--- a/src/courseware/course/sequence/Unit.jsx
+++ b/src/courseware/course/sequence/Unit.jsx
@@ -99,6 +99,13 @@ function Unit({
 
   const dispatch = useDispatch();
 
+  const checkForHash = () => {
+    const { hash } = window.location;
+    if (hash) {
+      const unitIframe = document.getElementById('unit-iframe');
+      unitIframe.contentWindow.postMessage({ hashName: hash }, `${getConfig().LMS_BASE_URL}`);
+    }
+  };
   // Do not remove this hook.  See function description.
   useLoadBearingHook(id);
 
@@ -215,6 +222,7 @@ function Unit({
             scrolling="no"
             referrerPolicy="origin"
             onLoad={() => {
+              checkForHash();
               window.onmessage = function handleResetDates(e) {
                 if (e.data.event_name) {
                   dispatch(processEvent(e.data, fetchCourse));
--- a/src/courseware/course/sequence/Unit.jsx
+++ b/src/courseware/course/sequence/Unit.jsx
@@ -221,8 +221,7 @@ function Unit({
             scrolling="no"
             referrerPolicy="origin"
             onLoad={() => {
-              const unitIframe = document.getElementById('unit-iframe');
-              checkForHash(unitIframe);
+              checkForHash(document.getElementById('unit-iframe'));
               window.onmessage = function handleResetDates(e) {
                 if (e.data.event_name) {
                   dispatch(processEvent(e.data, fetchCourse));
--- a/src/courseware/course/sequence/Unit.jsx
+++ b/src/courseware/course/sequence/Unit.jsx
@@ -70,9 +70,10 @@ function useLoadBearingHook(id) {
   }, [id]);
 }
 
-export function checkForHash(frame) {
+export function sendHash(frame) {
   const { hash } = window.location;
   if (hash) {
+    // The hash will be sent to LMS so that the location of the hash can be found.
     frame.contentWindow.postMessage({ hashName: hash }, `${getConfig().LMS_BASE_URL}`);
   }
 }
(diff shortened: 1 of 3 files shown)


src/courseware/course/sequence/Unit.jsx
```diff
@@ -70,6 +70,13 @@ function useLoadBearingHook(id) {
   }, [id]);
 }
 
+export function checkForHash(frame) {
+  const { hash } = window.location;
+  if (hash) {
+    frame.contentWindow.postMessage({ hashName: hash }, `${getConfig().LMS_BASE_URL}`);
```
kdmccormick, line 78: What will the frame do with this `hashName` message?

(I know what it'll do, but that's only because I saw your edx-platform PR :)

src/courseware/course/sequence/Unit.jsx
```diff
@@ -70,6 +70,13 @@ function useLoadBearingHook(id) {
   }, [id]);
 }
 
+export function checkForHash(frame) {
```
kdmccormick: I like that you named this function in an imperative way (ie, "check for hash" instead of "hash check") 👍🏻 

This function isn't just "checking", though, is it? Consider a name that makes it clearer that this function has an noticeable side-effect -- sending a message.

src/courseware/course/sequence/Unit.jsx
```diff
@@ -98,13 +105,13 @@ function Unit({
...
 
   const dispatch = useDispatch();
-
   // Do not remove this hook.  See function description.
   useLoadBearingHook(id);
 
   // We use this ref so that we can hold a reference to the currently active event listener.
   const messageEventListenerRef = useRef(null);
   useEffect(() => {
+    checkForHash(document.getElementById('unit-iframe'));
```
kdmccormick: Is there a reason you chose to call your function here in particular? (genuine question -- I'm fuzzy on the details of React Hooks and exactly where to put React code with side effects)

kdmccormick (commented):
Some comments on the implementation specifics. I tried it out on my devstack and it works great!

src/courseware/course/sequence/Unit.jsx
KristinAoki: `useEffect` is called only once when the page is rendered. This helps prevent infinite checks and renders when messages are constantly being sent. I had `checkForHash` orginially in the `onLoad` of the iframe, but could not access it from testing purposes so I moved it to the useEffect because it is the only other place that would wait for all the info to be loaded. It actually works out better here because the function is not dependent on the iframe loading, but instead the MFE page and getting the hash from that URL.

src/courseware/course/sequence/Unit.jsx
```diff
@@ -70,6 +70,14 @@ function useLoadBearingHook(id) {
   }, [id]);
 }
 
+export function sendHash(frame) {
```
kdmccormick: `send` is definitely a better verb, but consider that `hash` on its own is a super ambiguous word, and the function name doesn't make it clear where we sending said hash. Try to name functions such that someone wouldn't need to look at the function's definition in order to have a good idea of what the function does.

How about `sendUrlHashToFrame` or `sendUrlFragmentToFrame`?

src/courseware/course/sequence/Unit.jsx
```diff
@@ -70,6 +70,14 @@ function useLoadBearingHook(id) {
...
+  const { hash } = window.location;
+  if (hash) {
+    // The hash will be sent to LMS so that the location of the hash can be found.
```
kdmccormick: > The hash will be sent to LMS

I know that you mean the LMS-served iframe, but someone could interpret this as meaning "send via HTTP to LMS, the service".

> so that the location of the hash can be found.

Can you be more specific?

kdmccormick (commented):
More nitpicking on your function name and comment, sorry 😛

src/courseware/course/sequence/Unit.jsx
kdmccormick: Makes sense to me 👍🏻

kdmccormick (approved):
Send it!

PR #488, opened 2021-06-15
fix: remove sequences we shouldn't see by using learning_sequences

ormsbee:
Removes sequences we shouldn't see by using the Learning Sequences API (TNL-8377).

It works by adding a call to the Learning Sequences API and (if that endpoint is enabled, i.e. returns 200 for this user+course), uses the results of that endpoint to remove sequences from the Course Blocks API call. Learning Sequences knows how to do things like bubble up the content group settings of units to sequences for the case where all units have the same restrictions and the user would see an empty sequence.

Depends on https://github.com/edx/frontend-app-learning/pull/488#:~:text=edx/edx-platform%2327955

### DIFF
--- a/src/courseware/data/api.js
+++ b/src/courseware/data/api.js
@@ -10,6 +10,7 @@ export function normalizeBlocks(courseId, blocks) {
     sequences: {},
     units: {},
   };
+
   Object.values(blocks).forEach(block => {
     switch (block.type) {
       case 'course':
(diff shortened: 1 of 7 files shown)


ormsbee:
This depends on https://github.com/edx/edx-platform/pull/27955

ormsbee:
@davidjoy: Made the changes you suggested yesterday. May I trouble you for a review of that refactoring?

src/courseware/data/api.js
```diff
@@ -105,6 +106,29 @@ export async function getCourseBlocks(courseId) {
...
+  } catch (error) {
+    // This is not a critical API to use at the moment. If it errors for any
+    // reason, just send back a null so the higher layers know to ignore it.
+    if (error.response) {
+      if (error.response.status === 403) {
+        logInfo('Learning Sequences API not enabled for this user.');
+      } else {
+        logInfo(`Unexpected error calling Learning Sequences API (${error.response.status}). Ignoring.`);
+      }
+      return null;
```
davidjoy, line 126: So any error with the response that isn't a 403 will be swallowed?  That intentional?

davidjoy (approved):

src/courseware/data/api.js
ormsbee: Yeah. I'll be monitoring that on the server side. The overriding goal of this PR is to do no harm to the existing experience.

PR #489, opened 2021-06-15
fix: guard against courseGoals being undefined

mikix:
No idea why this would happen honestly - it looks always defined from the backend, and api.js doesn't transorm it. But we are seeing JS errors related to it. So trying this as a first pass.

https://openedx.atlassian.net/browse/AA-848

### DIFF
--- a/src/course-home/outline-tab/OutlineTab.jsx
+++ b/src/course-home/outline-tab/OutlineTab.jsx
@@ -52,7 +52,7 @@ function OutlineTab({ intl }) {
     courseGoals: {
       goalOptions,
       selectedGoal,
-    },
+    } = {},
     datesBannerInfo,
     datesWidget: {
       courseDateBlocks,


ciduarte (approved):
LGTM. Yeah I'm not sure why this was happening either because it should be populated 😕

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -52,7 +52,7 @@ function OutlineTab({ intl }) {
     courseGoals: {
       goalOptions,
       selectedGoal,
-    },
+    } = {},
```
MatthewPiatetsky, line 55: is there an explanation of this syntax somewhere? (i hadn't seen it yet)

mikix: It's just how you specify a default value when destructuring:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#default_values_2

PR #490, opened 2021-06-15

bmedx:

### DIFF
--- a/src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
+++ b/src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
@@ -1,31 +1,15 @@
-import React, { useState } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
-import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 import {
   FormattedMessage, FormattedDate, injectIntl, intlShape,
 } from '@edx/frontend-platform/i18n';
-import { Hyperlink } from '@edx/paragon';
 
 import { Alert, ALERT_TYPES } from '../../generic/user-messages';
-import messages from './messages';
-import AccessExpirationAlertMMP2P from './AccessExpirationAlertMMP2P';
 
-function AccessExpirationAlert({ intl, payload }) {
-  /** [MM-P2P] Experiment */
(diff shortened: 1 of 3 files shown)


src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
```diff
@@ -1,31 +1,15 @@
...
   FormattedMessage, FormattedDate, injectIntl, intlShape,
 } from '@edx/frontend-platform/i18n';
-import { Hyperlink } from '@edx/paragon';
 
 import { Alert, ALERT_TYPES } from '../../generic/user-messages';
-import messages from './messages';
-import AccessExpirationAlertMMP2P from './AccessExpirationAlertMMP2P';
-
-function AccessExpirationAlert({ intl, payload }) {
-  /** [MM-P2P] Experiment */
```
julianajlk, line 14: Could this (and other `MM-P2P` code) perhaps be in a separate commit since it's clean up work from another team? There is also a lot more code related to this experiment that are not included in this PR, if it's too much I would recommend the original team to do the clean up so we don't end up with a partial clean up.

src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
```diff
@@ -63,98 +45,7 @@ function AccessExpirationAlert({ intl, payload }) {
...
-              value={expirationDate}
-              {...timezoneFormatArgs}
-            />
-          ),
-        }}
-      />
-      {deadlineMessage}
-    </Alert>
-  );
+  return null;
```
julianajlk, line 48: I see that for this component, the main use case would no longer be used. Instead of returning `null`, I would reformat the code so that it would return some JSX, likely what `masqueradingExpiredCourse` is returning. It looks like this component is used in courseware as well, so make sure it would still work for that use case (I'm assuming there are tests for that?).

src/courseware/course/Course.test.jsx
```diff
@@ -112,60 +112,6 @@ describe('Course', () => {
     expect(sidebarOpenButton).toBeInTheDocument();
   });
 
-  it('displays offer and expiration alert', async () => {
```
julianajlk, line 115: Are these tests for courseware and not for Course Home?

PR #493, opened 2021-06-16

bmedx:

### DIFF
--- a/src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
+++ b/src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
@@ -9,6 +9,7 @@ import { Hyperlink } from '@edx/paragon';
 import { Alert, ALERT_TYPES } from '../../generic/user-messages';
 import messages from './messages';
 import AccessExpirationAlertMMP2P from './AccessExpirationAlertMMP2P';
+import AccessExpirationAlertMasquerade from './AccessExpirationAlertMasquerade';
 
 function AccessExpirationAlert({ intl, payload }) {
   /** [MM-P2P] Experiment */
@@ -42,24 +43,7 @@ function AccessExpirationAlert({ intl, payload }) {
 
   if (masqueradingExpiredCourse) {
     return (
-      <Alert type={ALERT_TYPES.INFO}>
-        <FormattedMessage
-          id="learning.accessExpiration.expired"
-          defaultMessage="This learner does not have access to this course. Their access expired on {date}."
-          values={{
-            date: (
-              <FormattedDate
-                key="accessExpirationExpiredDate"
-                day="numeric"
-                month="short"
-                year="numeric"
-                value={expirationDate}
-                {...timezoneFormatArgs}
-              />
-            ),
-          }}
-        />
-      </Alert>
+      <AccessExpirationAlertMasquerade payload={payload} />
     );
   }
 
--- a/src/alerts/access-expiration-alert/AccessExpirationAlertMasquerade.jsx
+++ b/src/alerts/access-expiration-alert/AccessExpirationAlertMasquerade.jsx
@@ -1,0 +1,62 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage, FormattedDate } from '@edx/frontend-platform/i18n';
+
+import { Alert, ALERT_TYPES } from '../../generic/user-messages';
+
+function AccessExpirationAlertMasquerade({ payload }) {
+  const {
+    accessExpiration,
+    userTimezone,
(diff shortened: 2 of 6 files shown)


src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
```diff
@@ -42,24 +43,7 @@ function AccessExpirationAlert({ intl, payload }) {
...
-                value={expirationDate}
-                {...timezoneFormatArgs}
-              />
-            ),
-          }}
-        />
-      </Alert>
+      <AccessExpirationAlertMasquerade payload={payload} />
     );
   }
```
julianajlk, line 49: Is L50 and beyond still here? The return of the `Alert` component that would be removed in your previous PR https://github.com/edx/frontend-app-learning/pull/490/

src/alerts/access-expiration-alert/AccessExpirationAlertMasquerade.jsx
```diff
@@ -0,0 +1,62 @@
...
+  if (!accessExpiration) {
+    return null;
+  }
+
+  const {
+    expirationDate,
+    masqueradingExpiredCourse,
+  } = accessExpiration;
+
+  if (!masqueradingExpiredCourse) {
```
julianajlk: This means if `masqueradingExpiredCourse` is `false` or `undefined` it will return nothing, right? I wonder if this is necessary? Another option is to have that logic in `AccessExpirationAlert` component and only render `AccessExpirationAlertMasquerade` if it's `true` (so no need to return `null`).

src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
julianajlk: Asking bc I'm seeing collapsed but cannot click on the `+` to commend on the line

bmedx: Yes, the rest of the file is still hiding under here. The only difference in this file is to move the Alert code out to the reusable AccessExpirationAlertMasquerade. You should be able to go to "..." on this diff and "view file" to see the whole thing.

src/alerts/access-expiration-alert/AccessExpirationAlertMasquerade.jsx
bmedx: It's also used directly in OutlineTab, so it seemed like a better idea to consolidate the check in one place, but I guess that check is handled in the hook? If so I totally agree and can change this.

bmedx: I've removed the duplicate checks

src/alerts/access-expiration-alert/AccessExpirationAlertMasquerade.jsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+AccessExpirationAlertMasquerade.propTypes = {
+  payload: PropTypes.shape({
+    accessExpiration: PropTypes.shape({
+      expirationDate: PropTypes.string.isRequired,
+      masqueradingExpiredCourse: PropTypes.bool.isRequired,
+      upgradeDeadline: PropTypes.string,
+      upgradeUrl: PropTypes.string,
+    }).isRequired,
```
dianekaplan, line 46: These upgradeDeadline and upgradeUrl props were used by the old AccessExpirationAlert for the upsell links... are they still needed for this masquerade case? Or could potentially be left off?

dianekaplan (approved):

src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
julianajlk: I can see the entire file but wanted to comment on them, was looking for a way to do that under the "files changed" view so I could click on `+` and comment, even though it's not a line that changed 🙂

Will continue here - wasn't that part of the code supposed to be removed in the other PR? Or it's still in use in courseware, so should be kept for now?

bmedx: I'm not sure which part of the code you're referring to. The MMP2P code below? If so, the experiment is still running so I'm leaving it for them to remove. Perhaps that's the wrong thing if it's displaying a type of banner this ticket is intended to remove, though? I'll ask Emma.

src/alerts/access-expiration-alert/AccessExpirationAlertMasquerade.jsx
bmedx: Nice catch! I added a commit to remove those and it seems good.

PR #494, opened 2021-06-17
AA-815: ui and a11y progress tab fixes

ciduarte:


src/course-home/data/slice.js
```diff
@@ -37,6 +38,9 @@ const slice = createSlice({
       state.toastBodyText = linkText;
       state.toastHeader = header;
     },
+    setGradesFeatureStatus: (state, { payload }) => {
+      state.gradesFeatureIsLocked = payload.gradesFeatureIsLocked;
+    },
```
ciduarte, line 43: It felt better to me to use state for this value rather than to pass it down to every single component it might have to touch... let me know if you have objections/more clear naming suggestions

mikix: I like passing data around, but I recognize that that seems to be not-favored by folks that live and breath react.

mikix (approved):

PR #495, opened 2021-06-21
feat: Remove course sock from OutlineTab

bmedx:
REV-2122: As part of the Value Prop implementation, remove the course sock from Course Home

### DIFF
--- a/src/course-home/outline-tab/OutlineTab.jsx
+++ b/src/course-home/outline-tab/OutlineTab.jsx
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from 'react';
+import React, { useState } from 'react';
 import { useSelector } from 'react-redux';
 import { sendTrackEvent, sendTrackingLogEvent } from '@edx/frontend-platform/analytics';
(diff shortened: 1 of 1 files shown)


julianajlk (approved):

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -245,16 +241,6 @@ function OutlineTab({ intl }) {
...
-        <CourseSock
-          courseId={courseId}
-          offer={offer}
-          orgKey={org}
-          pageLocation="Home Page"
-          ref={courseSock}
-          verifiedMode={verifiedMode}
-        />
-      )}
     </>
```
dianekaplan, line 244: Is course sock still used elsewhere? Or are we generally killing it now? If the latter, should we remove the relevant code (CourseSock.jsx, CourseSock.test, not sure if the translation strings go away on their own, etc)?

Edit: nevermind, I see that it's still used from Course.jsx, so that cleanup would likely happen in that removal ticket, REV-2220

dianekaplan (approved):

PR #499, opened 2021-06-22

bmedx:

### DIFF
--- a/src/alerts/access-expiration-alert/AccessExpirationAlertMasquerade.jsx
+++ b/src/alerts/access-expiration-alert/AccessExpirationAlertMasquerade.jsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage, FormattedDate } from '@edx/frontend-platform/i18n';
+
+import { Alert, ALERT_TYPES } from '../../generic/user-messages';
+
+function AccessExpirationAlertMasquerade({ payload }) {
+  const {
+    accessExpiration,
+    userTimezone,
+  } = payload;
+
+  if (!accessExpiration) {
+    return null;
+  }
+
+  const {
+    expirationDate,
+    masqueradingExpiredCourse,
+  } = accessExpiration;
+
+  if (!masqueradingExpiredCourse) {
+    return null;
+  }
+
+  const timezoneFormatArgs = userTimezone ? { timeZone: userTimezone } : {};
+
+  return (
+    <Alert type={ALERT_TYPES.INFO}>
+      <FormattedMessage
+        id="learning.accessExpiration.expired"
+        defaultMessage="This learner does not have access to this course. Their access expired on {date}."
+        values={{
+          date: (
+            <FormattedDate
+              key="accessExpirationExpiredDate"
+              day="numeric"
+              month="short"
+              year="numeric"
+              value={expirationDate}
+              {...timezoneFormatArgs}
+            />
+          ),
+        }}
+      />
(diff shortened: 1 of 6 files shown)


src/alerts/access-expiration-alert/AccessExpirationAlertMasquerade.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+import { Alert, ALERT_TYPES } from '../../generic/user-messages';
+
+function AccessExpirationAlertMasquerade({ payload }) {
+  const {
+    accessExpiration,
+    userTimezone,
+  } = payload;
+
+  if (!accessExpiration) {
```
bmedx, line 13: Removing this check and the masqueradingExpiredCourse check below were what caused the production issue.

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -531,60 +531,22 @@ describe('Outline Tab', () => {
           },
         });
         await fetchAndRender();
-        await screen.findByText('This learner does not have access to this course.', { exact: false });
+        const check = await screen.queryByText('This learner does not have access to this course.', { exact: false });
```
bmedx, line 534: I changed this test to mirror the one below to make sure they are actually both working.

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -531,60 +531,22 @@ describe('Outline Tab', () => {
...
+        expect(check).toBeInTheDocument();
       });
 
-      it('shows expiration', async () => {
+      it('does not have special masquerade text', async () => {
```
bmedx, line 538: This is the new test to make sure the masquerade banner doesn't show up if we're not masquerading.

src/alerts/access-expiration-alert/AccessExpirationAlertMasquerade.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+  const {
+    expirationDate,
+    masqueradingExpiredCourse,
+  } = accessExpiration;
+
+  if (!masqueradingExpiredCourse) {
+    return null;
+  }
+
```
dianekaplan, line 25: Not a comment on the code, but on the cause of the prod issue: I'm confused why not including that last bit (return null if not masqueradingExpiredCourse) made the bug. *I think* we only include this component `if (masqueradingExpiredCourse)`.  How can we have if masqueradingExpiredCourse and also !masqueradingExpiredCourse?

Answer from Brian's slack (all clear now): There are two paths to display the masquerade “user lost access” banner:
1- Via the AccessExpirationAlert on non-course home pages
2- Via the hook on the course home page
This bit above worked for #1 since AccessExpirationAlert already performs that check. The hook on #2 didn’t perform that check (but does perform a different check for accessExpiration). So the course home case showed the banner.

src/alerts/access-expiration-alert/hooks.js
```diff
@@ -22,4 +23,20 @@ function useAccessExpirationAlert(accessExpiration, courseId, org, userTimezone,
   return { clientAccessExpirationAlert: AccessExpirationAlert };
 }
 
+export function useAccessExpirationAlertMasquerade(accessExpiration, userTimezone, topic) {
+  const isVisible = !!accessExpiration; // If it exists, show it.
```
dianekaplan, line 27: Thank you for the comment here- I had to look up "!!"

dianekaplan (approved):

PR #502, opened 2021-06-23
[BD-26] Timer bar on non-sequence pages

UvgenGen:
Timer bar on non-sequence pages

**Jira**: [EDUCATOR-5857](https://openedx.atlassian.net/browse/EDUCATOR-5857)
NOTE: need to update version for frontend-lib-special-exams after merging https://github.com/edx/frontend-lib-special-exams/pull/30

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1383,9 +1383,9 @@
       }
     },
     "@edx/frontend-lib-special-exams": {
-      "version": "1.8.3",
-      "resolved": "https://registry.npmjs.org/@edx/frontend-lib-special-exams/-/frontend-lib-special-exams-1.8.3.tgz",
-      "integrity": "sha512-062mLzG0jzAjgfIi6ARG9kcI4RbarsjYmB71p1hsD6x5uBDBulb1biAbj+x3HmDCKc8k6+/4cMvzKpN+Z1Vxwg==",
+      "version": "1.9.0",
+      "resolved": "https://registry.npmjs.org/@edx/frontend-lib-special-exams/-/frontend-lib-special-exams-1.9.0.tgz",
+      "integrity": "sha512-dG9iW/vru/6Rh9+1kkqDu3gdiSVF49/9JvdjPb+7H1mdZPs4/FdP43HicPgaJ4r6PKfd9Ssam+3rq1NhxpWzvw==",
       "requires": {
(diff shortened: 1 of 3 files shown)


openedx-webhooks:
Thanks for the pull request, @UvgenGen! I've created [BLENDED-897](https://openedx.atlassian.net/browse/BLENDED-897) to keep track of it in Jira.   More details are on the [BD-26](https://openedx.atlassian.net/wiki/spaces/COMM/pages/2523627980/BD-26+Proctoring+MFE+and+configuration) project page.


zacharis278:
this seems to work fine but in terms of component structure it seems a bit odd that we're wrapping the entire course outline in our "Exam" component even though all we really need here is the timer. I'm wondering if we can export a lightweight wrapper component that just handles state and the timer from the library. That, or just to standalone way to include the timer without the need for a wrapping all of the page content (I don't think we need to do that on the outline page right right?)

UvgenGen:
@Zacharis278 this wrapper component does not make unnecessary requests. We won't get into the Instructions component when we don't pass a 'sequence' props into 'SequenceExamWrapper'. Maybe we should rename our wrapper to avoid confusion?

zacharis278:
it's the passing of null props to cut out most behavior of the component that seems odd to me. I think the name is appropriate for its original purpose, including on the course home page seems to be a different use case to me. Is there a reason we even need a wrapper in the first place to include the timer?

UvgenGen:
In the wrapper, an exam state is created, which is used to render the timer block. Active attempt and handlers such as stopExam, expireExam, pollAttempt, pingAttempt are used. Also in the wrapper the API is called to get an active attempt.

zacharis278:
looks good but can you look into the test failures? @ViktorRusakov

viktorrusakov:
> looks good but can you look into the test failures? @ViktorRusakov

I'm not sure why this PR would cause so many test errors
I guess it's because we still haven't merged special exam lib PR and here we try to export new component from the package which is not published yet
can you try to rerun the tests once the special exams package is updated?

zacharis278:
@ViktorRusakov oh of course that totally makes sense. The new version of special exams has been published, you'll just need to manually update package.json in this PR

zacharis278 (approved):

PR #503, opened 2021-06-24
chore: update special exams library

zacharis278:
update to most recent version of the special exams library. Includes several bugfixes for timed exams and limited support for proctored/onboarding. These features are still disabled by default in production.


schenedx (approved):
👍

PR #505, opened 2021-06-25
[AA-874] feat: add user id parameter to progress page

MatthewPiatetsky:
This allows users who have the proper permissions to view the progress of another user.

[Platform PR](https://github.com/edx/edx-platform/pull/28033)
[Ticket](https://openedx.atlassian.net/browse/AA-874)


src/course-home/progress-tab/messages.js
```diff
@@ -5,6 +5,10 @@ const messages = defineMessages({
     id: 'progress.header',
     defaultMessage: 'Your progress',
   },
+  progressHeaderDifferentUser: {
+    id: 'progress.header.differentUser',
+    defaultMessage: 'Course progress for',
```
mikix: You might want a description here, saying what kind of thing comes after the `for` - might be helpful for translators. And/instead you should probably put the username in the string somewhere as a format string. So like `Course progress for {user}` - they might need to move it around in the sentence.

src/tab-page/TabContainer.jsx
```diff
@@ -14,10 +14,16 @@ export default function TabContainer(props) {
...
   const dispatch = useDispatch();
   useEffect(() => {
     // The courseId from the URL is the course we WANT to load.
-    dispatch(fetch(courseIdFromUrl));
+    if (progressStudentId) {
+      progressStudentId = parseInt(progressStudentId, 10);
+      dispatch(fetch(courseIdFromUrl, progressStudentId));
+    } else {
+      dispatch(fetch(courseIdFromUrl));
+    }
```
mikix: This feels like you're breaking the abstractions of these generic components.

Can you just have `fetchProgressTab` get the param?

src/course-home/data/thunks.js
```diff
@@ -74,8 +74,8 @@ export function fetchDatesTab(courseId) {
   return fetchTab(courseId, 'dates', getDatesTabData);
 }
 
-export function fetchProgressTab(courseId) {
-  return fetchTab(courseId, 'progress', getProgressTabData);
+export function fetchProgressTab(courseId, progressStudentId) {
+  return fetchTab(courseId, 'progress', getProgressTabData, progressStudentId);
```
mikix: This is breaking a generic abstraction like `fetchTab` with details of one single specific usage.

You could use a closure here to avoid that. Something like:

```
return fetchTab(courseId, 'progress', (courseId) => {
    return getProgressTabData(courseId, progressStudentId);
});
```

src/course-home/progress-tab/ProgressHeader.jsx
```diff
@@ -14,14 +14,21 @@ function ProgressHeader({ intl }) {
...
+  const { studioUrl, progressStudentId, username } = useModel('progress', courseId);
+
+  const viewingOtherStudentsProgressPage = (progressStudentId && progressStudentId !== userId);
 
   return (
     <>
       <div className="row w-100 m-0 mt-3 mb-4 justify-content-between">
-        <h1>{intl.formatMessage(messages.progressHeader)}</h1>
+        {viewingOtherStudentsProgressPage && (
+          <h1>{intl.formatMessage(messages.progressHeaderDifferentUser)} {username}</h1>
```
mikix: As mentioned elsewhere - this should be a formatted message that takes username as an argument.

Also, this is marked as not covered by a test - this is the sort of code path that we won't notice breaking while doing normal dev work. Probably worth a test.

src/course-home/data/api.js
```diff
@@ -211,12 +211,22 @@ export async function getDatesTabData(courseId) {
...
+  if (progressStudentId) {
+    url += `/${progressStudentId}/`;
+  }
   try {
     const { data } = await getAuthenticatedHttpClient().get(url);
     const camelCasedData = camelCaseObject(data);
 
+    if (progressStudentId) {
+      camelCasedData.progressStudentId = progressStudentId;
+    }
```
mikix: This strikes me as a little odd - to save this here so that future components can easily grab it. For `courseId`, we use a `const { courseId }  = useSelector(state => state.courseHome)` line to grab that detail from the store. Wonder if we could grab the student id from there - but I'm forgetting what that `state.courseHome` store even is - @ciduarte what's the right pattern here?

src/index.jsx
```diff
@@ -44,7 +44,12 @@ subscribe(APP_READY, () => {
               <DatesTab />
             </TabContainer>
           </PageRoute>
-          <PageRoute path="/course/:courseId/progress">
+          <PageRoute
+            path={[
+              '/course/:courseId/progress/:progressStudentId/',
```
ciduarte: nit: I think this should just generally be named `studentId` and not `progressStudentId`

src/tab-page/TabContainer.jsx
```diff
@@ -14,10 +14,16 @@ export default function TabContainer(props) {
   } = props;
 
   const { courseId: courseIdFromUrl } = useParams();
+  let { progressStudentId } = useParams();
```
ciduarte: Why not just add `studentId` to line 16?

Oh I see you're reassigning it below. I feel like you could just pass the `parseInt()` into the fetch rather than reassigning it.

src/course-home/data/api.js
ciduarte: That store gets updated in the `fetchTab` thunk through dispatching `fetchTabSuccess({ courseId });`

I think it would make sense to save `studentId` as part of the state.

src/course-home/data/thunks.js
ciduarte: Oooh I like that.

I'll just note that if we don't pull `studentId` into `fetchTab`, we won't be able to use it to `dispatch(fetchTabSuccess...)` i.e. save `studentId` to the store. Could possibly dispatch that action elsewhere 🤷

ciduarte (commented):
Just some comments and responses to MT.

The biggest thing that stands out to me is that I don't love the use of `progressStudentId`, I think it should just be called `studentId` because it's a student ID regardless of the progress tab context.

src/index.jsx
MatthewPiatetsky: i would be concerned about simply naming the variable studentId (since in other places the variable is used the context is less clear)
people could think the studentId variable represents the studentId of the requesting user, or the studentId of the user being masqueraded as, neither of which is the case
i'm flexible as far as the name goes, though i'd be concerned that specific name could lead to the above mentioned confusion?

src/tab-page/TabContainer.jsx
MatthewPiatetsky: i think if I need to use the useParams hook it won't work inside fetchProgressTab because it is a function and not a functional component and hooks need to be called from a functional component (if i understood the error i got correctly) but open to suggestions here

src/course-home/data/api.js
MatthewPiatetsky: i removed this and am updating progressStudentId as part of fetchTabSuccess - not sure if this is what you were envisioning

src/course-home/progress-tab/ProgressTab.test.jsx
```diff
@@ -1034,4 +1034,30 @@ describe('Progress Tab', () => {
...
+          completion_summary: {
+            complete_count: 1,
+            incomplete_count: 1,
+            locked_count: 1,
+          },
+        });
+
+      const otherUserId = 10;
+      const progressUrlWithId = `${getConfig().LMS_BASE_URL}/api/course_home/v1/progress/${courseId}/${otherUserId}/`;
+      axiosMock.onGet(progressUrlWithId).reply(200, progressTabData);
```
MatthewPiatetsky, line 1055: for some reason this mock isn't working - still trying to figure it out

src/tab-page/TabContainer.jsx
mikix: Hmm, well then you can likely avoid this abstraction breakage by using a closure with the `fetch` passed down from `index.js` when the matching first happens.

MatthewPiatetsky: can you clarify this suggestion?

mikix: Sure. So just for context and clarity - the reason I don't love the current solution is that it injects tab-specific quirks into a generic tab container. Lets keep that code clean!

Javascript should be able to give us the tools to do so. Either by grabbing the param argument in the lower data method (my first suggestion above) or by using closures of some form above this generic container. But in both cases, my aim is to keep progress-tab quirks in progress-tab code.

So the closure suggestion is very similar to [my suggestion](https://github.com/edx/frontend-app-learning/pull/505#discussion_r659827333) for fetchProgressTab. In the main `src/index.js` file, where we pass `TabContainer` the `fetch` argument, we can give it a closure like:

```
<TabContainer ... fetch={(courseId) => fetchProgressTab(courseId, student_id_from_routing_match)}>
```

That way, we bundle the student id down from the routing all the way through TabContainer, but TabContainer doesn't have to know anything about it. Now, the trick is how to write the above. `student_id_from_routing_match` isn't an easy thing to get the way it is currently written. But that's because those routes are all written in the new fancy "children" style. By using maybe the [render prop](https://reactrouter.com/web/api/Route/render-func), you could have access to the `match` directly?

But the fact that we have to go to the older pre-hook style makes me think that a person that lives and breathes react would find some way to make `useParams` work, that we both missed? But I think this is a fine workaround if there are difficulties doing that. At least, I think it's better than having `TabContainer` know anything about this.

src/course-home/progress-tab/ProgressTab.test.jsx
```diff
@@ -1034,4 +1034,22 @@ describe('Progress Tab', () => {
       expect(screen.queryByTestId('certificate-status-component')).not.toBeInTheDocument();
     });
   });
+
+  describe('Viewing progress page of other students by changing url', () => {
+    it('Changing the url changes the header', async () => {
+      const axiosMockOther = new MockAdapter(getAuthenticatedHttpClient());
```
MatthewPiatetsky: this was conflicting with the axiosMock in the other tests, this seems to work (but let me know if there is a better way to avoid the conflicts)

src/tab-page/TabContainer.jsx
MatthewPiatetsky: thank you for the detailed explanation and the additional help! that makes sense, i believe I've updated the code to your suggestion

src/course-home/progress-tab/ProgressHeader.jsx
MatthewPiatetsky: i believe this should be covered by a test now

src/course-home/data/slice.js
```diff
@@ -22,6 +22,9 @@ const slice = createSlice({
     },
     fetchTabSuccess: (state, { payload }) => {
       state.courseId = payload.courseId;
+      if (payload.progressStudentId) {
+        state.progressStudentId = payload.progressStudentId;
+      }
```
mikix: This state can get overridden by future dispatches yeah? Maybe clear it out if there isn't a current student id? In which case you could probably just remove the `if`?

src/course-home/data/thunks.js
```diff
@@ -27,12 +27,12 @@ const eventTypes = {
   POST_EVENT: 'post_event',
 };
 
-export function fetchTab(courseId, tab, getTabData) {
+export function fetchTab(courseId, tab, getTabData, progressStudentId) {
```
mikix: You don't need to have this parameter anymore - that was the whole point of using a closure in `fetchProgressTab`. This generic code in `fetchTab` doesn't need to know anything about progress tab logic.

You can drop all changes to `fetchTab` here and stop passing this arg to it.

src/index.jsx
mikix: I hear your concern there about just `studentId` not being clear in all contexts, but I agree with Carla that `progressStudentId` is a little narrow also.

Like, you could imagine us enabling support for a similar feature on the Dates tab (we don't plan to, but it's possible). This is basically like a little mini-masquerade, but for coaches, yeah? It would be odd if each specific usage of masquerading in platform was prefixed with a specific view namespace like this. Harder to search for, harder to understand what's happening when reading the code.

```
 _, progress_user = setup_masquerade(...)
```

vs

```
 _, masquerade_user = setup_masquerade(...)
```

(But to Carla's point, in platform, we actually just reset `user` in these cases, we don't actually use `masquerade_user`)

Not a huge deal; also just a nit for me. But if it's easy to think of a slightly more generic name, might be nice. Target user? View user? User-for-whom-this-data-is-true?

src/index.jsx
```diff
@@ -44,11 +44,23 @@ subscribe(APP_READY, () => {
...
-              <ProgressTab />
-            </TabContainer>
-          </PageRoute>
+          <PageRoute
+            path={[
+              '/course/:courseId/progress/:progressStudentId/',
+              '/course/:courseId/progress',
+            ]}
+            render={({ match }) => (
+              <>
```
mikix: You shouldn't need these `<>` grouping markers - you just have one `TabContainer` child.

Also, these aren't covered by tests, because we don't have any tests that just hit this URL and let the routing work? We just have tests with their own routing? Is that easy? Might be worth a few lines. But if it's a pain in the butt, I'm fine without it.

src/course-home/data/thunks.js
MatthewPiatetsky: how would fetchTabSuccess know about the targetUserId then?
that's where i'm currently setting the state to contain the targetUserId, or should I be setting the state somewhere else?
currently fetchTabSuccess is just imported, but i guess i could pull it in to the getProgressTabData (and other methods) and have it passed as an argument, and then make another closure for the fetchTabSuccess method?
though there is probably a cleaner way to set the state to have the targetUserId?

src/index.jsx
MatthewPiatetsky: sure, i'm fine with targetUser - that makes it clear that it is not just your current user which was my main concern

src/index.jsx
MatthewPiatetsky: sure more of an integration test would make sense, do you know if we have examples of this pattern in the learning mfe? looking through a couple test files i'm not seeing it and want to make sure i do it correctly

src/course-home/data/thunks.js
mikix: Ah, yes - now that you set targetUserId in the state, you need to pass it to `fetchTabSuccess`. I had forgotten that aspect and was relying on my memory of an earlier version of the code, where you set the user id in the data handler.

So... if `fetchTab` is going to know about `targetUserId`, you don't really need the closure in `fetchProgressTab`, because its whole purpose was to avoid leaking progress-tab concepts into `fetchTab`. But if you're going to do that anyway, what's the point.

So I think you'd either want to always pass the `getTabData` and `fetchTabSuccess` methods the `targetUserId` and avoid the closure, or pass it to neither, keep the closure, and find a different way to let the view know what the `targetUserId` is, besides the `fetchTabSuccess` state-based approach.

Now that the parameter is a little less progress specific (a more generic `targetUserId` variable name), I could be OK with passing it to `getTabData` and `fetchTabSuccess` and keeping the state-setting you do in `fetchTabSuccess`. But the fact remains that this is a feature only one tab is using - if another tab grows a similar but different feature, are we going to keep adding one-off ad-hoc data to the `fetchTabSuccess` method and make it a swiss-army-knife state setter? Or should we think about that ahead of time here and figure another solution to the state-setting down to the component.

I had originally suggested keeping that variable in the redux state because I thought it might make things easier, not because that is the One Good Place for it. Like, we could go back to stuffing it into the model data? Or maybe I'm overthinking it and just having `fetchTab` learn about this one new feature is not a slippery slope and is just fine.

@ciduarte got any opinions here?

ciduarte: I feel okay with having `fetchTab` learn about `targetUserId` in order to set the state. This didn't feel like a slippery slope to me at first, and it might be worth crossing that bridge if/when we get there?

MatthewPiatetsky: ok, i've removed the closure in getProgressTabData

src/tab-page/TabContainer.test.jsx
```diff
@@ -42,4 +42,27 @@ describe('Tab Container', () => {
       .toHaveBeenCalledWith(courseId);
     expect(screen.getByTestId('TabPage')).toBeInTheDocument();
   });
+
+  it('Should handle passing in a targetUserId', () => {
+    history.push(`/course/${courseId}/1/`);
+    const mockFetch2 = jest.fn().mockImplementation((x) => x);
```
ciduarte: Curious about the addition of `mockFetch2`. Did you try switching the `beforeAll` to a `beforeEach` and creating the mockFetch there?

src/tab-page/TabContainer.test.jsx
```diff
@@ -42,4 +42,27 @@ describe('Tab Container', () => {
...
+
+    const mockData2 = {
+      children: [],
+      fetch: () => mockFetch2(courseId, 1),
```
ciduarte: Why is this being called with `(courseId, 1)`?

It looks like the original implementation just uses the `history.push()` and somehow the `courseId` gets passed to `mockFetch`... does that not work here?

src/tab-page/TabContainer.test.jsx
```diff
@@ -42,4 +42,27 @@ describe('Tab Container', () => {
       .toHaveBeenCalledWith(courseId);
     expect(screen.getByTestId('TabPage')).toBeInTheDocument();
   });
+
+  it('Should handle passing in a targetUserId', () => {
+    history.push(`/course/${courseId}/1/`);
```
ciduarte: Shouldn't this be `/course/${courseId}/progress/1`?

src/tab-page/TabContainer.test.jsx
```diff
@@ -42,4 +42,27 @@ describe('Tab Container', () => {
...
+
+    const mockData2 = {
+      children: [],
+      fetch: () => mockFetch2(courseId, 1),
+      tab: 'dummy',
+      slice: 'courseware',
+    };
+
+    render(
+      <Route path="/course/:courseId/:targetUserId">
```
ciduarte: Similar comment as above, shouldn't this route say `progress` in it?

src/course-home/progress-tab/ProgressTab.test.jsx
```diff
@@ -1034,4 +1034,22 @@ describe('Progress Tab', () => {
...
+      const courseMetadata = Factory.build('courseHomeMetadata', { id: courseId }, {});
+      axiosMockOther.onGet(courseMetadataUrl).reply(200, courseMetadata);
```
ciduarte: Did you try using the `setMetadata()` function in this file?

src/course-home/progress-tab/ProgressTab.test.jsx
ciduarte: I'm not sure what the conflicts were so it's hard for me to say. You shouldn't have to set it here because it's being re-assigned in the `beforeEach`

src/course-home/progress-tab/ProgressHeader.jsx
```diff
@@ -12,16 +12,33 @@ import messages from './messages';
...
+
+  const viewingOtherStudentsProgressPage = (targetUserId && targetUserId !== userId);
 
   return (
     <>
       <div className="row w-100 m-0 mt-3 mb-4 justify-content-between">
-        <h1>{intl.formatMessage(messages.progressHeader)}</h1>
+        {viewingOtherStudentsProgressPage && (
+          <h1>
+            <FormattedMessage
```
ciduarte: I'd rather not use a FormattedMessage here. How about something like this? https://github.com/edx/frontend-app-learning/blob/fdfb60bee803dc22c062865ed098e051714cf9cc/src/courseware/course/course-exit/messages.js#L145

You could also just do this before the return
```
const pageTitle = viewingOtherStudentsProgressPage ?
    intl.formatMessage(messages.progressHeaderForTargetUser, { username })
    : intl.formatMessage(messages.progressHeader);
```

and then have `<h1>{pageTitle}</h1>`.

src/tab-page/TabContainer.test.jsx
MatthewPiatetsky: i updated the code to make it work without redefining a new fetch by recreating the closure that exists in index.jsx within the test so that mockFetch has access to the targetUserId

src/tab-page/TabContainer.test.jsx
MatthewPiatetsky: same comment as above ...was doing that so the fetch would have access to the targetUserId but I fixed it now by recreating the closure within the test

src/tab-page/TabContainer.test.jsx
MatthewPiatetsky: fixed

src/tab-page/TabContainer.test.jsx
MatthewPiatetsky: fixed

src/course-home/progress-tab/ProgressTab.test.jsx
MatthewPiatetsky: i had created a different axios mock in order to mock the different progress url with the id and wasn't using setMetadata since it uses the existing axiosMock.
However, it seems like i found a way to reuse the mock for both by making it a regex.
The course id seems to break the regex even if i escape the + signs but if i just have a * after progress it seems like it works now without a separate axiosMock

src/course-home/progress-tab/ProgressHeader.jsx
MatthewPiatetsky: sure that's cleaner, updated

src/course-home/progress-tab/ProgressTab.test.jsx
MatthewPiatetsky: it seems like adding a regex to the url allowed me to avoid this (though i did remove the courseId as it was causing issues otherwise and just added a * to cover both the case with and without the targetUserId)

src/tab-page/TabContainer.test.jsx
```diff
@@ -5,31 +5,34 @@ import { initializeTestStore, render, screen } from '../setupTest';
...
+      tab: 'dummy',
+      slice: 'courseware',
+    };
     const store = await initializeTestStore({ excludeFetchSequence: true });
     courseId = store.getState().courseware.courseId;
   });
 
   it('renders correctly', () => {
-    history.push(`/course/${courseId}`);
+    history.push(`/course/${courseId}/progress`);
```
ciduarte: Is it necessary to add `/progress` here? I think this was originally just testing `<TabContainer />`, or did you add that in as an example of any tab that can be passed through?

ciduarte (approved):
Awesome looks good! Just one question, but not a blocker

src/tab-page/TabContainer.test.jsx
MatthewPiatetsky: no this wasn't necessary, fixed, thanks for the callout and the review!

PR #506, opened 2021-06-25
fix: Autoscroll moves element to center of page

KristinAoki:
Jira issue: [TNL-8312](https://openedx.atlassian.net/browse/TNL-8312)

This PR fixes the anchor tag's position on the page when autoscrolling is used. Previously, the scroll would move the element to the center of the page. Now the scroll moves the element to the top of the page. The only case where the element will not be at the top of the page is when the element is too close to the bottom of the page and there is not enough page remaining to force the element to the top.

### DIFF
--- a/src/courseware/course/sequence/Unit.jsx
+++ b/src/courseware/course/sequence/Unit.jsx
@@ -130,7 +130,8 @@ function Unit({
       } else if (event.data.offset) {
         // We listen for this message from LMS to know when the page needs to
         // be scrolled to another location on the page.
-        window.scrollTo(0, event.data.offset);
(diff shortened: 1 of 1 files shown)


kdmccormick (changes requested):
This does move the anchor tag a bit closer to the top viewport, but it doesn't seem to reliably put the anchor tag *right* at the top. I think the implementation is dependent on zoom level, as well.

Consider that `event.data.offset` is coming back with the height of the anchor tag *relative to the top of the unit iframe*, which is positioned noticeably below the top of the document. If you factor that in, you should be able to get the anchor right at the top of the screen.

kdmccormick (approved):
Works perfectly on my machine 👍🏻

PR #508, opened 2021-06-28
[MST-885] Add missing check for graded units

bseverino:
[MST-885](https://openedx.atlassian.net/browse/MST-885)

This fixes a bug where if the learner needs an integrity signature, but the unit is not graded, neither the honor code panel nor the unit content would display.


schenedx (approved):
👍

PR #510, opened 2021-06-30
test: add consumer pact tests for courseware

Ali-D-Akbar:
[PROD-2401](https://openedx.atlassian.net/browse/PROD-2401)
Adding contract based tests. The pact will be between frontend-app-learning and LMS.

Steps to run these tests on your local machine:
1. npm install
2. npm test -- src/courseware/data/pact-tests/lmsPact.test.jsx

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "@fortawesome/free-regular-svg-icons": "5.13.1",
     "@fortawesome/free-solid-svg-icons": "5.13.1",
     "@fortawesome/react-fontawesome": "0.1.14",
+    "@pact-foundation/pact": "^9.16.0",
     "@reduxjs/toolkit": "1.3.6",
     "classnames": "2.2.6",
     "core-js": "3.6.5",
--- a/src/courseware/data/pact-tests/frontend-app-learning-lms.json
+++ b/src/courseware/data/pact-tests/frontend-app-learning-lms.json
@@ -1,0 +1,423 @@
+{
+  "consumer": {
+    "name": "frontend-app-learning"
+  },
+  "provider": {
+    "name": "lms"
+  },
+  "interactions": [
+    {
+      "description": "a request to get course blocks",
+      "providerState": "Blocks data exists for course_id course-v1:edX+DemoX+Demo_Course",
+      "request": {
+        "method": "GET",
+        "path": "/api/courses/v2/blocks/",
+        "query": "course_id=course-v1%3AedX%2BDemoX%2BDemo_Course&username=Mock+User&depth=3&requested_fields=children%2Ceffort_activities%2Ceffort_time%2Cshow_gated_sections%2Cgraded%2Cspecial_exam_info%2Chas_scheduled_content"
+      },
+      "response": {
+        "status": 200,
+        "headers": {
+        },
+        "body": {
+          "root": "block-v1:edX+DemoX+Demo_Course+type@course+block@course",
+          "blocks": {
+            "block-v1:edX+DemoX+Demo_Course+type@course+block@course": {
+              "id": "block-v1:edX+DemoX+Demo_Course+type@course+block@course",
+              "block_id": "course",
+              "lms_web_url": "/courses/course-v1:edX+DemoX+Demo_Course/jump_to/block-v1:edX+DemoX+Demo_Course+type@course+block@course",
+              "legacy_web_url": "/courses/course-v1:edX+DemoX+Demo_Course/jump_to/block-v1:edX+DemoX+Demo_Course+type@course+block@course?experience=legacy",
+              "student_view_url": "/xblock/block-v1:edX+DemoX+Demo_Course+type@course+block@course",
+              "type": "course",
+              "display_name": "Demonstration Course"
+            }
+          }
+        },
+        "matchingRules": {
+          "$.body.root": {
+            "match": "type"
+          },
+          "$.body.blocks": {
+            "match": "type"
+          }
+        }
+      }
+    },
+    {
+      "description": "a request to get course metadata",
+      "providerState": "course metadata exists for course_id course-v1:edX+DemoX+Demo_Course",
+      "request": {
+        "method": "GET",
+        "path": "/api/courseware/course/course-v1:edX+DemoX+Demo_Course",
+        "query": "browser_timezone=Asia%2FKarachi"
+      },
+      "response": {
+        "status": 200,
+        "headers": {
+        },
+        "body": {
+          "access_expiration": {
+            "expiration_date": "2013-02-05T05:00:00Z",
+            "masquerading_expired_course": false,
+            "upgrade_deadline": "2013-02-05T05:00:00Z",
+            "upgrade_url": "link"
+          },
+          "can_show_upgrade_sock": false,
+          "content_type_gating_enabled": false,
+          "end": "2013-02-05T05:00:00Z",
+          "enrollment": {
+            "mode": "audit",
+            "is_active": true
+          },
+          "enrollment_start": "2013-02-05T05:00:00Z",
+          "enrollment_end": "2013-02-05T05:00:00Z",
+          "id": "course-v1:edX+DemoX+Demo_Course",
+          "license": "all-rights-reserved",
+          "name": "Demonstration Course",
+          "number": "DemoX",
+          "offer": {
+            "code": "code",
+            "expiration_date": "2013-02-05T05:00:00Z",
+            "original_price": "$99",
+            "discounted_price": "$99",
+            "percentage": 50,
+            "upgrade_url": "url"
+          },
+          "org": "edX",
+          "related_programs": null,
+          "short_description": "",
+          "start": "2013-02-05T05:00:00Z",
+          "tabs": [
+            {
+              "title": "Course",
+              "slug": "courseware",
+              "priority": 0,
+              "type": "courseware",
+              "url": "http://localhost:2000/course/course-v1:edX+DemoX+Demo_Course/home"
+            }
+          ],
+          "user_timezone": null,
+          "verified_mode": {
+            "access_expiration_date": null,
+            "currency": "USD",
+            "currency_symbol": "$",
+            "price": 149,
+            "sku": "8CF08E5",
+            "upgrade_url": "http://localhost:18130/basket/add/?sku=8CF08E5"
+          },
+          "show_calculator": false,
+          "original_user_is_staff": true,
+          "can_view_legacy_courseware": true,
+          "is_staff": true,
+          "course_access": {
+            "has_access": true,
+            "error_code": null,
+            "developer_message": null,
+            "user_message": null,
+            "additional_context_user_message": null,
+            "user_fragment": null
+          },
+          "notes": {
+            "enabled": false,
+            "visible": true
+          },
+          "marketing_url": null,
+          "celebrations": {
+            "irst_section": false,
+            "streak_length_to_celebrate": null,
+            "streak_discount_experiment_enabled": false
+          },
+          "user_has_passing_grade": false,
+          "course_exit_page_is_active": false,
+          "certificate_data": {
+            "cert_status": "audit_passing",
+            "cert_web_view_url": null,
+            "download_url": null,
+            "certificate_available_date": null
+          },
+          "verify_identity_url": null,
+          "verification_status": "none",
+          "linkedin_add_to_profile_url": null,
+          "is_mfe_special_exams_enabled": false,
+          "is_mfe_proctored_exams_enabled": false,
+          "user_needs_integrity_signature": false
+        },
+        "matchingRules": {
+          "$.body.access_expiration.expiration_date": {
+            "match": "regex",
+            "regex": "^(?:[1-9]\\d{3}-(?:(?:0[1-9]|1[0-2])-(?:0[1-9]|1\\d|2[0-8])|(?:0[13-9]|1[0-2])-(?:29|30)|(?:0[13578]|1[02])-31)|(?:[1-9]\\d(?:0[48]|[2468][048]|[13579][26])|(?:[2468][048]|[13579][26])00)-02-29)T(?:[01]\\d|2[0-3]):[0-5]\\d:[0-5]\\d(?:Z|[+-][01]\\d:[0-5]\\d)$"
+          },
+          "$.body.access_expiration.masquerading_expired_course": {
+            "match": "type"
+          },
+          "$.body.access_expiration.upgrade_deadline": {
+            "match": "regex",
+            "regex": "^(?:[1-9]\\d{3}-(?:(?:0[1-9]|1[0-2])-(?:0[1-9]|1\\d|2[0-8])|(?:0[13-9]|1[0-2])-(?:29|30)|(?:0[13578]|1[02])-31)|(?:[1-9]\\d(?:0[48]|[2468][048]|[13579][26])|(?:[2468][048]|[13579][26])00)-02-29)T(?:[01]\\d|2[0-3]):[0-5]\\d:[0-5]\\d(?:Z|[+-][01]\\d:[0-5]\\d)$"
+          },
+          "$.body.access_expiration.upgrade_url": {
+            "match": "type"
+          },
+          "$.body.can_show_upgrade_sock": {
+            "match": "type"
+          },
+          "$.body.content_type_gating_enabled": {
+            "match": "type"
+          },
+          "$.body.end": {
+            "match": "regex",
+            "regex": "^(?:[1-9]\\d{3}-(?:(?:0[1-9]|1[0-2])-(?:0[1-9]|1\\d|2[0-8])|(?:0[13-9]|1[0-2])-(?:29|30)|(?:0[13578]|1[02])-31)|(?:[1-9]\\d(?:0[48]|[2468][048]|[13579][26])|(?:[2468][048]|[13579][26])00)-02-29)T(?:[01]\\d|2[0-3]):[0-5]\\d:[0-5]\\d(?:Z|[+-][01]\\d:[0-5]\\d)$"
+          },
+          "$.body.enrollment.mode": {
+            "match": "regex",
+            "regex": "^(audit|verified)$"
+          },
+          "$.body.enrollment.is_active": {
+            "match": "type"
+          },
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -78,6 +78,7 @@
     "husky": "3.1.0",
     "jest": "24.9.0",
     "jest-chain": "1.1.5",
+    "jest-pact": "^0.9.1",
```
DawoudSheraz: why jest-pact? how is it different from pact-js? and what's the support level for jest-pact as compared to pact-js?

src/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,100 @@
+import { pactWith } from 'jest-pact';
+import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
+// import axios from 'axios';
+// import { initializeTestStore } from '../setupTest';
+// import { getCourseBlocks } from '../courseware/data/api';
+// import { Matchers } from '@pact-foundation/pact';
+
+pactWith({ consumer: 'MyConsumer', provider: 'MyProvider' }, provider => {
```
DawoudSheraz: use appropriate consumer and provider names. the consumer is frontend-app-learning and the provider is lms.

src/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,100 @@
...
+          },
+          root: 'block-v1:edX+DemoX+Demo_Course_1+type@course+block@bcdabcdabcdabcdabcdabcdabcdabcd4',
+        },
+      },
+      withRequest: {
+        method: 'GET',
+        path: '/api/courses/v2/blocks/',
+      },
+    }));
+    it('returns course blocks', () => getAuthenticatedHttpClient().get(`${baseUrl}/api/courses/v2/blocks/`).then(response => {
```
DawoudSheraz: You should not be making API calls directly in test. Use the method written in api.js. This is an incorrect way of writing contract tests. See https://docs.pact.io/consumer/contract_tests_not_functional_tests

src/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,100 @@
...
+              due: null,
+              graded: false,
+              icon: null,
+              showLink: true,
+              display_name: 'bcdabcdabcdabcdabcdabcdabcdabcd3',
+              id: 'block-v1:edX+DemoX+Demo_Course_1+type@chapter+block@bcdabcdabcdabcdabcdabcdabcdabcd3',
+              student_view_url: 'http://localhost:18000/xblock/block-v1:edX+DemoX+Demo_Course_1+type@chapter+block@bcdabcdabcdabcdabcdabcdabcdabcd3',
+              legacy_web_url: 'http://localhost:18000/courses/course-v1:edX+DemoX+Demo_Course_1/jump_to/block-v1:edX+DemoX+Demo_Course_1+type@chapter+block@bcdabcdabcdabcdabcdabcdabcdabcd3?experience=legacy',
+            },
+            'block-v1:edX+DemoX+Demo_Course_1+type@sequential+block@bcdabcdabcdabcdabcdabcdabcdabcd2': {
```
DawoudSheraz: don't just add random data. Remember that this data or matching data will need to be setup properly in provider verification. Additionally, only add those fields which are used within learning MFE.

package.json
Ali-D-Akbar: jest-pact is exactly like pact-js with some more utility as `pactWith` handles variable ports, starts and ends the mock server beforeAll and afterAll interactions/tests and `provider.verify()` after each interaction/test. Since it works like pact-js, support can be found by pact foundation related to pact-js.

src/pact-tests/lmsPact.test.jsx
Ali-D-Akbar: yes, I'll rename once the tests pass.

Ali-D-Akbar: yes, I'll rename once the tests pass.

src/courseware/data/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,397 @@
...
+      .setup()
+      .then((options) => mergeConfig({
+        LMS_BASE_URL: `http://localhost:${options.port}`,
+      }, 'Custom app config for pact tests'));
+  });
+
+  afterEach(() => provider.verify());
+  afterAll(async () => {
+    await provider.finalize();
+  });
```
DawoudSheraz: async/await can be removed.

src/courseware/data/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,397 @@
...
+
+  describe('When a request to get course blocks is made', () => {
+    it('should return the correct data', async () => {
```
DawoudSheraz: make this statement more explicit.

src/courseware/data/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,397 @@
...
+      expect(response).toBeTruthy();
+      expect(response.courses).toEqual(normalizedCourseBlock);
+      expect(response.sections).toEqual({});
+      expect(response.sequences).toEqual({});
+      expect(response.units).toEqual({});
+    });
+  });
+
+  describe('When a request to get course metadata is made', () => {
+    it('should return the correct data', async () => {
```
DawoudSheraz: same here

src/courseware/data/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,397 @@
...
+            access_expiration: null,
+            can_show_upgrade_sock: false,
+            content_type_gating_enabled: false,
+            effort: null,
+            end: null,
+            enrollment: { mode: null, is_active: false },
+            enrollment_start: null,
+            enrollment_end: null,
+            id: 'course-v1:edX+DemoX+Demo_Course',
+            license: null,
```
DawoudSheraz: don't just use random values. Understand how the api returns the information. e.g. you have mentioned enrollment mode as null. This will cause issues during provider verification.

src/courseware/data/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,397 @@
...
+        enrollmentMode: null,
+        isEnrolled: false,
+        courseAccess: {
+          hasAccess: false,
+          errorCode: 'authentication_required',
+          developerMessage: 'User must be authenticated to view the course',
+          userMessage: 'You must be logged in to see this course',
+          additionalContextUserMessage: null,
+          userFragment: null,
+        },
```
DawoudSheraz: we will assume user is logged(and setup this in provider states). Since you have mentioned correct data, don't add any auth related error. Instead, add what will be the success value for this.

src/courseware/data/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,397 @@
...
+          path: `/courses/${courseId}/xblock/${sequenceId}/handler/goto_position`,
+        },
+        willRespondWith: {
+          status: 200,
+          body:
+          {
+            success: true,
+          },
+        },
+      });
```
DawoudSheraz: description and interactions don't match.

DawoudSheraz (changes requested):

src/courseware/data/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,397 @@
...
+  dir: path.resolve(process.cwd(), 'src/courseware/data/pact-tests'),
+  logLevel: 'DEBUG',
+  cors: true,
+});
+
+describe('Courseware Service', () => {
+  const courseId = 'course-v1:edX+DemoX+Demo_Course';
+  const sequenceId = 'block-v1:edX+DemoX+Demo_Course+type@sequential+block@basic_questions';
+  const usageId = 'block-v1:edX+DemoX+Demo_Course+type@vertical+block@47dbd5f836544e61877a483c0b75606c';
+  beforeAll(async () => {
```
azanbinzahid, line 30: maybe async/await can be removed from here as well, [ref](https://github.com/pact-foundation/pact-js/blob/7c91c33a7a81f74c801f03a957b06905839d8743/examples/e2e/test/consumer.spec.js#L87).

src/courseware/data/pact-tests/lmsPact.test.jsx
azanbinzahid: A few examples can be found from [here](https://github.com/pact-foundation/pact-js/blob/master/examples/e2e/test/consumer.spec.js) especially describe, it and state statements.

src/courseware/data/pact-tests/lmsPact.test.jsx
Ali-D-Akbar: They're using one-liner syntax where the promise is returned directly.
We have to explicitly add async/await in our example since the promise won't be returned as a result.

src/courseware/data/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,421 @@
...
+  dir: path.resolve(process.cwd(), 'src/courseware/data/pact-tests'),
+  logLevel: 'DEBUG',
+  cors: true,
+});
+
+describe('Courseware Service', () => {
+  const courseId = 'course-v1:edX+DemoX+Demo_Course';
+  const sequenceId = 'block-v1:edX+DemoX+Demo_Course+type@sequential+block@basic_questions';
+  const usageId = 'block-v1:edX+DemoX+Demo_Course+type@vertical+block@47dbd5f836544e61877a483c0b75606c';
+  const reDate = '^(?:[1-9]\\d{3}-(?:(?:0[1-9]|1[0-2])-(?:0[1-9]|1\\d|2[0-8])|(?:0[13-9]|1[0-2])-(?:29|30)|(?:0[13578]|1[02])-31)|(?:[1-9]\\d(?:0[48]|[2468][048]|[13579][26])|(?:[2468][048]|[13579][26])00)-02-29)T(?:[01]\\d|2[0-3]):[0-5]\\d:[0-5]\\d(?:Z|[+-][01]\\d:[0-5]\\d)$';
```
DawoudSheraz: nit: dateRegex. Is this taken from pact's formatter?

src/courseware/data/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,421 @@
...
+      .setup()
+      .then((options) => mergeConfig({
+        LMS_BASE_URL: `http://localhost:${options.port}`,
+      }, 'Custom app config for pact tests'));
+  });
+
+  afterEach(() => provider.verify());
+  afterAll(() => {
+    provider.finalize();
+  });
```
DawoudSheraz: remove curly braces. The presence of braces can change the return format. Correct me I missed something with finalize block.

src/courseware/data/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,421 @@
...
+
+  describe('When a request to get course blocks is made', () => {
+    it('returns normalized course blocks', async () => {
+      await provider.addInteraction({
+        state: `Blocks data exists for course_id ${courseId}`,
+        uponReceiving: 'a request to get course blocks',
+        withRequest: {
+          method: 'GET',
+          path: '/api/courses/v2/blocks/',
```
DawoudSheraz, line 50: add query params that are added by api.js method when doing api call

src/courseware/data/pact-tests/lmsPact.test.jsx
DawoudSheraz: You can use common matchers from https://github.com/pact-foundation/pact-js/#match-common-formats

src/courseware/data/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,421 @@
...
+        },
+        willRespondWith: {
+          status: 200,
+          body:
+          {
+            root: like('block-v1:edX+DemoX+Demo_Course+type@course+block@course'),
```
DawoudSheraz: using like here does not make sense. like is used with objects. You can use string here https://github.com/pact-foundation/pact-js/#match-based-on-type

src/courseware/data/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,421 @@
...
+  });
+
+  describe('When a request to get course metadata is made', () => {
+    it('returns normalized course metadata', async () => {
+      await provider.addInteraction({
+        state: `course metadata exists for course_id ${courseId}`,
+        uponReceiving: 'a request to get course metadata',
+        withRequest: {
+          method: 'GET',
+          path: `/api/courseware/course/${courseId}`,
```
DawoudSheraz, line 101: again, if there are query params added before API call, add them here.

src/courseware/data/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,421 @@
...
+              expiration_date: term({
+                generate: '2013-02-05T05:00:00Z',
+                matcher: reDate,
+              }),
+              masquerading_expired_course: boolean(false),
+              upgrade_deadline: term({
+                generate: '2013-02-05T05:00:00Z',
+                matcher: reDate,
+              }),
+              upgrade_url: string('link'),
```
DawoudSheraz, line 120: you can use the `term` to do certain checks again link https://github.com/pact-foundation/pact-js/#match-by-regular-expression

src/courseware/data/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,421 @@
...
+            },
+            enrollment_start: term({
+              generate: '2013-02-05T05:00:00Z',
+              matcher: reDate,
+            }),
+            enrollment_end: term({
+              generate: '2013-02-05T05:00:00Z',
+              matcher: reDate,
+            }),
+            id: like('course-v1:edX+DemoX+Demo_Course'),
```
DawoudSheraz: instead of like, use term to validate the regex checks. You can use opaque keys regex for this purpose.

src/courseware/data/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,421 @@
...
+            tabs: eachLike({
+              title: 'Course', slug: 'courseware', priority: 0, type: 'courseware', url: `${getConfig().BASE_URL}/course/course-v1:edX+DemoX+Demo_Course/home`,
+            }),
+            user_timezone: null,
+            verified_mode: like({
+              access_expiration_date: null,
+              currency: 'USD',
+              currency_symbol: '$',
+              price: 149,
+              sku: '8CF08E5',
```
DawoudSheraz, line 177: instead of hard coding these values, use appropriate matchers.

src/courseware/data/pact-tests/lmsPact.test.jsx
```diff
@@ -0,0 +1,421 @@
...
+        uponReceiving: 'a request to get sequence metadata',
+        withRequest: {
+          method: 'GET',
+          path: `/api/courseware/sequence/${sequenceId}`,
+        },
+        willRespondWith: {
+          status: 200,
+          body:
+          {
+            items: eachLike({
```
DawoudSheraz, line 316: add minimum value for eachLike.

src/courseware/data/pact-tests/lmsPact.test.jsx
Ali-D-Akbar: I took it from here: https://www.myintervals.com/blog/2009/05/20/iso-8601-date-validation-that-doesnt-suck/

DawoudSheraz (commented):
added some nits and updates to address

src/courseware/data/pact-tests/lmsPact.test.jsx
Ali-D-Akbar: Either is fine Since the testing environment eventually waits for every open promise to be closed.
But yes, changing this for the sake of consistency.

src/courseware/data/pact-tests/lmsPact.test.jsx
Ali-D-Akbar: using like on an object applies all appropriate matchers to each value: https://github.com/pact-foundation/pact-js/#match-based-on-type 

`Note that you can wrap a like around a single value or an object. When wrapped around an object, all values and child object values will be matched according to types, unless overridden by something more specific like a term.`

src/courseware/data/pact-tests/lmsPact.test.jsx
Ali-D-Akbar: Min value cannot be set to 0 and by default, the min value is 1.

src/courseware/data/pact-tests/frontend-app-learning-lms.json
```diff
@@ -0,0 +1,414 @@
...
+    "name": "lms"
+  },
+  "interactions": [
+    {
+      "description": "a request to get course blocks",
+      "providerState": "Blocks data exists for course_id course-v1:edX+DemoX+Demo_Course",
+      "request": {
+        "method": "GET",
+        "path": "/api/courses/v2/blocks/"
+      },
```
azanbinzahid: If we cross reference it from lms api-docs, we can see this endpoint requires at-least one param. Therefore it fails to get a matching response from LMS. What can be done to ensure param requirements for all such api calls?

image

src/courseware/data/pact-tests/lmsPact.test.jsx
Ali-D-Akbar: This is kind of an unsurety what this field will have. It can be either a whole http url, just an api endpoint for frontend-app-lms or can be null. Applying regex might complicate a lot of things here.

src/courseware/data/pact-tests/lmsPact.test.jsx
DawoudSheraz: @Ali-D-Akbar For followup PR, replace this regex with the matchers provided by pact

DawoudSheraz (approved):
ok to merge for the first run.

azanbinzahid (approved):

PR #521, opened 2021-07-05
[BD-26] Fix special exam redirect check

viktorrusakov:
Fix and refactor how special exam redirect is checked

[JIRA](https://openedx.atlassian.net/browse/EDUCATOR-5869)

### DIFF
--- a/src/courseware/CoursewareContainer.jsx
+++ b/src/courseware/CoursewareContainer.jsx
@@ -59,9 +59,11 @@ const checkUnitToSequenceUnitRedirect = memoize((courseStatus, courseId, sequenc
   }
 });
 
-const checkSpecialExamRedirect = memoize((sequenceStatus, sequence) => {
+const checkSpecialExamRedirect = memoize((sequenceStatus, sequence, specialExamsEnabled, proctoredExamsEnabled) => {
   if (sequenceStatus === 'loaded') {
-    if (sequence.isTimeLimited && sequence.legacyWebUrl !== undefined) {
+    const shouldRedirectTimeLimited = sequence.isTimeLimited && !specialExamsEnabled;
+    const shouldRedirectProctored = sequence.isProctored && !proctoredExamsEnabled;
+    if ((shouldRedirectTimeLimited || shouldRedirectProctored) && sequence.legacyWebUrl !== undefined) {
       global.location.assign(sequence.legacyWebUrl);
     }
   }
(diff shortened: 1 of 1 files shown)


openedx-webhooks:
Thanks for the pull request, @ViktorRusakov! I've created [BLENDED-911](https://openedx.atlassian.net/browse/BLENDED-911) to keep track of it in Jira.   More details are on the [BD-26](https://openedx.atlassian.net/wiki/spaces/COMM/pages/2523627980/BD-26+Proctoring+MFE+and+configuration) project page.


src/courseware/CoursewareContainer.jsx
```diff
@@ -59,9 +59,11 @@ const checkUnitToSequenceUnitRedirect = memoize((courseStatus, courseId, sequenc
   }
 });
 
-const checkSpecialExamRedirect = memoize((sequenceStatus, sequence) => {
+const checkSpecialExamRedirect = memoize((sequenceStatus, sequence, specialExamsEnabled, proctoredExamsEnabled) => {
   if (sequenceStatus === 'loaded') {
-    if (sequence.isTimeLimited && sequence.legacyWebUrl !== undefined) {
+    const shouldRedirectTimeLimited = sequence.isTimeLimited && !specialExamsEnabled;
+    const shouldRedirectProctored = sequence.isProctored && !proctoredExamsEnabled;
+    if ((shouldRedirectTimeLimited || shouldRedirectProctored) && sequence.legacyWebUrl !== undefined) {
```
zacharis278, line 66: does memoize look at a passed in objects properties? I'm wondering if the sequence is 'loaded' but the webUrl is undefined this function won't recalculate when webUrl is resolved (since it isn't an argument).

If that's not an issue this looks good to me

viktorrusakov: hm, didn't think about it, and judging by the documentation it uses simple === to check for equality

I'm not sure if it's an issue since the only way sequence object can be updated is by calling the API which will trigger sequenceLoaded to change from 'loaded' to 'loading' and that will trigger checkSpecialExamRedirect to be evaluated again, and then after the new sequence is fetched the status will again change to 'loaded' and checkSpecialExamRedirect will get called again (since memoize only remembers last values of the arguments)

I might be wrong though, I can try to somehow update the function so we know for sure it is not an issue if you want

I also don't see it being addressed in checkSequenceToSequenceUnitRedirect function, which might have the same issue

zacharis278: I thought I had seen that occurring before but looking at the code again I'm coming to the same conclusion as you. So let's just merge this.

zacharis278 (approved):

PR #527, opened 2021-07-07
REV-2132: remove FPD and audit upgrade banners from courseware page

dianekaplan:
## Description
The Value Prop redesign streamlines our messaging so we won't need so many different upsell links. Here we remove the first purchase discount and the 'audit access expires' banners (aka offerAlert and accessExpirationAlert in the learning MFE, and EDXWELCOME banner aka course_home_welcome upsell link**) from the learning MFE courseware page because they'll be redundant once the UpgradeNotification box appears there.  offerAlert is no longer used, so I'm removing that component, but leaving accessExpirationAlert in place since it's still used in masquerade mode. 

** see upsell link reference [here](https://openedx.atlassian.net/wiki/spaces/RS/pages/1675100377/How+to+find+upsell+link+click+events)

## Deadline
None- this upsell link will be redundant once the new expiration box is live on the courseware page (this sprint), so removing it will avoid clutter/duplication

## Supporting information
Relevant Jira ticket: https://openedx.atlassian.net/browse/REV-2132

See screenshots below for before-and-after with 'access expires' banner. The EDXWELCOME banner wasn't working locally, but we already know each offer's name- I set offerAlert to the other banner for local testing, but I have a prod example showing both to verify when it's time. 

Before: 
before

After: 
after


src/courseware/course/Course.jsx
```diff
@@ -82,17 +75,6 @@ function Course({
       <Helmet>
         <title>{`${pageTitleBreadCrumbs.join(' | ')} | ${getConfig().SITE_NAME}`}</title>
       </Helmet>
-      { /** This conditional is for the [MM-P2P] Experiment */}
-      { !MMP2P.state.isEnabled && (
```
julianajlk, line 83: I believe the MMP2P code was not removed from Brian's previous PR, and I would imagine it should not be removed here either? I would check with them. If you have already, disregard!

src/courseware/course/Course.test.jsx
```diff
@@ -112,92 +111,6 @@ describe('Course', () => {
...
-      org_key: 'edX',
-      courserun_key: courseMetadata.id,
-      linkCategory: 'FBE_banner',
-      linkName: 'in_course_audit_access_expires',
-      linkType: 'link',
-      pageName: 'in_course',
-    });
-  });
-
-  it('sends analytics event onClick of offer alert link', async () => {
```
julianajlk, line 157: These onClick even tracking (offer and access expiration) are being tested in the `UpgradeNotification` (aka `UpgradeCard`)? I'm assuming removing the tests from here it would still test that in the separate `UpgradeNotification` component tests.

julianajlk (commented):
Are we still keeping the function `useAccessExpirationAlert`? Is it being used somewhere else other than `Course.jsx`?

src/courseware/course/Course.jsx
dianekaplan: Good call- I'll ask! My thought process was if customAlerts no longer has anything in it, then <AlertList> can be removed, and if AlertList can be removed then this conditional can be removed too- but I'll confirm!

src/courseware/course/Course.test.jsx
dianekaplan: These tests are specifically for the banners that appear in the courseware ("in_course_audit_access_expires" and "in_course_welcome")- they wouldn't have any impact on any tests covering other upsell links. 

As for the UpgradeNotification, good question! The upgrade notification on the course home has its own test (can find by searching 'name course_home_green'), but that's a good point that we'll need to name it/track it from the in-course page too.

src/courseware/course/Course.jsx
dianekaplan: Update: Ben W confirmed this is true/correct

dianekaplan:
> Are we still keeping the function `useAccessExpirationAlert`? Is it being used somewhere else other than `Course.jsx`?

yep! As per the description this is still used in masquerade mode (to tell admins what date the learner's access expired on), so I left accessExpirationAlert in place

julianajlk (approved):

PR #532, opened 2021-07-08
REV-2304: add upsell events for progress tab upsell links

dianekaplan:
The new learning MFE version of the progress tab has two new upsell links: 
1) “Upgrade now” link in ‘Preview of a locked feature’ banner (currently sends: edx.ui.lms.course_progress.grades_upgrade.clicked)
2) “Upgrade now” link in ‘Earn a certificate’ square (currently sends edx.ui.lms.course_progress.certificate_status.clicked with a certificate_status_variant of audit_passing)

When a learner clicks either of those links, we want them to also send our typical "upsell click" event: edx.bi.ecommerce.upsell_links_clicked

image

### DIFF
--- a/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
+++ b/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
@@ -46,6 +46,12 @@ function CertificateStatus({ intl }) {
     isEnrolled,
     userHasPassingGrade,
   );
+
+  const eventProperties = {
+    org_key: org,
+    courserun_key: courseId,
+  };
+
   const dispatch = useDispatch();
   const { administrator } = getAuthenticatedUser();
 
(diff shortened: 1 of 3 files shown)


src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
```diff
@@ -198,6 +204,15 @@ function CertificateStatus({ intl }) {
       is_staff: administrator,
       certificate_status_variant: certEventName,
     });
+    if (certCase === 'upgrade') {
```
dianekaplan, line 207: This card/button has 7 possible states and sets of verbiage, etc. We only want to count a click as an 'upsell click' if we're in the state of showing 'Upgrade now', etc.

emma-green (approved):

PR #533, opened 2021-07-09
ignore for now

MatthewPiatetsky:
feat: disable course goals UI as part of transition to new goals feature


openedx-webhooks:
Although this pull request is already closed, I've created [OSPR-6501](https://openedx.atlassian.net/browse/OSPR-6501) so that we can track it in Jira.

PR #538, opened 2021-07-12
feat: add problem scores to progress tab (AA-875)

ciduarte:
differs from mock, but cleared by Kavya
Screen Shot 2021-07-15 at 1 24 33 PM


MatthewPiatetsky (commented):
does this need any tests?

src/course-home/progress-tab/grades/detailed-grades/SubsectionTitleCell.jsx
```diff
@@ -0,0 +1,74 @@
...
+import { ArrowDropDown, ArrowDropUp } from '@edx/paragon/icons';
+
+import messages from '../messages';
+import { useModel } from '../../../../generic/model-store';
+import ProblemScoreDrawer from './ProblemScoreDrawer';
+
+function SubsectionTitleCell({ intl, subsection }) {
+  const {
+    courseId,
+    gradesFeatureIsLocked,
```
mikix: This still feels like a weird place to pull this from.

src/course-home/progress-tab/grades/detailed-grades/SubsectionTitleCell.jsx
```diff
@@ -0,0 +1,74 @@
...
+        >
+          <Collapsible.Visible whenClosed><Icon src={ArrowDropDown} /></Collapsible.Visible>
+          <Collapsible.Visible whenOpen><Icon src={ArrowDropUp} /></Collapsible.Visible>
+        </Collapsible.Trigger>
+        <a
+          href={url}
+          className="muted-link small"
+          onClick={() => {
+            logSubsectionClicked(blockKey);
+          }}
```
mikix: Minor nit: you could include `blockKey` in logSubsectionClicked's closure and have it not take any arguments. Then you could simply do `onClick={logSubsectionClicked}` here.

src/course-home/progress-tab/grades/detailed-grades/ProblemScoreDrawer.jsx
```diff
@@ -0,0 +1,27 @@
...
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+
+import messages from '../messages';
+
+function ProblemScoreDrawer({ intl, problemScores }) {
+  return (
+    <span className="w-100 m-0 x-small ml-4 pt-2 pl-1 text-gray-700">
+      {intl.formatMessage(messages.problemScoreLabel)}
+      {problemScores.map(problemScore => (
+        <span className="ml-3.5">{problemScore.earned}/{problemScore.possible}</span>
```
mikix: These are never undefined or can cause division by zero?

mikix (commented):
To Matthew's question about tests, will we see this stuff in our day to day dev work? Probably not with grade data? I could see an argument for a test that ensures we show reasonable values?

src/course-home/progress-tab/grades/detailed-grades/ProblemScoreDrawer.jsx
MatthewPiatetsky: i think this is just displaying a ratio in text form not actually doing the math

mikix: Ah yes - I hadn't noticed the internal braces, good call.

ciduarte: Oh there isn't any math happening here, this is just text in the DOM

src/course-home/progress-tab/grades/detailed-grades/SubsectionTitleCell.jsx
ciduarte: I'm not sure what you mean, do you mind elaborating?

mikix: I just remember when someone (you?) added that bit, we had a minor convo about where it should live - and it living in state rather than models bumped me oddly. But it's fine as is. I just bumped against it again when scanning this file.

ciduarte: Ooooh okay I revisited that PR to refresh my own memory.

The conversation at the time was about refactoring this field from passing it down through all of the components, to storing it in state so it could be grabbed from one spot (storing it in state was my gut reaction). We didn't talk about state vs models, but I agree with you that this should probably live in the `progress` model (will update to reflect that)

src/course-home/data/api.js
```diff
@@ -245,6 +245,8 @@ export async function getProgressTabData(courseId, targetUserId) {
     // in order to preserve a course team's desired grade formatting.
     camelCasedData.gradingPolicy.gradeRange = data.grading_policy.grade_range;
 
+    camelCasedData.gradesFeatureIsLocked = camelCasedData.completionSummary.lockedCount > 0;
```
ciduarte, line 248: Slight refactor to pull this out of the state and into the progress model.

MatthewPiatetsky (approved):
looks good, a test would be good
(also screenshots would be nice)

PR #539, opened 2021-07-13
[AA-886] feat: handle FBE Exception cases on new progress page

MatthewPiatetsky:
implementing these [mocks](https://www.figma.com/file/KFz3P2Is0nmIBzy4nf4gJy/Progress-tab?node-id=935%3A5661)
for this [ticket](https://openedx.atlassian.net/browse/AA-886)
Limited Access
download (6)

Also the other states:

Locked Access
download (5)
Full Access
download (4)

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@edx/frontend-enterprise": "4.2.3",
     "@edx/frontend-lib-special-exams": "1.9.0",
     "@edx/frontend-platform": "1.11.0",
-    "@edx/paragon": "15.2.2",
+    "@edx/paragon": "16.2.0",
     "@fortawesome/fontawesome-svg-core": "1.2.34",
     "@fortawesome/free-brands-svg-icons": "5.13.1",
     "@fortawesome/free-regular-svg-icons": "5.13.1",
--- a/src/course-home/data/slice.js
+++ b/src/course-home/data/slice.js
@@ -10,7 +10,6 @@ const slice = createSlice({
   initialState: {
     courseStatus: 'loading',
     courseId: null,
-    gradesFeatureIsLocked: false,
     toastBodyText: null,
     toastBodyLink: null,
     toastHeader: '',
@@ -39,9 +38,6 @@ const slice = createSlice({
       state.toastBodyText = linkText;
       state.toastHeader = header;
     },
-    setGradesFeatureStatus: (state, { payload }) => {
-      state.gradesFeatureIsLocked = payload.gradesFeatureIsLocked;
-    },
   },
 });
 
@@ -50,7 +46,6 @@ export const {
   fetchTabSuccess,
   fetchTabFailure,
   setCallToActionToast,
-  setGradesFeatureStatus,
 } = slice.actions;
 
 export const {
--- a/src/course-home/progress-tab/ProgressTab.jsx
+++ b/src/course-home/progress-tab/ProgressTab.jsx
@@ -18,10 +18,10 @@ function ProgressTab() {
   } = useSelector(state => state.courseHome);
 
   const {
-    gradesFeatureIsLocked,
+    gradesFeatureIsFullyLocked,
   } = useModel('progress', courseId);
 
-  const applyLockedOverlay = gradesFeatureIsLocked ? 'locked-overlay' : '';
+  const applyLockedOverlay = gradesFeatureIsFullyLocked ? 'locked-overlay' : '';
 
   const layout = layoutGenerator({
     mobile: 0,
@@ -41,7 +41,7 @@ function ProgressTab() {
             <CertificateStatus />
           </OnMobile>
           <CourseGrade />
-          <div className={`grades my-4 p-4 rounded shadow-sm ${applyLockedOverlay}`} aria-hidden={gradesFeatureIsLocked}>
+          <div className={`grades my-4 p-4 rounded shadow-sm ${applyLockedOverlay}`} aria-hidden={gradesFeatureIsFullyLocked}>
             <GradeSummary />
             <DetailedGrades />
           </div>
--- a/src/course-home/progress-tab/grades/grade-summary/AssignmentTypeCell.jsx
+++ b/src/course-home/progress-tab/grades/grade-summary/AssignmentTypeCell.jsx
@@ -1,18 +1,26 @@
 import React from 'react';
 import { useSelector } from 'react-redux';
 import PropTypes from 'prop-types';
+import { Blocked } from '@edx/paragon/icons';
+import { Icon } from '@edx/paragon';
 import { useModel } from '../../../../generic/model-store';
 
-function AssignmentTypeCell({ assignmentType, footnoteMarker, footnoteId }) {
+function AssignmentTypeCell({
+  assignmentType, footnoteMarker, footnoteId, locked,
+}) {
   const {
     courseId,
   } = useSelector(state => state.courseHome);
+
   const {
-    gradesFeatureIsLocked,
+    gradesFeatureIsFullyLocked,
   } = useModel('progress', courseId);
+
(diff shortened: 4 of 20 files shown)


src/course-home/progress-tab/ProgressTab.jsx
```diff
@@ -22,14 +22,32 @@ function ProgressTab() {
...
 
-  const gradesFeatureIsLocked = lockedCount > 0;
-  const applyLockedOverlay = gradesFeatureIsLocked ? 'locked-overlay' : '';
+  let gradesFeatureIsFullyLocked = lockedCount > 0;
+  let gradesFeatureIsPartiallyLocked = false;
+  if (gradesFeatureIsFullyLocked) {
+    sectionScores.forEach((chapter) => {
+      chapter.subsections.forEach((subsection) => {
+        // If something is eligible to be gated by content type gating,
+        if (subsection.assignmentType !== null && subsection.hasGradedAssignment && subsection.numPointsPossible > 0) {
```
MatthewPiatetsky, line 34: I think this should cover most cases where content could be content type gated - but i'm not convinced that this condition covers all the cases.
it may be necessary to pass through additional data from the backend on whether the problem would be eligible for content type gating (though this wouldn't change much besides this condition)

package.json
```diff
@@ -40,7 +40,7 @@
     "@edx/frontend-enterprise": "4.2.3",
     "@edx/frontend-lib-special-exams": "1.8.3",
     "@edx/frontend-platform": "1.11.0",
-    "@edx/paragon": "15.2.2",
+    "@edx/paragon": "16.2.0",
```
ciduarte, line 43: Did you check the repo for any impacts from these breaking changes? https://github.com/edx/paragon/releases/tag/v16.0.0

MatthewPiatetsky: I don't see Pagination/buttonLabel/buttonLabels appearing in frontend-app-learning, so I don't think there is an impact based on that message

src/course-home/data/slice.js
```diff
@@ -40,7 +40,8 @@ const slice = createSlice({
       state.toastHeader = header;
     },
     setGradesFeatureStatus: (state, { payload }) => {
-      state.gradesFeatureIsLocked = payload.gradesFeatureIsLocked;
+      state.gradesFeatureIsFullyLocked = payload.gradesFeatureIsFullyLocked;
+      state.gradesFeatureIsPartiallyLocked = payload.gradesFeatureIsPartiallyLocked;
```
ciduarte: Realized an oops from my PR, all of the grades feature code should have been removed from this state slice when I refactored. Would you mind removing this?

Shouldn't impact anything in your PR as you're pulling `gradesFeatureIsPartiallyLocked` from the progress model.

src/course-home/progress-tab/grades/grade-summary/AssignmentTypeCell.jsx
```diff
@@ -7,20 +7,22 @@ function AssignmentTypeCell({ assignmentType, footnoteMarker, footnoteId }) {
...
+
   const {
-    gradesFeatureIsLocked,
+    gradesFeatureIsFullyLocked,
   } = useModel('progress', courseId);
+
   return (
     <div className="small">
-      {assignmentType}
+      <span className="row ml-0 mr-0 d-inline-flex">{assignmentType}</span>
```
ciduarte: Adding the class name `row` makes this `d-flex` under the hood, with negative margins on the left and right side, so using it w/ another `d-*` class is redundant. Just the `d-inline-flex` should do (and then you should be able to remove the `ml-0` and `mr-0` as well).

Also, random tip - the class name `mx-0` could replace `ml-0` and `mr-0` because it applies the margin on the x-axis (right and left), and `my-0` also exists for the y-axis (top and bottom, `mt-0` and `mb-0` respectively)

src/course-home/progress-tab/grades/grade-summary/GradeSummaryTable.jsx
```diff
@@ -44,11 +60,14 @@ function GradeSummaryTable({ intl }) {
...
+
     return {
-      type: { footnoteId, footnoteMarker, type: assignment.type },
-      weight: `${(assignment.weight * 100).toFixed(0)}%`,
-      grade: `${(assignment.averageGrade * 100).toFixed(0)}%`,
-      weightedGrade: `${(assignment.weightedGrade * 100).toFixed(0)}%`,
+      type: { footnoteId, footnoteMarker, type: <>{lockedIcon}{assignment.type}</> },
+      weight: <span className={lockedClass}>{(assignment.weight * 100).toFixed(0)}%</span>,
+      grade: <span className={lockedClass}>{(assignment.averageGrade * 100).toFixed(0)}%</span>,
+      weightedGrade: <span className={lockedClass}>{(assignment.weightedGrade * 100).toFixed(0)}%</span>,
```
ciduarte: I think I'd rather see the lockedClass styling applied within the DataTable arrangement. I'll show you what I mean

```
isLockedAssignmentType = !gradesFeatureIsFullyLocked && hasNoAccessToAssignmentsOfType(assignment.type);

return {
    type: { footnoteId, footnoteMarker, type: assignment.type, locked: isLockedAssignmentType },
    weight: { weight: (assignment.weight * 100).toFixed(0)}%, locked:  isLockedAssignmentType, },
    ...
}
```

Then in the DataTable, for the **assignment type**
```
Cell: ({ value }) => (
  <AssignmentTypeCell
    locked={value.locked}
    ...
  />
),
```

Then you'd have to update `AssignmentTypeCell` to apply the icon accordingly.

And for the rest of the cells:
```
Cell: ({ value }) => (
  <span className={value.locked ? 'locked-overlay' : ''}>{value.weight}</span>
),
```

This way the values passed to the DataTable don't have to be concerned with their own styling (which is why something like `AssignmentTypeCell` was created in the first place).

src/course-home/progress-tab/grades/grade-summary/GradeSummaryTable.jsx
```diff
@@ -14,12 +15,14 @@ import messages from '../messages';
 function GradeSummaryTable({ intl }) {
   const {
     courseId,
+    gradesFeatureIsFullyLocked,
```
ciduarte: Just pointing out that this is coming from the state, and once we remove that state logic, it should be coming from the progress model on line 26.

ciduarte (commented):
A few comments but overall looks really good so far!

src/course-home/data/slice.js
ciduarte: Noticed at least one spot actually where you pull from state rather than progress model

MatthewPiatetsky: removed the function and fixed that spot

src/course-home/progress-tab/grades/grade-summary/AssignmentTypeCell.jsx
MatthewPiatetsky: thanks good to know! fixed

src/course-home/progress-tab/grades/grade-summary/GradeSummaryTable.jsx
MatthewPiatetsky: updated this section per these comments

src/course-home/progress-tab/grades/grade-summary/GradeSummaryTable.jsx
MatthewPiatetsky: thanks! removed

src/course-home/data/slice.js
```diff
@@ -10,7 +10,7 @@ const slice = createSlice({
   initialState: {
     courseStatus: 'loading',
     courseId: null,
-    gradesFeatureIsLocked: false,
+    gradesFeatureIsFullyLocked: false,
```
ciduarte: I think we should remove this as well!

ciduarte (approved):
Just one more thing that needs to be removed from the state but otherwise looks good to go!

PR #540, opened 2021-07-13
feat: Make courseware URLs more human-readable

KristinAoki:
This PR updates the format of MFE URLs. Previously the URLs were extremely long and had repetitive information.  This made it difficult for users to read and easily understand where they are within a course based on the URL. Now, the URL removes all the repetitive information and replaces it with a unique identifier. Updated all API call URLs with a decoded version of block id since unique identifier is being pulled from URL, but does not translate to any API in LMS.

This PR is connected to the edx-plarform [PR-28174](https://github.com/edx/edx-platform/pull/28174/)


docs/decisions/0009-courseware-url-shortening.md
```diff
@@ -0,0 +1,58 @@
...
+
+After exploring different URL patterns and possible redundancies in the current URL format. The course, run, and organization is stated in every portion of the URL. We also do not need the URL to tell us the type of block since it has been determined that all URLs will follow the path` /course/:courseId/:sequenceId/:unitId`.
+
+## Decision
+
+The courseware URL will format to the following structure:
+
+```
+
+https://learning.edx.org/c/:courseId/:sequenceHash/:unitHash/:sectionSlug/:sequenceSlug/:unitSlug/
```
ormsbee, line 29: Are we ever going to allow partial paths? Or will all other path types forward to this?

docs/decisions/0009-courseware-url-shortening.md
```diff
@@ -0,0 +1,58 @@
...
+
+```
+
+Example URL:
+
+```
+
+https://learning.edx.org/c/course-v1:edX+DemoX.1+2T2019/YmxvY2/njuRCq/Optional_Example_Problem_Types/STEM_Problems/Code_Grader
```
ormsbee: What is used to generate the slug fields? Django's [slugify](https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.text.slugify)? If so, it's going to move things to lowercase and dashes, so the slug portion would look like: `optional-example-problem-types/stem-problems/code-grader`.

Are Unicode identifiers going to be accepted in the slug? (Django supports it).

docs/decisions/0009-courseware-url-shortening.md
```diff
@@ -0,0 +1,58 @@
...
+The fields definition and requirements ar as follows:
+
+* :courseId (required) - same as the previous `courseId`.
+* :sequenceHash (required) - a `urlsafe_b64encode` version of the `sequenceId`.
+* :unitHash (required) - a `urlsafe_b64encode` version of the `unitId`.
+* :sectionSlug (optional) - `display_name` of the current sequence's parent section.
+* :sequenceSlug (optional) - `display_name` of the current sequence.
+* :unitSlug (optional) - `display_name` of the current unit
+
+The slugs based on `display_name` are optional because not all blocks have an associated `display_name` attributes, most likely to occur in OLX imports. The `sequenceHash` and `unitHash` will shorten their respective ids using `hashlib.blake2b`. `Blake2b` will reduce the length of the id so the encoded version can also be short. Hashing will be handled by `blake2b` because it is the fastest hashing function in the `hashlib` library. The hash will be generated and mapped in LMS.
```
ormsbee: We have a `display_name_with_default` attribute that will guarantee that something gets put there, FWIW.

docs/decisions/0009-courseware-url-shortening.md
```diff
@@ -0,0 +1,58 @@
...
+https://learning.edx.org/c/course-v1:edX+DemoX.1+2T2019/YmxvY2/njuRCq/Optional_Example_Problem_Types/STEM_Problems/Code_Grader
+
+```
+
+The fields definition and requirements ar as follows:
+
+* :courseId (required) - same as the previous `courseId`.
+* :sequenceHash (required) - a `urlsafe_b64encode` version of the `sequenceId`.
+* :unitHash (required) - a `urlsafe_b64encode` version of the `unitId`.
+* :sectionSlug (optional) - `display_name` of the current sequence's parent section.
```
ormsbee, line 46: Does the Course name get its own slug?

docs/decisions/0009-courseware-url-shortening.md
ormsbee: > blake2b

Please specify the digest size (6 bytes?).

docs/decisions/0009-courseware-url-shortening.md
```diff
@@ -0,0 +1,58 @@
...
+* :sequenceSlug (optional) - `display_name` of the current sequence.
+* :unitSlug (optional) - `display_name` of the current unit
```
ormsbee, line 48: When slugs are omitted (or they're old slugs because the name has changed since the link was created) do they redirect to the canonical version with slugs?

docs/decisions/0009-courseware-url-shortening.md
ormsbee: If we're going to allow partial paths (e.g. a sequence hash without a unit hash), please consider combining the significant parts with a different delimiter, like a period:

```
https://learning.edx.org/c/course-v1:edX+DemoX.1+2T2019/YmxvY2.njuRCq/Optional_Example_Problem_Types/STEM_Problems/Code_Grader
```

That way there's no ambiguity–we know that the segment after the course key is going to be the significant ID portion of it, and everything that follows is for readability. We don't have to worry that maybe "njuRCq" is a hash ID or maybe it's the first part of the decorative-stuff-to-be-ignored.

docs/decisions/0009-courseware-url-shortening.md
KristinAoki: I was not originally going to include it, but if you think I should, I will.

ormsbee: @mondiaz, @JAAkana: Any thoughts on whether a slug of the course name belongs in courseware URLs, or is that just making it longer without providing sufficient value?

JAAkana: I have a mild opinion to rely first on  courseID since the org/number/run are unique and convey human-ish info. Course names can be long and not-unique (i.e. multiple runs), I believe names  re-cover ground captured in the courseID.

docs/decisions/0009-courseware-url-shortening.md
```diff
@@ -0,0 +1,58 @@
...
+
+This is what the URLs currently look like:
+
+```
+
+https://learning.edx.org/course/course-v1:edX+DemoX.1+2T2019/block-v1:edX+DemoX.1+2T2019+type@sequential+block@e0a61b3d5a2046949e76d12cac5df493/block-v1:edX+DemoX.1+2T2019+type@vertical+block@52dbad5a28e140f291a476f92ce11996
+
+```
+
+After exploring different URL patterns and possible redundancies in the current URL format. The course, run, and organization is stated in every portion of the URL. We also do not need the URL to tell us the type of block since it has been determined that all URLs will follow the path` /course/:courseId/:sequenceId/:unitId`.
```
kdmccormick: > After exploring different URL patterns and possible redundancies in the current URL format. 

Consider rewording this incomplete sentence.

src/course-home/data/api.js
```diff
@@ -167,7 +168,7 @@ export function normalizeOutlineBlocks(courseId, blocks) {
     if (Array.isArray(section.sequenceIds)) {
       section.sequenceIds.forEach(sequenceId => {
         if (sequenceId in models.sequences) {
-          models.sequences[sequenceId].sectionId = section.id;
+          models.sequences.[sequenceId].sectionId = section.id;
```
kdmccormick: Huh, I've never seen the `.[` syntax before. What does that do?

docs/decisions/0009-courseware-url-shortening.md
```diff
@@ -0,0 +1,58 @@
+# Courseware URL shortening
```
kdmccormick, line 1: Thanks for including an ADR! It's already made this change easier for me to review, and it should be helpful for devs who want to understand this change in the future.

docs/decisions/0009-courseware-url-shortening.md
```diff
@@ -0,0 +1,58 @@
...
+
+https://learning.edx.org/c/course-v1:edX+DemoX.1+2T2019/YmxvY2/njuRCq/optional-example-problem-types/stem-problems/code-grader
+
+```
+
+The fields definition and requirements ar as follows:
+
+* :courseId (required) - same as the previous `courseId`.
+* :sequenceHash (required) - a `urlsafe_b64encode` version of the `sequenceId`.
+* :unitHash (required) - a `urlsafe_b64encode` version of the `unitId`.
```
kdmccormick: Are these actually the urlsafe_b64encoded versions *of the blake2b hash* of the ids?

docs/decisions/0009-courseware-url-shortening.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Consequences
+
+If old URLs are not properly routed then the content and those links will no longer be accessible to the user. The old URLs could include, but not limited to, bookmarks and exams.
```
kdmccormick, line 54: From the ticket's acceptance criteria:

> * First, implement the proposal above, but redirect the new URLs to old URLs. So, there will be no user-facing change yet, since users will still only know about the old URLs.
> * When we have confidence that the new URLs are working, make it so old URLs redirect new URLs instead.
> * ...
> * Finally, update https://github.com/edx/edx-platform/blob/c194b3fa573ef730c5a228acbea0ce6823519b8c/openedx/features/course_experience/url_helpers.py#L147, which builds URLs for frontend redirection, to use the new URLs.
> * It may be smart to leave in support for the old URLs (redirecting them to the new ones) forever. This would avoid breaking bookmarks, old links, etc. It would also allow us to link to the Learning MFE without computing the unit_key_hash , for debugging or whatever.

It seems like this PR jumps to Step 2, in which the new URLs become the canonical URLs. That's OK, but I do think it's important that the old URLs continue to work, at least for now. Ensuring that the old URLs don't break will significantly de-risk the rollout of this change.

src/courseware/course/sequence/content-lock/ContentLock.jsx
```diff
@@ -12,7 +12,7 @@ function ContentLock({
   intl, courseId, prereqSectionName, prereqId, sequenceTitle,
 }) {
   const handleClick = useCallback(() => {
-    history.push(`/course/${courseId}/${prereqId}`);
+    history.push(`/c/${courseId}/${prereqId}`);
```
kdmccormick, line 15: Wow, half of the tests in this repo must have broken when you made this change! Kudos to you for working through those.

docs/decisions/0009-courseware-url-shortening.md
```diff
@@ -0,0 +1,58 @@
...
+
```
kdmccormick, line 55: The changed URLs may have an impact on currently-running experiments, which often depend on the URL scheme. I'll follow up on this.

kdmccormick (changes requested):
Good work so far. Of my comments below, the only one that's a real concern to me is the one about the old URLs still working.

Do you feel like you'll be able to integrate that requirement into your project? Happy to chat if you think it would increase the scope too much--maybe there is a way the team can help.

src/course-home/data/api.js
KristinAoki: Definitely a typo, thanks for catching1

bradenmacdonald:
I'm gonna close this PR - it's definitely not going to be merged after all this time. Would still be nice to see something like this done eventually though.

PR #544, opened 2021-07-15
fix: Corrected status message with incorrect mention of email

staubina:


src/course-home/progress-tab/certificate-status/messages.js
```diff
@@ -61,10 +61,6 @@ const messages = defineMessages({
     id: 'progress.certificateStatus.notAvailableHeader',
     defaultMessage: 'Certificate status',
   },
-  notAvailableBody: {
```
staubina, line 64: I do not see this string being used anywhere

crice100 (approved):

PR #548, opened 2021-07-19
[feat] MB-1192: Add not passing, course ended status to certificate status alert

Tj-Tracy:
image

mobile:
image

### DIFF
--- a/src/course-home/data/__snapshots__/redux.test.js.snap
+++ b/src/course-home/data/__snapshots__/redux.test.js.snap
@@ -461,6 +461,7 @@ Object {
           "canEnroll": true,
           "extraText": "Contact the administrator.",
         },
+        "enrollmentMode": undefined,
         "handoutsHtml": "<ul><li>Handout 1</li></ul>",
         "hasEnded": undefined,
         "hasScheduledContent": null,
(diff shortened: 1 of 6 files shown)


src/course-home/outline-tab/alerts/certificate-status-alert/CertificateStatusAlert.jsx
```diff
@@ -118,6 +120,24 @@ function CertificateStatusAlert({ intl, payload }) {
     return alertProps;
   };
 
+  const renderNotPassingCourseEneded = () => {
```
oliviaruizknott: Typo? `Eneded` -> `Ended`

Tj-Tracy: fiexed :)

src/course-home/data/__snapshots__/redux.test.js.snap
```diff
@@ -461,6 +461,7 @@ Object {
           "canEnroll": true,
           "extraText": "Contact the administrator.",
         },
+        "enrollmentMode": undefined,
```
justinhynes, line 464: As someone who still isn't comfortable/proficient with Javascript... Is there any reason some things are defined with a default of `undefined` vs. `null`? (I realize this looks to be test data, I was curious)

justinhynes (approved):
LGTM!

src/course-home/data/__snapshots__/redux.test.js.snap
Tj-Tracy: the short of it is: `null` is an assigned state of the absence of value. Its usually used to explicitly state "this is deemed to have no value by some code". Where as, `undefined` is the default state given by javascript to things that implicitly have no value. But like everything else in JS, non of this is really that strict, and they basically mean the same thing.

justinhynes: Thanks!

PR #549, opened 2021-07-19
fix: LockPaywall content gating layout issue

julianajlk:
[REV-2307](https://openedx.atlassian.net/browse/REV-2307). 

There was a change in Paragon that caused the below layout bug (prod screenshots in [this course](https://learning.edx.org/course/course-v1:BerkeleyX+CS198.1x+3T2019/block-v1:BerkeleyX+CS198.1x+3T2019+type@sequential+block@c6e867c00e404e1095bcd91f9d145440/block-v1:BerkeleyX+CS198.1x+3T2019+type@vertical+block@f62bb9067471453c87533218f632470c)):
Screen Shot 2021-07-19 at 3 04 55 PM
Screen Shot 2021-07-19 at 3 04 41 PM

**This PR fixes this spacing issue in the gated content value prop banner:**
- Changed the size of the `UpgradeButton`.
- Used the `Alert` embedded icon i/o using an `Icon` in a separate `div`.
- Added column sizes and logic depending on whether the `NotificationTray` is open/closed and whether it's in the responsive screensize to determine if it should have 2 or 3 columns (the bullet points goes under the certificate image for 2 columns). 

**Option 1: following "expected" screenshot from the ticket with 2 columns.** 
Screen Shot 2021-07-19 at 3 39 03 PM
Screen Shot 2021-07-19 at 3 38 48 PM

**Option 2: 3 columns following the [Figma doc](https://www.figma.com/file/1EaMYJ8FecwzbbSv5YIpEC/Value-Prop-Final-Designs?node-id=1067%3A89):** **--> APPROVED BY UX**
Screen Shot 2021-07-19 at 5 35 08 PM
Screen Shot 2021-07-19 at 5 34 55 PM

**Responsive views:**
There are various screen sizes and whether the `NotificationTray` is open/closed, as well as if there's an `offer` (the `UpgradeButton` text will be longer). To account for all of these, several variables were defined and added on the last commit of this PR. 

- Extra large screen sizes (>1200px)
Screen Shot 2021-07-21 at 9 26 31 AM
Screen Shot 2021-07-21 at 9 24 56 AM

- Large screen sizes (>992px)
Screen Shot 2021-07-21 at 9 36 01 AM
Screen Shot 2021-07-21 at 9 35 12 AM

- Medium screen sizes (>768 px)
Screen Shot 2021-07-21 at 9 36 56 AM

- Small screen sizes (>576 px)
Screen Shot 2021-07-21 at 9 38 20 AM

- Extra small screen sizes (>375 px) Here if there's an `offer` the text will wrap
Screen Shot 2021-07-20 at 4 59 15 PM
Screen Shot 2021-07-21 at 9 39 26 AM

### DIFF
--- a/src/courseware/course/sequence/lock-paywall/LockPaywall.scss
+++ b/src/courseware/course/sequence/lock-paywall/LockPaywall.scss
@@ -1,3 +1,6 @@
+.lock-paywall-container svg {
+    color: #002121;
+}
 
 // Temporary CSS intervention until paragon list items will support icons (PAR-429)
 .fa-li {
--- a/src/courseware/course/sequence/lock-paywall/LockPaywall.scss
+++ b/src/courseware/course/sequence/lock-paywall/LockPaywall.scss
@@ -1,8 +1,7 @@
 .lock-paywall-container svg {
-    color: #002121;
+    color: $primary-700;
 }
 
-// Temporary CSS intervention until paragon list items will support icons (PAR-429)
 .fa-li {
     left: -31px !important;
     padding-right: 22px;
--- a/src/courseware/course/NotificationIcon.jsx
+++ b/src/courseware/course/NotificationIcon.jsx
@@ -5,7 +5,6 @@ import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import { Icon } from '@edx/paragon';
 import { WatchOutline } from '@edx/paragon/icons';
 
-import './NotificationIcon.scss';
 import messages from './messages';
 
 function NotificationIcon({ intl, status, notificationColor }) {
--- a/src/courseware/course/sequence/Sequence.jsx
+++ b/src/courseware/course/sequence/Sequence.jsx
@@ -204,6 +204,7 @@ function Sequence({
             sequenceId={sequenceId}
             unitId={unitId}
             unitLoadedHandler={handleUnitLoaded}
+            notificationTrayVisible={notificationTrayVisible}
             /** [MM-P2P] Experiment */
             mmp2p={mmp2p}
           />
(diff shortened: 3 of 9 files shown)


src/courseware/course/sequence/lock-paywall/LockPaywall.scss
```diff
@@ -1,3 +1,6 @@
+.lock-paywall-container svg {
+    color: #002121;
```
julianajlk: I should use `$primary-700` instead of hard coding the hex color here, but it's currently breaking even though I've added this SCSS file to index.scss imports

bmedx:
In the "after" screenshots it looks like the bullet point text is still not bumping up next to the certificate image. In the final image I would expect the bullet text to be between the certificate and the button. Is that just the size of the page in the screenshot or is that currently not working?

julianajlk:
> In the "after" screenshots it looks like the bullet point text is still not bumping up next to the certificate image. In the final image I would expect the bullet text to be between the certificate and the button. Is that just the size of the page in the screenshot or is that currently not working?

I'll check with UX, since the "expected" in the ticket has it as 2 columns (with the bullet points under the certificate), but the Figma doc has 3 columns. 
**_Edit: UX approved Option 2 (above screenshots), commit already has the code to follow this option_**

dianekaplan (approved):

src/courseware/course/sequence/lock-paywall/LockPaywall.scss
julianajlk: The reason was because `LockPaywall.scss` file was imported twice - once at `index.scss` and another at the `LockPaywall.jsx` component file. I removed the import in the component file.

bmedx (approved):

PR #550, opened 2021-07-20
fix: handle new 403 "redirect" responses from backend

mikix:
To make it easier to redirect the user to an another page or an access denied error page, this PR deprecates the old custom 401 & 404 error handling, replacing them both with a generic 403 error code that contains a URL to redirect the user to.

Before this change, we weren't handling all the error cases, and when we didn't, we ended up showing a broken page to the user.

Related to https://github.com/edx/edx-platform/pull/28235

### DIFF
--- a/src/course-home/data/api.js
+++ b/src/course-home/data/api.js
@@ -182,6 +182,26 @@ export function normalizeOutlineBlocks(courseId, blocks) {
   return models;
 }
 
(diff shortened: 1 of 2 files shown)


mikix:
Closing in favor of https://github.com/edx/frontend-app-learning/pull/558

PR #552, opened 2021-07-20
add edxwelcome code info to unhappy path

emma-green:
This pr is to solve an issue where if courses/users are not eligible for full FBE (content gating and course duration limits), the upsell messaging did not include a coupon code for the first purchase discount.
Here is the old state: 
image

Here is the new state:
Screen Shot 2021-07-21 at 10 05 34 AM


dianekaplan:
Adding for posterity- here's the relevant ticket: 
https://openedx.atlassian.net/browse/REV-2313

dianekaplan (approved):
This looks good to me!

PR #553, opened 2021-07-21
Merge Learning Sequences API results with Course Blocks API

ormsbee:
Switching basic structural data to come from Learning Sequences API, but still relying on the Course Blocks API for:

* Sequence legacy URL (do we still need this, since we're running the MFE for proctoring now?)
* Unit information (because of a perf regression if we switched entirely to SequenceMetadata API as it stands now)
* Effort Estimation (currently implemented through Course Blocks).

In order to see this code working, you need to enable the course waffle flag `learning_sequences.use_for_outlines` on your devstack.

### DIFF
--- a/src/courseware/data/thunks.js
+++ b/src/courseware/data/thunks.js
@@ -23,17 +23,113 @@ import {
   fetchSequenceFailure,
 } from './slice';
 
-// Make a copy of the sectionData and return it, but with the sequences filtered
-// down to only those sequences in allowedSequences
-function filterSequencesFromSection(sectionData, allowedSequences) {
-  return Object.fromEntries(
-    Object.entries(sectionData).map(
-      ([key, value]) => [
-        key,
-        (key === 'sequenceIds') ? value.filter(seqId => seqId in allowedSequences) : value,
-      ],
-    ),
-  );
+/**
+ * Combines the models from the Course Blocks and Learning Sequences API into a
+ * new models obj that is returned. Does not mutate the models passed in.
+ *
+ * For performance and long term maintainability, we want to switch as much of
+ * the Courseware MFE to use the Learning Sequences API as possible, and
+ * eventually remove calls to the Course Blocks API. However, right now, certain
+ * data still has to come form the Course Blocks API. This function is a
+ * transitional step to help build out some of the data from the new API, while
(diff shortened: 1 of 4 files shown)


ormsbee:
Tests still need to be written/updated, but this is the main code change I'm looking at.

src/courseware/data/thunks.js
```diff
@@ -23,17 +23,111 @@ import {
...
+  // course. I might end up doing that before my refactoring is done here. >_<
+  mergedModels.courses[courseBlockId] = {
+    // Learning Sequences API Data
+    id: lsCourse.id,
+    title: lsCourse.title,
+    sectionIds: lsCourse.sectionIds,
+    hasScheduledContent: lsCourse.hasScheduledContent,
+
+    // Still pulling from Course Blocks API
+    effortActivities: courseBlock.effortActivities,
```
connorhaugh, line 84: Is effort estimates something the Learning Sequences api will soon do? Or will another api (courseblocks?) have to do this in future?

connorhaugh: This in general looks good as a method to combine Sequences & Blocks apis. I'm a little confused by why we need to do so for Sequence legacy URL and Effort Estimation. See question RE: the latter above, and for the former, how do we know if and when Sequence legacy URL will no longer be required? Still getting context as to what is going on!

ormsbee: I don't think Learning Sequences should ever do those. This is needed for the course outline tab, and it should only come through there. We should probably have a separate call that gets effort estimation for a single sequence, which might call a Learning Sequence Python API on the backend (once Units are supported).

connorhaugh: gotcha sounds good

kenclary (commented):
Overall looks good. Waiting for tests, of course.

Is "remove calls to Course Blocks API" documented somewhere (maybe an upcoming epic?), or is there any other place we can put a pointer to these changes as tech debt?

ormsbee:
@kenclary:

> Is "remove calls to Course Blocks API" documented somewhere (maybe an upcoming epic?), or is there any other place we can put a pointer to these changes as tech debt?

That should be part of the ADR writeup in TNL-7326.

kenclary (approved):
Still looks good.

PR #554, opened 2021-07-21
fix: AA-912: If the URL is not provided, only show display name

Dillon-Dumesnil:
In https://github.com/edx/edx-platform/pull/28233, the logic was updated
to only return a URL if the content was still accessible to the learner.
This handles the case of the URL being null


Dillon-Dumesnil:
I know it all looks the same, but it doesn't get an underline on hover /shrug
image

ciduarte (approved):
Beautiful

PR #555, opened 2021-07-21
differentiate upsell link eventing between in_course and course_home

emma-green:
The Upgrade Notification component appears on two pages in the course: course home and inside of the notification tray inside the course. This component contains an upgrade link. Prior to this change, it was sending the same LinkName in upsell_link_clicked segment tracking events. This made it impossible to know where the user was upgrading from. This pr attempts to remedy that.

I also changed the order of the props in the UpgradeNotification component so that they would be in alphabetical order.

### DIFF
--- a/src/generic/upgrade-notification/UpgradeNotification.jsx
+++ b/src/generic/upgrade-notification/UpgradeNotification.jsx
@@ -265,10 +265,11 @@ function UpgradeNotification({
   courseId,
   offer,
   org,
+  shouldDisplayBorder,
   timeOffsetMillis,
+  upsellPageName,
   userTimezone,
   verifiedMode,
-  shouldDisplayBorder,
 }) {
   const timezoneFormatArgs = userTimezone ? { timeZone: userTimezone } : {};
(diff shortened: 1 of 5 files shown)


src/generic/upgrade-notification/UpgradeNotification.jsx
```diff
@@ -305,7 +306,7 @@ function UpgradeNotification({
     sendTrackEvent('edx.bi.ecommerce.upsell_links_clicked', {
       ...eventProperties,
       linkCategory: 'green_upgrade',
-      linkName: 'course_home_green',
+      linkName: upsellLinkName,
       linkType: 'button',
       pageName: 'course_home',
     });
```
dianekaplan, line 312: In addition to customizing the linkName, we'll also need the pageName (I think here it's still sending 'course_home' in both cases)

dianekaplan (changes requested):
Aside from pageName- (see comment)- everything looks great!

dianekaplan (approved):

PR #556, opened 2021-07-22
feat: Remove upgrade sock from course pages

bmedx:
REV-2220: The upgrade sock is being removed from the remaining course pages in favor of the new Value Prop work.


dianekaplan (approved):
Looks good to me! I wasn't able to verify that the sock is removed (it's missing with your branch, but also missing in master for me locally), but the code looks good, and here's a link in stage that's showing the sock for my user now and can be used for verification in stage before deploying: 
https://learning.stage.edx.org/course/course-v1:MichiganX+py4e101x+3T2020/block-v1:MichiganX+py4e101x+3T2020+type@sequential+block@05810886c7384a099d056e48f2ba7dd4/block-v1:MichiganX+py4e101x+3T2020+type@vertical+block@25ace819c4084517bb846f20e3274d0b

PR #558, opened 2021-07-22
fix: handle course access errors in Course Home side of things too

mikix:
The courseware was properly reading the access errors and redirecting the user as appropriate (like to the dashboard or whatever).

This requires a backend change to push the error along.

### DIFF
--- a/src/course-home/dates-tab/DatesTab.test.jsx
+++ b/src/course-home/dates-tab/DatesTab.test.jsx
@@ -23,19 +23,24 @@ jest.mock('@edx/frontend-platform/analytics');
 
 describe('DatesTab', () => {
   let axiosMock;
+  let store;
+  let component;
 
-  const store = initializeStore();
-  const component = (
-    <AppProvider store={store}>
-      <UserMessagesProvider>
-        <Route path="/course/:courseId/dates">
-          <TabContainer tab="dates" fetch={fetchDatesTab} slice="courseHome">
-            <DatesTab />
-          </TabContainer>
-        </Route>
-      </UserMessagesProvider>
-    </AppProvider>
-  );
+  beforeEach(() => {
+    axiosMock = new MockAdapter(getAuthenticatedHttpClient());
+    store = initializeStore();
+    component = (
+      <AppProvider store={store}>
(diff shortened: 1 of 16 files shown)


src/course-home/dates-tab/DatesTab.test.jsx
```diff
@@ -295,4 +298,45 @@ describe('DatesTab', () => {
...
+      render(component);
+      await waitForElementToBeRemoved(screen.getByRole('status'));
+    }
+
+    beforeEach(() => {
+      axiosMock.onGet(datesUrl).reply(200, datesTabData);
+      history.push(`/course/${courseId}/dates`); // so tab can pull course id from url
+    });
+
+    it('should go to course survey for an survey_required error code', async () => {
```
ciduarte: nit: `an` should be `a`

ciduarte: Another nit, can we change `should go` in all of these tests to `redirects`? Feels a little more clear when scanning

src/courseware/CoursewareContainer.jsx
```diff
@@ -338,10 +296,9 @@ const sectionShape = PropTypes.shape({
 });
 
 const courseShape = PropTypes.shape({
-  canLoadCourseware: PropTypes.shape({
-    errorCode: PropTypes.string,
-    additionalContextUserMessage: PropTypes.string,
-  }).isRequired,
+  celebrations: PropTypes.shape({
+    firstSection: PropTypes.bool,
+  }),
```
ciduarte, line 301: Did you add this in because there was a props warning or something?

ciduarte (approved):
Just some nits and a question! LGTM

src/courseware/CoursewareContainer.jsx
mikix: I didn't notice one - just was updating the shape to be accurate to what we currently use in this file.

src/course-home/dates-tab/DatesTab.test.jsx
mikix: :+1:

PR #560, opened 2021-07-26
fix: remove unused portions of effort estimate experiment

mikix:
We are sticking with the sequence version, and abandoning the section version. This commit also marks the strings for translation, as it is a real feature now, not just an experiment.

https://openedx.atlassian.net/browse/AA-659

### DIFF
--- a/src/shared/effort-estimate/EffortEstimate.jsx
+++ b/src/shared/effort-estimate/EffortEstimate.jsx
@@ -1,9 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import { FormattedMessage, injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+
+import messages from './messages';
 
 // This component shows an effort estimate provided by the backend block data. Either time, activities, or both.
-// Right now it is an experiment, and AA-659 is its cleanup ticket.
 
 function EffortEstimate(props) {
   const {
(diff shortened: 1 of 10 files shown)


src/shared/effort-estimate/EffortEstimate.jsx
```diff
@@ -12,27 +14,34 @@ function EffortEstimate(props) {
...
+  const activities = intl.formatMessage(messages.activities, { activityCount: effortActivities });
   let content = null;
 
   if (effortTime && effortActivities) {
-    content = <>{minutes} + {activities}</>;
+    content = (
+      <FormattedMessage
+        id="learning.effortEstimation.combinedEstimate"
+        defaultMessage="{minutes} + {activities}"
+        description="You can likely leave this alone, unless you want to use a full width plus or similar change"
```
MatthewPiatetsky, line 37: what does this mean

MatthewPiatetsky (commented):
it looks like codecov is expecting more tests?

src/shared/effort-estimate/EffortEstimate.jsx
mikix: So this is a translation of basically, "x + y" - which normally does not need a translation. But some languages (notably CJK languages that are full width) might want to do something with it. Like use the "fullwidth plus" sign: ＋ (which has different spacing than the normal plus: +). Or some other modification similar to that, that I don't know about.

mikix: This is also why you should try to include colons and other punctuation if possible inside the translation string. That is, preferring `_('Due date:')` to `_('Due date') + ':'` or whatever. Because CJK will want to use a fullwidth colon or period or whatever.

mikix:
OK, think I improved coverage a bit - at least project coverage is up now. Patch coverage is down because we don't explicitly test the "only activities" or "only minutes" code paths. But I don't think that's super important - those are trivial paths and something that devs will see in the normal course of viewing the page, if the paths do break or whatever.

MatthewPiatetsky:
> those are trivial paths and something that devs will see in the normal course of viewing the page, if the paths do break or whatever.

is there a flag that exists for this feature? i've found that most devs aren't aware that flags need to be enabled for features they're not working on, so then those features tend to stay off in devstack

mikix:
> is there a flag that exists for this feature? i've found that most devs aren't aware that flags need to be enabled for features they're not working on, so then those features tend to stay off in devstack

The context of this PR and this ticket is that we are removing the experiment and making it a default feature. We may leave an opt-out, but it will be on by default, is the plan.

MatthewPiatetsky (approved):

PR #563, opened 2021-07-27
fix: temporarily disable course-home denied redirects

mikix:
They caused a regression when not logged in or enrolled. Will hopefully re-enable after developing a fix.

### DIFF
--- a/src/course-home/data/thunks.js
+++ b/src/course-home/data/thunks.js
@@ -17,7 +17,7 @@ import {
 } from '../../generic/model-store';
 
 import {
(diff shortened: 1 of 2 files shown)


MatthewPiatetsky (approved):
assuming this was manually tested

mikix:
> assuming this was manually tested

Yeah on the unauthenticated path - was doing the redirect loop, then this fix stopped it (didn't do a full revert PR because the backend serialization format has also changed a bit - this was an easy hook to stop just the new code path)

PR #566, opened 2021-07-27
fix: Swap the prev and next icons in rtl

cdeery:
Right To Left (RTL) languages need to reverse the
direction of the icons in navigation.

Fixes: AA-891


ciduarte (approved):
Didn't know `isRtl()` existed! Nice find. Looks good to me 👍

PR #569, opened 2021-07-30
REV-2297: red dot functionality

dianekaplan:
[REV-2297](https://openedx.atlassian.net/browse/REV-2297) captures the following behavior for the red dot functionality: 

1) After the sidebar has shown for 3 seconds, red dot goes away
2) If the learner re-loads the browser or comes back, their state should be remembered (save in localStorage)
3) Then when the user reaches one of the next phases, the red dot should reappear so they read the new message, disappear after 3 seconds, and be remembered upon reload (like above). The  7 defined phases do not overlap, so whenever the current phase is different from the last seen, we bring back the red dot (and then as above it disappears after 3 seconds). See Figma docs linked in ticket- first we have 3 phases for first purchase discount (having days, hours, or minutes left til it expires), then 4 for audit access: more than a week left (show the expiration date), or days, hours, or minutes until expiration (reusing the same components for display). 

**I've added new variables to do this:** 
- _notificationStatus_ ('active' or 'inactive') for whether or not to show red dot. This is a prop for the current experience, and saved to local storage for persistence when the user comes back. 
- _upgradeNotificationCurrentState_: prop to capture the learner's notification phase (ex: days until first purchase discount expires, hours until audit access expires, etc)
- _upgradeNotificationLastSeen_: last phase they've had an upgrade notification for (localStorage-only, for checking when they come back)

**How it works:** 
1) notificationStatus starts as 'active' so a red dot will show on the NotificationTray icon. Then in NotificationTray, 3 seconds after it loads we call onNotificationSeen(), which sets the notification status to 'inactive' (the prop and in local storage)
2) The UpgradeNotification component is where we know what phase the learner is in, so we save it from there (localStorage for persistence, and the prop to trigger re-mounting). Note: where components are shared for both types of messages (first purchase discount and access expiration), I pass in a 'type' to differentiate ('offer' or 'access'). 
3) The NotificationTrigger checks the state when this mounts: if there's a upgradeNotificationCurrentState that's different from upgradeNotificationLastSeen, we set notificationStatus back to 'active', and update upgradeNotificationLastSeen with the value from upgradeNotificationCurrentState.  (This prop change makes it re-mount, and since notificationStatus is active, its NotificationIcon shows the red dot). 

**Testing:** 
Automated: npm test -- src/courseware/course/NotificationTrigger.test.js

Manual: 
1. log in with edx@example.com/edit, enroll as audit learner in demo course, and navigate to course home. 
2. From there, "View this course as Audit"
3. From outline, click a lesson to go to courseware
4. Result: NotificationTrigger shows the red dot icon at first, then switches to no-dot after three seconds
5. Masquerade mode doesn't have a way to mimic the various phases yet (see [TNL-8327](https://openedx.atlassian.net/browse/TNL-8327)), so I updated the code manually to see them

Screen Shot 2021-08-23 at 4 47 36 PM
Screen Shot 2021-08-23 at 4 48 08 PM

### DIFF
--- a/src/courseware/course/Course.jsx
+++ b/src/courseware/course/Course.jsx
@@ -15,6 +15,7 @@ import NotificationTrigger from './NotificationTrigger';
 
 import { useModel } from '../../generic/model-store';
 import useWindowSize, { responsiveBreakpoints } from '../../generic/tabs/useWindowSize';
+import { getLocalStorage, setLocalStorage } from '../../data/localStorage';
 
 /** [MM-P2P] Experiment */
 import { initCoursewareMMP2P, MMP2PBlockModal } from '../../experiments/mm-p2p';
@@ -60,6 +61,22 @@ function Course({
     if (notificationTrayVisible) { setNotificationTray(false); } else { setNotificationTray(true); }
   };
 
+  if (!getLocalStorage('notificationStatus')) {
+    setLocalStorage('notificationStatus', 'active'); // Show red dot on notificationTrigger until seen
+  }
+
+  if (!getLocalStorage('currentState')) {
+    setLocalStorage('currentState', 'initialize');
+  }
+
+  const [notificationStatus, setNotificationStatus] = useState(getLocalStorage('notificationStatus'));
+  const [currentState, setCurrentState] = useState(getLocalStorage('currentState'));
+
+  const onNotificationSeen = () => {
+    setNotificationStatus('inactive');
+    setLocalStorage('notificationStatus', 'inactive');
+  };
+
   /** [MM-P2P] Experiment */
   const MMP2P = initCoursewareMMP2P(courseId, sequenceId, unitId);
 
@@ -81,6 +98,9 @@ function Course({
           <NotificationTrigger
             toggleNotificationTray={toggleNotificationTray}
             isNotificationTrayVisible={isNotificationTrayVisible}
+            notificationStatus={notificationStatus}
+            setNotificationStatus={setNotificationStatus}
+            currentState={currentState}
           />
         ) : null}
       </div>
@@ -96,6 +116,11 @@ function Course({
         toggleNotificationTray={toggleNotificationTray}
         isNotificationTrayVisible={isNotificationTrayVisible}
         notificationTrayVisible={notificationTrayVisible}
+        notificationStatus={notificationStatus}
+        setNotificationStatus={setNotificationStatus}
+        onNotificationSeen={onNotificationSeen}
+        currentState={currentState}
+        setCurrentState={setCurrentState}
         //* * [MM-P2P] Experiment */
         mmp2p={MMP2P}
       />
--- a/src/courseware/course/Course.jsx
+++ b/src/courseware/course/Course.jsx
@@ -65,12 +65,12 @@ function Course({
     setLocalStorage('notificationStatus', 'active'); // Show red dot on notificationTrigger until seen
   }
 
-  if (!getLocalStorage('currentState')) {
-    setLocalStorage('currentState', 'initialize');
+  if (!getLocalStorage('upgradeNotificationCurrentState')) {
+    setLocalStorage('upgradeNotificationCurrentState', 'initialize');
   }
 
   const [notificationStatus, setNotificationStatus] = useState(getLocalStorage('notificationStatus'));
-  const [currentState, setCurrentState] = useState(getLocalStorage('currentState'));
+  const [upgradeNotificationCurrentState, setupgradeNotificationCurrentState] = useState(getLocalStorage('upgradeNotificationCurrentState'));
 
   const onNotificationSeen = () => {
     setNotificationStatus('inactive');
@@ -100,7 +100,7 @@ function Course({
             isNotificationTrayVisible={isNotificationTrayVisible}
             notificationStatus={notificationStatus}
             setNotificationStatus={setNotificationStatus}
-            currentState={currentState}
+            upgradeNotificationCurrentState={upgradeNotificationCurrentState}
           />
         ) : null}
       </div>
@@ -119,8 +119,8 @@ function Course({
         notificationStatus={notificationStatus}
         setNotificationStatus={setNotificationStatus}
         onNotificationSeen={onNotificationSeen}
-        currentState={currentState}
-        setCurrentState={setCurrentState}
+        upgradeNotificationCurrentState={upgradeNotificationCurrentState}
+        setupgradeNotificationCurrentState={setupgradeNotificationCurrentState}
         //* * [MM-P2P] Experiment */
         mmp2p={MMP2P}
       />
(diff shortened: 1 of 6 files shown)


src/courseware/course/Course.jsx
```diff
@@ -60,6 +61,22 @@ function Course({
...
   };
 
+  if (!getLocalStorage('notificationStatus')) {
+    setLocalStorage('notificationStatus', 'active'); // Show red dot on notificationTrigger until seen
+  }
+
+  if (!getLocalStorage('currentState')) {
+    setLocalStorage('currentState', 'initialize');
+  }
+
```
dianekaplan, line 71: 'initialize' is an arbitrary value to start, but we need *something* in there because it's used to initialize the currentState prop two lines later. Open to ideas on a better value to start with

src/courseware/course/NotificationTrigger.jsx
```diff
@@ -1,29 +1,50 @@
...
+   The currentState prop will be available after UpgradeNotification mounts. Once available,
+  compare with the last state they've seen, and if it's different then set dot back to red */
+  function updateLastSeen() {
+    if (currentState) {
+      if (getLocalStorage('lastSeen') !== currentState) {
+        setNotificationStatus('active');
+        setLocalStorage('notificationStatus', 'active');
+        setLocalStorage('lastSeen', currentState);
+      }
+    }
```
dianekaplan, line 24: Not sure if this is more comment-heavy than folks like; being new to react I love a clear signpost for what's going on, and here we have sibling components affecting each other. Should this be trimmed down?

src/courseware/course/NotificationTrigger.jsx
```diff
@@ -1,29 +1,50 @@
...
 import messages from './messages';
 
-function NotificationTrigger({ intl, toggleNotificationTray, isNotificationTrayVisible }) {
+function NotificationTrigger({
+  intl, toggleNotificationTray, isNotificationTrayVisible, notificationStatus, setNotificationStatus,
+  currentState,
+}) {
+  /* Re-show a red dot beside the notification trigger for each of the 7 UpgradeNotification stages
+   The currentState prop will be available after UpgradeNotification mounts. Once available,
+  compare with the last state they've seen, and if it's different then set dot back to red */
```
dianekaplan, line 16: Not sure if this is more comment-heavy than folks like; being new to react I love a clear signpost for what's going on, and here we have sibling components affecting each other so I've been comment-y throughout. Should this be trimmed down?

src/courseware/course/NotificationTrigger.test.jsx
```diff
@@ -4,33 +4,71 @@ import {
...
-  let mockData;
+  let mockDataBaseline;
+  let mockDataSameState;
+  let mockDataDifferentState;
   const courseMetadata = Factory.build('courseMetadata');
 
   beforeAll(async () => {
     await initializeTestStore({ courseMetadata, excludeFetchCourse: true, excludeFetchSequence: true });
-    mockData = {
+    mockDataBaseline = {
```
julianajlk: Instead of creating 3 different mock data objects that gets set up at first, is it possible to create one mockData and just modify what you need to be different in each test? You will need to use `beforeEach()` instead of `beforeAll()` to reset the states/statuses back to the default one

julianajlk: And in each test, it could be `render(<NotificationTrigger {...mockData} currentState={whatever it needs to be for the specific test} />

src/courseware/course/NotificationTrigger.test.jsx
```diff
@@ -4,33 +4,71 @@ import {
...
-    // REV-2297 TODO: update below test once the status=active or inactive is implemented
-    // expect(screen.getByTestId('notification-dot')).toBeInTheDocument();
+  it('renders notification trigger WITHOUT icon 3 seconds later', async () => {
+    const { container } = render(<NotificationTrigger {...mockDataBaseline} />);
+    expect(container).toBeInTheDocument();
+    jest.useFakeTimers();
+    setTimeout(() => {
+      expect(screen.queryByRole('notification-dot')).not.toBeInTheDocument();
+    }, 3000);
+    jest.runAllTimers();
```
julianajlk, line 41: After this, should it test for the red dot?

src/courseware/course/NotificationTrigger.test.jsx
```diff
@@ -4,33 +4,71 @@ import {
...
+  });
+
+  it('renders notification trigger WITHOUT icon within the same phase', async () => {
```
julianajlk: Just a wording comment, I believe it's not without the icon but without the red dot

src/courseware/course/NotificationTrigger.test.jsx
```diff
@@ -40,4 +78,14 @@ describe('Notification Trigger', () => {
     fireEvent.click(notificationTrigger);
     expect(toggleNotificationTray).toHaveBeenCalledTimes(1);
   });
+
+  it('we make the right updates when rendering a new phase (before -> after)', async () => {
+    const { container } = render(<NotificationTrigger {...mockDataDifferentState} />);
+    expect(container).toBeInTheDocument();
+
+    // rendering NotificationTrigger has the effect of calling updateLastSeen()
+    // Verify that local storage was updated accordingly
```
julianajlk: Instead of this comment here I would update the test to say this on `it('we make the right updates...')`. This comment is more clear to me of what the test is about than the name of the test

julianajlk (approved):
Well done Diane! This is a tricky feature :) Just made some comments, let me know if you have questions

src/courseware/course/NotificationTrigger.jsx
julianajlk: My assumption was that they would be removed before you merge, but I don't have strong feelings about either. Just as a pattern in `frontend-app-learning`, it doesn't have too many comments (perhaps it would make developers do less discovery in a new codebase if it did?)

src/courseware/course/NotificationTrigger.test.jsx
dianekaplan: good point! I'll update

src/courseware/course/NotificationTrigger.test.jsx
dianekaplan: I agree- it'll look neater that way too- will update

src/courseware/course/NotificationTrigger.test.jsx
dianekaplan: Not sure I follow- the test above covers that the red dot shows when notificationStatus is 'active', then here we check that it disappears after 3 seconds. Do you mean add a line into the last test to check again for the red dot?

src/courseware/course/NotificationTrigger.test.jsx
dianekaplan: Thanks for the info on how to do that! Update made!

src/courseware/course/NotificationTrigger.test.jsx
julianajlk: Nevermind, I missed the expect on L55

dianekaplan: no worries! Will close

PR #570, opened 2021-07-30
fix: re-enable access error redirects for course home

mikix:
These redirects are already in place for the courseware, and will redirect to the outline page, or the dashboard, or wherever, based on the type of access error (unenrolled, access expired, survey needed, etc).

This commit stops the course home tabs from paying attention to the 401 error messages coming from the backend - course_access in the metadata API handles that now.

This commit also changes our useModel hook to more gracefully handle not being able to find the model - it is a valid case we want to allow (still will cause problems if you actually try to use the data, but will at least provide an object you can inspect).

### DIFF
--- a/src/generic/model-store/hooks.js
+++ b/src/generic/model-store/hooks.js
@@ -2,7 +2,7 @@ import { useSelector, shallowEqual } from 'react-redux';
 
 export function useModel(type, id) {
   return useSelector(
-    state => (state.models[type] !== undefined ? state.models[type][id] : undefined),
+    state => (state.models[type] !== undefined ? state.models[type][id] : {}),
     shallowEqual,
   );
 }
@@ -10,7 +10,7 @@ export function useModel(type, id) {
 export function useModels(type, ids) {
   return useSelector(
     state => ids.map(
-      id => (state.models[type] !== undefined ? state.models[type][id] : undefined),
+      id => (state.models[type] !== undefined ? state.models[type][id] : {}),
     ),
     shallowEqual,
   );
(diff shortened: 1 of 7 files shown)


src/shared/access.js
```diff
@@ -48,20 +33,9 @@ function AccessDeniedRedirect(props) {
     case 'authentication_required':
     case 'enrollment_required':
     default:
+      if (activeTabSlug !== 'outline') {
+        url = `/redirect/course-home/${courseId}`;
+      }
```
mikix, line 38: This stops the looping.

src/generic/model-store/hooks.js
```diff
@@ -2,15 +2,15 @@ import { useSelector, shallowEqual } from 'react-redux';
 
 export function useModel(type, id) {
   return useSelector(
-    state => (state.models[type] !== undefined ? state.models[type][id] : undefined),
+    state => (state.models[type] !== undefined ? state.models[type][id] : {}),
```
ciduarte, line 5: I'm not sure what the implications of this switch from `undefined` to `{}` are, but it seems like it would operate in the same way?

ciduarte (approved):
LGTM

src/generic/model-store/hooks.js
mikix: Well since the most common way to use `useModel` is like:
```
const {
  var1,
  var2,
} = useModel(...);
```

When it was undefined, it blew up trying to deconstruct `undefined`. But with `{}` it doesn't throw an exception yet, it just sets `var1` and `var2` as undefined (which might later blow something up). But at least the `useModel` call works.

PR #577, opened 2021-08-04
fix: ignore grade for allowlist certs

tuchfarber:
Allowlist certs were hidden on the progress view because the grade
wasn't passing. This gives "downloadable" certs preference over the
grade.

### DIFF
--- a/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
+++ b/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
@@ -83,17 +83,21 @@ function CertificateStatus({ intl }) {
   const idVerificationSupportLink = <IdVerificationSupportLink />;
   const profileLink = <ProfileLink />;
 
+  // Some learners have a valid ("downloadable") certificate without being in a passing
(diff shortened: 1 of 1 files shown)


src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
```diff
@@ -83,17 +83,21 @@ function CertificateStatus({ intl }) {
   const idVerificationSupportLink = <IdVerificationSupportLink />;
   const profileLink = <ProfileLink />;
 
+  // Some learners have a valid ("downloadable") certificate without being in a passing
+  // state (e.g. learners who have been added to a course's allowlist), so we need to
+  // skip grade validation for these learners
+  const certIsDownloadable = certStatus === 'downloadable';
```
tuchfarber, line 89: Does this feel a little silly? Yeah. But I think it's _slightly_ easier to grok that this is an overriding thing rather than having to read `&& !(certStatus === 'downloadable`))` in the branching below. Can be convinced otherwise.

Tj-Tracy (approved):
LGTM, I actually like the explicit variable 👍

PR #600, opened 2021-08-17
fix: media query max-width effect on course content pages

awais-ansari:
[TNL-8650](https://openedx.atlassian.net/browse/TNL-8650)
1. Remove max-width 1440px from sequence-container media query for non mobile screens.

Before fix
Screenshot 2021-08-17 at 5 01 23 PM

After fix
Screenshot 2021-08-17 at 5 00 26 PM
Screenshot 2021-08-17 at 5 00 35 PM

### DIFF
--- a/src/index.scss
+++ b/src/index.scss
@@ -5,7 +5,6 @@
 
 @import "~@edx/frontend-component-footer/dist/footer";
 
-
 #root {
   display: flex;
   flex-direction: column;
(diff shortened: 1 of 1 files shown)


mehaknasir (approved):
LGTM

saadyousafarbi (approved):

awaisdar001 (approved):

PR #606, opened 2021-08-19
feat: add page banner to masquerade (AA-877)

ciduarte:
- Introduces `AccessExpirationMasqueradeBanner` into `InstructorToolbar`
- Replaces `AccessExpirationAlertMasquerade` on Outline Tab with the new banner, subsequently removing the need for the original alert's code
- Introduces `CourseStartMasqueradeBanner` into `InstructorToolbar`
- Refactors `course-start-alert` into `src/alerts`
- Updates the `CourseStartAlert` hook logic to use `start` from `CourseHomeCourseMetadata` rather than the Outline Tab's date blocks
- Updates `userTimezone` to come from `CourseHomeCourseMetadata` rather than the individual tabs

**Access Expiration Banner**
<div>
Screen Shot 2021-08-19 at 10 56 59 AM
Screen Shot 2021-08-19 at 10 56 50 AM
</div>

**Course Not Yet Started Banner**
<div>
Screen Shot 2021-08-19 at 10 45 16 AM
Screen Shot 2021-08-19 at 10 45 25 AM
</div>


src/instructor-toolbar/InstructorToolbar.jsx
```diff
@@ -54,7 +57,47 @@ export default function InstructorToolbar(props) {
     courseId,
     unitId,
     canViewLegacyCourseware,
+    tab,
   } = props;
+
+  let masqueradeUserAccessExpirationDate = '';
+  let masqueradeUserCourseStartDate = '';
+  if (tab === 'progress') {
```
mikix: I worry about the scalablity here? What determines which tabs want these banners?

src/course-home/progress-tab/ProgressTab.test.jsx
```diff
@@ -1186,6 +1189,32 @@ describe('Progress Tab', () => {
...
+    });
+    it('renders course not yet started page banner on masquerade', async () => {
+      setMetadata({ is_enrolled: true, original_user_is_staff: true, start: '2999-01-01T00:00:00Z' });
+      await executeThunk(thunks.fetchProgressTab(courseId), store.dispatch);
+      await act(async () => render(<InstructorToolbar courseId={courseId} tab="progress" />, { store }));
+      expect(screen.getByTestId('instructor-toolbar')).toBeInTheDocument();
+      expect(screen.getByText('This learner does not yet have access to this course. The course starts on', { exact: false })).toBeInTheDocument();
+      expect(screen.getByText('1/1/2999')).toBeInTheDocument();
+    });
+  });
```
Dillon-Dumesnil: Love that you're testing this, but it feels a little weird with this test content being in the ProgressTab tests since it doesn't really have anything to do with the Progress tab. Is it possible for this to be closer to where InstructorToolbar is defined?

src/instructor-toolbar/InstructorToolbar.jsx
```diff
@@ -54,7 +57,47 @@ export default function InstructorToolbar(props) {
...
+      userTimezone,
+    } = useModel('progress', courseId);
+    const {
+      start,
+    } = useModel('courseHomeMeta', courseId);
+
+    const timezoneFormatArgs = userTimezone ? { timeZone: userTimezone } : {};
+
+    if (accessExpiration) {
+      if (accessExpiration.masqueradingExpiredCourse) {
```
Dillon-Dumesnil: Nit: Combine these into an &&

src/instructor-toolbar/InstructorToolbar.jsx
ciduarte: The use case for these banners was to match the alerts we had on legacy Progress Tab but didn't initially get carried over. I don't believe they exist/are intended to exist elsewhere. (edit: these both already exist on the outline tab as alerts) (another edit: Kaitlin said we're not going to be applying these two banners to any other tabs)

When we groomed this ticket, I asked if we should just include them everywhere the instructor toolbar exists (outline tab and dates tab), because I could see that where this would live in the code would have access to all tabs, but the decision at the time was no.

Dillon-Dumesnil (approved):
Mostly just a comment on test location. I see that this logic only applies on the progress course tab right now, but if that logic ever changes, it would make sense for these tests to exist more generically. I could see these banners being generally helpful on all pages so I'm curious if that update will eventually happen.

src/instructor-toolbar/InstructorToolbar.jsx
ciduarte: I considered creating two individual page banner components and wrapping all of this logic in both of them, similar to how our Alerts work but w/o the hooks and complexity. That way they'd be more of a plug in to the Instructor Toolbar (or wherever you want to put them) rather than having the Instructor Toolbar handle so much of the logic to actually show the banners... I didn't know if that was overkill. I'm not sure if that solves the scalability problem so much, but could help clean this up. Thoughts?

src/course-home/progress-tab/ProgressTab.test.jsx
ciduarte: I see what you're saying. My first instinct was to put these tests inside of the Instructor Toolbar tests, but I was running into a lot of walls w/ the way those tests were setup because it's specific to courseware, and it felt like I'd have to upend that setup for this specific Progress Tab use case. I say Progress Tab specific because in grooming I asked if we should include these alerts on other tabs since they'd be living somewhere that has access to all tabs and the answer was no. Also, a version of both of these alerts already exists on Outline Tab.

Separately, I was running into issues w/ testing in general because of the new Paragon component, so once I got it working, I didn't think to see if those issues had been impacting my initial attempt to thread these into the Instructor Toolbar tests... TLDR I can spend some more time writing these into the Instructor Toolbar tests, but it felt okay to put include it in Progress Tab tests because it's a feature specific to Progress Tab

src/instructor-toolbar/InstructorToolbar.jsx
mikix: Some sort of clean-up might be nice. And/or unification with what the outline tab is doing. For example, you re-implement `AccessExpirationAlertMasquerade` below (with a slightly different user string). That, at minimum, should be consolidated (you could probably just replace the below code with `<AccessExpirationAlertMasquerade>`). But maybe also, some of that logic could be shared between the two tabs?

And maybe? (you might have better opinions here) the course starts alert should follow the same pattern as the expiration alert - show either the normal or a masquerade version, and abstract that distinction a bit. Though I guess here, you're just interested in the masquerade version. But the outline tab might also be interested in showing the masquerade version too, instead of the normal `<CourseStartAlert>` it uses, when masquerading.

ciduarte: Yessss great idea about consolidating AccessExpirationAlertMasquerade

And sure I'll see if I can do something similar w/ CourseStartAlert. Kaitlin's said she doesn't want us to implement these on the Outline Tab for fear of crowding info. But it might be harmless to show the masquerade version and hide the normal CourseStartAlert when masquerading...

mikix: Yeah I agree any CourseStartAlert consolidation is less obvious than the expiration alert. These various alerts feel like they're on the cusp of a logical layout/abstraction, but aren't quite there yet. And I just want us to be there already.  :(

src/alerts/access-expiration-alert/AccessExpirationAlert.jsx
```diff
@@ -36,17 +35,10 @@ function AccessExpirationAlert({ intl, payload }) {
...
-    masqueradingExpiredCourse,
     upgradeDeadline,
     upgradeUrl,
   } = accessExpiration;
 
-  if (masqueradingExpiredCourse) {
-    return (
-      <AccessExpirationAlertMasquerade payload={payload} />
-    );
-  }
```
ciduarte, line 48: Removing this in favor of the `AccessExpirationMasqueradePageBanner` (what a name)

ciduarte: I delete the entire component (below) because it's going unused.

It was removed by [this commit](https://github.com/edx/frontend-app-learning/commit/554e63d65352a76bcb088df03aff5b1b311f1c79), and they chose to leave the component in tact for it's masquerade message. But now the masquerade message is handled by the new page banner. So this component is useless

src/course-home/data/api.js
```diff
@@ -372,6 +372,7 @@ export async function getOutlineTabData(courseId) {
     resumeCourse,
     timeOffsetMillis, // This should move to a global time correction reference
     userHasPassingGrade,
+    userTimezone: datesWidget.userTimezone,
```
ciduarte: Okay so this was an interesting realization. Outline Tab is the only API that doesn't cleanly surface the `userTimezone` (i.e. it's hidden under the `datesWidget` key). I wanted to be able to access the userTimezone across tabs w/o checking what tab we were on, but I didn't want to bust up all our APIs, so this was my solution.

I would love to see `userTimezone` come through the course home metadata call, but not sure if it would be appropriate to introduce breaking changes across three APIs to make this happen.

src/alerts/course-start-alert/hooks.js
```diff
@@ -0,0 +1,59 @@
...
+
+function isStartDateInFuture(courseId) {
+  const {
+    start,
+  } = useModel('courseHomeMeta', courseId);
+
+  const today = new Date();
+  const startDate = new Date(start);
+  return startDate > today;
+}
```
ciduarte, line 16: idk if this abstraction was necessary I just didn't like seeing the logic twice :)

src/alerts/course-start-alert/CourseStartAlert.jsx
```diff
@@ -9,14 +9,20 @@ import {
 import { Alert } from '@edx/paragon';
 import { Info } from '@edx/paragon/icons';
 
+import { useModel } from '../../generic/model-store';
+
 const DAY_MS = 24 * 60 * 60 * 1000; // in ms
 
 function CourseStartAlert({ payload }) {
```
Dillon-Dumesnil, line 16: Does this still need to say `{ payload }` or can it just be `(courseId)`?

src/course-home/progress-tab/ProgressTab.test.jsx
```diff
@@ -1186,6 +1189,32 @@ describe('Progress Tab', () => {
     });
   });
 
+  describe('Instructor toolbar', () => {
```
Dillon-Dumesnil: Nit: Change the describe to be more about masquerading and not about the Instructor Toolbar. That just happens to be where it shows up, not the functionality you're testing here.

Dillon-Dumesnil (approved):
LGTM

src/alerts/course-start-alert/CourseStartAlert.jsx
ciduarte: Looking into our alert hook logic I think this needs to remain as `{ payload }`

mikix (approved):
I'm not going to re-review in depth, since Dillon already did and looks like my original comments were addressed. But love the cleanup here, thanks for that!

PR #612, opened 2021-08-24
feat: add a new course goal unsubscribe landing page

mikix:
URL format: /goal-unsubscribe/UUID

This is designed to be used in the new course goals feature, where emails will be sent to learners and those emails will include a link to this landing page, as an unsubscribe link.

Also, update calls to the LMS course home API to not include the /v1/ fragment anymore, as we're moving to an unversioned API.

https://openedx.atlassian.net/browse/AA-907

Screenshot from 2021-08-24 12-10-36
Screenshot from 2021-08-24 12-10-57
Screenshot from 2021-08-24 13-37-13


src/index.jsx
```diff
@@ -73,7 +75,6 @@ subscribe(APP_READY, () => {
             component={CoursewareContainer}
           />
         </Switch>
-        <Footer />
```
mikix, line 76: Kavya specifically didn't want a footer on this page. So I figured I'd move it down a step (like Header is) to each component that does want it. Not that much work, since most pages use TabPage.

src/generic/PageLoading.jsx
```diff
@@ -23,9 +25,9 @@ export default class PageLoading extends Component {
             height: '50vh',
           }}
         >
-          <div className="spinner-border text-primary" role="status">
+          <Spinner animation="border" variant="primary" role="status">
```
mikix, line 28: No visible / dom changes here - just a drive-by cleanup.

src/tab-page/LoadedTabPage.jsx
```diff
@@ -82,6 +83,7 @@ function LoadedTabPage({
           {children}
         </div>
       </main>
+      <Footer />
```
Dillon-Dumesnil: I suppose you're just copying the pattern as Header, but I wonder if this Header/Footer should just live in TabPage. It's already imported and used there and becomes one less spot to care about. And it already shows up in every use case anyway there, just via this file for this use case. Any thoughts on this?

src/course-header/Header.jsx
```diff
@@ -29,7 +29,7 @@ LinkedLogo.propTypes = {
 };
 
 function Header({
-  courseOrg, courseNumber, courseTitle, intl,
+  courseOrg, courseNumber, courseTitle, intl, showUserDropdown,
```
Dillon-Dumesnil, line 32: Is this necessary? I suppose the use case is that even though you hit the endpoint through your email, you're already an authenticated user and we just don't want to show anything. I guess I don't understand why not show it? Does it matter if a user is able to navigate to other parts of edX from the unsubscribe? Maybe just a weird user experience?

src/course-home/goal-unsubscribe/GoalUnsubscribe.jsx
```diff
@@ -0,0 +1,52 @@
...
+import messages from './messages';
+import ResultPage from './ResultPage';
+
+function GoalUnsubscribe({ intl }) {
+  const { token } = useParams();
+  const [error, setError] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
+  const [data, setData] = useState({});
+
+  // We don't need to bother with redux thing for this simple page. We're not sharing state with other pages at all.
```
Dillon-Dumesnil: Can you clear this up a little bit? I don't think I understand `We don't need to bother with redux thing`

src/course-home/goal-unsubscribe/ResultPage.jsx
```diff
@@ -0,0 +1,40 @@
+import PropTypes from 'prop-types';
+import { getConfig } from '@edx/frontend-platform';
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import { Button } from '@edx/paragon';
+
+import messages from './messages';
+import { ReactComponent as UnsubscribeIcon } from './unsubscribe.svg';
+
+function ResultPage({ courseTitle, error, intl }) {
```
Dillon-Dumesnil, line 9: Nit: I feel like this could just be pulled up into the parent file

src/course-home/goal-unsubscribe/ResultPage.jsx
```diff
@@ -0,0 +1,40 @@
...
+  const header = error
+    ? intl.formatMessage(messages.errorHeader)
+    : intl.formatMessage(messages.header);
+  const description = error
+    ? intl.formatMessage(messages.errorDescription)
+    : intl.formatMessage(messages.description, { courseTitle });
+
+  return (
+    <>
+      <UnsubscribeIcon className="text-primary" alt="" />
```
Dillon-Dumesnil, line 37: Should this have alt text? What does `alt=""` mean? Could it just be removed altogether if it's empty?

src/course-home/goal-unsubscribe/ResultPage.jsx
```diff
@@ -0,0 +1,40 @@
...
+      <div role="heading" aria-level="1" className="h2">{header}</div>
```
Dillon-Dumesnil, line 38: We use class names here? Why not use h2? I feel like I've seen this before, just blanking on it.

Dillon-Dumesnil (approved):
Nothing major, just some clarifications for my sanity.

src/course-home/goal-unsubscribe/ResultPage.jsx
mikix: Someone needs to redo our a11y training. No `alt` is arguably an error that linters might complain about. An empty `alt` is flagging this image as just a visual thing, not conveying any real info. Screen readers will ignore it.

src/course-home/goal-unsubscribe/GoalUnsubscribe.jsx
mikix: I'm not super familiar with the current thinking of the React community, but my understanding is that they are slowly moving away from redux. Though I don't think it's an obvious "redux is dead" thing, just typical slow-moving-but-frequent frontend churn.

https://dev.to/alexandrudanpop/why-react-projects-still-use-redux-in-2020-395p

Looks like [Context](https://reactjs.org/docs/context.html) is the new built-in thing to solve similar issues?

But also in general, this component I added doesn't need to share state with a lot of other components or do any fancy ajax/data processing. It can simply just call the endpoint and use the result. No need for the `Context` or `redux` big guns.

src/course-header/Header.jsx
mikix: That was the design :shrug: -- similar to no footer, right? Like, why not have a footer? I think they just wanted a really trim landing page. I can ask Kavya for ya?

src/tab-page/LoadedTabPage.jsx
mikix: Sure, makes sense.

src/course-home/goal-unsubscribe/ResultPage.jsx
mikix: I hear you - it's not too long. But I personally like the division:

- parent file: container, ajax logic, and spinner (all the business/logic of the page, plus the tiny wrapping DOM that doesn't change)
- child file: actual content renderer

I dunno, the division made sense to me, and kept the logic forking more readable in the parent class, personally.

src/course-home/goal-unsubscribe/GoalUnsubscribe.jsx
mikix: The last paragraph is the actual answer to your question, I realize. The first bits are just background for redux, but not at all relevant to your question, given that the learning MFE uses redux instead of Context.  :)

src/course-home/goal-unsubscribe/ResultPage.jsx
mikix: Yeah this was copy/paste by me. I think the rationale is: we want the visual look of h2, but this is in fact the main header of the page. So using `<h2>` would be misleading / confusing to screen readers (though I imagine aria-level="1" would solve the reader problem?).

Carla might be able to answer better. I was just pasting around.

src/course-home/goal-unsubscribe/GoalUnsubscribe.jsx
mikix: Oh also - the `thing` in `We don't need to bother with redux thing` was a typo that I thought I had gotten rid of already - I can see why that was adding confusion.

src/course-header/Header.jsx
mikix: Asked, but no answer yet.

src/tab-page/LoadedTabPage.jsx
mikix: Done.

.env
```diff
@@ -4,6 +4,7 @@
 NODE_ENV='production'
 ACCESS_TOKEN_COOKIE_NAME=''
 BASE_URL=''
+CONTACT_URL=''
```
mikix, line 7: This is a standard frontend variable, already set by our configuration - just not used in the learning MFE before.

src/course-home/goal-unsubscribe/GoalUnsubscribe.jsx
```diff
@@ -0,0 +1,52 @@
...
+import messages from './messages';
+import ResultPage from './ResultPage';
+
+function GoalUnsubscribe({ intl }) {
+  const { token } = useParams();
+  const [error, setError] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
+  const [data, setData] = useState({});
+
+  // We don't need to bother with redux for this simple page. We're not sharing state with other pages at all.
```
Dillon-Dumesnil, line 18: Is it worth calling this out at all? Why explicitly state something we aren't doing vs just do the thing?

src/course-home/goal-unsubscribe/ResultPage.jsx
Dillon-Dumesnil: Yeah, I'm not an expert here. Defer to you

Dillon-Dumesnil (approved):

src/course-header/Header.jsx
mikix: Yeah - she says same reason as no footer - just wanted a bare bones page. :shrug:

src/course-home/goal-unsubscribe/GoalUnsubscribe.jsx
mikix: Just cause I was bucking the trend a bit - this page is unique in a lot of ways, and I wanted to keep it as self contained as possible. Most other pages we have in the MFE go through our normal redux layers.

mikix: You're right that it's not strictly necessary to call it out though. If you think it's more confusing to mention redux at all, I'll drop it. A 3rd party eyeball on this is better than my own.

Dillon-Dumesnil: Eh I think it's fine. Just my original confusion with the extra "thing" threw me off

src/course-home/goal-unsubscribe/GoalUnsubscribe.test.jsx
```diff
@@ -0,0 +1,61 @@
...
+    );
+    history.push('/goal-unsubscribe/TOKEN'); // so we can pull token from url
+  });
+
+  it('starts with a spinner', () => {
+    render(component);
+    screen.getByRole('status');
+  });
+
+  it('loads a real token', async () => {
```
Dillon-Dumesnil, line 41: Nit: have this test check for the existence of "My Sample Course" to ensure proper receipt and display of the data

Dillon-Dumesnil (approved):
Tests LGTM

src/course-home/goal-unsubscribe/GoalUnsubscribe.test.jsx
mikix: Good idea - done.

PR #616, opened 2021-08-25
fix: add accent to e in resumé

pshiu:
# Description
Resumé is spelled with an accented letter E.

The UI currently spells resume without an accent on the letter E.

UX requested this be corrected.

Fixes: https://openedx.atlassian.net/browse/REV-2214
See: https://github.com/edx/edx-platform/pull/27435
See: https://openedx.atlassian.net/browse/REV-2129

# Screenshots

* upgrade-notification

 Screen Shot 2021-08-25 at 12 32 51 PM

* lock-paywall-container / Gated Content

Screen Shot 2021-08-25 at 4 30 54 PM


brianhw:
(Interesting.   I thought that the word was "resume" in English and "résumé" in French.  I didn't know that "resumé" was even acceptable, let alone preferred.   TIL.)

pshiu:
*sigh* Oh bother, you’re correct, good catch. The ticket says resumé but the dictionary says résumé (albeit admits resume and less commonly resumé as an alternative spelling).

I’ll double-check with UX!

pshiu:
Per @hilaryg123, we'll center around resumé, so this is good to go as-is.

inventhouse (approved):

PR #623, opened 2021-08-30
Feat: [AA-950] Streak discount productization

cdeery:
feat: [AA-950] Productize Streak discount

Convert experiment to feature.

Rename variables to remove references to ticket AA-759

Make expiration two weeks from date modal is shown.


src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -125,26 +125,26 @@ function StreakModal({
...
               <div className="col-11 factoid-wrapper">
                 <b>{intl.formatMessage(messages.congratulations)}</b>
                 &nbsp;{intl.formatMessage(messages.streakDiscountMessage)}&nbsp;
                 <FormattedMessage
-                  id="learning.streakCelebration.streakAA759EndDateMessage"
+                  id="learning.streakCelebration.streakCelebrationCouponEndDateMessage"
                   defaultMessage="Ends {date}."
                   values={{
-                    date: new Date('2021-7-20 00:00').toLocaleDateString({ timeZone: 'UTC' }),
+                    date: new Date(Date.now() + 14 * 24 * 60 * 60 * 1000).toLocaleDateString({ timeZone: 'UTC' }),
```
MatthewPiatetsky, line 147: Is the coupon expiring in two weeks something we actually need to enforce?

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -125,26 +125,26 @@ function StreakModal({
...
+        { !StreakCelebrationCouponEnabled && (
           <div className="d-flex py-3 bg-light-300">
             <Icon className="col-small ml-3" src={Lightbulb} />
             <div className="col-11 factoid-wrapper">
               {randomFactoid}
             </div>
           </div>
         )}
-        { AA759ExperimentEnabled && (
+        { StreakCelebrationCouponEnabled && (
```
MatthewPiatetsky: here you call it Streak celebration coupon enabled but in the other PR you call it streak discount enabled, might be good to choose streak discount or streak coupon for consistency

MatthewPiatetsky (commented):
1. Could you include a screenshot?
2. Are there any business logic changes we needed to make when implementing this? (besides the requirement about the coupon being one-use per user)
can we still have the same coupon for everyone?
3. You will need to rename that coupon (in the messages) and create a new coupon in ecommerce
4. There is also the email portion of this feature within Braze that you will need to modify

src/shared/streak-celebration/StreakCelebrationModal.jsx
cdeery: No, my understanding is that we are saying it in order to create urgency, but we don't actually have a way of enforcing it.

src/shared/streak-celebration/StreakCelebrationModal.jsx
cdeery: Here's the screenshot. The date will be two weeks from whenever the dialog is displayed.
image

MatthewPiatetsky (approved):
nit: you're using streak discount enabled in some places and streak discount coupon enabled in others

PR #624, opened 2021-08-30
chore: update frontend-lib-special-exams version

alangsto:
The latest version of this library contains a fix for the special exams masquerading message. Previously this message would show for any content if a user was masquerading, while the fix limits this message to only show for time gated content while masquerading.


schenedx (approved):
👍

PR #628, opened 2021-09-03
doc: ADR for "Direction of Courseware APIs"

ormsbee:
```
Describing the removal of calls to Course Blocks API for courseware
rendering, how those responsibilities would be split, and the motivation
for doing so. TNL-7326
```

### DIFF
--- a/docs/decisions/0009-courseware-api-direction.md
+++ b/docs/decisions/0009-courseware-api-direction.md
@@ -1,0 +1,62 @@
+# Direction of Courseware APIs
+
+In order to allow for greater flexibility and separation of concerns, we're going to stop using the Course Blocks API for navigational data, and pull that data from the Learning Sequences Outlines API instead. The intention is to give us four distinct layers of courseware that can eventually be composed in different ways:
+
+* Learning Context Metadata
+* Learning Context Navigation
+* Sequence Navigation
+* Unit Rendering
+
+Note that "Learning Context" is a generalization of "Course" that includes other things like Content Libraries, Learning Pathways, and potentially other logical groupings of content.
+
+This is a refinement of [0002-courseware-page-decisions.md](0002-courseware-page-decisions.md). The fundamental layers remain the same, but this document tries to better clarify the boundaries and path forward for these layers. We're not making these layers completely swappable/pluggable now, but we should separate the data access in a way that allows for that in the future.
+
+## Background
+
+We currently make four primary requests to the LMS when rendering courseware instructional content:
+
+1. Course Metadata: `/api/courseware/course/{courseId}` (REST API)
+2. Course Blocks API: `/api/courses/v2/blocks/?course_id={courseId}` (REST API)
+3. Sequence Metadata: `/api/courseware/sequence/{sequenceUsageKey}` (REST API)
+4. Unit: `/xblock/{unitBlockUsageKey}` (rendered in an iframe)
+
+There is a significant amount of overlap between the Course Blocks API and the others at the moment, since Course Blocks takes a static snapshot of the entire tree of course content at once. There are a few problems with the current arrangement:
+
+* It's slow and complex. The Course Blocks API can be difficult to maintain and reason about, and trickier to optimize.
+* Assuming that all course structures are the same makes it difficult to support other content types, like LabXchange Learning Pathways or adaptive content.
+* The overlap between Course Blocks and the other APIs means that there can be conflicts about the state.
+
+## Motivating Vision
+
+We have seen a desire to extend or enhance the courseware experience in various ways:
+
+Learning Context Navigation
+* Allowing for shorter, human-readable URLs in courseware.
+* Smaller courses that do not need the current navigational hierarchy.
+* LabXchange pathways.
+
+Sequence Navigation
+* Adaptive content, where the full list of units is not known up front.
+* More limited navigation, where content is pushed linearly, without the ability to jump ahead.
+* Different layouts for content browsing.
+
+Unit Rendering
+* Use of QTI content (currently serviced by cc2olx conversion).
+* Desire to experiment with a next-gen version of XBlock.
+* Use of entirely LTI units.
+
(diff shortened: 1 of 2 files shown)


docs/decisions/0009-courseware-api-direction.md
```diff
@@ -0,0 +1,60 @@
...
+This is a refinement of [0002-courseware-page-decisions.md](0002-courseware-page-decisions.md). The fundamental layers remain the same, but this document tries to better clarify the boundaries and path forward for these layers. We're not making these layers completely swappable/pluggable now, but we should separate the data access in a way that allows for that in the future.
+
+## Background
+
+We currently make four primary requests to the LMS when rendering courseware instructional content:
+
+1. Course Metadata: `/api/courseware/course/{courseId}` (REST API)
+2. Course Blocks API: `/api/courses/v2/blocks/?course_id={courseId}` (REST API)
+3. Sequence Metadata: `/api/courseware/sequence/{sequenceUsageKey}` (REST API)
+4. Unit: `/xblock//{unitBlockUsageKey}` (rendered in an iframe)
```
Dillon-Dumesnil: Nit: Looks like you accidentally added an extra `/`

docs/decisions/0009-courseware-api-direction.md
```diff
@@ -0,0 +1,60 @@
...
+* Smaller courses that do not need the current navigational hierarchy.
+* LabXchange pathways.
+
+Sequence Navigation
+* Adaptive content, where the full list of units is not known up front.
+* More limited navigation, where content is pushed linearly, without the ability to jump ahead.
+* Different layouts for content browsing.
+
+Unit Rendering
+* Use of QTI content (currently serviced by cc2olx conversion).
```
Dillon-Dumesnil, line 44: What is QTI? I haven't seen this acronym before

docs/decisions/0009-courseware-api-direction.md
```diff
@@ -0,0 +1,60 @@
...
+* Desire to experiment with a next-gen version of XBlock.
+* Use of entirely LTI units.
+
+The idea would be to insulate each layer from the layers above and below it. Sequence rendering shouldn't be affected by whether or not it's in a two level hiearchy (Course → Section → Sequence), or a flat one (Course → Sequence). Learning Context Navigation should be able to reference Sequences without caring if a Sequence is an adaptive one or not. Sequences should be able to have a common interface to call Unit iframes, whether those Units are rendering XBlocks or QTI content.
```
Dillon-Dumesnil: I understand the sentiment here, but I do fear this might be encouraging course structures that we don't want to encourage (and actively do not support in many areas of the code base) with the two level hierarchy vs flat

Also `hiearchy` -> `hierarchy`

docs/decisions/0009-courseware-api-direction.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Next Step: Removing use of the Course Blocks API.
```
Dillon-Dumesnil, line 52: This reads like it should be:
```
## Next Steps:

### Removing use of the Course Blocks API

**Review by davidjoy — APPROVED** [review]
No warning flags at all. This sounds right. 👍👍

File: docs/decisions/0009-courseware-api-direction.md [hunk 5]
```diff
@@ -0,0 +1,60 @@
...
+
+The next step in this process is to remove the call to the Course Blocks API, and split its responsiblities across just the existing Learning Sequences Outline and Sequence Metadata APIs.
```
mikix: nit: responsibilities

mikix (approved):
Sure

docs/decisions/0009-courseware-api-direction.md
ormsbee: It's a widely used problem format. Sorta like XBlock + Capa in terms of its goals. If we were building Open edX today from scratch, we'd probably have used it as the starting point for assessments: https://www.imsglobal.org/spec/qti/v3p0/guide#introduction

A brief example to give you an idea:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<qti-assessment-item
xmlns="http://www.imsglobal.org/xsd/qti/imsqtiasi_v3p0" 
xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
xsi:schemaLocation="http://www.imsglobal.org/xsd/imsqtiasi_v3p0 
https://purl.imsglobal.org/spec/qti/v3p0/schema/xsd/imsqti_asiv3p0_v1p0.xsd"
identifier="firstexample"
time-dependent="false" 
xml:lang="en-US">

<qti-response-declaration base-type="identifier" cardinality="single" identifier="RESPONSE">
   <qti-correct-response>
     <qti-value>A</qti-value>
   </qti-correct-response>
 </qti-response-declaration>

<qti-outcome-declaration base-type="float" cardinality="single" identifier="SCORE">
   <qti-default-value>
     <qti-value>1</qti-value>
   </qti-default-value>
 </qti-outcome-declaration>

 <qti-item-body>

<p>Of the following hormones, which is produced by the adrenal glands?</p>
  <qti-choice-interaction max-choices="1" min-choices="1" 
  response-identifier="RESPONSE">
    <qti-simple-choice identifier="A">Epinephrine</qti-simple-choice>
    <qti-simple-choice identifier="B">Glucagon</qti-simple-choice>
    <qti-simple-choice identifier="C">Insulin</qti-simple-choice>
    <qti-simple-choice identifier="D">Oxytocin</qti-simple-choice>
  </qti-choice-interaction>

</qti-item-body>

<qti-response-processing
template="https://purl.imsglobal.org/spec/qti/v3p0/rptemplates/match_correct"/>

</qti-assessment-item>
```

docs/decisions/0009-courseware-api-direction.md
ormsbee: Good point. I'll put a little more warning language around this. It's been an explicitly stated goal on the pedagogy side for a while, so I don't want to back down from it. But I should call out that it will have knock-on effects in other systems.

Dillon-Dumesnil: > It's been an explicitly stated goal on the pedagogy side for a while

Sorry, do you mean it's an explicitly stated goal to allow flat structures?

ormsbee: To allow unbundling of courses into more granular content chunks. LabXchange has leaf-node things like Simulations/Videos, which can roll up into Pathways, which can roll up into Clusters (which are sort of like small, unordered courses without section hierarchy). We are probably not going to adopt exactly that model for edx.org, but I think it demonstrates where we should be building flexibility into the content platform.

PR #636, opened 2021-09-09
feat: notify the user if a sequence is hidden because of due date

mikix:
Normally, these sequences are skipped. But if the user manually goes to the section, they should be notified why they can't access it. That can easily happen if they bookmarked the page or something.

Backend side: https://github.com/edx/edx-platform/pull/28698

https://openedx.atlassian.net/browse/AA-1000

## Old Notice

Screenshot from 2021-09-09 09-25-20

## New Notice

Screenshot from 2021-09-09 09-26-00

## Old masquerading as specific learner banner

Screenshot from 2021-09-09 09-29-41

## New masquerading as specific learner banner

Screenshot from 2021-09-09 09-27-05


src/courseware/course/sequence/hidden-after-due/HiddenAfterDue.jsx
```diff
@@ -0,0 +1,52 @@
...
+    <Alert variant="info" icon={Info}>
+      <h3>{intl.formatMessage(messages.header)}</h3>
+      <p>
+        {intl.formatMessage(messages.description)}
+        {progressLink && (
+          <>
+            <br />
+            <FormattedMessage
+              id="learn.hiddenAfterDue.gradeAvailable"
+              defaultMessage="If you have completed this assignment, your grade is available on the {progressPage}."
```
Dillon-Dumesnil, line 35: Is there any concern with this message if the assignment is also hidden from the progress tab? Can we detect that? Is that even a real scenario?

src/courseware/course/sequence/Sequence.jsx
```diff
@@ -144,6 +145,12 @@ function Sequence({
     );
   }
 
+  if (sequenceStatus === 'loaded' && sequence.isHiddenAfterDue) {
+    // Shouldn't even be here - these sequences are normally stripped out of the navigation.
+    // But we are here, so render a notice instead of the normal content.
```
Dillon-Dumesnil, line 150: 👏

src/courseware/course/sequence/Sequence.test.jsx
```diff
@@ -81,6 +81,37 @@ describe('Sequence', () => {
...
+    const sequenceMetadata = [Factory.build(
+      'sequenceMetadata',
+      { is_hidden_after_due: true },
+      { courseId: courseMetadata.id, unitBlocks, sequenceBlock: sequenceBlocks[0] },
+    )];
+    const testStore = await initializeTestStore(
+      {
+        courseMetadata, unitBlocks, sequenceBlocks, sequenceMetadata,
+      }, false,
+    );
```
Dillon-Dumesnil, line 99: Do these not already exist for the test suite? Feels like a lot of setup for one test. If not, should this kind of setup be defined in a `beforeEach` or something?

Dillon-Dumesnil (approved):
Small comments, but lgtm

src/courseware/course/sequence/hidden-after-due/HiddenAfterDue.jsx
mikix: It should be shown there, just without the link to the sequence (i.e. without a link back to this page).

@ciduarte - are there cases where we don't show a grade and text like this would be a lie?

src/courseware/course/sequence/Sequence.test.jsx
mikix: I'll be honest, I don't fully understand this test file. A lot of other tests do exactly this kind of setup and it could probably be unified in a fancy way. But I wasn't in the mood.

src/courseware/course/sequence/hidden-after-due/HiddenAfterDue.jsx
ciduarte: We have two types of visibility settings. Content visibility and grade visibility. The two function independently. It seems like you're just accounting for content visibility here, which wouldn't impact whether grades are shown in progress tab.

We use a `show_grades()` function in the progress tab api that determines whether `show_correctness` is true. Cases where we don't show grades on the progress tab include:
- `show_correctness` == `false` (i.e. grades are set to never shown or only shown after due date has passed)
- the assignment is a practice assignment, not a graded assignment
- the assignment is staff only (that doesn't apply here)
- the assignment has a future release date (doesn't apply here)
and I think that's it...

PR #637, opened 2021-09-10
fix: stop logging course-blocks 403 responses as errors

mikix:
They are benign and normal for logged out users. Instead, log them as info messages, so we can still track them if we need to.

`logInfo` can take an error object or a string. Passing the object will hopefully make it easy to continue tracking this condition in newrelic if we want to.

https://openedx.atlassian.net/browse/AA-1011

### DIFF
--- a/src/courseware/data/thunks.js
+++ b/src/courseware/data/thunks.js
@@ -1,4 +1,4 @@
-import { logError } from '@edx/frontend-platform/logging';
+import { logError, logInfo } from '@edx/frontend-platform/logging';
 import {
   getBlockCompletion,
   postSequencePosition,
@@ -188,7 +188,14 @@ export function fetchCourse(courseId) {
       // Log errors for each request if needed. Course block failures may occur
       // even if the course metadata request is successful
       if (!fetchedBlocks) {
-        logError(courseBlocksResult.reason);
+        const { response } = courseBlocksResult.reason;
+        if (response && response.status === 403) {
+          // 403 responses are normal - they happen when the learner is logged out.
+          // We'll redirect them in a moment to the outline tab by calling fetchCourseDenied() below.
+          logInfo(courseBlocksResult.reason);
+        } else {
+          logError(courseBlocksResult.reason);
+        }
       }
       if (!fetchedMetadata) {
         logError(courseMetadataResult.reason);


MatthewPiatetsky (commented):
i think we should check if the user is logged out and only logInfo in that case?  
because in some of the other cases we were getting alerted for there were 403s for logged in users

Dillon-Dumesnil (approved):
LGTM

mikix:
> i think we should check if the user is logged out and only logInfo in that case?
because in some of the other cases we were getting alerted for there were 403s for logged in users

Do you think those are actionable alerts? Like, a user can get a 403 if they don't have permission to view a page. That also strikes me as a normal thing though, yeah? Appropriate for logInfo but not logError.

mikix:
I suppose that same argument could apply to any 4xx error. Which I'm not opposed to doing...

MatthewPiatetsky:
you'd want to update the comment to be more generic then than just the logout case
it is probably normal behavior, but one thing that i wonder if we should consider is the following - what if we break user auth for this endpoint and suddenly everyone gets a forbidden error when they would normally have the right permissions - we may want to be alerted in that case and i wonder how that would happen if these weren't errors..

mikix:
> you'd want to update the comment to be more generic then than just the logout case

You mean if changed the code to handle all 4xx errors the same? Yes, I agree I should also update the comment for the changed code.

> it is probably normal behavior, but one thing that i wonder if we should consider is the following - what if we break user auth for this endpoint and suddenly everyone gets a forbidden error when they would normally have the right permissions - we may want to be alerted in that case and i wonder how that would happen if these weren't errors..

I think you are asking this `logError` to function as a canary in the coal mine in spite of its inappropriateness, not because it is appropriate.

Consider an example method `getEnrollmentTrack` and you notice it is logging an error with New Relic whenever it returns `'audit'`. Sometimes it wakes up engineers if we happen to get more audit users than normal that day. Would you argue to keep it because the enrollment code might break some day?

This 403 is an entirely normal event (like most 4xx) - just the backend communicating status to the frontend. This 403 could easily be a json field like `user_can_access: false` with a status code of 200 instead (in fact, some things that the Course Home side uses 4xx for, the Courseware side uses json for -- like whether the MFE waffles are enabled).

If we are worried about the authn service going down, we have the support team as a backstop. But also, that service should have its own health checks and New Relic alerting - edX shouldn't be relying on a higher than normal 403 line in the Learning MFE's New Relic panel. I mean, it's not bad to for that to also notify us. But only if it makes sense to.

> [mikix] I suppose that same argument could apply to any 4xx error. Which I'm not opposed to doing...

I had not wanted to grow scope last Friday, so we could ideally land this before your weekend on-call. But even now, I'm inclined to keep scope small and be cautious here. If there are other 4xx codes that alert us, we can investigate em and expand scope then.

PR #641, opened 2021-09-13
feat: Breadcrumb Jump Navigation

connorhaugh:
feat: Breadcrumb Jump Navigation Per [TNL-7107](https://openedx.atlassian.net/secure/RapidBoard.jspa?rapidView=580&projectKey=TNL&modal=detail&selectedIssue=TNL-7107)

Enable faster movement through the course content for learners and course instructors familiar with their course structure using jump navigation selectors in dropdown menus that augment our existing breadcrumbs in the learner sequence experience. When learners/instructors click on sections or subsections these menus are revealed and can be selected to jump to this part of the course. 

Implemented using paragon's Selectmenu component, and data from the learning_sequences API.

Note: as the L_S api does not yet have completion data, we are holding off on accepting the completion ACs.

Smoke testing and QA testing will be required, as this feature is prominent in the learner experience

### DIFF
--- a/src/courseware/course/CourseBreadcrumbs.jsx
+++ b/src/courseware/course/CourseBreadcrumbs.jsx
@@ -5,29 +5,49 @@ import { FormattedMessage } from '@edx/frontend-platform/i18n';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faHome } from '@fortawesome/free-solid-svg-icons';
 import { useSelector } from 'react-redux';
-import { useModel } from '../../generic/model-store';
+import {
+  Hyperlink, MenuItem, SelectMenu,
+} from '@edx/paragon';
+import { useModel, useModels } from '../../generic/model-store';
 
 /** [MM-P2P] Experiment */
 import { MMP2PFlyoverTrigger } from '../../experiments/mm-p2p';
 
 function CourseBreadcrumb({
-  url, children, withSeparator, ...attrs
+  content, withSeparator,
 }) {
+  const defaultMessage = content.filter(destination => destination.default)[0].label;
   return (
     <>
       {withSeparator && (
         <li className="mx-2 text-primary-500" role="presentation" aria-hidden>/</li>
       )}
-      <li {...attrs}>
-        <a className="text-primary-500" href={url}>{children}</a>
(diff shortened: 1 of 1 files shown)


src/courseware/course/CourseBreadcrumbs.jsx
```diff
@@ -5,29 +5,49 @@ import { FormattedMessage } from '@edx/frontend-platform/i18n';
...
+          {content.map(item => (
+            <MenuItem
+              as={Hyperlink}
+              defaultSelected={item.default}
+              href={item.url}
+            >
+              {item.label}
+            </MenuItem>
+          ))}
+        </SelectMenu>
```
connorhaugh: Open question: if there is only one piece of content (select-menu is just one item), should it be a link instead of a selectmenu?

connorhaugh: Asking product about this

connorhaugh: Product's verdict: it should just render a link. Fix incoming.

kenclary (approved):

src/courseware/course/CourseBreadcrumbs.jsx
```diff
@@ -43,51 +68,55 @@ export default function CourseBreadcrumbs({
...
     if (courseStatus === 'loaded' && sequenceStatus === 'loaded') {
-      return [section, sequence].filter(node => !!node).map((node) => ({
-        id: node.id,
-        label: node.title,
-        url: `${getConfig().LMS_BASE_URL}/courses/${course.id}/course/#${node.id}`,
-      }));
+      temp.push(course.sectionIds.map(id => ({
+        id,
+        label: sections[id].title,
+        default: !!(id === sectionId),
```
mikix: I'm used to seeing `!!` as a coersion into a boolean. But wouldn't `===` already yield a bool?

(code is fine, who cares - just asking in case I can learn something about Javascript oddities)

connorhaugh: You are absolutely correct-- it seems my perspective was lets be REALLY sure we get a boolean here :). I'll trim it out.

PR #643, opened 2021-09-15
feat: add learner type to course homepage

zainab-amir:
#### context:
We want to run a post enrolment Hotjar survey that should be visible to all B2C users on first visit to the course home page with `Start Course` button.

More details about the survey can be found on the related ticket: https://openedx.atlassian.net/browse/VAN-701

#### development notes:
Hotjar survey will be triggered by GTM. In order to show this survey to only B2C user I have added an attribute that can be read from GTM.

This will be removed once the survey has run successfully.

### DIFF
--- a/src/course-home/outline-tab/OutlineTab.jsx
+++ b/src/course-home/outline-tab/OutlineTab.jsx
@@ -1,6 +1,7 @@
 import React, { useState } from 'react';
 import { useSelector } from 'react-redux';
 import { sendTrackEvent, sendTrackingLogEvent } from '@edx/frontend-platform/analytics';
+import { getAuthenticatedUser } from '@edx/frontend-platform/auth';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 
 import { Button, Toast } from '@edx/paragon';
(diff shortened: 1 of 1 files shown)


src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -107,9 +108,19 @@ function OutlineTab({ intl }) {
     });
   };
 
+  const isEnterpriseUser = () => {
+    const { authenticatedUser } = useContext(AppContext);
+    const userRoleNames = authenticatedUser.roles.map(role => role.split(':')[0]);
```
ciduarte: I believe we're trying to move away from getting the authenticatedUser from `useContext`, and we want to lean towards using the frontend-platform mechanism instead:
```
import { getAuthenticatedUser } from '@edx/frontend-platform/auth';

const { roles } = getAuthenticatedUser();
```

zainab-amir: I normally use getAuthenticatedUser from `@edx/frontend-platform/auth` but saw that almost everywhere in learning MFE `useContext` was used. I used it for consistency purpose but I can update it.

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -117,7 +128,7 @@ function OutlineTab({ intl }) {
       >
         {goalToastHeader}
       </Toast>
-      <div className="row w-100 mx-0 my-3 justify-content-between">
+      <div learner-type={learnerType} className="row w-100 mx-0 my-3 justify-content-between">
```
waheedahmed: It would be great if we could set it as a data attribute instead of a custom attribute.

waheedahmed (approved):

ciduarte (approved):

PR #646, opened 2021-09-17
[AA-906] number of days goals front end

cdeery:
WIP: Looking for feedback on overall structure and productization for [AA-906].

Feature work for https://openedx.atlassian.net/browse/AA-906
WeeklyLearningGoals


src/course-home/data/api.js
```diff
@@ -390,6 +390,15 @@ export async function postCourseGoals(courseId, goalKey) {
   return getAuthenticatedHttpClient().post(url.href, { course_id: courseId, goal_key: goalKey });
 }
 
+export async function postWeeklyCourseGoals(courseId, daysPerWeek, subscribedToReminders) {
```
MatthewPiatetsky, line 394: it looks like throughout you're calling our new version of goals WeeklyCourseGoals and calling the previous version CourseGoals, which i think makes the previous version seem more like the de facto state.
perhaps it would be better to call the new goals CourseGoals and call the old goals DeprecatedCourseGoals or something?
or even if we keep the Weekly in the name adding a deprecated to the calls such as the one above this might be nice?
but open to other opinions on this

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -31,6 +32,7 @@ import AccountActivationAlert from '../../alerts/logistration-alert/AccountActiv
 
 /** [MM-P2P] Experiment */
 import { initHomeMMP2P, MMP2PFlyover } from '../../experiments/mm-p2p';
+import WeeklyLearningGoal from './widgets/WeeklyLearningGoal';
```
MatthewPiatetsky: i would definitely pick if you want to call this WeeklyCourseGoal or WeeklyLearningGoal and try to stay consistent throughout

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -173,6 +176,13 @@ function OutlineTab({ intl }) {
               setGoalToastHeader={(newHeader) => { setGoalToastHeader(newHeader); }}
             />
           )}
+          {resumeCourseUrl && numberOfDaysGoalsEnabled && (
```
MatthewPiatetsky: what are these changes?

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -212,6 +222,13 @@ function OutlineTab({ intl }) {
                 setGoalToastHeader={(newHeader) => { setGoalToastHeader(newHeader); }}
               />
             )}
+            {numberOfDaysGoalsEnabled && (
```
MatthewPiatetsky: lets stay consistent on whether this is WeeklyCourseGoals, WeeklyLearningGoals or numberOfDaysGoals

src/course-home/outline-tab/widgets/FlagButton.jsx
```diff
@@ -0,0 +1,53 @@
...
+  const [isHighlight, setHighlight] = useState(false);
+
+  function getOutlineStyle() {
+    return isEnabled || isHighlight ? selectedOutlineStyle : baseOutlineStyle;
+  }
+
+  return (
+    <button
+      type="button"
+      className={getOutlineStyle()}
```
MatthewPiatetsky: you can use classNames here instead of className, so that the shared classes will only need to be written once

src/course-home/outline-tab/widgets/FlagButton.jsx
```diff
@@ -0,0 +1,53 @@
...
+function FlagButton({
+  icon,
+  title,
+  text,
+  isEnabled,
+  handleSelect,
+}) {
+  const baseOutlineStyle = 'col flex-grow-1 p-3 border border-light rounded bg-white';
+  const selectedOutlineStyle = 'col flex-grow-1 p-3 border border-dark rounded bg-white';
+  const [isHighlight, setHighlight] = useState(false);
```
MatthewPiatetsky: is this just setting the styling for the active button in the selector group?
i wonder if you can just set the :active css property on the buttons and then you could do this part without javascript
i know this might not exactly fit the radio buttons that might do this for free, but it does feel odd that all this would be necessary for just some styling

src/course-home/outline-tab/widgets/FlagButton.jsx
```diff
@@ -0,0 +1,53 @@
+import React, { useState } from 'react';
+
+import PropTypes from 'prop-types';
+
+function FlagButton({
```
MatthewPiatetsky, line 6: i think there could be a clearer name than FlagButton

Maybe GoalSelectionButton or something along those lines?

src/course-home/outline-tab/widgets/StartResumeCard.jsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Button, Card } from '@edx/paragon';
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+
+import messages from '../messages';
+
+function StartResumeCard({
```
MatthewPiatetsky: Is this the name because of the text? Maybe it could be called something like EnterCourseButtonCard?

src/course-home/outline-tab/widgets/StartResumeCard.jsx
```diff
@@ -0,0 +1,43 @@
...
+        <div className="row w-100 m-0 ">
+          <h2 className="h4 col-auto flex-grow-1">{intl.formatMessage(messages.startBlurb)}</h2>
+          <div className="col col-auto p-0 justify-content-end">
+            <Button
+              variant="brand"
+              block
+              href={resumeCourseUrl}
+              onClick={() => logResumeCourseClick()}
+            >
+              {hasVisitedCourse ? intl.formatMessage(messages.resume) : intl.formatMessage(messages.start)}
```
MatthewPiatetsky: what mock changes is this implementing?

src/course-home/outline-tab/widgets/WeeklyLearningGoal.jsx
```diff
@@ -0,0 +1,134 @@
...
+
+  function handleSelect(days) {
+    setDaysPerWeekGoal(days);
+    saveWeeklyCourseGoal(courseId, days, isGetReminderChecked).then((response) => {
+      const { data } = response;
+      const {
+        header,
+        message,
+      } = data;
+      // TODO: add Toast?, remove console.log
```
MatthewPiatetsky: TODO

src/course-home/outline-tab/widgets/WeeklyLearningGoal.jsx
```diff
@@ -0,0 +1,119 @@
...
+import FlagButton from './FlagButton';
+
+import { saveWeeklyCourseGoal } from '../../data';
+
+function WeeklyLearningGoal({
+  selectedGoal,
+  courseId,
+  intl,
+}) {
+  // eslint-disable-next-line react/prop-types
```
MatthewPiatetsky: what's this disable doing?

src/course-home/outline-tab/widgets/WeeklyLearningGoal.jsx
```diff
@@ -0,0 +1,119 @@
...
+  const [daysPerWeekGoal, setDaysPerWeekGoal] = useState('daysPerWeek' in selectedGoal ? selectedGoal.daysPerWeek : 0);
```
MatthewPiatetsky: maybe null would be a clearer unset state than 0?

src/course-home/outline-tab/widgets/WeeklyLearningGoal.jsx
```diff
@@ -0,0 +1,119 @@
...
+    CASUAL: 3,
+    REGULAR: 4,
+    INTENSE: 5,
+  };
+  Object.freeze(LevelToDays);
+
+  function handleSelect(days) {
+    setDaysPerWeekGoal(days);
+    saveWeeklyCourseGoal(courseId, days, isGetReminderChecked);
+    // TODO: add Toast? Or is that just for the previous feature because the UI disappears?
```
MatthewPiatetsky: could you clarify the question?

src/course-home/outline-tab/widgets/WeeklyLearningGoal.jsx
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  function handleSubscribeToReminders(event) {
+    const isGetReminders = event.target.checked;
```
MatthewPiatetsky: this variable name can be improved imo

src/course-home/outline-tab/widgets/WeeklyLearningGoal.jsx
```diff
@@ -0,0 +1,119 @@
...
+    setGetReminderChecked(isGetReminders);
+    saveWeeklyCourseGoal(courseId, daysPerWeekGoal, isGetReminders);
+  }
+  const buttonRowStyle = 'row w-100 m-0 p-0 justify-content-around'; // 'row w-100 m-0 flex-grow-1 p-0 justify content-end';
```
MatthewPiatetsky: commented classes?

MatthewPiatetsky (commented):
could you add a video or gif of you using your added UI?
personally i use licecap for that

also a docstring somewhere describing how this feature is working on the front-end might be nice

looks like there are also some failing tests

src/course-home/outline-tab/widgets/StartResumeCard.jsx
ciduarte: I'd suggest `ResumeCourseCard`

src/course-home/outline-tab/messages.js
```diff
@@ -74,6 +82,14 @@ const messages = defineMessages({
     id: 'learning.outline.setGoal',
     defaultMessage: 'To start, set a course goal by selecting the option below that best describes your learning plan.',
   },
+  setWeeklyGoal: {
+    id: 'learning.outline.setWeeklyGoal',
+    defaultMessage: 'Set a weekly learning goal',
+  },
+  setWeeklyGoalDetail: {
+    id: 'learning.outline.setWeeklyGoalDetail',
+    defaultMessage: 'Setting a goal motivates you to finish the course. You can always change it later',
```
ciduarte: There should be a period after "later"

src/course-home/outline-tab/messages.js
```diff
@@ -66,6 +66,14 @@ const messages = defineMessages({
     defaultMessage: 'Open',
     description: 'A button to open the given section of the course outline',
   },
+  startBlurb: {
+    id: 'learning.outline.startBlurb',
+    defaultMessage: 'Begin your course today',
+  },
+  resumeBlurb: {
+    id: 'learning.outline.resumeBlurb',
+    defaultMessage: 'Pick up where you left off',
```
ciduarte, line 75: Sometimes the meaning of text can be lost in translation, but we can add descriptions to avoid that. Example:
```
{
  id: '',
  defaultMessage: '',
  description: '',
}
```

I would add a description here that says "Text describing to the learner that they can return to the last section they visited within the course".

src/course-home/outline-tab/messages.js
```diff
@@ -112,6 +128,38 @@ const messages = defineMessages({
...
+    id: 'learning.outline.goalButtonTextIntense',
+    defaultMessage: '5 days a week',
+  },
+  setGoalReminder: {
+    id: 'learning.outline.setGoalReminder',
+    defaultMessage: 'Set a goal reminder',
+  },
+  goalReminderDetail: {
+    id: 'learning.outline.goalReminderDetail',
+    defaultMessage: 'If we notice you\'re not quite at your goal, we\'ll send you an email reminder.',
```
ciduarte: We don't always get this right, but we're trying to use this character `’` rather than this character `'` around frontend-app-learning for apostrophes. If you replace your apostrophes with `’`, you can avoid the escaping :)

src/course-home/data/api.js
cdeery: Will do. Good call

src/course-home/outline-tab/messages.js
```diff
@@ -112,6 +128,38 @@ const messages = defineMessages({
     defaultMessage: 'Welcome to',
     description: 'This precedes the title of the course',
   },
+  goalButtonTitleCasual: {
+    id: 'learning.outline.goalButtonTitleCasual',
```
ciduarte: nit: I would update this and the subsequent id's to follow this format: `learning.outline.goalButton.casual.title` and `learning.outline.goalButton.casual.text`

and I'd update the key to be `casualGoalButtonTitle`, `casualGoalButtonText`, etc. I think this read more fluidly

src/course-home/outline-tab/OutlineTab.jsx
cdeery: The new UI goes in the place where the StartCourse button was. The StartCourse button moves to just above the Outline

src/course-home/outline-tab/messages.js
```diff
@@ -112,6 +128,38 @@ const messages = defineMessages({
...
+    defaultMessage: 'Casual',
```
ciduarte, line 139: I would add a description to these goal button title strings. Not sure if the text strings also need them...

src/course-home/outline-tab/widgets/FlagButton.jsx
cdeery: I was trying to make it generic, since it didn't have anything specific to goal setting in it.

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -123,7 +126,7 @@ function OutlineTab({ intl }) {
         <div className="col-12 col-sm-auto p-0">
           <div role="heading" aria-level="1" className="h2">{title}</div>
         </div>
-        {resumeCourseUrl && (
+        {resumeCourseUrl && !numberOfDaysGoalsEnabled && (
```
ciduarte: I was under the impression we were replacing the resume course button with the resume course card regardless of the goals UI. I don't think this should be impacted by our goals feature

src/course-home/outline-tab/widgets/FlagButton.jsx
cdeery: Cool, I will check it out. I didn't know about :active

src/course-home/outline-tab/widgets/StartResumeCard.jsx
```diff
@@ -0,0 +1,43 @@
...
+  resumeCourseUrl,
+  logResumeCourseClick,
+  intl,
+}) {
+  return (
+    <Card className="mb-3" data-testid="start-resume-panel">
+      <Card.Body>
+        <div className="row w-100 m-0 ">
+          <h2 className="h4 col-auto flex-grow-1">{intl.formatMessage(messages.startBlurb)}</h2>
+          <div className="col col-auto p-0 justify-content-end">
```
ciduarte: I think you can replace these rows and columns with
```
<div className="d-flex justify-content-between">
    <h2 className="h4">{intl.formatMessage(messages.startBlurb)}</h2>
    <Button>...</Button>
</div>
```

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -173,6 +176,13 @@ function OutlineTab({ intl }) {
               setGoalToastHeader={(newHeader) => { setGoalToastHeader(newHeader); }}
             />
           )}
+          {resumeCourseUrl && numberOfDaysGoalsEnabled && (
+          <StartResumeCard
+            hasVisitedCourse={hasVisitedCourse}
+            resumeCourseUrl={resumeCourseUrl}
+            logResumeCourseClick={logResumeCourseClick}
```
ciduarte: Rather than passing these three props through from `OutlineTab`, I would just grab them within this component.

You can use this to grab the course ID inside of `StartResumeCard` and go from there:
```
const {
  courseId,
} = useSelector(state => state.courseHome);
```

src/course-home/outline-tab/OutlineTab.jsx
ciduarte: I'd also consider whether we want it to include the word `Weekly`. Seems to me like this could just be `LearningGoal`, `postLearningGoal`... Maybe that's not specific enough? 🤷‍♀️

src/course-home/outline-tab/widgets/WeeklyLearningGoal.jsx
```diff
@@ -0,0 +1,119 @@
...
+import { Email } from '@edx/paragon/icons';
+import { ReactComponent as FlagIntenseIcon } from '@edx/paragon/icons/svg/flag.svg';
+import { ReactComponent as FlagCasualIcon } from './flag_outline.svg';
+import { ReactComponent as FlagRegularIcon } from './flag_gray.svg';
+import messages from '../messages';
+import FlagButton from './FlagButton';
+
+import { saveWeeklyCourseGoal } from '../../data';
+
+function WeeklyLearningGoal({
```
ciduarte: Whatever name we land on for this, I'd argue this component should have `Card` at the end of it. ex. `WeeklyLearningGoalCard`, `LearningGoalCard`

src/course-home/outline-tab/widgets/WeeklyLearningGoal.jsx
```diff
@@ -0,0 +1,119 @@
...
+  selectedGoal,
+  courseId,
+  intl,
+}) {
+  // eslint-disable-next-line react/prop-types
+  const [daysPerWeekGoal, setDaysPerWeekGoal] = useState('daysPerWeek' in selectedGoal ? selectedGoal.daysPerWeek : 0);
+  // eslint-disable-next-line react/prop-types
+  const [isGetReminderChecked, setGetReminderChecked] = useState('subscribedToReminders' in selectedGoal ? selectedGoal.subscribedToReminders : false);
```
ciduarte: nit: I'd prefer if these were called `isReminderSelected` and `setReminderSelected`

src/course-home/outline-tab/widgets/WeeklyLearningGoal.jsx
```diff
@@ -0,0 +1,119 @@
...
+  const LevelToDays = {
```
ciduarte: This variable should start with a lowercase letter. Also, this variable name isn't super clear to me

src/course-home/outline-tab/widgets/WeeklyLearningGoal.jsx
```diff
@@ -0,0 +1,119 @@
...
+  function handleSubscribeToReminders(event) {
+    const isGetReminders = event.target.checked;
+    setGetReminderChecked(isGetReminders);
+    saveWeeklyCourseGoal(courseId, daysPerWeekGoal, isGetReminders);
+  }
+  const buttonRowStyle = 'row w-100 m-0 p-0 justify-content-around'; // 'row w-100 m-0 flex-grow-1 p-0 justify content-end';
+  const flagButtonStyle = 'col-auto col-md-12 col-xl-auto m-0 p-0 pb-md-3 pb-xl-0'; // 'col-auto flex-grow-1 p-0';
+
+  return (
+    <div className="row w-100 m-0 p-0">
```
ciduarte: does this need to be wrapped in a row?

src/course-home/outline-tab/widgets/WeeklyLearningGoal.jsx
```diff
@@ -0,0 +1,119 @@
...
+      <Card className="mb-3" data-testid="course-goal-card">
+        <Card.Body>
+          <Card.Title>
+            <div className="h4 m-0">{intl.formatMessage(messages.setWeeklyGoal)}</div>
```
ciduarte: can this just be an `<h4>`?

src/course-home/outline-tab/widgets/WeeklyLearningGoal.jsx
```diff
@@ -0,0 +1,119 @@
...
+          </Card.Title>
+          <Card.Text>
+            {intl.formatMessage(messages.setWeeklyGoalDetail)}
+          </Card.Text>
+          <div className={buttonRowStyle}>
+            <div className={flagButtonStyle}>
```
ciduarte: I'd argue for readability, these shouldn't be variables. Although repetitive, the class name strings should just be inline

src/course-home/outline-tab/widgets/FlagButton.jsx
ciduarte: yeah I'd like to see this be `WeeklyLearningGoalButton` and treated like a component with variants. so we pass in a variant and a number of days, and it handles the icon and strings that are used.

```
<WeeklyLearningGoalButton
  variant="casual"
  daysPerWeek="1"
  /* and whatever other props are necessary */
/>
```

and within the component you could handle the string variant like this:
```
`learning.outline.goalButton.${variant}.title`
`learning.outline.goalButton.${variant}.text`
```

src/course-home/outline-tab/widgets/FlagButton.jsx
```diff
@@ -0,0 +1,53 @@
...
+  icon,
+  title,
+  text,
+  isEnabled,
+  handleSelect,
+}) {
+  const baseOutlineStyle = 'col flex-grow-1 p-3 border border-light rounded bg-white';
+  const selectedOutlineStyle = 'col flex-grow-1 p-3 border border-dark rounded bg-white';
```
ciduarte: A few points here
1. We can extract the common styles here and use a ternary for the border: `col flex-grow-1 p-3 border rounded bg-white ${isSelected ? 'border-dark ': 'border-light'}`
2. I'd put this styling right inside of the component instead of assigning it to a variable
3. I think you can replace the `col flex-grow-1` with `col-12`

src/course-home/outline-tab/widgets/FlagButton.jsx
ciduarte: I don't know off the top of my head if there's an accessibility concern here with using `active` or not. I think we should try to use it, but feels like a Jeff Witt question.

Also, I'd prefer if these were renamed to `isSelected` and `setIsSelected`

src/course-home/outline-tab/widgets/FlagButton.jsx
```diff
@@ -0,0 +1,53 @@
...
+  const [isHighlight, setHighlight] = useState(false);
+
+  function getOutlineStyle() {
+    return isEnabled || isHighlight ? selectedOutlineStyle : baseOutlineStyle;
+  }
```
ciduarte: For the `isEnabled` prop, you could pull the `levelToDays` enum into this component and match the variant with the `daysPerWeek`? Rather than setting the state to false on render.

```
const [isSelected, setIsSelected] = useState(daysPerWeek === levelToDays[variant]);
```

src/course-home/outline-tab/widgets/FlagButton.jsx
```diff
@@ -0,0 +1,53 @@
...
+      className={getOutlineStyle()}
+      onMouseEnter={() => setHighlight(true)}
+      onMouseLeave={() => setHighlight(false)}
+      onClick={() => handleSelect()}
+    >
+      <div className=" justify-content-center">
+        {icon}
+      </div>
+      <div className="text-center small">
+        {title}
```
ciduarte, line 29: Can we ask Jeff Witt/Kavya if this just reading "Casual" to screen reader users is okay? Currently this button would read "Casual 1 day a week"

I think it should read "Set a Casual learning goal of 1 day a week"
So we'd prepend and append screen reader only text:
```
<span className="sr-only">Set a </span>
{title}
<span className="sr-only"> learning goal of </span>
/*TODO: translate the sr-only text*/
```

ciduarte (commented):
We should get an a11y review of this feature before it's released. Curious about the active state that Matthew mentioned, and if all these interactive parts are correctly associated w/ one another to avoid confusion amongst screen reader users.

src/course-home/outline-tab/OutlineTab.jsx
cdeery: The previous feature used CourseGoal and I want to be able to distinguish them, for my own sake. NumberOfDaysGoal or DaysPerWeekGoal seem a little long. I will try to be consistent with WeeklyLearningGoal

src/course-home/outline-tab/OutlineTab.jsx
cdeery: I was under the impression that it only had to move because of the space it takes up in the side column, but it's definitely simpler to just move it.

src/alerts/logistration-alert/AccountActivationAlert.jsx
```diff
@@ -29,22 +32,12 @@ function AccountActivationAlert() {
...
     Cookies.remove('show-account-activation-popup', { path: '/', domain: process.env.SESSION_COOKIE_DOMAIN });
     // extra check to make sure cookie was removed before updating the state. Updating the state without removal
-    // of cookie would make it infinit rendering
+    // of cookie would make it infinite rendering
     if (Cookies.get('show-account-activation-popup') === undefined) {
       setShowModal(true);
     }
   }
 
-  const title = (
```
cdeery, line 38: This was filling the console with errors, so I refactored it to fit the propTypes for the AlertModal.

src/course-home/outline-tab/widgets/FlagButton.jsx
cdeery: I had something like that originally, and it felt cleaner to make a component that was less clever.

src/course-home/outline-tab/widgets/FlagButton.jsx
cdeery: This definitely needs more work to appear correctly as an active button to screen readers.

src/course-home/outline-tab/widgets/FlagButton.jsx
cdeery: I added screen reader text. Should I make the rest of it invisible to screen readers so it's clear?

PR #647, opened 2021-09-20
feat:  breadcrumb rolloutout flag + analytics

connorhaugh:
As an addendum to https://openedx.atlassian.net/browse/TNL-7107, we want to hide rollout behind a frontend feature flag added in https://github.com/edx/edx-internal/pull/5489. We also want to report these events to the events api with name `edx.ui.lms.jump_nav.selected`. Documentation to add this event is listed at the following PR: https://github.com/edx/edx-documentation/pull/1982


connorhaugh:
Still need to add unit tests, but defaults for the environments seem to be working.

.env.development
```diff
@@ -37,3 +37,5 @@ TWITTER_HASHTAG='myedxjourney'
 TWITTER_URL='https://twitter.com/edXOnline'
 USER_INFO_COOKIE_NAME='edx-user-info'
 SESSION_COOKIE_DOMAIN='localhost'
+ENABLE_JUMPNAV=''
```
kdmccormick: is this really the value you want in development mode?

.env.test
```diff
@@ -36,3 +36,4 @@ TERMS_OF_SERVICE_URL='https://www.edx.org/edx-terms-service'
 TWITTER_HASHTAG='myedxjourney'
 TWITTER_URL='https://twitter.com/edXOnline'
 USER_INFO_COOKIE_NAME='edx-user-info'
+ENABLE_JUMPNAV=''
```
kdmccormick: same question

README.rst
```diff
@@ -109,3 +109,9 @@ TWITTER_URL
   unless this is set. Optional.
 
   Example: https://twitter.com/edXOnline
+
+ENABLE_JUMPNAV
+  Enables the new Jump Navigation feature in the course breadcrumbs, defaulted to 'true'.
+  Disable to have simple hyperlinks for breadcrumbs.
```
kdmccormick: What would I set it to to disable it?

(yeah, I know 'false' is kinda obvious, but can booleans-as-strings can get confusing sometimes. best to be explicit.)

.env
```diff
@@ -37,3 +37,4 @@ TWITTER_HASHTAG=''
 TWITTER_URL=''
 USER_INFO_COOKIE_NAME=''
 SESSION_COOKIE_DOMAIN=''
+ENABLE_JUMPNAV='false'
```
kdmccormick: this is the default value -- so you're actually defaulting it to 'false' right now.

src/index.jsx
```diff
@@ -92,6 +92,7 @@ initialize({
         CONTACT_URL: process.env.CONTACT_URL || null,
         CREDENTIALS_BASE_URL: process.env.CREDENTIALS_BASE_URL || null,
         ENTERPRISE_LEARNER_PORTAL_HOSTNAME: process.env.ENTERPRISE_LEARNER_PORTAL_HOSTNAME || null,
+        ENABLE_JUMPNAV: process.env.ENABLE_JUMPNAV || 'true',
```
kdmccormick: As counterintuitive as it is, I'd go with the grain here and go with `|| null` instead of `|| 'true'`. The best place to set the default is in the `.env` file, which is where folks will look to see what variables are available. This `mergeConfig` block is just a bit of boilerplate that we have in all MFEs for some reason.

kdmccormick (changes requested):
Tracking stuff looks good.

Some change requests are around the config stuff. I feel like we should have some better docs around all of this...

kdmccormick:
@connorhaugh Here's how the frontend config process works, since it'll probably clarify my suggestions.

* When an MFE is built, `NODE_ENV` is checked.
  * If there is a matching `.env.$NODE_ENV`, then that file is used to populated `process.ENV`.
  * Otherwise, `.env` is used to populate `process.ENV`.
  * Additionally, build configuration values can be explicitly passed in on the command line. These values override what's in the `.env*` files.
* Once the frontend is loaded, frontend-platform calls `initialize`, and merges the config from `process.ENV` into its internal configuration dictionary, which is exposed through `getConfig()`.

How does this work in production? In our prod-frontend-app-* and stage-frontend-app-* pipelines, each
 MFE is built by invoking `npm run build`.  Since NODE_ENV is `production`, the default `.env` is used. However, the pipelines also load the corresponding `*-internal/frontends/frontend-app-*/*_config.yaml` file and pass the values in on the command line there, overriding the default values.

src/index.jsx
connorhaugh: Ahh, that makes sense, just kind of a jumbled "where is the thing that happens first" situtation. I'm planning to turn your reviews, etc into a doc or confluence page.

README.rst
connorhaugh: I have added additional info.

.env
connorhaugh: I have made this "true", as true should be default, then it can be overwritten in prod/edge.

.env.development
connorhaugh: set to true.

.env.test
connorhaugh: set to true.

connorhaugh:
@kdmccormick I've addressed you concerns, making "true" the default, and letting the config overrides for prod and edge be "false". That should work. I'd love to finish off the review/ approval pending code coverage (which is taking a while, bc jest).

README.rst
```diff
@@ -109,3 +109,9 @@ TWITTER_URL
   unless this is set. Optional.
 
   Example: https://twitter.com/edXOnline
+
+ENABLE_JUMPNAV
+  Enables the new Jump Navigation feature in the course breadcrumbs, defaulted to  the string 'true'.
+  Disable to have simple hyperlinks for breadcrumbs. Setting it to any other value but 'true' ('false','I love flags', 'etc' would disable the Jumpnav).
```
kdmccormick, line 115: I appreciate the specificity here 💯

kdmccormick (approved):
LGTM pending codecov 👍🏻 

Heads up -- someone could deploy stage frontend-app-learning to production at any point. If this isn't merged yet, then that would push the jumpnav out to everyone, right? If so, I would recommending either pausing the frontend-app-learning pipeline (with a message referencing this PR) or temporarily reverting your earlier PR. If you're not able to merge this by EOD today, then I'd recommend the latter option, as to not clog up the pipeline for everyone for too long.

PR #652, opened 2021-09-24
fix: hardcode the newrelic agent, to load it earlier

mikix:
This is a test, before making a more proper fix in frontend-build. But I'd like to confirm this fixes some issues we've seen with newrelic metrics.

https://openedx.atlassian.net/browse/AA-1015


public/index.html
```diff
@@ -5,6 +5,16 @@
...
+        ;NREUM.loader_config={accountID:"<%=process.env.NEW_RELIC_ACCOUNT_ID%>",trustKey:"<%=process.env.NEW_RELIC_TRUST_KEY%>",agentID:"<%=process.env.NEW_RELIC_AGENT_ID%>",licenseKey:"<%=process.env.NEW_RELIC_LICENSE_KEY%>",applicationID:"<%=process.env.NEW_RELIC_APP_ID%>"}
+        ;NREUM.info={beacon:"bam-cell.nr-data.net",errorBeacon:"bam-cell.nr-data.net",licenseKey:"<%=process.env.NEW_RELIC_LICENSE_KEY%>",applicationID:"<%=process.env.NEW_RELIC_APP_ID%>",sa:1}
+      </script>
```
mikix, line 15: This is agent version 1209, not 1210 (the latest). I'm using that for now to stick with what we currently use in frontend-build, to reduce variables when confirming the fix.

webpack.prod.config.js
```diff
@@ -0,0 +1,18 @@
+const { getBaseConfig } = require('@edx/frontend-build');
+
+const config = getBaseConfig('webpack-prod');
+
+// Filter plugins in the preset config that we don't want
+function filterPlugins(plugins) {
+  const pluginsToRemove = [
+    'a', // "a" is the constructor name of HtmlWebpackNewRelicPlugin
```
mikix, line 8: You may say, "what the heck is this about, Mike?"

To which I'd point at [frontend-payment doing the same thing](https://github.com/edx/frontend-app-payment/blob/3e924e8/webpack.prod.config.js#L31). Or the names of all the plugins we load:

```
CleanWebpackPlugin
MiniCssExtractPlugin
HtmlWebpackPlugin
Dotenv
a
NewRelicPlugin
BundleAnalyzerPlugin
```

Notice an odd-plugin out there?

MatthewPiatetsky (approved):
seems like something that has potential to break things, but as long as we're monitoring new relic when this goes out seems good

public/index.html
cdeery: Would it make sense to include this as a file with the agent version number in it, for sanity? I don't think this is easy to look at and say "oh, right, version 1209". If we ever have trouble with NewRelic, the first thing they are likely to ask is "are you running the latest version?"

mikix: This is hopefully only a very short term thing, to validate it helps, before a proper fix goes into frontend-build. That fix be nicer hopefully, and yeah, will likely include a comment on version.

mikix: Also how to update it, etc

mikix: (But also just for flavor - the way NewRelic says to check your agent version is to search through that blob for `nr-` and find the one (there's multiple hits!) that looks like a filename and look at the version there. Madness, versus them just having it in a comment at the beginning or something.)

PR #653, opened 2021-09-24
fix: correct some escaping on the new relic agent string

mikix:


public/index.html
```diff
@@ -7,9 +7,10 @@
     <link rel="shortcut icon" href="<%=htmlWebpackPlugin.options.FAVICON_URL%>" type="image/x-icon" />
 
     <% if (process.env.NEW_RELIC_ACCOUNT_ID) { %>
+      <!-- NewRelic agent 1209 -->
```
cdeery, line 10: Looks good (as far as a wall of script can look)

cdeery (approved):

PR #658, opened 2021-09-28
fix: don't show 3-day streak discount if it won't provide a discount

mikix:
Also, this will pull the actual discount percent from ecommerce, instead of hardcoding it.

When a once-only coupon is already claimed by a user, ecommerce will tell us there isn't a discount to be had.

Here is the other PRs related to this: https://github.com/edx/edx-platform/pull/28868

https://openedx.atlassian.net/browse/AA-1012

### DIFF
--- a/src/shared/streak-celebration/StreakCelebrationModal.jsx
+++ b/src/shared/streak-celebration/StreakCelebrationModal.jsx
@@ -1,11 +1,14 @@
 import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
+import { camelCaseObject, getConfig } from '@edx/frontend-platform';
+import { sendTrackEvent } from '@edx/frontend-platform/analytics';
+import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 import {
   FormattedMessage, injectIntl, intlShape,
 } from '@edx/frontend-platform/i18n';
 import { Lightbulb, MoneyFilled } from '@edx/paragon/icons';
 import {
-  Alert, Icon, ModalDialog,
+  Alert, Icon, ModalDialog, Spinner,
 } from '@edx/paragon';
 import { layoutGenerator } from 'react-break';
 import { useDispatch } from 'react-redux';
@@ -41,17 +44,30 @@ function getRandomFactoid(intl, streakLength) {
   return factoids[Math.floor(Math.random() * (factoids.length))];
 }
 
+async function calculateVoucherDiscount(voucher, sku, username) {
+  const urlBase = `${getConfig().ECOMMERCE_BASE_URL}/api/v2/baskets/calculate`;
+  const url = `${urlBase}/?code=${voucher}&sku=${sku}&username=${username}`;
+  return getAuthenticatedHttpClient().get(url)
+    .then(res => camelCaseObject(res));
+}
+
 function StreakModal({
   courseId, metadataModel, streakLengthToCelebrate, intl, isStreakCelebrationOpen,
-  closeStreakCelebration, StreakDiscountCouponEnabled, verifiedMode, ...rest
+  closeStreakCelebration, streakDiscountCouponEnabled, verifiedMode, ...rest
 }) {
   if (!isStreakCelebrationOpen) {
     return null;
   }
-  const { org, celebrations } = useModel(metadataModel, courseId);
+  const { org, celebrations, username } = useModel(metadataModel, courseId);
   const factoid = getRandomFactoid(intl, streakLengthToCelebrate);
   // eslint-disable-next-line no-unused-vars
   const [randomFactoid, setRandomFactoid] = useState(factoid); // Don't change factoid on re-render
+
+  // Open edX Folks: if you create a voucher with this code, the MFE will notice and show the discount
+  const discountCode = 'ZGY11119949';
+  // Negative means "we don't know yet" vs zero meaning no discount available
+  const [discountPercent, setDiscountPercent] = useState(-1);
+  const queryingDiscount = discountPercent < 0;
 
   const layout = layoutGenerator({
     mobile: 0,
@@ -68,6 +84,37 @@ function StreakModal({
     }
   }, [isStreakCelebrationOpen, org, courseId]);
 
+  // Ask ecommerce to calculate discount savings
+  useEffect(() => {
+    if (streakDiscountCouponEnabled && verifiedMode && getConfig().ECOMMERCE_BASE_URL) {
+      calculateVoucherDiscount(discountCode, verifiedMode.sku, username)
+        .then(
+          (result) => {
+            const { totalInclTax, totalInclTaxExclDiscounts } = result.data;
+            if (totalInclTaxExclDiscounts && totalInclTax !== totalInclTaxExclDiscounts) {
+              // Just store the percent (rather than using these values directly), because ecommerce doesn't give us
+              // the currency symbol to use, so we want to use the symbol that LMS gives us. And I don't want to assume
+              // ecommerce's currency is the same as the LMS. So we'll keep using the values in verifiedMode, just
+              // multiplied by the calculated percentage.
+              setDiscountPercent(1 - totalInclTax / totalInclTaxExclDiscounts);
+              sendTrackEvent('edx.bi.course.streak_discount_enabled', {
+                course_id: courseId,
+                sku: verifiedMode.sku,
+              });
+            } else {
+              setDiscountPercent(0);
+            }
+          },
+          () => {
+            // ignore any errors - we just won't show the discount to the user then
+            setDiscountPercent(0);
+          },
+        );
+    } else {
+      setDiscountPercent(0);
+    }
+  }, [streakDiscountCouponEnabled, username, verifiedMode]);
+
   function CloseText() {
     return (
       <span>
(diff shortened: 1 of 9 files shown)


src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -41,18 +44,31 @@ function getRandomFactoid(intl, streakLength) {
...
-  const { org, celebrations } = useModel(metadataModel, courseId);
+  const { org, celebrations, username } = useModel(metadataModel, courseId);
   const factoid = getRandomFactoid(intl, streakLengthToCelebrate);
   // eslint-disable-next-line no-unused-vars
   const [randomFactoid, setRandomFactoid] = useState(factoid); // Don't change factoid on re-render
 
+  // Open edX Folks: if you create a voucher with this code, the MFE will notice and show the discount
+  const discountCode = 'ZGY11119949';
+  // Negative means "we don't know yet" vs zero meaning no discount available
+  const [discountPercent, setDiscountPercent] = useState(-1);
```
MatthewPiatetsky, line 69: couldn't you just make the default state null and make the if statement be if (discountPercent)?

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -68,6 +84,37 @@ function StreakModal({
...
+              // the currency symbol to use, so we want to use the symbol that LMS gives us. And I don't want to assume
+              // ecommerce's currency is the same as the LMS. So we'll keep using the values in verifiedMode, just
+              // multiplied by the calculated percentage.
+              setDiscountPercent(1 - totalInclTax / totalInclTaxExclDiscounts);
+              sendTrackEvent('edx.bi.course.streak_discount_enabled', {
+                course_id: courseId,
+                sku: verifiedMode.sku,
+              });
+            } else {
+              setDiscountPercent(0);
```
MatthewPiatetsky, line 105: maybe we should use null here instead of 0?

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -68,6 +84,37 @@ function StreakModal({
     }
   }, [isStreakCelebrationOpen, org, courseId]);
 
+  // Ask ecommerce to calculate discount savings
+  useEffect(() => {
+    if (streakDiscountCouponEnabled && verifiedMode && getConfig().ECOMMERCE_BASE_URL) {
+      calculateVoucherDiscount(discountCode, verifiedMode.sku, username)
```
MatthewPiatetsky, line 90: in the past we've had an issue where because lms has a lot more traffic than ecommerce, adding requests from lms to ecommerce caused ecommerce to crash
so we might want to consider having a flag to toggle this off for safety?
as is this might be the most frequent call to ecommerce?
also perhaps only making the request when the celebration is open?

src/tab-page/LoadedTabPage.jsx
```diff
@@ -59,7 +59,7 @@ function LoadedTabPage({
         streakLengthToCelebrate={streakLengthToCelebrate}
         isStreakCelebrationOpen={!!isStreakCelebrationOpen}
         closeStreakCelebration={closeStreakCelebration}
-        StreakDiscountCouponEnabled={StreakDiscountCouponEnabled}
+        streakDiscountCouponEnabled={streakDiscountCouponEnabled}
```
MatthewPiatetsky, line 62: i guess same question as on the backend PR, what is the reasoning for not checking this value on the backend and passing it through this variable
(also not sure if we checked if this information is available within the lms or only in ecommerce)

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -82,21 +129,25 @@ function StreakModal({
...
       price: verifiedMode.price,
       upgradeUrl,
     };
 
-    offer = {
-      discountedPrice: `${verifiedMode.currencySymbol}${(mode.price * 0.85).toFixed(2).toString()}`,
-      originalPrice: `${verifiedMode.currencySymbol}${mode.price.toString()}`,
-      upgradeUrl: mode.upgradeUrl,
-    };
+    if (discountPercent > 0) {
```
MatthewPiatetsky, line 139: aren't values less than 0 already falsey?

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -41,18 +44,31 @@ function getRandomFactoid(intl, streakLength) {
...
+  const urlBase = `${getConfig().ECOMMERCE_BASE_URL}/api/v2/baskets/calculate`;
+  const url = `${urlBase}/?code=${voucher}&sku=${sku}&username=${username}`;
+  return getAuthenticatedHttpClient().get(url)
+    .then(res => camelCaseObject(res));
+}
+
 function StreakModal({
   courseId, metadataModel, streakLengthToCelebrate, intl, isStreakCelebrationOpen,
-  closeStreakCelebration, StreakDiscountCouponEnabled, verifiedMode, ...rest
+  closeStreakCelebration, streakDiscountCouponEnabled, verifiedMode, ...rest
```
cdeery, line 56: Thanks for catching that. My bad.

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -125,21 +176,26 @@ function StreakModal({
             <img src={StreakDesktopImage} alt="" className="img-fluid" />
           </OnDesktop>
         </p>
-        { !StreakDiscountCouponEnabled && (
+        { queryingDiscount && (
```
cdeery, line 179: So if there is no available discount, and the query takes a little while, the user will get a big alert with a spinner that then goes away? Would this work without the spinner at all? When the discount gets updated, wouldn't the alert be triggered?

src/shared/streak-celebration/StreakCelebrationModal.jsx
mikix: I currently have two "non-discount" states: negative and zero (waiting on answer and no discount available, respectively). I'm setting the discount to zero here to exit the "waiting on answer" state and just show the dialog without a discount.

Are you suggesting a third null state? Or just thought I was doing something else here?

src/shared/streak-celebration/StreakCelebrationModal.jsx
mikix: We only call this if the celebration is open and the discount should be shown (isStreakCelebrationOpen and streakDiscountCouponEnabled respectively).

So we'll hit ecommerce every time the 3-day streak dialog is shown for a user.

I don't know if that will take down ecommerce, but I'd hope not? What do you think?

There isn't a waffle for this code, but it'd be easy to turn off with a one liner if there is a problem.

src/tab-page/LoadedTabPage.jsx
mikix: Reproducing my answer from that PR below:

Yeah the architecture vibe at edX is to not do backend-to-backend blocking requests.

An alternative design for this could have been to have LMS pull coupon usage data from ecommerce and store it locally on the LMS side. (async, eventually consistent data flows are OK, whereas blocking requests are not).

Then the backend could tell the frontend if the coupon was still valid up front. But that seemed like a lot of bother compared to just having the MFE ask ecommerce directly. The backend wouldn't use that coupon data anywhere else... shrug

src/shared/streak-celebration/StreakCelebrationModal.jsx
mikix: No. Only zero is falsy.

mikix: (or maybe some NaN nonsense too, I dunno. But negatives are truthy)

src/shared/streak-celebration/StreakCelebrationModal.jsx
mikix: Maybe... I do want to distinguish between the null and zero state in a few places though. So not sure it buys us too much clarity.

I had been thinking I didn't like mixing "types" by having either a null or a number.

src/shared/streak-celebration/StreakCelebrationModal.jsx
mikix: Ah yeah so the spinner. Should have explained that in the PR description.

The dialog has two different states: discount or factoid mode. Discount mode shows a little box and a button with a price and a button to dismiss. Factoid mode shows a fact like "Users who learn 3 days in a row complete 5x as much course content on average vs. those who don’t" or something. With a button to dismiss (with different text than the discount dismiss button).

So they look very different on the bottom half of the modal.

My first pass at a fix was to show the factoid mode until we got an answer back from ecommerce - then switch modal to the discount view. But it was too jarring. We could also wait to finish loading the page until we heard back from ecommerce - but that makes the wait for any UI even longer. Kavya and I figured a spinner on the bottom part of the modal while we figured out which version to show was best option we had. (User can still dismiss the whole modal by clicking outside of it or on the x in the upper right)

Does that make sense?

> When the discount gets updated, wouldn't the alert be triggered?

Sorry, what do you mean?

src/shared/streak-celebration/StreakCelebrationModal.jsx
MatthewPiatetsky: ok, i guess i have a different personal preference as to what i find clearer but this is fine

src/shared/streak-celebration/StreakCelebrationModal.jsx
MatthewPiatetsky: right, i guess i meant if you had used 0 and null then you could do if (discountPercent) but this makes sense since you're using -1

src/shared/streak-celebration/StreakCelebrationModal.jsx
MatthewPiatetsky: ok i think this is fine, though i'd probably let whoever is on call and revenue team know when you deploy as an fyi in case ecommerce does go down

MatthewPiatetsky (approved):

src/shared/streak-celebration/StreakCelebrationModal.jsx
cdeery: ok, thanks

PR #659, opened 2021-09-29
fix: UpgradeNotification bullet point padding override

julianajlk:
[REV-2188](https://openedx.atlassian.net/browse/REV-2188).

The class `fa-ul` `padding-left: 0` overrides my changes in prod on the bullet points for the `UpgradeNotification` (but not locally). This PR adds an `!important` to the left padding.

Before (prod):
Screen Shot 2021-09-29 at 11 11 58 AM

After (local, both with and without `!important` looks the same):
Screen Shot 2021-09-29 at 11 14 12 AM

### DIFF
--- a/src/generic/upgrade-notification/UpgradeNotification.scss
+++ b/src/generic/upgrade-notification/UpgradeNotification.scss
@@ -19,7 +19,9 @@
 }
 
 .upgrade-notification-ul {
-  padding: 0.875rem 1.25rem 0 1.25rem;
+  padding-left: 1.25rem !important;
+  padding-top: 0.875rem;
+  padding-right: 1.25rem;
 }
 
 .upgrade-notification-li {


dianekaplan (approved):
LGTM! One (probably obvious) note since the issue doesn't show up locally: you'll want to find a case of the issue in stage before deploy, so you can confirm the fix there, and also click around in case there are any unintended side effects.

julianajlk:
> LGTM! One (probably obvious) note since the issue doesn't show up locally: you'll want to find a case of the issue in stage before deploy, so you can confirm the fix there, and also click around in case there are any unintended side effects.

Thank you Diane! In stage it's so difficult to test since I don't end up having the right settings to see the Upgrade Notification version that has the bullet points (specific hours to access expiration deadline) 🙁

PR #663, opened 2021-10-01
Feat  lazy load jump nav destinations

connorhaugh:
In order to finish off [TNL-7107](https://openedx.atlassian.net/secure/RapidBoard.jspa?rapidView=580&projectKey=TNL&modal=detail&selectedIssue=TNL-7107) I needed to meet the acceptance criteria: `When learners or educators select a section dropdown item they are taken to the first subsection within that section that is not completed by default. If all subsections are completed they should be taken to the first(subsection) in that section. `

This reimagining of Jumpnav does that by lazy loading in the menuItem's destinations and routing the user using React-Router.


kenclary (approved):
looks good to me.

PR #665, opened 2021-10-04
[MST-973] Prevent integrity signature creation while masquerading

bseverino:
[MST-973](https://openedx.atlassian.net/browse/MST-973)

If a staff user encounters the Honor Code panel while masquerading as a learner, accepting the conditions should not actually create a signature for that user in the database; it should only remove the panel and display content. If the masquerading user refreshes, the panel will reappear.

This also fixes a minor issue where `TERMS_OF_SERVICE_URL` was being accessed with `process.env` and not `getConfig()`.

### DIFF
--- a/src/courseware/data/thunks.js
+++ b/src/courseware/data/thunks.js
@@ -303,10 +303,15 @@ export function saveSequencePosition(courseId, sequenceId, activeUnitIndex) {
   };
 }
 
-export function saveIntegritySignature(courseId) {
+export function saveIntegritySignature(courseId, isMasquerading) {
   return async (dispatch) => {
     try {
-      await postIntegritySignature(courseId);
+      // If the request is made by a staff user masquerading as a learner,
+      // don't actually create a signature for them on the backend. Only
+      // frontend state will be updated.
+      if (!isMasquerading) {
+        await postIntegritySignature(courseId);
+      }
       dispatch(updateModel({
         modelType: 'coursewareMeta',
         model: {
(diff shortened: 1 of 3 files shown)


src/courseware/data/thunks.js
```diff
@@ -303,10 +303,15 @@ export function saveSequencePosition(courseId, sequenceId, activeUnitIndex) {
   };
 }
 
-export function saveIntegritySignature(courseId) {
+export function saveIntegritySignature(courseId, isMasquerading) {
   return async (dispatch) => {
     try {
-      await postIntegritySignature(courseId);
+      // If the request is made by a staff user masquerading as a learner,
```
schenedx, line 309: With this change, the next time masquerading staff user loads this view, or encounters this, they'd have to face the honor code again, right? What can we do to help in that situation?

schenedx (approved):
I'll approve this change because I think this is a good enough short term solution.
However, I think we should discuss the long term solution with rest of the team and @spencertiberi.

src/courseware/data/thunks.js
bseverino: An alternative solution I can think of is to only display the honor code when masquerading as a _specific_ user who hasn't signed it yet, and don't display it otherwise.

schenedx: Certainly, that solution combined with this one would minimize the impact. But my worry that the honor code still shows remains?

PR #667, opened 2021-10-05
[feat] Add notices redirect to learning MFE

Tj-Tracy:
To support the notices plugin on platform, this adds a redirect to the course home page. If a user lands on that page and has not acknowledged a notice, the user will be redirected to notice instead of the course home. The user can back out of the notice, or acknowledge it. They will continue to see the notice every time they hit any course until it is acknowledged.


src/generic/notices/NoticesProvider.jsx
```diff
@@ -0,0 +1,35 @@
...
+ */
+const NoticesProvider = ({ children }) => {
+  const [isRedirected, setIsRedirected] = useState();
+  useEffect(async () => {
+    if (getConfig().ENABLE_NOTICES) {
+      const data = await getNotices();
+      if (data && data.results) {
+        const { results } = data;
+        setIsRedirected(true);
+        window.location.replace(`${results[0]}?next=${window.location.href}`);
```
MatthewPiatetsky, line 19: i'm not a react expert but i feel like there is probably a cleaner way to do this? Carla may know

src/generic/notices/api.js
```diff
@@ -0,0 +1,18 @@
...
+
+export const getNotices = async () => {
+  const authenticatedUser = getAuthenticatedUser();
+  const url = new URL(`${getConfig().LMS_BASE_URL}/notices/api/v1/unacknowledged`);
+  if (authenticatedUser) {
+    try {
+      const { data } = await getAuthenticatedHttpClient().get(url.href, {});
+      return data;
+    } catch (e) {
+      // we will just swallow error, as that probably means the notices app is not installed.
```
MatthewPiatetsky, line 14: maybe we should be logging something here?
also are we catching all errors? should we be catching only certain errors?

MatthewPiatetsky (commented):
the description of this feature makes me rather scared
what kind of manual testing have we done with the feature disabled to ensure users don't get redirected when:
1. the feature is disabled
2. they've acknowledged the notice

src/generic/notices/NoticesProvider.jsx
ciduarte: This line specifically? 🤷‍♀️

ciduarte (approved):

src/generic/notices/api.js
Tj-Tracy: I've adding logging - if its a 404 (probably meaning the app is not installed) its at the info level. Otherwise, it logs as an error.

src/generic/notices/NoticesProvider.jsx
Tj-Tracy: React helps us with in-domain redirects, but since this is routing outside of the react app, I just used javascript to accomplish it. I'm not aware of any other way we've done this in the past.

MatthewPiatetsky (approved):

PR #671, opened 2021-10-06
fix: Unsubscribe audit users from goal remindners on the course exit page

MatthewPiatetsky:
[edx-platform PR](https://github.com/edx/edx-platform/pull/28952)

### DIFF
--- a/src/courseware/course/course-exit/CourseExit.jsx
+++ b/src/courseware/course/course-exit/CourseExit.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 
 import { getConfig } from '@edx/frontend-platform';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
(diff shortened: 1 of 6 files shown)


src/courseware/course/course-exit/CourseExit.jsx
```diff
@@ -32,6 +36,15 @@ function CourseExit({ intl }) {
     courseExitPageIsActive,
   );
 
+  // Audit users cannot fully complete a course, so we will
+  // unsubscribe them from goal reminders once they reach the course exit page
+  // to avoid spamming them with goal reminder emails
+  if (courseGoals && enrollmentMode === 'audit' && !isMasquerading) {
```
MatthewPiatetsky, line 42: this is the same as the PR that was already merged except i'm not checking that courseGoals are defined (to avoid sending the request if goals are disabled)
i've also added a check to not send the request if the user is masquerading

ciduarte (approved):

PR #672, opened 2021-10-06
feat: rollout jump nav to course staff

connorhaugh:
WATING ON COMMS TO PARTNERS TO MERGE. As per https://openedx.atlassian.net/browse/TNL-7107 we need to rollout jumpnav to course staff.

### DIFF
--- a/src/courseware/course/Course.jsx
+++ b/src/courseware/course/Course.jsx
@@ -90,7 +90,7 @@ function Course({
           courseId={courseId}
           sectionId={section ? section.id : null}
           sequenceId={sequenceId}
-          isStaff={course ? course.isStaff : null} // for TNL-7107 Rollout
+          isStaff={course ? course.isStaff : null}
           unitId={unitId}
           //* * [MM-P2P] Experiment */
           mmp2p={MMP2P}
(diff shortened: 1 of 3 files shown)


kdmccormick:
I don't know what the best practice is, but I feel like passing `isStaff` around isn't something that needs to be annotated with TNL-7107. It's completely reasonable for a component to want to know if the user is course staff, and maybe it's fine to leave the `isStaff` props in place even after we roll jump nav out to learners.

Btw - unless `isStaff` is already established in the MFE to mean "course staff", consider naming calling these variables `isCourseStaff` to avoid any ambiguity.

Otherwise, looks good!

connorhaugh:
> I don't know what the best practice is, but I feel like passing `isStaff` around isn't something that needs to be annotated with [TNL-7107](https://openedx.atlassian.net/browse/TNL-7107). It's completely reasonable for a component to want to know if the user is course staff, and maybe it's fine to leave the `isStaff` props in place even after we roll jump nav out to learners.

Ah, this is mainly to flag that it will need to be removed once jump nav is not gated to staff. I have no problem removing the annotation

> Btw - unless isStaff is already established in the MFE to mean "course staff", consider naming calling these variables isCourseStaff to avoid any ambiguity.

isStaff is established in the MFE insofar as it is the name of the variable in the course metadata which relates that a user is staff. `administrator` is used for edx staff in the MFE.

kdmccormick (approved):

PR #675, opened 2021-10-07
fix: move NewRelic testing script even higher

mikix:
Per their docs, this really should be the first script. And we were hitting some cloudflare scripts being inserted ahead of us. This might fix that.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1,11 +1,6 @@
 <!doctype html>
 <html lang="en-us">
   <head>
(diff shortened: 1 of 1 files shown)


MatthewPiatetsky (approved):

MatthewPiatetsky:
looks like tests are failing?

mikix:
> looks like tests are failing?

Looks like npm is having a day: https://status.npmjs.org/

PR #693, opened 2021-10-19
fix: catch and redirect URLs with spaces in them

mikix:
In an attempt to debug some odd LMS errors (which would happen if you loaded an MFE page with spaces instead of plus signs), this commit notices page requests with spaces in the course key and switches it out for plus signs, logging the incident.

https://openedx.atlassian.net/browse/AA-1075

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -29,56 +29,59 @@ import { fetchDatesTab, fetchOutlineTab, fetchProgressTab } from './course-home/
 import { fetchCourse } from './courseware/data';
 import initializeStore from './store';
 import NoticesProvider from './generic/notices';
+import PathFixesProvider from './generic/path-fixes';
 
 subscribe(APP_READY, () => {
   ReactDOM.render(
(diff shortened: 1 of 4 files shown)


src/index.jsx
```diff
@@ -29,56 +29,59 @@ import { fetchDatesTab, fetchOutlineTab, fetchProgressTab } from './course-home/
 import { fetchCourse } from './courseware/data';
 import initializeStore from './store';
 import NoticesProvider from './generic/notices';
+import PathFixesProvider from './generic/path-fixes';
 
 subscribe(APP_READY, () => {
   ReactDOM.render(
     <AppProvider store={initializeStore()}>
-      <NoticesProvider>
```
mikix, line 36: No change here - just wrapping whole thing around `PathFixesProvider`.

Dillon-Dumesnil (approved):
Sure.
Hope it's temporary. 🤞 It's nice, but I hope we solve the root cause instead

PR #694, opened 2021-10-19
fix: [AA-906] WeeklyLearningGoals

cdeery:
Fix incorrect message for returning students. The message was about starting course regardless of status.


mikix (approved):
I assume that message exists

MatthewPiatetsky (approved):

PR #695, opened 2021-10-20
feat: product tour components

ciduarte:
**On desktop:**

https://user-images.githubusercontent.com/25124041/140201418-3fdb4552-3753-4dcd-8e5b-b6be80a066a5.mov

**On mobile:**

https://user-images.githubusercontent.com/25124041/140201300-5e428e2f-0565-4a79-b866-37179e3e2e23.mov

**Testing out extreme sizing scenarios:**

https://user-images.githubusercontent.com/25124041/140201050-5f2ca1a9-0656-4e9c-8ec0-6b85bedf4e88.mov


src/tour/Checkpoint.scss
```diff
@@ -0,0 +1,58 @@
...
+  max-width: $popover-max-width;
+  @media (max-width: map-get($grid-breakpoints, 'md')) {
+    min-width: 90%;
+    max-width: 90%;
+  }
+
+  #arrow,
+  #arrow::before {
+    position: absolute;
+    width: 8px;
```
MatthewPiatetsky: are the px in this file supposed to be rem? not sure about the rules

src/tour/Checkpoint.scss
```diff
@@ -0,0 +1,58 @@
...
+    height: 8px;
+    background: inherit;
+  }
+
+  #arrow {
```
MatthewPiatetsky: is there risk of another component having the id arrow? (a next arrow for example) and having conflicts. 
i was previously told by Alasdair that css rules should target classes rather than ids but not sure if this team agrees with that or not

src/tour/CheckpointActionRow.jsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button } from '@edx/paragon';
+
+export default function CheckpointActionRow({
+ advanceButtonText,
```
MatthewPiatetsky: github actions seems angry

src/tour/Tour.jsx
```diff
@@ -0,0 +1,159 @@
...
+    setIsEnabled(tourValue && tourValue.enabled);
+  }, [tourValue]);
+
+  useEffect(() => {
+    if (tourValue) {
+      setCheckpointData(tourValue.checkpoints[index]);
+    }
+  }, [index, isEnabled]);
+
+  const isMobile = useMediaQuery({ query: '(max-width: 768px)' });
```
MatthewPiatetsky, line 34: is this something that should be in a more general place?

src/tour/Tour.jsx
```diff
@@ -0,0 +1,161 @@
...
+      onDismiss={handleDismiss}
+      onEnd={handleEnd}
+      title={checkpointData.title}
+      totalCheckpoints={tourValue.checkpoints.length}
+    />
+  );
+}
+
+Tour.defaultProps = {
+//  TODO: define default props
```
MatthewPiatetsky: TODO?

MatthewPiatetsky (commented):
would it be worthwhile to include the example use as a block comment somewhere?  
or is that not worth it because this is documented in paragon and we expect people to go look it up themselves in paragonn?

src/tour/Checkpoint.scss
ciduarte: yes great point, copy pasta got me

src/tour/Checkpoint.scss
```diff
@@ -0,0 +1,58 @@
...
+  @media (max-width: map-get($grid-breakpoints, 'md')) {
+    min-width: 90%;
+    max-width: 90%;
+  }
+
+  #checkpoint-arrow,
+  #checkpoint-arrow::before {
+    position: absolute;
+    width: 8px;
+    height: 8px;
```
cdeery: The arrow in the spec is 28px by 13px. Is this 8x8? It looks a bit small in the video

ciduarte: ooh yeah good catch. this was just some default styling from the popper library. this is actually a square that's been rotated. looks like this is closer to 19px for each side of the triangle

src/tour/Tour.jsx
ciduarte: Yes, but for this implementation I wanted the component to be contained because it'll eventually be ported over to Paragon.

ciduarte:
> would it be worthwhile to include the example use as a block comment somewhere? or is that not worth it because this is documented in paragon and we expect people to go look it up themselves in paragonn?

I could write up a README. This will be documented in Paragon eventually.

src/tour/Checkpoint.scss
ciduarte: I updated some of these, and I tried to mirror where we use px vs rems in Paragon.

Dillon-Dumesnil:
> > would it be worthwhile to include the example use as a block comment somewhere? or is that not worth it because this is documented in paragon and we expect people to go look it up themselves in paragonn?
> 
> I could write up a README. This will be documented in Paragon eventually.

This is what I was about to recommend. Lol. Maybe as a good way to capture this information while also being readily used when it moves to Paragon is write up a README in the `tours` directory and it can all be ported over to Paragon at the same time.

src/tour/Checkpoint.jsx
```diff
@@ -0,0 +1,79 @@
...
+      className="checkpoint-popover p-4 bg-light-300"
+      aria-labelledby="checkpoint-title"
+      role="dialog"
+      style={{ display: hideCheckpoint ? 'none' : 'block' }}
+    >
+      {/* This text is not translated due to Paragon's lack of i18n support */}
+      <span className="sr-only">Top of step {index + 1}</span>
+      {(title || !isOnlyCheckpoint) && (
+        <div className="d-flex justify-content-between mb-2.5">
+          <CheckpointTitle>{title}</CheckpointTitle>
```
Dillon-Dumesnil: I don't know if this is maybe not a real thing, but what goes into `CheckpointTitle` if title is null, but `!isOnlyCheckpoint` is true?
Looking at `CheckpointTitle`, it looks like it will just return null. What does that visually look like? Should title be required?

src/tour/Checkpoint.jsx
```diff
@@ -0,0 +1,79 @@
...
+      )}
+      <CheckpointBody>{body}</CheckpointBody>
+      <CheckpointActionRow
+        advanceButtonText={advanceButtonText}
+        dismissButtonText={dismissButtonText}
+        endButtonText={endButtonText}
+        isLastCheckpoint={isLastCheckpoint}
+        onAdvance={onAdvance}
+        onDismiss={onDismiss}
+        onEnd={onEnd}
```
Dillon-Dumesnil: Should these just be dropped into this via `{...props}`? It feels like most of the props of this component are pass through

src/tour/Tour.jsx
```diff
@@ -0,0 +1,188 @@
...
+import PropTypes from 'prop-types';
+import { useMediaQuery } from 'react-responsive';
+import { createPopper } from '@popperjs/core';
+
+import Checkpoint from './Checkpoint';
+
+function Tour({
+  tours,
+}) {
+  const tourValue = tours.filter((tour) => tour.enabled)[0];
```
Dillon-Dumesnil, line 11: If there are indeed multiple tours, why does this component just select the 0th tour? Does that mean it should be pre-filtered by anyone choosing to use this component to put the tour they want first?

src/tour/Tour.jsx
```diff
@@ -0,0 +1,188 @@
...
+
+        let targetOffset = targetElement.getBoundingClientRect().top;
+        if (checkpointData.placement && checkpointData.placement.includes('top')) {
+          if (targetOffset < 0) {
+            targetOffset *= -1;
+          }
+          targetOffset -= 280;
+        } else {
+          targetOffset -= 80;
+        }
```
Dillon-Dumesnil, line 70: sure

src/tour/Tour.jsx
```diff
@@ -0,0 +1,188 @@
...
+          top: targetOffset, behavior: 'smooth',
+        });
+      }
+    }
+  }, [checkpointData, index, isMobile]);
+
+  useEffect(() => {
+    const handleEsc = (event) => {
+      if (isEnabled) {
+        if (event.keyCode === 27) {
```
Dillon-Dumesnil: I think you can combine these conditions

src/tour/Tour.jsx
```diff
@@ -0,0 +1,188 @@
...
+            tourValue.onEnd();
+          }
+        }
+      }
+    };
+    window.addEventListener('keydown', handleEsc);
+
+    return () => {
+      window.removeEventListener('keydown', handleEsc);
+    };
```
Dillon-Dumesnil, line 92: Sorry, for my own clarity, when is this code executed? Is it going to remove the Event Listener when the effect is triggered or only after a certain condition?

src/tour/Tour.jsx
```diff
@@ -0,0 +1,188 @@
...
+    }
+  };
+
+  const handleDismiss = () => {
+    setIndex(0);
+    setIsEnabled(false);
+    if (checkpointData.onDismiss) {
+      checkpointData.onDismiss();
+    } else {
+      tourValue.onDismiss();
```
Dillon-Dumesnil, line 112: Is this guaranteed to exist? In most spots, you check existence of `tourValue` or a method before calling it.

Dillon-Dumesnil (commented):
Couple questions, but looks good to me.

src/tour/Tour.jsx
ciduarte: If there are multiple tours defined, only one tour should be enabled at a time. The Tour component restricts multiple enabled tours from rendering at once.

src/tour/Tour.jsx
ciduarte: There's an early exit on line 95, so this will always be true

src/tour/Checkpoint.jsx
ciduarte: We don't want `title` to be required because there are use cases for a `Checkpoint` with no title. This is what the `Checkpoint` is meant to look like with no title. The breadcrumbs would still render on the righthand side. When there's only one `Checkpoint` in a `Tour`, the whole header row goes away.

Screen Shot 2021-11-03 at 4 04 54 PM
Screen Shot 2021-11-03 at 4 07 44 PM

src/tour/Tour.jsx
ciduarte: A return statement within a `useEffect` hook functions like a cleanup method when the component is unmounted

src/tour/Checkpoint.jsx
```diff
@@ -0,0 +1,69 @@
...
+  const isOnlyCheckpoint = totalCheckpoints === 1;
+  return (
+    <div
+      id="checkpoint"
+      className="checkpoint-popover p-4 bg-light-300"
+      aria-labelledby="checkpoint-title"
+      role="dialog"
+      style={{ display: hideCheckpoint ? 'none' : 'block' }}
+    >
+      {/* This text is not translated due to Paragon's lack of i18n support */}
```
MatthewPiatetsky, line 27: curious if a ticket of some sort exists on a paragon roadmap? (does paragon have a board?)

src/tour/Tour.jsx
```diff
@@ -0,0 +1,189 @@
...
+
+  useEffect(() => {
+    if (checkpointData && isEnabled) {
+      const targetElement = document.querySelector(checkpointData.target);
+      const checkpoint = document.querySelector('#checkpoint');
+      if (!targetElement) {
+        setHideCheckpoint(true);
+      } else {
+        setHideCheckpoint(false);
+        createPopper(targetElement, checkpoint, {
```
MatthewPiatetsky, line 44: what is this doing?

src/tour/Tour.jsx
```diff
@@ -0,0 +1,189 @@
...
+                padding: 5,
+              },
+            },
+          ],
+        });
+
+        let targetOffset = targetElement.getBoundingClientRect().top;
+        if (checkpointData.placement && checkpointData.placement.includes('top')) {
+          if (targetOffset < 0) {
+            targetOffset *= -1;
```
MatthewPiatetsky, line 65: what is this doing? can it switch multiple times from negative to positive? a comment might be nice here

MatthewPiatetsky (approved):
love the readme
at least my previous feedback has been addressed, not sure about others' feedback

src/tour/Checkpoint.jsx
ciduarte: A ticket does not exist yet because the component hasn't been formally proposed. We're asking Kavya to create a design proposal to bring to Paragon Working Group, then we'll be able to create a ticket in Paragon's Jira for development.

src/tour/Tour.jsx
ciduarte: From my understanding, this is taking the rendered Checkpoint DOM element and placing it (i.e. performing a translation to) wherever the target element is within the DOM. There's some additional behaviors that the Popper library provides in using this method (flipping the tooltip, placement orientation, preventing overflow, etc.).

https://popper.js.org/docs/v2/constructors/
https://popper.js.org/docs/v2/modifiers/

src/tour/Checkpoint.jsx
Dillon-Dumesnil: Chill vibes. Thank you!

src/tour/Tour.jsx
Dillon-Dumesnil: Thank you for the reminder!

src/tour/README.md
```diff
@@ -0,0 +1,86 @@
+Tour
+=========================
+
+Basic Usage
+------------
+
+A `Tour` takes a list of tour objects. `Tour` will only support one enabled tour at a time. If multiple
+tours are enabled, Tour will only render the first in the `tours` list.
```
Dillon-Dumesnil: Nit: first -> first enabled

src/tour/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+`Checkpoints` are rendered in the order they're listed in the checkpoint array.
+The checkpoint objects themselves have additional props that can override the props defined in a `Tour`.
```
Dillon-Dumesnil, line 11: Would you mind adding an example of an override in your below example in one of your checkpoints?

src/tour/README.md
```diff
@@ -0,0 +1,86 @@
...
+    <Container>
+  </>
+);
+```
+
+Tour Props API
+------------
+tours `array` comprised of objects with the following values:
+- advanceButtonText `string` The text displayed on all buttons used to advance the tour.
+- checkpoints `array` An array comprised of checkpoint objects:
```
Dillon-Dumesnil: Nit: objects: -> objects supporting the following props:

src/tour/README.md
```diff
@@ -0,0 +1,86 @@
...
+        <div id="checkpoint-3">...</div>
+      </Row>
+    <Container>
+  </>
+);
+```
+
+Tour Props API
+------------
+tours `array` comprised of objects with the following values:
```
Dillon-Dumesnil: Nit: Call out which props are required (or optional) in your documentation here. 
Edit: I actually see that in some spots, but not everywhere. Make sure you're consistent with this. Also, related to my below point, try and have some visual separator to the "required" is easier to spot. Bold was a nice touch, but it still just looks like a line of text. Perhaps something like 

advanceButtonText (string) **required**: The text displayed on all buttons used to advance the tour.
advanceButtonText (string - **required**): The text displayed on all buttons used to advance the tour.

Dillon-Dumesnil: Nit: Place the types in brackets or parentheses. It's a little hard to read when it's all straight text. Or perhaps add in colons instead. Or both if you choose. Examples:

advanceButtonText (`string`) The text displayed on all buttons used to advance the tour.
advanceButtonText `string`: The text displayed on all buttons used to advance the tour.
advanceButtonText [`string`]: The text displayed on all buttons used to advance the tour.

Dillon-Dumesnil: Also ensure what's here matches with what are listed in your propTypes. I see `target` listed for checkpoint, but don't see that existing inside of the component
edit: Oh I see, it's not a prop used in the JSX, but it's still a prop type used. Okay, cool cool cool

src/tour/Tour.jsx
Dillon-Dumesnil: Ah interesting. Okay, I think it was just difficult for me to visualize, but I get what you're saying.

src/tour/Tour.jsx
Dillon-Dumesnil: Ah thanks!

src/tour/tests/Tour.test.jsx
```diff
@@ -0,0 +1,328 @@
...
+import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
+import { fireEvent, render, screen } from '@testing-library/react';
+import * as popper from '@popperjs/core';
+
+import Tour from '../Tour';
+
+// This can be removed once the component is ported over to Paragon
+Enzyme.configure({ adapter: new Adapter() });
+
+describe('Tour', () => {
```
Dillon-Dumesnil, line 15: Nit: One more test for multiple tours and correctly grabbing the one enabled one/the first enabled one.

Dillon-Dumesnil (approved):
Don't think I have any blocking feedback. Just some nits and personal styling things. Do with them as you please.
Great job on this Carla!

src/tour/README.md
ciduarte: Thanks for the notes here

I was trying not to over style this document because Paragon does a lot of automated doc styling for free. Was mostly trying to mirror what those docs look like without going overboard. But I can update this to be a bit more readable

About the `required` consistency note - there are only three required props. The convention here is only to write required and to not write anything if the prop is optional.

Dillon-Dumesnil: Ah yeah, all nits so feel free to make your own best judgements with all these. Don't want you to do unnecessary work.

Ah okay, perfect. Hard for me referencing back and forth to keep it all in my head, but glad you got it all. And yeah, I meant either required or optional, not both. Totally cool with that. Thanks!

PR #716, opened 2021-10-29
fix: skip inaccessible learning sequence data

mikix:
When normalizing learning sequences, skip inaccessible sequences
and also skip sections with only inaccessible sequences.

This both imitates the legacy course block behavior and also
avoids a failure when merging course block data with LS data
when they disagree about which sequences exist.

### DIFF
--- a/src/courseware/data/api.js
+++ b/src/courseware/data/api.js
@@ -98,12 +98,40 @@ export function normalizeLearningSequencesData(learningSequencesData) {
     sequences: {},
   };
 
+  // Sequences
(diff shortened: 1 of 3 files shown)


MatthewPiatetsky (approved):
test seems like a good idea

mikix:
OK - test added using the new pact stuff. For both simple response and one with inaccessible sequences.

mikix:
For posterity: after doing some digging, I believe this was introduced to fix https://openedx.atlassian.net/browse/AA-1085

And this PR seems to have introduced a bug by skipping too many sequences. See https://github.com/openedx/frontend-app-learning/pull/885 for the follow-on fix.

PR #718, opened 2021-11-01
chore: update special exams lib

alangsto:
Update frontend-special-exams-lib to latest release version, which removes the use of the `ready_to_resume` status.


schenedx (approved):
👍

PR #723, opened 2021-11-02
refactor: [REV-2262] create generic bullets to reuse in all upsell messaging

pshiu:
(Note: This is a reopen of https://github.com/edx/frontend-app-learning/pull/689 that was reverted by https://github.com/edx/frontend-app-learning/pull/696.

The only change since https://github.com/edx/frontend-app-learning/pull/689 is a merge from master and [this commit](https://github.com/edx/frontend-app-learning/commit/e98f7ab277e6585f15016b78559cff87193b1166) to fix the issue that caused the revert.)

# Description
This PR implements [REV-2262](https://openedx.atlassian.net/browse/REV-2262) to refactor text bullets in frontend-app-learning that are duplicated between components `<UpgradeNotification>` and `<GatedContent>` into new React components.

# Purpose
* Duplicated strings cause a headache for our translators, who have to translate the same string multiple times. 
* In preparation for possible future uses of these bullets elsewhere in our application.

# Implementation
This PR creates 4 new components called:

* `<VerifiedCertBullet>`
* `<UnlockGradedBullet>`
* `<FullAccessBullet>`
* `<SupportMissionBullet>`

taking code out of `<UpgradeNotification>` and `<GatedContent>` for the text bullets:

* "Earn a verified certificate of completion to showcase on your resumé"
* "Unlock your access to all course activities, including graded assignments"
* "Full access to course content and materials, even after the course ends"
* "Support our mission at edX"

# Testing
Since this is a refactor, the goal is not to significantly change the look and feel of anything. To prove this, this is a screenshot of after (in devstack, on top half of screenshot) and before (in production, on bottom half of screenshot):

For `<GatedContent>`:

course-lock-paywall-comparison

For `<UpgradeNotification>`:

upgrade-notification-comparison

Also tested for non-FBE and in mobile. Screenshots available on request.

## Examples in Production
* FBE: https://learning.edx.org/course/course-v1:MichiganX+py4e101x+3T2020/home
* Non-FBE: https://learning.edx.org/course/course-v1:JaverianaX+TDPx+2T2021/home

# Notes

This PR modifies the names of many translation strings. While the overall number of translation strings has dropped from 17 to 8, the new strings do not use old names and will need to be re-translated. This was discussed and accepted as a consequence of this ticket.

This PR performs some work to remove language regarding edX being a non-profit in [REV-2388](https://openedx.atlassian.net/browse/REV-2388). Spreadsheet for REV-2388 updated.

# Attachments

* [Diff of translation strings](https://github.com/edx/frontend-app-learning/files/7368499/transifex_input.json.diff.txt)
* [Implementation notes](https://openedx.atlassian.net/wiki/spaces/~931919476/pages/3186098448/REV-2262)


dianekaplan:
This looks like a clean solution for making sure the special new bullets get the correct styling- love it! Just wanted to confirm: were you able to find a way to reproduce the issue locally to confirm that this change applies the style correctly where it didn't before? (Should we be checking that each instance of the new bullets has both the "upgrade-notification" and "fa-ul" classes?)

pshiu:
> Just wanted to confirm: were you able to find a way to reproduce the issue locally to confirm that this change applies the style correctly where it didn't before?

Yup! Good question. I have a hack to make my devstack act more like production/stage, and I see both red (with the hack, I see the same as production) and green (with the hack and the fix, I see the desired result on my local).

> Should we be checking that each instance of the new bullets has both the "upgrade-notification" and "fa-ul" classes?

I've checked both the upgrade notification and the gated content bullets in FBE and non-FBE modes, and they look correct to me. However, I'll be performing more in-depth testing on stage like last time ([notes](https://github.com/pshiu/edx-pensieve/tree/master/REV-2262_UpsellBulletComponent) [go to section "On Stage"], [screenshots](https://github.com/pshiu/edx-pensieve/tree/master/REV-2262_UpsellBulletComponent/screenshots)).

src/generic/upsell-bullets/UpsellBullets.jsx
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// Must be child of a <ul className="fa-ul">
+export function UnlockGradedBullet() {
+  const gradedAssignmentsInBoldText = (
+    <span className="font-weight-bold">
+      <FormattedMessage
+        id="learning.generic.upsell.unlockGradedBullet.gradedAssignments"
+        defaultMessage="graded assignments"
+        description="Bolded words 'graded assignment', which is the name of unlocked feature with upgrade."
```
inventhouse: Typo 'graded assignment*s*'; more context may be helpful here to explain what "graded assignments" are

src/generic/upsell-bullets/UpsellBullets.jsx
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// Must be child of a <ul className="fa-ul">
+export function FullAccessBullet() {
+  const fullAccessInBoldText = (
+    <span className="font-weight-bold">
+      <FormattedMessage
+        id="learning.generic.upsell.fullAccessBullet.fullAccess"
+        defaultMessage="Full access"
+        description="Bolded phrase 'Full access'."
```
inventhouse: More context about where the phrase is used would help translators

src/generic/upsell-bullets/UpsellBullets.jsx
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// Must be child of a <ul className="fa-ul">
+export function SupportMissionBullet() {
+  const missionInBoldText = (
+    <span className="font-weight-bold">
+      <FormattedMessage
+        id="learning.generic.upsell.supportMissionBullet.mission"
+        defaultMessage="mission"
+        description="Bolded word 'mission'."
```
inventhouse: A little more about the context of the word "mission" would be helpful here (to say that it supports our goal rather than some other meaning, such as buildings that house or were built by missionaries)

src/generic/upsell-bullets/UpsellBullets.jsx
```diff
@@ -0,0 +1,107 @@
...
+      />
+    </span>
+  );
+  return (
+    <li className="upsell-bullet">
+      <CheckmarkBullet />
+      <FormattedMessage
+        id="learning.generic.upsell.supportMissionBullet"
+        defaultMessage="Support our {missionInBoldText} at edX"
+        description="Bullet encouraging user to support edX."
```
inventhouse: I would add "Bullet encouraging user to support edX's goals."

inventhouse (approved):
A few places where more context would help translators, otherwise looks good.

julianajlk:
Ben leaves a good point about translations descriptions, which Phil and I have chatted about before on the original PR - the way these bullet points were previously done prior to this ticket is that it separates the translation from any styled text to the rest of the phrase, which is not helpful for translators (this worked out ok since I am the translator but would not be ideal for the future). 

Will discuss at standup!

dianekaplan (approved):

src/generic/upsell-bullets/UpsellBullets.jsx
pshiu:
```suggestion
        description="Bolded words 'graded assignments', which are the bolded portion of a bullet point highlighting that course content is unlocked when purchasing an upgrade. Graded assignments are any course content that is graded and are unlocked by upgrading to verified certificates."
```

src/generic/upsell-bullets/UpsellBullets.jsx
pshiu:
```suggestion
        description="Bolded phrase 'Full access', which is the bolded portion of a bullet point highlighting that access to course content will not have time limits."
```

src/generic/upsell-bullets/UpsellBullets.jsx
pshiu: Good point. Apparently, TIL it can also mean an embassy too. We wouldn't want to cause a diplomatic incident.

```suggestion
        description="Bolded word 'mission', which is the bolded portion of a bullet point encouraging the learner to support the goals of the website."
```

src/generic/upsell-bullets/UpsellBullets.jsx
pshiu:
```suggestion
        description="Bullet encouraging user to support edX's goals."
```

julianajlk:
> Ben leaves a good point about translations descriptions, which Phil and I have chatted about before on the original PR - the way these bullet points were previously done prior to this ticket is that it separates the translation from any styled text to the rest of the phrase, which is not helpful for translators (this worked out ok since I am the translator but would not be ideal for the future).
> 
> Will discuss at standup!

Update: for now we'll add a detailed description, and [REV-2452](https://openedx.atlassian.net/browse/REV-2452) has been created for future improvements on i18n, since this would be a company wide fedX/Paragon effort.

julianajlk (approved):

PR #727, opened 2021-11-02
fix: [AA-1087] defer goals widget for proctoring

cdeery:
- pause showing goals widget while proctor info panel awaits callback
AA-1087 Proctor Info Panel

### DIFF
--- a/src/course-home/data/api.js
+++ b/src/course-home/data/api.js
@@ -345,6 +345,7 @@ export async function getOutlineTabData(courseId) {
   const datesWidget = camelCaseObject(data.dates_widget);
   const enrollAlert = camelCaseObject(data.enroll_alert);
   const enrollmentMode = data.enrollment_mode;
+  const enableProctoredExams = data.enable_proctored_exams;
   const handoutsHtml = data.handouts_html;
   const hasScheduledContent = data.has_scheduled_content;
   const hasEnded = data.has_ended;
@@ -366,6 +367,7 @@ export async function getOutlineTabData(courseId) {
     datesWidget,
     enrollAlert,
     enrollmentMode,
+    enableProctoredExams,
     handoutsHtml,
     hasScheduledContent,
     hasEnded,
--- a/src/course-home/data/api.js
+++ b/src/course-home/data/api.js
@@ -343,9 +343,9 @@ export async function getOutlineTabData(courseId) {
   const courseTools = camelCaseObject(data.course_tools);
   const datesBannerInfo = camelCaseObject(data.dates_banner_info);
   const datesWidget = camelCaseObject(data.dates_widget);
+  const enableProctoredExams = data.enable_proctored_exams;
   const enrollAlert = camelCaseObject(data.enroll_alert);
   const enrollmentMode = data.enrollment_mode;
-  const enableProctoredExams = data.enable_proctored_exams;
   const handoutsHtml = data.handouts_html;
   const hasScheduledContent = data.has_scheduled_content;
   const hasEnded = data.has_ended;
--- a/src/course-home/outline-tab/OutlineTab.jsx
+++ b/src/course-home/outline-tab/OutlineTab.jsx
@@ -49,7 +49,6 @@ function OutlineTab({ intl }) {
 
   const {
     accessExpiration,
-    enableProctoredExams,
     courseBlocks: {
       courses,
       sections,
@@ -63,10 +62,11 @@ function OutlineTab({ intl }) {
     datesWidget: {
       courseDateBlocks,
     },
+    enableProctoredExams,
+    offer,
     resumeCourse: {
       url: resumeCourseUrl,
     },
-    offer,
     timeOffsetMillis,
     verifiedMode,
   } = useModel('outline', courseId);
@@ -74,6 +74,8 @@ function OutlineTab({ intl }) {
   const [deprecatedCourseGoalToDisplay, setDeprecatedCourseGoalToDisplay] = useState(selectedGoal);
   const [goalToastHeader, setGoalToastHeader] = useState('');
   const [expandAll, setExpandAll] = useState(false);
+  // Defer showing the goal widget until the ProctoringInfoPanel is either shown or determined as not showing
+  // to avoid components bouncing around too much as screen is displayed
   const [proctorPanelResolved, setProctorPanelResolved] = useState(!enableProctoredExams);
 
   const eventProperties = {
--- a/src/course-home/outline-tab/OutlineTab.jsx
+++ b/src/course-home/outline-tab/OutlineTab.jsx
@@ -49,6 +49,7 @@ function OutlineTab({ intl }) {
 
   const {
     accessExpiration,
+    enableProctoredExams,
     courseBlocks: {
       courses,
       sections,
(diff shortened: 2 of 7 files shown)


src/course-home/outline-tab/widgets/ProctoringInfoPanel.jsx
```diff
@@ -82,6 +84,7 @@ function ProctoringInfoPanel({ courseId, username, intl }) {
             if (Object.keys(response).length > 0) {
               setShowInfoPanel(true);
             }
+            isResolved();
```
Dillon-Dumesnil: My only concern with this PR is will this code path always be hit? What if `getProctoringInfoData` errors? I don't see a `.catch` method, but it seems like in any unsuccessful call, `isResolved` will never resolve. Same with if there isn't a response for some reason. 
Is there a way we can ensure we will always resolve this since it will block the goals widget otherwise.

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -49,6 +49,7 @@ function OutlineTab({ intl }) {
 
   const {
     accessExpiration,
+    enableProctoredExams,
```
Dillon-Dumesnil: Nit: keep these alphabetized. And if you're updating, maybe also update offer to be higher than resumeCourse

src/course-home/data/api.js
```diff
@@ -366,6 +367,7 @@ export async function getOutlineTabData(courseId) {
     datesWidget,
     enrollAlert,
     enrollmentMode,
+    enableProctoredExams,
```
Dillon-Dumesnil, line 370: Nit: keep alphabetized

Dillon-Dumesnil (commented):
Couple questions to ensure we resolve the proctoring request.

As a last point too, should we have a test around this? Ensuring goals widget doesn't display until the correct state change?

src/course-home/outline-tab/widgets/ProctoringInfoPanel.jsx
MatthewPiatetsky: agree, i think you might be able to use a .finally( block just for this part as this approach wouldn't show the goals panel if the proctoring request fails

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -73,6 +74,7 @@ function OutlineTab({ intl }) {
   const [deprecatedCourseGoalToDisplay, setDeprecatedCourseGoalToDisplay] = useState(selectedGoal);
   const [goalToastHeader, setGoalToastHeader] = useState('');
   const [expandAll, setExpandAll] = useState(false);
+  const [proctorPanelResolved, setProctorPanelResolved] = useState(!enableProctoredExams);
```
MatthewPiatetsky, line 79: i think a comment would be worthwhile here to enable what we're doing and why

MatthewPiatetsky (commented):
agree that a test could be a good idea since this is very edge casey

cdeery:
I am a little stumped about how to test this. When I mock in a 404 response from the proctoringInfoURL, the test framework fails.
Cannot read property 'httpErrorStatus' of undefined
TypeError: Cannot read property 'httpErrorStatus' of undefined
    at _callee4$ (/Users/cdeery/src/frontend-app-learning/src/course-home/data/api.js:292:13)
    at tryCatch (/Users/cdeery/src/frontend-app-learning/node_modules/axios-cache-adapter/dist/webpack:/axiosCacheAdapter/node_modules/regenerator-runtime/runtime.js:63:15)
    at Generator.invoke [as _invoke] (/Users/cdeery/src/frontend-app-learning/node_modules/axios-cache-adapter/dist/webpack:/axiosCacheAdapter/node_modules/regenerator-runtime/runtime.js:293:1)
    at Generator.throw (/Users/cdeery/src/frontend-app-learning/node_modules/axios-cache-adapter/dist/webpack:/axiosCacheAdapter/node_modules/regenerator-runtime/runtime.js:118:1)
    at asyncGeneratorStep (/Users/cdeery/src/frontend-app-learning/src/course-home/data/api.js:29:103)
    at _throw (/Users/cdeery/src/frontend-app-learning/src/course-home/data/api.js:31:291)

MatthewPiatetsky:
maybe fedx will have suggestions on how to test?
i'm not very familiar with testing in mfes

src/course-home/outline-tab/widgets/ProctoringInfoPanel.jsx
cdeery: Added some tests, it turned out to be tricky because the proctoring info API returns a 404 if the course does not have proctoring info. I modified ProctoringInfoPanel to eat the error. This probably needs a better long term solution, I saw that the API threw 170K 404s into Splunk in 24 hours.

cdeery:
Carla helped me get the testing resolved.

src/course-home/outline-tab/OutlineTab.jsx
Dillon-Dumesnil: Nice comment!

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -529,6 +529,47 @@ describe('Outline Tab', () => {
...
+    const onboardingReleaseDate = new Date();
+    onboardingReleaseDate.setDate(new Date().getDate() - 7);
+
+    beforeEach(async () => {
+      setTabData({
+        course_goals: {
+          weekly_learning_goal_enabled: true,
+        },
+      });
+    });
```
Dillon-Dumesnil: Nit: So there are only 3 tests in this file and you change the value of course_goals in two of them already. I'd say just drop this beforeEach call and add the setTabData into the test itself.

Also, I just realized all of these tests have `weekly_learning_goal_enabled` set to `true`. Do we have any tests where it is false? If not, would you mind adding one? And looking at this now, I realize all of these tests are related to the proctored exam stuff (which fair), but do we have tests for the actual component itself? Do those exist and are just in a different file? Sorry for dumping all of these testing questions on ya last minute.

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -1172,6 +1213,7 @@ describe('Outline Tab', () => {
 
     it('does not appear for 404', async () => {
       axiosMock.onGet(proctoringInfoUrl).reply(404);
+      await fetchAndRender();
```
Dillon-Dumesnil, line 1211: 👏

Dillon-Dumesnil (commented):
Some questions about tests, but otherwise this PR looks good to me! Great improvements being made here. Nice job!

src/course-home/outline-tab/OutlineTab.test.jsx
cdeery: I can drop the beforeEach. It is an artifact of experimentation as I was trying to figure out tests. Good call.
There are tests for weekly learning goal in this file from lines 424 to 531, directly above these tests. You may have misread the diff - there are many more than 3 tests in the file.
The test at line 424 tests that the goal panel is not shown if the the flag is false.

cdeery: You must mean there are only 3 tests in this block. That's fair

Dillon-Dumesnil: Yes, sorry, I meant this block. 
Ah I see the other tests now. Okay, new request because y'boy is picky. So the describe here is "weekly learning goal is displayed". I think if you want to keep that, you should move the other display based tests into this describe. If not, I think the describe should maybe be updated to call out that this section of tests about weekly learning goals is specifically how it interacts with the ProctoringInfoPanel and does not encompass all display behaviors.

cdeery: Folded the proctoring tests into the other goals widget tests

Dillon-Dumesnil (approved):
LGTM

PR #728, opened 2021-11-03
build: use shared browserslist config and remove is-es5 check

adamstankiewicz:
* Removes custom `browserslist` configuration in favor of using a [shared configuration](https://github.com/edx/browserslist-config).
* Removes `is-es5` check in CI since ES5 was only needed for IE 11 support which is dropped. The supported browsers defined by the shared configuration all [natively support ES6](https://caniuse.com/?search=es6).

This change reduces the resultant asset bundle size.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,6 @@ validate:
 	npm run lint -- --max-warnings 0
 	npm run test
 	npm run build
-	npm run is-es5
 
 .PHONY: validate.ci
 validate.ci:
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,6 +41,7 @@
         "util": "0.12.4"
       },
       "devDependencies": {
+        "@edx/browserslist-config": "1.0.2",
         "@edx/frontend-build": "9.1.4",
         "@edx/reactifex": "1.1.0",
         "@pact-foundation/pact": "9.17.3",
@@ -49,7 +50,6 @@
         "@testing-library/user-event": "13.5.0",
         "axios-mock-adapter": "1.20.0",
         "codecov": "3.8.3",
-        "es-check": "6.2.1",
         "husky": "7.0.4",
         "jest": "27.5.1",
         "rosie": "2.1.0"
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -60,7 +60,6 @@ validate:
 	npm run lint -- --max-warnings 0
 	npm run test
 	npm run build
-	npm run is-es5
```
mikix, line 61: Is there not value in some replacement check like `is-es6`? Or is the idea that all our supported browsers all update regularly enough that we don't stress about that anymore, and ECMA updates slowly enough that the last release (ES9 in 2018) can be assumed - and that any future update will be supported by last 2 browser updates as well. Which is totally reasonable.

mikix (approved):

mikix:
I had asked Adam about this PR in the past, and here's the reason we haven't landed it yet: we use fancy new JS features in the learning MFE that our current browserslist config polyfills for us. But some of these fancy features are newer than node 12. So we can't really land this until we get onto node 16 or those fancy new features would break on us without the polyfill. (correct me if I'm wrong on any of this Adam)

I believe that effort (node16-ification) is in https://github.com/openedx/frontend-app-learning/pull/771

mikix:
@adamstankiewicz actually, can this PR just sub in `node 12` into the browserslist config to avoid that problem? That way we get the benefit of no IE11 support, and can just drop that line (or really, update that line) when we switch to node16. (It's possible `current node` would dynamically use whatever we are running at the moment - not entirely clear from reading their one line in the doc about that)

Seems wise to list our node version in the browserslist config anyway.

mikix:
The Learning MFE now only supports and uses node 16. I rebased this on master for ya - you cool with me pressing merge @adamstankiewicz?

mikix:
Oh. Once I fix that react-intl error. I think we now need to include it as a toplevel dependency.

adamstankiewicz:
> The Learning MFE now only supports and uses node 16. I rebased this on master for ya - you cool with me pressing merge @adamstankiewicz?

@mikix The only thing that's still a bit of an open question to me is whether we want to keep any sort of `es-check` here. That command does actually support checking against ES6, so I'm tempted to say we should keep it and make it validate `is-es6` instead.

mikix:
I'd be down with `is-es6`. I'll fix this dumb react-intl thing, because (A) I caused it and (B) it's related to another fight I'm having with our dependencies now that we're on npm8.

So while here, I can bring back is-es6

src/setupTest.js
```diff
@@ -5,14 +5,13 @@ import '@testing-library/jest-dom/extend-expect';
 import './courseware/data/__factories__';
 import './course-home/data/__factories__';
 import { getConfig, mergeConfig } from '@edx/frontend-platform';
-import { configure as configureI18n } from '@edx/frontend-platform/i18n';
+import { configure as configureI18n, IntlProvider } from '@edx/frontend-platform/i18n';
```
mikix, line 8: This was because we don't directly depend on react-intl, so some linting was complaining about the below direct use of it. Instead, grab the pass-through component from frontend-platform. Dunno how it got through to this point.

mikix:
After talking in slack, we figured the es check wasn't worth it. I rebased on master and these are basically the original approved changes again. Will merge shortly, to make it for nutmeg.

PR #742, opened 2021-11-08
feat: Update notification feature to be course specific

julianajlk:
[REV-2360](https://openedx.atlassian.net/browse/REV-2360).

When the notification feature was updated to have the "red dot" functionality in https://github.com/edx/frontend-app-learning/pull/569 it was not course specific, meaning that while it persisted whether the learner had seen the notification or not for a course, when in another course that information was applied as if it was the same course.

This PR makes the "red dot" functionality be course specific and persist by a course, no matter how many courses a learner is enrolled in.

**Below is a screenshot of 2 different courses (local):**
Screen Shot 2021-11-08 at 3 55 38 PM

Screen Shot 2021-11-08 at 3 53 44 PM

**Note on Codecov:**
For the `codecov/patch` coverage, it's below the target since in the original implementation it was below (it was at 53%,  it's now at 70%). 
It's understood that as long as the `codecov/project` coverage is within target we're good to go. 

**How to test:**
- Checkout my branch `julianajlk/REV-2360/vp-notification-red-dot`
- Create a new course other than the Demo course with verified mode available, via [Ecommerce Course Admin](http://localhost:18130/courses/).
- Make sure you enroll in one course first, go to courseware, check the "red dot" notification goes away after in 3 seconds when the `NotificationTray` is open (it should load open). 
- Now enroll in the second course you're created, and you should see the "red dot" notification for this course.

### DIFF
--- a/src/courseware/course/Course.jsx
+++ b/src/courseware/course/Course.jsx
@@ -74,7 +74,7 @@ function Course({
 
   const onNotificationSeen = () => {
     setNotificationStatus('inactive');
-    setLocalStorage('notificationStatus', 'inactive');
+    setLocalStorage(`notificationStatus.${courseId}`, 'inactive');
   };
 
   /** [MM-P2P] Experiment */
--- a/src/courseware/course/Course.jsx
+++ b/src/courseware/course/Course.jsx
@@ -61,16 +61,16 @@ function Course({
     if (notificationTrayVisible) { setNotificationTray(false); } else { setNotificationTray(true); }
   };
 
-  if (!getLocalStorage('notificationStatus')) {
-    setLocalStorage('notificationStatus', 'active'); // Show red dot on notificationTrigger until seen
+  if (!getLocalStorage(`notificationStatus.${courseId}`)) {
+    setLocalStorage(`notificationStatus.${courseId}`, 'active'); // Show red dot on notificationTrigger until seen
   }
 
-  if (!getLocalStorage('upgradeNotificationCurrentState')) {
-    setLocalStorage('upgradeNotificationCurrentState', 'initialize');
+  if (!getLocalStorage(`upgradeNotificationCurrentState.${courseId}`)) {
+    setLocalStorage(`upgradeNotificationCurrentState.${courseId}`, 'initialize');
   }
 
-  const [notificationStatus, setNotificationStatus] = useState(getLocalStorage('notificationStatus'));
-  const [upgradeNotificationCurrentState, setupgradeNotificationCurrentState] = useState(getLocalStorage('upgradeNotificationCurrentState'));
+  const [notificationStatus, setNotificationStatus] = useState(getLocalStorage(`notificationStatus.${courseId}`));
+  const [upgradeNotificationCurrentState, setupgradeNotificationCurrentState] = useState(getLocalStorage(`upgradeNotificationCurrentState.${courseId}`));
 
   const onNotificationSeen = () => {
     setNotificationStatus('inactive');
@@ -98,6 +98,7 @@ function Course({
 
         { shouldDisplayNotificationTrigger ? (
           <NotificationTrigger
+            courseId={courseId}
             toggleNotificationTray={toggleNotificationTray}
             isNotificationTrayVisible={isNotificationTrayVisible}
             notificationStatus={notificationStatus}
--- a/src/courseware/course/NotificationTrigger.test.jsx
+++ b/src/courseware/course/NotificationTrigger.test.jsx
@@ -4,10 +4,11 @@ import {
   render, initializeTestStore, screen, fireEvent,
 } from '../../setupTest';
 import NotificationTrigger from './NotificationTrigger';
-import { getLocalStorage } from '../../data/localStorage';
 
 describe('Notification Trigger', () => {
   let mockData;
+  let getItemSpy;
+  let setItemSpy;
   // let mockDataSameState;
   // let mockDataDifferentState;
   const courseMetadata = Factory.build('courseMetadata');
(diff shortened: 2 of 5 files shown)


src/courseware/course/NotificationTrigger.test.jsx
```diff
@@ -4,76 +4,126 @@ import {
   render, initializeTestStore, screen, fireEvent,
 } from '../../setupTest';
 import NotificationTrigger from './NotificationTrigger';
-import { getLocalStorage } from '../../data/localStorage';
 
 describe('Notification Trigger', () => {
   let mockData;
-  // let mockDataSameState;
-  // let mockDataDifferentState;
```
julianajlk, line 12: @dianekaplan I'm assuming these are not needed and were just left unintentionally from your PR? I removed them, but let me know if not!

inventhouse (approved):

src/courseware/course/Course.jsx
```diff
@@ -61,16 +61,16 @@ function Course({
...
+  if (!getLocalStorage(`upgradeNotificationCurrentState.${courseId}`)) {
+    setLocalStorage(`upgradeNotificationCurrentState.${courseId}`, 'initialize');
   }
 
-  const [notificationStatus, setNotificationStatus] = useState(getLocalStorage('notificationStatus'));
-  const [upgradeNotificationCurrentState, setupgradeNotificationCurrentState] = useState(getLocalStorage('upgradeNotificationCurrentState'));
+  const [notificationStatus, setNotificationStatus] = useState(getLocalStorage(`notificationStatus.${courseId}`));
+  const [upgradeNotificationCurrentState, setupgradeNotificationCurrentState] = useState(getLocalStorage(`upgradeNotificationCurrentState.${courseId}`));
 
   const onNotificationSeen = () => {
```
pshiu, line 75: On line 77 below, should:

    setLocalStage('notificationStatus', 'inactive');

instead be:

    setLocalStage(`notificationStatus.${courseId}`, 'inactive');

?

pshiu:
While checking for all usages of components with a new `courseId` prop:

```
% git grep -l -e "<NotificationTrigger" -e "<ExpirationCountdown" -e "<AccessExpirationDateBanner"
src/courseware/course/Course.jsx
src/courseware/course/NotificationTrigger.test.jsx
src/courseware/course/sequence/Sequence.jsx
src/generic/upgrade-notification/UpgradeNotification.jsx
```

Oddly, it appears `<Sequence>` uses `<NotificationTrigger>` ([link](https://github.com/edx/frontend-app-learning/blob/2460992f0779a00429ef0b976c027f4fd188a62d/src/courseware/course/sequence/Sequence.jsx#L184-L190)). Not sure why that's the case, but could it use a `courseId` prop?

julianajlk:
> While checking for all usages of components with a new `courseId` prop:
> 
> ```
> % git grep -l -e "<NotificationTrigger" -e "<ExpirationCountdown" -e "<AccessExpirationDateBanner"
> src/courseware/course/Course.jsx
> src/courseware/course/NotificationTrigger.test.jsx
> src/courseware/course/sequence/Sequence.jsx
> src/generic/upgrade-notification/UpgradeNotification.jsx
> ```
> 
> Oddly, it appears `<Sequence>` uses `<NotificationTrigger>` ([link](https://github.com/edx/frontend-app-learning/blob/2460992f0779a00429ef0b976c027f4fd188a62d/src/courseware/course/sequence/Sequence.jsx#L184-L190)). Not sure why that's the case, but could it use a `courseId` prop?

That's correct Phil, thank you for catching that - NotificationTrigger is displayed in the Sequence component when the screensize is less than `small` (576px per [bootstrap's breakpoints](https://getbootstrap.com/docs/5.0/layout/breakpoints/)), so the button that triggers the tray is right after the > button (Next).

julianajlk:
> Oddly, it appears `<Sequence>` uses `<NotificationTrigger>` ([link](https://github.com/edx/frontend-app-learning/blob/2460992f0779a00429ef0b976c027f4fd188a62d/src/courseware/course/sequence/Sequence.jsx#L184-L190)). Not sure why that's the case, but could it use a `courseId` prop?

That's correct Phil, thank you for catching that - NotificationTrigger is displayed in the Sequence component when the screensize is less than `small` (576px per [bootstrap's breakpoints](https://getbootstrap.com/docs/5.0/layout/breakpoints/)), so the button that triggers the tray is right after the > button (Next).

src/courseware/course/Course.jsx
julianajlk: Yes! I had a different way to manually test the "red dot" with a click for testing purposes and forgot to update it here, thanks for catching

pshiu (approved):
Thanks for the great work, Ju!

src/courseware/course/NotificationTrigger.test.jsx
dianekaplan: Yes that's right! You'd given the suggestion to use one mockData and update in each test, and afterward I didn't notice this cruft- thanks!

src/courseware/course/NotificationTrigger.test.jsx
```diff
@@ -4,76 +4,105 @@ import {
...
+    const testData = {
+      ...mockData,
+      toggleNotificationTray,
+    };
+    render(<NotificationTrigger {...testData} />);
+
+    const notificationTrigger = screen.getByRole('button', { name: /Show notification tray/i });
+    expect(notificationTrigger).toBeInTheDocument();
+    fireEvent.click(notificationTrigger);
+    expect(toggleNotificationTray).toHaveBeenCalledTimes(1);
```
dianekaplan, line 45: Shall we also check that after it's been toggled, it no longer shows? 
expect(notificationTrigger).not.toBeInTheDocument();

src/courseware/course/NotificationTrigger.test.jsx
```diff
@@ -4,76 +4,105 @@ import {
...
   });
 
   it('renders notification trigger icon WITHOUT red dot 3 seconds later', async () => {
-    const { container } = render(<NotificationTrigger {...mockData} />);
+    const { container } = render(<NotificationTrigger {...mockData} notificationStatus="active" />);
     expect(container).toBeInTheDocument();
+    expect(screen.getByTestId('notification-dot')).toBeInTheDocument();
     jest.useFakeTimers();
     setTimeout(() => {
+      expect(localStorage.setItem).toHaveBeenCalledTimes(2);
```
dianekaplan, line 62: For this test and the ones below, could you say more about why we're checking that we've called localStorage.setItem during the course of the test? (If I'm understanding correctly, that seems like an implementation detail of how we're getting the test to work, as opposed to the behavior we want to cover in the test). Are you imagining a specific problem case where the existing test line below (notification-dot is not in the document) does pass but where we didn't call localStorage.setItem?

src/courseware/course/NotificationTrigger.test.jsx
julianajlk: The `NotificationTrigger` should still be there after it was clicked (it's the button that open/closes the `NotificationTray`, it doesn't go anywhere), here we're just testing in the onClick of the trigger is happening as expected. I believe in `NotificationTray` tests we test the opening/closing of the tray more thoroughly. 

FYI just moving this test's location, did not add it in this PR.

dianekaplan: Ah thanks for the clarification- that makes sense!

src/courseware/course/NotificationTrigger.test.jsx
julianajlk: That's a good question! I'm checking if set/get localStorage are called/changed accordingly - partly to keep codecov happy, in a way this is also testing that the custom functions in `localStorage.js` are doing what they're supposed to do.

julianajlk: But I'm not actually calling them in the test, just checking if they are being called when `NotificationTrigger` is rendered

dianekaplan: that makes sense- thanks Juliana!

dianekaplan (approved):

PR #750, opened 2021-11-23
feat: engage product tour

ciduarte:
**What I did in this PR:**
1. Cleaned up target elements for Checkpoints (added IDs, restructured some div's to be proper anchors for Popper)
2. Added new components and data models to support the following tours: NewUserCourseHomeTour, ExistingUserCourseHomeTour, AbandonTour, CoursewareTour
3. Added eventing for the above tours
4. Updated some Checkpoint/Tour logic (explained in a comment on this PR)
5. Addressed some UI bugs from UX Review

**Screenshots**

**New user course home tour**
This video shows the full course home tour experience. I'll note that you won't find a distinction between the audit and verified versions of the tour in the code. The tour does not change based on whether the learner is audit or verified because the tour is able to adapt according to which elements have rendered in the DOM.

For example, if the WeeklyLearningGoalCard has not rendered (possibly due to an inactive waffle flag), the tour will skip over the Weekly Learning Goal checkpoint. Similarly, when the UpgradeNotification does not render (i.e. a user is verified), the tour will adapt and skip over the UpgradeNotification Checkpoint.

https://user-images.githubusercontent.com/25124041/143145608-6886237d-ea83-42a4-ac2c-555b07392723.mov

**Abandon tour**

https://user-images.githubusercontent.com/25124041/143145614-25674a04-7942-48f7-855c-ded68aa6bbe1.mov

**Courseware tour**
<div>
coursewareTour1
coursewareTour2
</div>

**Existing user course home tour**
existingUserTour


src/courseware/course/NotificationTray.scss
```diff
@@ -16,6 +16,7 @@
         margin: 0;
         border: none;
         border-radius: 0;
+        z-index: 1100;
```
ciduarte, line 19: Needed to bump this z-index on the mobile view because the Checkpoint within courseware was appearing above it.

src/course-home/outline-tab/widgets/CourseDates.jsx
```diff
@@ -29,21 +29,23 @@ function CourseDates({
...
-            userTimezone={userTimezone}
-            /** [MM-P2P] Experiment */
-            mmp2p={mmp2p}
-          />
-        ))}
-      </ol>
-      <a className="font-weight-bold ml-4 pl-1 small" href={datesTabLink}>
-        {intl.formatMessage(messages.allDates)}
-      </a>
+      <div id="courseHome_dates">
```
ciduarte: Just wrapped this in a div so that I could apply the target ID to the correct spot. The popper library wouldn't anchor to `<section>` elements.

src/tours/newUserCourseHomeTour/LaunchCourseHomeTourButton.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+  const dispatch = useDispatch();
+
+  const handleClick = () => {
+    dispatch(launchCourseHomeTour());
+  };
+
+  return (
+    <>
+      {toursEnabled && (
```
ciduarte: Showing this conditionally based on whether the tour waffle flag is active

src/tour/Tour.jsx
```diff
@@ -9,58 +9,69 @@ function Tour({
...
+  const [isTourEnabled, setIsTourEnabled] = useState(!!tourValue);
+  const [prunedCheckpoints, setPrunedCheckpoints] = useState([]);
+
+  const pruneCheckpoints = (checkpoints) => {
+    const checkpointsWithRenderedTargets = checkpoints.filter((checkpoint) => {
+      const targetElement = document.querySelector(checkpoint.target);
+      return !!targetElement;
+    });
+    setPrunedCheckpoints(checkpointsWithRenderedTargets);
+  };
```
ciduarte: Introduced this function to determine which targetIDs have actually been rendered in the DOM. This allows tours to continue and skip past Checkpoints with bad targetIDs.

Previously, the tour would just end unexpectedly.

src/generic/upgrade-notification/UpgradeNotification.jsx
```diff
@@ -366,22 +366,24 @@ function UpgradeNotification({
...
-        {upsellMessage}
-      </div>
-      <div className="upgrade-notification-button">
-        <UpgradeButton
-          offer={offer}
-          onClick={logClick}
-          verifiedMode={verifiedMode}
-          block
-        />
+      <div id="courseHome_upgradeNotification">
```
ciduarte: Also wrapped this in a div to define a proper anchor, similar to within CourseDates.jsx

src/tours/data/thunks.js
```diff
@@ -0,0 +1,31 @@
...
+  };
+}
+
+export function endCourseHomeTour(username) {
+  return async (dispatch) => {
+    try {
+      await patchCourseHomeTourData(username);
+      dispatch(disableCourseHomeTour(false));
+    } catch (error) {
+      // TODO: how to handle this error
```
ciduarte, line 17: I'm not sure if I should actually do anything here... or just `throw error;`

src/tours/AbandonTour.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
+
+import { OkayButtonFormattedMessage } from './GenericTourFormattedMessages';
+
+const abandonTour = (enabled, onEnd) => ({
```
ciduarte: I had to do some roundabout structural things with the tour objects. I'll try to explain my thought process incase this strikes you as odd.

I wanted a way for the tour objects to be self contained, but to populate the tours prop in `<Tour tours={} />`, the prop expects an array of objects. Okay. So I put each tour object (ex. `const abandonTour`) into its own file. But then, I ran into problems in translating the checkpoint and button strings.

Problem 1: You can't injectIntl into something that isn't a React component. So using our expected `{intl.formatMessage(messsages.foo)}` would not be possible.
Problem 2: You can't return normal objects from a React component, only React elements. So switching these from arrow functions to React function based components would not work either, because the `tours` prop expects objects.

My solution was to use `<FormattedMessage />` directly within the object. I also created shared FormattedMessage components for the button text inside of `GenericTourFormattedMessages.jsx` for use across the tours.

Let me know if you have any questions.

src/course-header/Header.jsx
```diff
@@ -60,6 +62,7 @@ function Header({
 
   return (
     <header className="course-header">
+      <LaunchCourseHomeTourButton srOnly />
```
ciduarte: This functions like a "Skip to main content" link. Just prompts users to launch the tour if they'd like because the "launch tour" button is pretty hidden in the DOM under "Course Tools". Might need to revisit this w/ Jeff Witt to take a second pass at the a11y here, but this is what we agreed on for now

src/tab-page/LoadedTabPage.jsx
```diff
@@ -40,8 +57,80 @@ function LoadedTabPage({
...
+  useEffect(() => {
+    if (metadataModel === 'courseHomeMeta' && showNewUserCourseHomeTour) {
+      setIsNewUserCourseHomeTourEnabled(true);
+    }
+  }, [showNewUserCourseHomeTour]);
+
   return (
     <>
+      <Tour
+        tours={[
```
cdeery: Should this be set up in a const to avoid building the whole structure in the props?

src/tour/Checkpoint.jsx
```diff
@@ -8,12 +10,67 @@ import CheckpointTitle from './CheckpointTitle';
...
+  target,
   title,
   totalCheckpoints,
   ...props
 }) {
+  const [checkpointVisible, setCheckpointVisible] = useState(false);
+  const isMobile = useMediaQuery({ query: '(max-width: 768px)' });
+
+  useEffect(() => {
+    // Translate the Checkpoint to it's target's coordinates
```
cdeery: nit: "its" :)

src/tour/Checkpoint.jsx
```diff
@@ -22,7 +79,7 @@ function Checkpoint({
       className="checkpoint-popover p-4 bg-light-300"
       aria-labelledby="checkpoint-title"
       role="dialog"
-      style={{ display: hideCheckpoint ? 'none' : 'block' }}
+      style={{ opacity: checkpointVisible ? 1 : 0 }}
     >
       {/* This text is not translated due to Paragon's lack of i18n support */}
       <span className="sr-only">Top of step {index + 1}</span>
```
cdeery, line 93: Could the text be passed in as an optional prop?

src/tours/AbandonTour.jsx
cdeery: Some of that explanation should be left as comments in the code to help the next person working on this feature.

src/tours/data/api.js
```diff
@@ -0,0 +1,41 @@
...
+
+export async function getTourData(username) {
+  const url = `${getConfig().LMS_BASE_URL}/api/user_tours/v1/${username}`;
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(url);
+    return camelCaseObject(data);
+  } catch (error) {
+    const { httpErrorStatus } = error && error.customAttributes;
+    if (httpErrorStatus === 401) {
+      // The API will return a 403 if the user is not authenticated
```
cdeery: Should that comment say 401?

src/tours/newUserCourseHomeTour/NewUserCourseHomeTour.jsx
```diff
@@ -0,0 +1,96 @@
...
+    id="tours.tabNavigationCheckpoint.title"
+    defaultMessage="Additional course resources"
+  />,
+};
+
+const upgradeCheckpoint = {
+  body: <FormattedMessage
+    id="tours.upgradeCheckpoint.body"
+    defaultMessage="Work towards a certificate and gain full access to course materials. Upgrade now!"
+  />, // TODO: add upgrade link
```
cdeery: TODO?

cdeery (commented):
Looks pretty good. I have a couple of little comments, and you are still working on the tests, correct?

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -185,7 +185,7 @@ function OutlineTab({ intl }) {
                   </Button>
                 </div>
               </div>
-              <ol className="list-unstyled">
+              <ol id="courseHome_outline" className="list-unstyled">
```
Dillon-Dumesnil: Nit: Should this be `courseHome-outline`? Or even just `courseHomeOutline` or `course-home-outline`? Seeing the `-` in the className right next to it is just wonky to me, but I know HTML doesn't really care

Don't want to write this comment everywhere, so this generally applies to all the `courseHome-{tab}` spots

src/courseware/course/NotificationTray.scss
Dillon-Dumesnil: I know this isn't you, but this is why I have mistrust in CSS. Thing isn't where you want, pop the z-index 1100 on it. Next person comes around, they pop in a 1200. It's madness!

src/courseware/course/sequence/sequence-navigation/SequenceNavigation.jsx
```diff
@@ -84,7 +84,7 @@ function SequenceNavigation({
   const prevArrow = isRtl(getLocale()) ? ChevronRight : ChevronLeft;
 
   return sequenceStatus === LOADED && (
-    <nav className={classNames('sequence-navigation', className)} style={{ width: shouldDisplayNotificationTrigger ? '90%' : null }}>
+    <nav id="courseware_sequenceNavigation" className={classNames('sequence-navigation', className)} style={{ width: shouldDisplayNotificationTrigger ? '90%' : null }}>
```
Dillon-Dumesnil: Nit: same as my comment above about `courseHome-{tab}`

src/tab-page/LoadedTabPage.jsx
```diff
@@ -40,8 +57,80 @@ function LoadedTabPage({
...
+
+  useEffect(() => {
+    if (metadataModel === 'courseHomeMeta' && showExistingUserCourseHomeTour) {
+      setIsExistingUserCourseHomeTourEnabled(true);
+    }
+  }, [showExistingUserCourseHomeTour]);
+
+  useEffect(() => {
+    if (metadataModel === 'courseHomeMeta' && showNewUserCourseHomeModal) {
+      setShowNewUserTourPrompt(true);
```
Dillon-Dumesnil: What is the difference of the modal and the prompt? It was confusing for me reading through this and not having a 1-1 mapping of all of the `useState`s above and the `useEffect`s below.

src/tab-page/LoadedTabPage.jsx
```diff
@@ -40,8 +57,80 @@ function LoadedTabPage({
...
+    }
+  }, [showCoursewareTour]);
+
+  useEffect(() => {
+    if (metadataModel === 'courseHomeMeta' && showExistingUserCourseHomeTour) {
+      setIsExistingUserCourseHomeTourEnabled(true);
+    }
+  }, [showExistingUserCourseHomeTour]);
+
+  useEffect(() => {
```
Dillon-Dumesnil: Can this entire useEffect just be dropped? It looks like it only exists because `showNewUserCourseHomeModal` might be true even when the metadataModel is courseHomeMeta, but you could just put that check below in the `isOpen` i.e. `isOpen={showNewUserCourseHomeModal && metadataModel === 'courseHomeMeta'}`. Then you can drop the whole thing about the "prompt"

src/tab-page/LoadedTabPage.jsx
```diff
@@ -72,6 +161,18 @@ function LoadedTabPage({
           }}
         />
         <CourseTabsNavigation tabs={tabs} className="mb-3" activeTabSlug={activeTabSlug} />
+        <NewUserCourseHomeTourModal
+          isOpen={showNewUserTourPrompt}
+          onDismiss={() => {
+            setShowNewUserTourPrompt(false);
```
Dillon-Dumesnil: If you end up keeping the tourPrompt stuff, does this setting it to false need to exist in both the onStartTour and the onDismiss? Er wait, does this `tourPrompt`stuff exist for the sole purpose of being able to be set to false and not needing to rely on the showModal? since you might want to keep that value around?

src/tour/Checkpoint.jsx
```diff
@@ -22,7 +79,7 @@ function Checkpoint({
       className="checkpoint-popover p-4 bg-light-300"
       aria-labelledby="checkpoint-title"
       role="dialog"
-      style={{ display: hideCheckpoint ? 'none' : 'block' }}
+      style={{ opacity: checkpointVisible ? 1 : 0 }}
```
Dillon-Dumesnil: For my own learning, why this change?

src/tour/Checkpoint.scss
```diff
@@ -43,20 +42,48 @@
 }
 
 .checkpoint-popover[data-popper-placement^='top'] > #checkpoint-arrow {
-  bottom: -6px;
+  left: -15px !important;
```
Dillon-Dumesnil: Why is the `!important` necessary here and below?

src/tour/Tour.jsx
Dillon-Dumesnil: It might be worth adding this more human readable description as a comment in the code/docstring to this function

src/tour/tests/Tour.test.jsx
```diff
@@ -78,6 +79,14 @@ describe('Tour', () => {
   };
 
   describe('multiple enabled tours', () => {
+    beforeEach(() => {
+      popperMock.mockImplementation(jest.fn());
+    });
+
+    afterEach(() => {
+      popperMock.mockReset();
+    });
```
Dillon-Dumesnil: Hard to see everything in github, but is it possible to move these lines outside into the highest describe so you don't need to repeat this for each test suite?

src/tours/data/api.js
```diff
@@ -0,0 +1,41 @@
...
+    show_courseware_tour: false,
+  });
+}
+
+export async function patchCourseHomeTourData(username) {
+  const url = `${getConfig().LMS_BASE_URL}/api/user_tours/v1/${username}`;
+  return getAuthenticatedHttpClient().patch(url, {
+    course_home_tour_status: 'no-tour',
+  });
+}
```
Dillon-Dumesnil: Nit: combine these and just pass in the data. I could see how this might be convenient though. /shrug up to you!

Although, looking in the thunks file, it would be easy to just move the data into there and then you could make this more generic of just patching tour data. I only say this because technically you could alter multiple tour attributes at once. Like say we implemented a "NO TOURS EVER" button. You could (and would want to) set that all in a single API call.

src/tours/newUserCourseHomeTour/NewUserCourseHomeTourModal.jsx
```diff
@@ -0,0 +1,73 @@
...
+      heroNode={(
+        <ModalDialog.Hero>
+          <ModalDialog.Hero.Background
+            backgroundSrc={heroImage}
+          />
+          <ModalDialog.Hero.Content style={{ maxWidth: '20rem' }}>
+            <ModalDialog.Title as="h2">
+              <FormattedMessage
+                id="tours.newUserModal.title"
+                defaultMessage="{welcome} {siteName} course!"
```
Dillon-Dumesnil: Why separately translate the welcome? Is it just for the text-accent? Do we not want that on the rest of the text?

Dillon-Dumesnil (commented):
I don't think there was anything major, but just holding off on approval until you have a chance to read through and respond

src/tours/newUserCourseHomeTour/NewUserCourseHomeTourModal.jsx
ciduarte: Yeah it's because "Welcome to your" needs to be in the accent color and "edX course!" needs to be white.

src/tours/data/api.js
ciduarte: I like this idea 👍

src/tour/tests/Tour.test.jsx
ciduarte: Good call!

src/tour/Checkpoint.scss
ciduarte: The popper library tries to center the arrow (which is usually just a square rotated about its center axis, so centering it makes sense). Due to the design constraints, I needed to use a triangle shape, and to create that and center it, we need to override the left and top alignments by the width of the triangle (15px).

Thanks for pointing this out, realized I should use a variable here to solidify the correlation between those values.

src/tour/Checkpoint.jsx
ciduarte: Actually the removal of the first line and addition of the second line are slightly unrelated. I took out the first line, then some iterations later realized I needed a way to control the Checkpoint's visibility again, then added this back in w/ opacity rather than display. Glad you called this out because for a11y, the better option here is `display: none` to hide the content from screen readers while it isn't visible. Technically, the checkpoint is intended to always become visible, but I think better safe than sorry (i.e. we should hide it from screen readers too until it's actually visible).

For context: For the Popper library to be able to anchor the Checkpoint to the correct spot, the Checkpoint needs to be rendered in the DOM. This change had to do w/ the order of operations of trying to render the Checkpoint, anchor it, and scroll to it.

src/tab-page/LoadedTabPage.jsx
ciduarte: Sorry I can see how this is confusing. For context:

`showNewUserTourPrompt` is the state used within `LoadedTabPage` to handle opening/closing the Modal. a state handler is necessary because we can't close the Modal w/o state.

`showNewUserCourseHomeModal` is a value that's populated asynchronously from the tour endpoint, which is what made this useEffect necessary (i.e. we needed a way update the default state of `showNewUserTourPrompt`. Basically prompt and modal were interchangeable here, and I named them poorly.

`showNewUserCourseHomeModal` exists separately from `showNewUserCourseHomeTour` because we want to avoid launching the modal when someone clicks the "Launch tour" button.

But you pointing this out made me realize we can just use `showNewUserCourseHomeModal` and add a new action to the slice to handle closing the modal (which would allow us to get rid of this useEffect). So I'm going to resolve all the comments about prompt vs modal because I'm getting rid of prompt.

Thanks 😁

src/course-home/outline-tab/OutlineTab.jsx
ciduarte: Yeah I was slightly adopting BEM naming conventions where you do `block__element--modifier` to visually denote that courseHome is kind of a parent of the element in question. But I don't have a strong preference here, and BEM is for class name styling anyways, so I'll switch these to hyphens

src/tour/Checkpoint.jsx
ciduarte: Oooh just read through the documentation and Popper does not like `display: none`. Need to use:
```
  visibility: hidden;
  pointer-events: none;
```

src/tour/Checkpoint.jsx
ciduarte: It could, but Jeff Witt's recommendation was to leave it in the component to avoid inconsistency in relying on developers to supply this text.

src/tour/Checkpoint.jsx
Dillon-Dumesnil: 👏

src/tour/Checkpoint.scss
Dillon-Dumesnil: Ah I see. I think I was just a bit confused on why `!important` rather than being able to just use `left: -15px;`. CSS is wild. I would have thought this CSS would have priority being applied last, but I guess it's not?

src/tours/newUserCourseHomeTour/NewUserCourseHomeTourModal.jsx
Dillon-Dumesnil: Well I don't love that, but _fine_

src/tour/Checkpoint.scss
ciduarte: Yeah the popper library adds its own `left` and `top` styling directly inline to the element, which trumps whats applied here in the CSS file in terms of order. So we use `!important` to say no thank you to popper's inline styling

src/tours/AbandonTour.jsx
ciduarte: Added an ADR to address this!

docs/decisions/0010-tour-structure-decisions.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+The reason we use a JSX file rather than a JS file is to allow for use of React components within the objects such as
+`<FormattedMessage />`.
+
+## Implementing i18n in tour objects
+The `<Tour />` component ingests a single prop called `tours` which expects a list of objects.
+Given the structure in which we organized tour objects, there were two considerations in working with i18n:
+- You can't injectIntl into something that isn't a React component without considerable adjustments,
+so using the familiar `{intl.formatMessage(messsages.foo)}` syntax would not be possible.
```
Dillon-Dumesnil: Nit: extra s in messages

docs/decisions/0010-tour-structure-decisions.md
```diff
@@ -0,0 +1,32 @@
...
+
+const exampleTour = (enabled) => ({
+    checkpoints: [],
+    enabled,
+    tourId: 'exampleTour',
+});
+```
+
+The reason we use a JSX file rather than a JS file is to allow for use of React components within the objects such as
+`<FormattedMessage />`.
```
Dillon-Dumesnil, line 20: Fantastic context here. I was wondering why not a JS file above and then you answer it here. Wonderful

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -1322,7 +1325,7 @@ describe('Outline Tab', () => {
     });
   });
 
-  describe('Accont Activation Alert', () => {
+  describe('Account Activation Alert', () => {
```
Dillon-Dumesnil, line 1328: :clap:

src/tab-page/LoadedTabPage.jsx
```diff
@@ -13,6 +15,18 @@ import InstructorToolbar from '../instructor-toolbar';
 import useEnrollmentAlert from '../alerts/enrollment-alert';
 import useLogistrationAlert from '../alerts/logistration-alert';
 
+import abandonTour from '../tours/AbandonTour';
+import coursewareTour from '../tours/CoursewareTour';
+import existingUserCourseHomeTour from '../tours/ExistingUserCourseHomeTour';
+import newUserCourseHomeTour from '../tours/newUserCourseHomeTour/NewUserCourseHomeTour';
+import NewUserCourseHomeTourModal from '../tours/newUserCourseHomeTour/NewUserCourseHomeTourModal';
```
Dillon-Dumesnil: Why does this have a capital `N`?

src/tab-page/LoadedTabPage.jsx
```diff
@@ -40,8 +63,78 @@ function LoadedTabPage({
...
+    }
+  }, [showExistingUserCourseHomeTour]);
+
+  useEffect(() => {
+    if (metadataModel === 'courseHomeMeta' && showNewUserCourseHomeTour) {
+      setIsNewUserCourseHomeTourEnabled(true);
+    }
+  }, [showNewUserCourseHomeTour]);
+
+  const tours = [
```
Dillon-Dumesnil: Maybe an inline comment here about how these are ordered? It looks like it is alphabetized, but if I remember correctly, don't we prefer the earliest enabled tour? So the order here could matter.

src/tab-page/LoadedTabPage.jsx
```diff
@@ -40,8 +63,78 @@ function LoadedTabPage({
...
+  const [isNewUserCourseHomeTourEnabled, setIsNewUserCourseHomeTourEnabled] = useState(false);
+
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    // Tours currently only exist on the Outline Tab and within Courseware, so we'll avoid
+    // calling the tour endpoint unnecessarily.
+    if (username && (activeTabSlug === 'outline' || metadataModel === 'coursewareMeta')) {
+      dispatch(fetchTourData(username));
+    }
```
Dillon-Dumesnil: Just want to confirm, will this end up calling every single courseware page or just the first time and is smart enough to know it already has the data after that?

src/tour/tests/Tour.test.jsx
```diff
@@ -345,34 +358,68 @@ describe('Tour', () => {
...
             {targets}
-          </>
-        ));
+          </>,
+        );
+
+        expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
+      });
+
+      it('advances to next valid Checkpoint', () => {
```
Dillon-Dumesnil, line 373: Maybe have this test also show how Checkpoint 1 is not in the document? Or not part of the tour?

src/tours/CourseHomeTours.test.jsx
```diff
@@ -0,0 +1,150 @@
...
+  ${403}
+  ${404}
+  `('does not render tour components for $errorStatus response', async (errorStatus) => {
+  setTourData({}, errorStatus, false);
+
+  // Verify no launch tour button
+  expect(await screen.queryByRole('button', { name: 'Launch tour' })).not.toBeInTheDocument();
+
+  // Verify no Checkpoint or MarketingModal has rendered
+  expect(await screen.queryByRole('dialog')).not.toBeInTheDocument();
```
Dillon-Dumesnil: That's dope

src/tours/data/api.js
```diff
@@ -0,0 +1,36 @@
...
+        toursDisabled: true,
+      };
+    }
+    if (httpErrorStatus === 404) {
+      // The API will return a 404 if no User Tour objects exist
+      return {
+        toursDisabled: true,
+      };
+    }
+    throw error;
```
Dillon-Dumesnil: Should this be in a switch or maybe just combine all of them into a single if? Feels like it could be more DRY.
Or maybe at minimum, reducing the return to `return { toursDisabled: true };`

src/tours/data/slice.js
```diff
@@ -0,0 +1,58 @@
...
+    setTourData: (state, { payload }) => {
+      const {
+        courseHomeTourStatus,
+        showCoursewareTour,
+        toursDisabled,
+      } = payload;
+      state.showCoursewareTour = showCoursewareTour;
+      state.showExistingUserCourseHomeTour = courseHomeTourStatus === 'show-existing-user-tour';
+      state.showNewUserCourseHomeModal = courseHomeTourStatus === 'show-new-user-tour';
+      state.toursEnabled = !toursDisabled;
```
Dillon-Dumesnil: Nit: Should the payload maybe just send `toursEnabled`? I can't remember all the code, but curious if we could be more consistent with these.

Dillon-Dumesnil (approved):
Some small questions and clarification questions, but I think overall, it looks great.

src/tab-page/LoadedTabPage.jsx
ciduarte: Because it's an actual React component (unlike the preceding tour objects)

src/tab-page/LoadedTabPage.jsx
ciduarte: I believe with the way courseware is set up this will only run on the initial render of a courseware page. So switching between units/sections/sequences shouldn't re-render the whole app and will not trigger fetching this data again.

src/tours/data/api.js
ciduarte: Yeah I can definitely clean this up 🙃

src/tab-page/LoadedTabPage.jsx
ciduarte: The Tour component will only render the first enabled tour, but in theory none of these should be enabled at the same time which is why I went with alphabetical order

docs/decisions/0010-tour-structure-decisions.md
cdeery: Agreed. Great documentation.

cdeery (approved):
Great job. This looks fantastic.

src/product-tours/ProductTours.jsx
```diff
@@ -0,0 +1,184 @@
...
+  const logUpgradeClick = () => {
+    sendTrackEvent('edx.bi.ecommerce.upsell_links_clicked', {
+      org_key: org,
+      courserun_key: courseId,
+      linkCategory: '(none)',
+      linkName: 'course_home_upgrade_product_tour',
+      linkType: 'link',
+      pageName: 'course_home',
+    });
+  };
```
Dillon-Dumesnil: Not a big deal, but why define this log function here rather than in the one file that uses it? My two guesses are in case we want to reuse it in other tours and/or so we don't have to pass down `org` and/or `courseId`. 

If there's no reason, I might advocate to move it down closer to the piece of code. If it's the first reason, I think it might be nice to make the `linkName` and `pageName` configurable so other tours actually could use them. If it's just the second reason, /shrug. looks fine then.

src/product-tours/ProductTours.jsx
```diff
@@ -0,0 +1,184 @@
...
+
+  // The <Tour /> component cannot handle rendering multiple enabled tours at once.
+  // I.e. when adding new tours, beware that if multiple tours are enabled,
+  // the first enabled tour in the following array will be the only one that renders.
+  // The suggestion for populating these tour objects is to ensure only one tour is enabled at a time.
```
Dillon-Dumesnil, line 90: 👏

src/product-tours/ProductTours.jsx
```diff
@@ -0,0 +1,184 @@
...
+  const tours = [
```
Dillon-Dumesnil, line 91: I know this is late in the game to request this, but would you mind adding in the parameter names to each of these tours? It would make reading this file so much easier if I knew
```
() => { ... },
() => { ... },
```
was actually
```
onStart=() => { ... },
onEnd=() => { ... }
```
It is true there are some context clues that can be used, but I think it would still be easier if we could see what param each of these was being mapped to since it is also not consistent across tours (some take 2 args, some take 5)

src/product-tours/ProductTours.test.jsx
```diff
@@ -0,0 +1,150 @@
+/**
```
Dillon-Dumesnil, line 1: Why rename this file if it is still only testing the CourseHomeTours? I feel like it might be more clear as the CourseHomeTours.test.jsx and then you can add in a docstring that it contains the tests for all possible CourseHomeTour states (new, existing, none) since those technically live in distinct files.

On a related note, I don't see any tests about the courseware tour. Should there maybe be one in here or perhaps broken off into its own test file?

src/product-tours/data/api.js
```diff
@@ -0,0 +1,26 @@
...
+  const url = `${getConfig().LMS_BASE_URL}/api/user_tours/v1/${username}`;
+  try {
+    const { data } = await getAuthenticatedHttpClient().get(url);
+    return { toursEnabled: true, ...camelCaseObject(data) };
+  } catch (error) {
+    const { httpErrorStatus } = error && error.customAttributes;
+    /** The API will return a
+     *    401 if the user is not authenticated
+     *    403 if the tour waffle flag is inactive
+     *    404 if no User Tour objects exist
```
Dillon-Dumesnil: tiny tiny nit: `404 if no User Tour object exists for the given username`

src/product-tours/data/thunks.js
```diff
@@ -0,0 +1,46 @@
...
+  disableCoursewareTour,
+  disableNewUserCourseHomeModal,
+  setTourData,
+} from './slice';
+
+export function closeNewUserCourseHomeModal() {
+  return async (dispatch) => dispatch(disableNewUserCourseHomeModal());
+}
+
+export function endCourseHomeTour(username) {
```
Dillon-Dumesnil, line 15: Small question, why have the patch endpoints catch and log their own errors, but the get endpoint handles it all for this thunks file?

src/product-tours/newUserCourseHomeTour/LaunchCourseHomeTourButton.jsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const {
+    toursEnabled,
+  } = useSelector(state => state.tours);
+
+  const dispatch = useDispatch();
+
+  const handleClick = () => {
+    const { administrator } = getAuthenticatedUser();
+    sendTrackEvent('edx.ui.lms.launch_tour.clicked', {
```
Dillon-Dumesnil, line 32: Should this specify something about this being the course home tour that is being launched as opposed to any other? Could maybe do this via a variant field in the payload if you'd like all tours to be able to use this same tracking id.

src/tab-page/LoadedTabPage.jsx
```diff
@@ -42,6 +45,12 @@ function LoadedTabPage({
 
   return (
     <>
+      <ProductTours
+        activeTab={activeTabSlug}
+        courseId={courseId}
+        metadataModel={metadataModel}
+        org={org}
+      />
```
Dillon-Dumesnil: 👨‍🍳 💋

Dillon-Dumesnil (commented):
Some questions, but overall, loving all of these improvements to the PR

src/product-tours/ProductTours.test.jsx
ciduarte: Good callout, this file is meant to also test the courseware test but you don't see one because I forgot to write it!

src/product-tours/ProductTours.jsx
ciduarte: Yeah this is a good idea. JS doesn't support that syntax directly so I had to update the function arguments to take objects, but now we'll be able to label the values using keys.

src/product-tours/ProductTours.jsx
ciduarte: I agree that it would be nice if this could be closer to the tour so I just created a new object called `upgradeData` to pass down.

src/product-tours/data/thunks.js
ciduarte: Another good callout. We're actually pretty inconsistent with how/where we handle errors in this MFE. I'm not sure if there is a valid reason for it, but I adjusted this so we'd error handle in one spot 👍

PR #759, opened 2021-12-02
fix: Only dismiss the modal when masquerading as specific learner

schenedx:
[MST-1109](https://openedx.atlassian.net/browse/MST-1109) If the user is masquerading as a specific learner, then dismiss the modal and do not post back and save the Honor Code signature

### DIFF
--- a/src/courseware/course/sequence/honor-code/HonorCode.jsx
+++ b/src/courseware/course/sequence/honor-code/HonorCode.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { useDispatch } from 'react-redux';
 import { getConfig, history } from '@edx/frontend-platform';
+import { getAuthenticatedUser } from '@edx/frontend-platform/auth';
 import { FormattedMessage, injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import { ActionRow, Alert, Button } from '@edx/paragon';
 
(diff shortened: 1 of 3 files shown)


alangsto (approved):
🔥

src/courseware/course/sequence/honor-code/HonorCode.jsx
```diff
@@ -11,13 +12,23 @@ import messages from './messages';
...
 
-  const handleAgree = () => dispatch(saveIntegritySignature(courseId, isMasquerading));
+  const handleAgree = () => dispatch(
+    // If the request is made by a staff user masquerading as a specific learner,
+    // don't actually create a signature for them on the backend.
+    // Only the modal dialog will be dismissed.
+    // Otherwise, even for staff users, we want to record the signature.
+    saveIntegritySignature(
+      courseId,
+      coursewareMetaData.isMasquerading && coursewareMetaData.username !== authUser.username,
```
michaelroytman, line 29: This is a nit.

This was a little unclear to me. I think this might be a little clearer.

```
if (coursewareMetaData.username !== authUser.username) {
    saveIntegritySignature(
          courseId,
          coursewareMetaData.isMasquerading,
)};
```
The `coursewareMetaData.username !== authUser.username` isn't really representing whether the user is masquerading or not, right? They're masquerading either way, so it feels a little unclear to me to pass it as the parameter `isMasquerading`.

I think what you have is good; it's just something I noticed that might improve the clarity a little.

michaelroytman (approved):
Looks great to me! I left one suggestion that I'm fine with you taking or leaving. I appreciate the time you took to write the tests. They're very clear.

I have one question.

In the ticket, there is the following description...

> This will require comparing the staff user's ID or username with the user they're masquerading as. This isn't currently possible with the way the data is set up on the frontend, so it may require some backend changes. (you can access the staff user's information with `getAuthenticatedUser()`, but there is no way to do the same for the masqueraded user)

Was this the case? It looks like the `coursewareMeta` model has information about the masquerading user, right? Was that already there? Or did you add that in another pull request?

schenedx:
@MichaelRoytman To answer your question, it was already there. The `coursewareMeta.username` already contain the username being masqueraded. So, if I am an edX staff masquerade someone, say `learner1`, the `coursewareMeta.username` would give me `learner1`

src/courseware/course/sequence/honor-code/HonorCode.jsx
schenedx: Thank you for the feedback. 
The truth is, I am too lazy to figure out how to put another function between `dispatch` and `saveIntegritySignature` so they work. I tried with an unnamed function and that didn't work. 
As for `coursewareMetaData.username !== authUser.username` isn't representing masquerading, I'd argue, that condition alone means it is masquerading. Frankly, in prod, that condition alone should suffice because we won't have a case where username is different and we are not masquerading. But I want to be pedantic to the code structure here. Thank you for your suggestion.

PR #772, opened 2021-12-10
feat: Add NotificationTray persistence by course

julianajlk:
[REV-2424](https://openedx.atlassian.net/browse/REV-2424).
When the `NotificationTray` was first implemented, it was not specific to a course - when the user closed the tray it would persist as they navigated through the units however if they reloaded the page or went to course home and then back to courseware, it would display the tray open even if the user had closed it.

This PR adds the functionality to have this open/closed persist by course and session. User session is saved for each course until the browser is closed. Once the browser is closed, when the user opens the browser again, courseware loads with the tray open for all courses.

**Note on unit tests:**
The tray touches 3 generations of components, which with Jest/jsdom and frontend testing by component makes it difficult to fully cover a real user behavior across multiple components.
Codecov was happy without any tests added, but I added some tests to cover the behavior when the button is clicked to open/close the tray and a test to see that with multiple courses.

**This PR:**
- Changes the `NotificationTray` functionality to be course specific. 
- Adds tests. 
- Renames some variables: `shouldDisplayNotificationTriggerInSequence` or `...InCourse` and `shouldDisplayNotificationTrayOpenOnLoad`

**How to test:**
- Checkout my branch `julianajlk/REV-2424/value-prop-notification-tray-course-specific`
- Create a new course other than the Demo course with at least verified mode available, via [Ecommerce Course Admin](http://localhost:18130/courses/). Make sure you create the course in studio first.
- Enroll in both courses, and open each course in a different tab.
- On the first course, go to courseware, and close the tray. It should load open at first.
- On the second course, go to courseware, and leave it open. 
- Reload both tabs.
- First course should load with the tray closed. Second course should load with the tray open.
- Close the tab for the first course to end the session.
- Open a new tab and go to the first course again. Tray should load open.

### DIFF
--- a/src/courseware/course/Course.jsx
+++ b/src/courseware/course/Course.jsx
@@ -16,6 +16,7 @@ import NotificationTrigger from './NotificationTrigger';
 import { useModel } from '../../generic/model-store';
 import useWindowSize, { responsiveBreakpoints } from '../../generic/tabs/useWindowSize';
 import { getLocalStorage, setLocalStorage } from '../../data/localStorage';
+import { getSessionStorage, setSessionStorage } from '../../data/sessionStorage';
 
 /** [MM-P2P] Experiment */
 import { initCoursewareMMP2P, MMP2PBlockModal } from '../../experiments/mm-p2p';
@@ -50,15 +51,31 @@ function Course({
     courseId, sequenceId, unitId, celebrateFirstSection, dispatch, celebrations,
   );
 
-  const shouldDisplayNotificationTrigger = useWindowSize().width >= responsiveBreakpoints.small.minWidth;
-
-  const shouldDisplayNotificationTrayOpen = useWindowSize().width > responsiveBreakpoints.medium.minWidth;
+  // Responsive breakpoints for showing the notificaiton button/tray
+  const shouldDisplayNotificationTriggerInCourse = useWindowSize().width >= responsiveBreakpoints.small.minWidth;
+  const shouldDisplayNotificationTrayOpenOnLoad = useWindowSize().width > responsiveBreakpoints.medium.minWidth;
+  // Course specific notification tray open/closed persistance by browser session
+  if (!getSessionStorage(`notificationTrayStatus.${courseId}`)) {
+    if (shouldDisplayNotificationTrayOpenOnLoad) {
+      setSessionStorage(`notificationTrayStatus.${courseId}`, 'open');
+    } else {
+      // responsive version displays the tray closed on initial load, set the sessionStorage to closed
+      setSessionStorage(`notificationTrayStatus.${courseId}`, 'closed');
+    }
+  }
 
   const [notificationTrayVisible, setNotificationTray] = verifiedMode
-    && shouldDisplayNotificationTrayOpen ? useState(true) : useState(false);
+    && shouldDisplayNotificationTrayOpenOnLoad && getSessionStorage(`notificationTrayStatus.${courseId}`) !== 'closed' ? useState(true) : useState(false);
+
   const isNotificationTrayVisible = () => notificationTrayVisible && setNotificationTray;
+
   const toggleNotificationTray = () => {
     if (notificationTrayVisible) { setNotificationTray(false); } else { setNotificationTray(true); }
+    if (getSessionStorage(`notificationTrayStatus.${courseId}`) === 'open') {
+      setSessionStorage(`notificationTrayStatus.${courseId}`, 'closed');
+    } else {
+      setSessionStorage(`notificationTrayStatus.${courseId}`, 'open');
+    }
   };
 
   if (!getLocalStorage(`notificationStatus.${courseId}`)) {
@@ -96,7 +113,7 @@ function Course({
           mmp2p={MMP2P}
         />
 
-        { shouldDisplayNotificationTrigger ? (
+        { shouldDisplayNotificationTriggerInCourse ? (
           <NotificationTrigger
             courseId={courseId}
             toggleNotificationTray={toggleNotificationTray}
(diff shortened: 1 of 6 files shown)


src/courseware/course/Course.test.jsx
```diff
@@ -91,6 +101,52 @@ describe('Course', () => {
     expect(notificationTrigger).not.toHaveClass('trigger-active');
   });
 
+  it('handles click to open/close notification tray', async () => {
+    render(<Course {...mockData} />);
+    expect(sessionStorage.getItem(`notificationTrayStatus.${mockData.courseId}`)).toBe('"closed"');
```
julianajlk: I'm running into an issue where on the first test, upon load of the `Course` component `sessionStorage` for `notificationTrayStatus` is closed but should be open when it loads. This affects the other tests after this one.

**_EDIT_**: clearing the `sessionStorage` before rendering the component only on the tests that handle `sessionStorage` fixed this issue

src/courseware/course/Course.jsx
```diff
@@ -50,15 +51,26 @@ function Course({
...
 
+  // Course specific notification tray open/closed persistance by browser session
+  if (!getSessionStorage(`notificationTrayStatus.${courseId}`)) {
+    setSessionStorage(`notificationTrayStatus.${courseId}`, 'open');
+  }
+  const isNotificationTrayClosed = getSessionStorage(`notificationTrayStatus.${courseId}`) === 'closed';
+
   const [notificationTrayVisible, setNotificationTray] = verifiedMode
-    && shouldDisplayNotificationTrayOpen ? useState(true) : useState(false);
+    && shouldDisplayNotificationTrayOpen && !isNotificationTrayClosed ? useState(true) : useState(false);
```
dianekaplan: I think the naming here could be a little confusing, especially after it's been a while and we come back; based on the names alone I'd imagine that shouldDisplayNotificationTrayOpen would map directly to isNotificationTrayClosed. 

Then looking back at the code, it looks like shouldDisplayNotificationTrayOpen comes from the window size, and then this new thing we're capturing is from user behavior: have they already actively closed the tray, is that right? 

If I'm understanding it correctly, I think the new variable may be clearer as something like 'userClosedTray'

julianajlk: Ah yes! Meant to change that when I started but forgot, I will change the first to something with `responsive` since it applies to the difference of where/when it shows depending if it's a smaller screen or not, and correct, the second is user related, will adjust. Thanks Diane!

Also, the rationality for the naming `shouldDisplayNotificationTrayOpen` being not named after its medium sized screen is that being medium doesn't mean anything for someone who is not familiar with UX's asks for this, so I named it after what it should do/behave. Will find a way to name them in a distinct way but still related to their behavior

julianajlk: Changed the naming to the below 
`shouldDisplayNotificationTrayOpen` -> `shouldDisplayNotificationTrayOpenOnLoad`
`isNotificationTrayClosed` -> `isNotificationTrayClosedByUser`

**_Edit:_** I decided to remove `isNotificationTrayClosed` since even if naming it `isNotificationTrayClosedByUser` would not be true since initially, if in mobile responsive version, the sessionStorage is "closed" even though it was not set by the user.

I did change `shouldDisplayNotificationTrigger` -> `shouldDisplayNotificationTriggerInSequence` or `...InCourse` to be more specific from this review (not related to this ticket)

dianekaplan: thanks Juliana- boy this stuff is confusing! So just to make sure I'm understanding these two variables: 
- shouldDisplayNotificationTriggerInCourse: true if window width >= small, determines whether to show NotificationTrigger on Course.jsx
- shouldDisplayNotificationTriggerInSequence: true if window width < small, determines whether to show NotificationTrigger on Sequence.jsx, plus some display details

Thanks for the info- will have a fresh look at the commits

src/courseware/course/Course.jsx
```diff
@@ -50,15 +51,31 @@ function Course({
     courseId, sequenceId, unitId, celebrateFirstSection, dispatch, celebrations,
   );
 
-  const shouldDisplayNotificationTrigger = useWindowSize().width >= responsiveBreakpoints.small.minWidth;
-
-  const shouldDisplayNotificationTrayOpen = useWindowSize().width > responsiveBreakpoints.medium.minWidth;
+  // Responsive breakpoints for showing the notificaiton button/tray
```
dianekaplan: (little typo in the comment: notificaiton)

src/courseware/course/sequence/Sequence.test.jsx
```diff
@@ -390,6 +392,19 @@ describe('Sequence', () => {
...
+        ...mockData,
+        toggleNotificationTray,
+        notificationTrayVisible: true,
+      };
+      render(<Sequence {...testData} />);
+      const notificationCloseIconButton = await screen.findByRole('button', { name: /Close notification tray/i });
+      fireEvent.click(notificationCloseIconButton);
+      expect(toggleNotificationTray).toHaveBeenCalledTimes(1);
+    });
+
```
dianekaplan, line 407: Should we also check here that the sessionStorage is "closed"? Is the expectation different in Sequence vs Course?

src/courseware/course/Course.jsx
julianajlk: Correct! It's because the button to trigger opening/closing the tray is in the Sequence component (after the > and omitting the word "next"), but in desktop is completely different.

src/courseware/course/sequence/Sequence.test.jsx
julianajlk: This test I just added as a "bonus" that should've been added before when the feature was implemented to check if the click is behaving as it should, calling `toggleNotificationTray`. That's all the test is intended to do. 
Adding more for clarity - my understanding is that since the toggle function is passed in as a prop (so it's not created in this Sequence component), I want to check that the function is called but not necessarily test what the function is doing, which would be tested in the component that the function was created in.

dianekaplan (approved):

PR #773, opened 2021-12-13
fix: [AA-1076] show grade override notice

cdeery:
- Progress page indicates if a grade has been overridden. This adds functionality that was in the legacy progress page but missing in the new Progress page.

image

Ticket 
https://github.com/edx/edx-platform/pull/29585

Backend changes are in https://openedx.atlassian.net/browse/AA-1076

### DIFF
--- a/src/course-home/progress-tab/ProgressTab.test.jsx
+++ b/src/course-home/progress-tab/ProgressTab.test.jsx
@@ -13,6 +13,7 @@ import * as thunks from '../data/thunks';
 import initializeStore from '../../store';
 import ProgressTab from './ProgressTab';
 import LoadedTabPage from '../../tab-page/LoadedTabPage';
+import messages from './grades/messages';
 
 initializeMockApp();
 jest.mock('@edx/frontend-platform/analytics');
@@ -702,6 +703,92 @@ describe('Progress Tab', () => {
       await fetchAndRender();
       expect(screen.getByTestId('gradeSummaryFooterTotalWeightedGrade').innerHTML).toEqual('50%');
     });
+
+    it('renders override notice', async () => {
+      setTabData({
+        grading_policy: {
+          assignment_policies: [
+            {
+              num_droppable: 1,
+              num_total: 2,
+              short_label: 'HW',
+              type: 'Homework',
+              weight: 0.5,
+            },
+            {
+              num_droppable: 0,
+              num_total: 1,
+              short_label: 'Ex',
+              type: 'Exam',
+              weight: 0.5,
+            },
+          ],
+          grade_range: {
+            pass: 0.75,
+          },
+        },
+        section_scores: [
+          {
+            display_name: 'First section',
+            subsections: [
+              {
+                assignment_type: 'Homework',
+                block_key: 'block-v1:edX+DemoX+Demo_Course+type@sequential+block@12345',
+                display_name: 'First subsection',
+                has_graded_assignment: true,
+                learner_has_access: true,
+                num_points_earned: 1,
+                num_points_possible: 2,
+                percent_graded: 1.0,
+                problem_scores: [{
+                  earned: 1,
+                  possible: 2,
+                }],
+                show_correctness: 'always',
+                show_grades: true,
+                url: 'http://learning.edx.org/course/course-v1:edX+Test+run/first_subsection',
+              },
+            ],
+          },
+          {
+            display_name: 'Second section',
+            subsections: [
+              {
+                assignment_type: 'Exam',
+                block_key: 'block-v1:edX+DemoX+Demo_Course+type@sequential+block@98765',
+                display_name: 'Second subsection',
+                learner_has_access: true,
+                has_graded_assignment: true,
+                num_points_earned: 0,
+                num_points_possible: 1,
+                override: {
+                  system: 'PROCTORING',
+                  reason: 'Suspicious activity',
+                },
+                percent_graded: 1.0,
+                problem_scores: [{
+                  earned: 0,
+                  possible: 1,
+                }],
+                show_correctness: 'always',
+                show_grades: true,
+                url: 'http://learning.edx.org/course/course-v1:edX+Test+run/second_subsection',
+              },
+            ],
+          },
+        ],
+      });
+
+      await fetchAndRender();
+
+      const problemScoreDrawerToggle = screen.getByRole('button', { name: 'Toggle individual problem scores for Second subsection' });
+      expect(problemScoreDrawerToggle).toBeInTheDocument();
+
+      // Open the problem score drawer
+      fireEvent.click(problemScoreDrawerToggle);
+
+      expect(screen.getByText(messages.sectionGradeOverridden.defaultMessage)).toBeInTheDocument();
+    });
   });
 
   describe('Detailed Grades', () => {
@@ -775,7 +862,7 @@ describe('Progress Tab', () => {
       Object.defineProperty(window, 'matchMedia', {
         writable: true,
         value: jest.fn().mockImplementation(query => {
-          const matches = !!(query === 'screen and (min-width: 992px)');
+          const matches = (query === 'screen and (min-width: 992px)');
           return {
             matches,
             media: query,
--- a/src/course-home/progress-tab/ProgressTab.test.jsx
+++ b/src/course-home/progress-tab/ProgressTab.test.jsx
@@ -706,27 +706,6 @@ describe('Progress Tab', () => {
 
     it('renders override notice', async () => {
       setTabData({
-        grading_policy: {
-          assignment_policies: [
-            {
-              num_droppable: 1,
-              num_total: 2,
-              short_label: 'HW',
-              type: 'Homework',
-              weight: 0.5,
-            },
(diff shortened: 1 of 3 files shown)


src/course-home/progress-tab/grades/detailed-grades/SubsectionTitleCell.jsx
```diff
@@ -6,7 +6,11 @@ import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 import { getAuthenticatedUser } from '@edx/frontend-platform/auth';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import { Collapsible, Icon, Row } from '@edx/paragon';
-import { ArrowDropDown, ArrowDropUp, Blocked } from '@edx/paragon/icons';
+import {
+  ArrowDropDown, ArrowDropUp, Blocked,
+} from '@edx/paragon/icons';
+import { faInfoCircle } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
```
mikix: We are moving away from using FontAwesome icons to paragon icons (which may or may not use FA under the covers, I'm not 100% - I believe it has some icons that FA does not?):
- https://edx.github.io/paragon/components/icon
- https://edx.github.io/paragon/foundations/icons

So something like:
```
import { Icon } from '@edx/paragon';
import { Info } from '@edx/paragon/icons';

<Icon src={Info} />
```

(which here, means you can add just Icon to the already imported list of icons above)

src/course-home/progress-tab/grades/detailed-grades/SubsectionTitleCell.jsx
```diff
@@ -79,7 +83,16 @@ function SubsectionTitleCell({ intl, subsection }) {
         </span>
       </Row>
       <Collapsible.Body className="d-flex w-100">
-        <ProblemScoreDrawer problemScores={problemScores} subsection={subsection} />
+        <div className="col w-100">
+          { subsection.override
+            && (
+            <div className="row w-100 m-0 x-small ml-4 pt-2 pl-1 text-gray-700 flex-nowrap">
+              <div><FontAwesomeIcon icon={faInfoCircle} className="fa-lg mr-1 text-primary-500" /></div>
+              <div data-testid="override-message">{intl.formatMessage(messages.sectionGradeOverridden)}</div>
```
mikix: The testid looks leftover from a previous test attempt - I'd drop it though, as it's unused and would end up in resulting DOM.

mikix (changes requested):

src/course-home/progress-tab/ProgressTab.test.jsx
```diff
@@ -702,6 +703,92 @@ describe('Progress Tab', () => {
       await fetchAndRender();
       expect(screen.getByTestId('gradeSummaryFooterTotalWeightedGrade').innerHTML).toEqual('50%');
     });
+
+    it('renders override notice', async () => {
+      setTabData({
```
Dillon-Dumesnil, line 708: Not sure if this is possible, but are you able to override a smaller subset of the tab data here? I feel bad for ya having to write all this when I think we only care about one small part. Maybe even by taking the existing tab data and just doing something like `tabData.section_scores[1].subsections.push(...)` and then just add the one thing of the override. Or maybe even `tabData.section_scores[1].subsections[0].override = ...;`.

I think it could help the readability of this test as well if that's possible so it's easier to see exactly what we care about in the tab data.

src/course-home/progress-tab/grades/messages.js
```diff
@@ -115,6 +115,14 @@ const messages = defineMessages({
...
       + "Your weighted grade is what's used to determine if you pass the course.",
   },
+  noAccessToAssignmentType: {
+    id: 'progress.noAcessToAssignmentType',
+    defaultMessage: 'You do not have access to assignments of type {assignmentType}',
+  },
+  noAccessToSubsection: {
+    id: 'progress.noAcessToSubsection',
+    defaultMessage: 'You do not have access to subsection {displayName}',
+  },
```
Dillon-Dumesnil, line 125: 👏

src/course-home/progress-tab/grades/detailed-grades/SubsectionTitleCell.jsx
Dillon-Dumesnil: I don't think you need the data-testid here. I don't think your test is using it. Maybe consult Carla on this point.

src/course-home/progress-tab/grades/detailed-grades/SubsectionTitleCell.jsx
```diff
@@ -79,7 +83,16 @@ function SubsectionTitleCell({ intl, subsection }) {
         </span>
       </Row>
       <Collapsible.Body className="d-flex w-100">
-        <ProblemScoreDrawer problemScores={problemScores} subsection={subsection} />
+        <div className="col w-100">
+          { subsection.override
+            && (
+            <div className="row w-100 m-0 x-small ml-4 pt-2 pl-1 text-gray-700 flex-nowrap">
```
Dillon-Dumesnil, line 86: Might be worth having Carla take a look at this because I'm definitely not the one to speak on it, but looks fine to me

Dillon-Dumesnil: Do you need the `w-100` here if you already have it on the parent div? And I just realized it's on the collapsible.body as well. I think you might be able to drop both.

src/course-home/progress-tab/grades/detailed-grades/SubsectionTitleCell.jsx
```diff
@@ -91,6 +104,10 @@ SubsectionTitleCell.propTypes = {
     blockKey: PropTypes.string.isRequired,
     displayName: PropTypes.string.isRequired,
     learnerHasAccess: PropTypes.bool.isRequired,
+    override: PropTypes.shape({
+      system: PropTypes.string,
+      reason: PropTypes.string,
+    }),
```
Dillon-Dumesnil, line 113: Do you know if we have plans on using these fields?

src/course-home/progress-tab/grades/detailed-grades/SubsectionTitleCell.jsx
Dillon-Dumesnil: Don't know why this didn't show up when I was reviewing the code, but +1 here

src/course-home/progress-tab/ProgressTab.test.jsx
cdeery: I kept adding stuff to make the warnings being thrown by the test go away.

Dillon-Dumesnil: Oof. Okay, took a look at the code and see that that method sets what the axios call will respond so yes, better to set it all there. That being said, I believe you only need to `section_scores` (see test "renders correct total weighted grade when showGrades is false" for an example) since I think the only part you care about is setting the override. Should at least help with some of the cruft.

src/course-home/progress-tab/grades/detailed-grades/SubsectionTitleCell.jsx
cdeery: It was leftover from a different testing approach. removed.

src/course-home/progress-tab/grades/detailed-grades/SubsectionTitleCell.jsx
cdeery: Got it. Done

src/course-home/progress-tab/grades/detailed-grades/SubsectionTitleCell.jsx
cdeery: Done

src/course-home/progress-tab/grades/messages.js
cdeery: I thought you would like that!

src/course-home/progress-tab/grades/detailed-grades/SubsectionTitleCell.jsx
cdeery: Thanks for the reminder. I was going to see if we wanted them in a tooltip for the info icon

src/course-home/progress-tab/grades/detailed-grades/SubsectionTitleCell.jsx
mikix: Did you forget to push up? I still see the testid.

src/course-home/progress-tab/grades/detailed-grades/SubsectionTitleCell.jsx
mikix: The paragon docs linked above have an example:
```
  <Icon src={Add} className="text-white" style={{ height: '48px', width: '48px' }} />
```

Which seems like a very manual way to do it, but :shrug:. Does that work?

mikix (approved):

PR #774, opened 2021-12-14
feat: use learning sequences even when masquerading

mikix:
The backend recently grew support for it, so we can use it directly instead of falling back to course blocks in that case.

[AA-1151](https://openedx.atlassian.net/browse/AA-1151)


Dillon-Dumesnil (approved):
LGTM

PR #778, opened 2021-12-17
fix: round grades on progress tab

ciduarte:


src/course-home/data/api.js
```diff
@@ -15,7 +15,10 @@ const calculateAssignmentTypeGrades = (points, assignmentWeight, numDroppable) =
   let averageGrade = 0;
   let weightedGrade = 0;
   if (points.length) {
-    averageGrade = points.reduce((a, b) => a + b, 0) / points.length;
+    // Calculate the average grade for the assignment and round it. This rounding is not ideal and does not accurately
+    // reflect what a learner's grade would be, however, we must have parity with the current grading behavior that
+    // exists in edx-platform.
```
ciduarte, line 20: Not sure if this comment has too much of my opinion in it 🙂

mikix: I think it's fine, but would actually like to see a little more detail about _what_ is wrong - just a throwaway line like "multiple rounding steps leads to an inaccurate final grade" or something. Like, maybe the "not ideal" rounding problem you mention is that Acadamia tends to actually prefer `floor` rounding (as a hypothetical) or maybe some engineer in two years will read this but is not familiar with the issue with rounding rounded values.

Anyway, just a nit - not enough to block this over the holidays

mikix (approved):

PR #790, opened 2022-01-04
fix: don't log errors when we ask for sequence metadata for units

mikix:
Related to (but not blocking or blocked by) https://github.com/edx/edx-platform/pull/29707

### DIFF
--- a/src/courseware/data/thunks.js
+++ b/src/courseware/data/thunks.js
@@ -229,7 +229,13 @@ export function fetchSequence(sequenceId) {
         dispatch(fetchSequenceSuccess({ sequenceId }));
       }
     } catch (error) {
-      logError(error);
+      // Some errors are expected - for example, CoursewareContainer may request sequence metadata for a unit and rely
+      // on the request failing to notice that it actually does have a unit (mostly so it doesn't have to know anything
+      // about the opaque key structure). In such cases, the backend gives us a 422.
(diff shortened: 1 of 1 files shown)


Dillon-Dumesnil (approved):

cdeery (approved):
This seems right, if a little ugly. Not your fault. 
Using the error codes as a way of sending implied information back to the caller definitely feels like an antipattern.

mikix:
> Using the error codes as a way of sending implied information back to the caller definitely feels like an antipattern.

I feel like this isn't the first time this has come up - I think you had a similar bump on a previous PR?

I think HTTP status codes have a reasonable place in inter-application communication. Sometimes it makes more sense or is more convenient than throwing json at the other side, at least in my mind. But reasonable people could disagree there.

Would your preferred solution be maybe a 200 status code, with a json bit like `{ is_sequence: false }` or `{ error: not-a-sequence }`? And all 4xx codes should be reserved for application errors that would get flagged?

I'm not sure I'm fully bought into that worldview, but I understand it, if that is your vibe.

cdeery:
I get it. There is a pattern for APIs that return a 200, but send back an error block with more information about what went wrong. It's a judgement call - you can argue that a 200 is also quite wrong. It depends on how much context you want to return with your error message.

mikix:
I won't change anything on this already finished PR, but I will give a harder think about json vs status code next time I do something similar.

PR #792, opened 2022-01-06
fix: add aria radiogroup role to goals widget

MatthewPiatetsky:
first bullet here https://openedx.atlassian.net/browse/AA-1100
i believe this implements the suggestion - thought not entirely sure if this is what we want, as i don't think the radiogroup is selectable so i don't think the label will be read, but will do some more investigation


ciduarte (approved):
As long as this didn't impact styling/behavior, this seems like what Jeff suggested we do!

PR #794, opened 2022-01-07
fix: update Tour components and product tour behavior

ciduarte:
Will update the PR and squash these commits before merging, but left it this way for more simplicity while reviewing 🤞

### DIFF
--- a/src/course-home/data/slice.js
+++ b/src/course-home/data/slice.js
@@ -11,11 +11,15 @@ const slice = createSlice({
   initialState: {
     courseStatus: 'loading',
     courseId: null,
+    proctoringPanelStatus: 'loading',
     toastBodyText: null,
     toastBodyLink: null,
     toastHeader: '',
   },
   reducers: {
+    fetchProctoringInfoResolved: (state) => {
+      state.proctoringPanelStatus = LOADED;
+    },
     fetchTabDenied: (state, { payload }) => {
       state.courseId = payload.courseId;
       state.courseStatus = DENIED;
@@ -47,6 +51,7 @@ const slice = createSlice({
 });
 
 export const {
+  fetchProctoringInfoResolved,
   fetchTabDenied,
   fetchTabFailure,
   fetchTabRequest,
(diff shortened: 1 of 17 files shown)


src/product-tours/ProductTours.jsx
```diff
@@ -52,12 +55,16 @@ function ProductTours({
   const administrator = getAuthenticatedUser() && getAuthenticatedUser().administrator;
 
   useEffect(() => {
+    const isCourseware = metadataModel === 'coursewareMeta';
```
mikix: This feels a little shaky. Like, if Chris's work to unify the metadata calls finishes, this would need retooling. Looks like you have `activeTab` available - could you just compare that? (I know you're just moving this check, but still.)

src/product-tours/ProductTours.jsx
```diff
@@ -52,12 +55,16 @@ function ProductTours({
...
+    const isOutlineTabAndProctorPanelIsResolved = activeTab === 'outline' && proctoringPanelStatus === 'loaded';
+    const userIsAuthenticated = !!username;
```
mikix, line 57: We check `getAuthenticatedUser() !== null` elsewhere in the MFE for this - is that 100% overlap with this check? Probably...? (I know you're just moving this check.)

mikix (approved):

src/course-home/data/slice.js
```diff
@@ -11,11 +11,15 @@ const slice = createSlice({
   initialState: {
     courseStatus: 'loading',
     courseId: null,
+    proctoringPanelStatus: 'loading',
     toastBodyText: null,
     toastBodyLink: null,
     toastHeader: '',
   },
   reducers: {
+    fetchProctoringInfoResolved: (state) => {
```
cdeery, line 20: This is elegant.

src/product-tours/ProductTours.jsx
ciduarte: I couldn't find any examples of us checking whether the user is authenticated that would overlap with this check. But this is a good callout to use `getAuthenticatedUser()` here instead of `username` from the API. Was able to consolidate some code

PR #795, opened 2022-01-07
feat: show error page when xblock fails to render

mikix:
Screenshot from 2022-01-07 15-34-58

[AA-1175](https://openedx.atlassian.net/browse/AA-1175)

### DIFF
--- a/src/courseware/course/sequence/Unit.jsx
+++ b/src/courseware/course/sequence/Unit.jsx
@@ -10,7 +10,7 @@ import { useDispatch } from 'react-redux';
 import PropTypes from 'prop-types';
 import { getConfig } from '@edx/frontend-platform';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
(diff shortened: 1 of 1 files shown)


src/courseware/course/sequence/Unit.jsx
```diff
@@ -235,7 +239,15 @@ function Unit({
             scrolling="no"
             referrerPolicy="origin"
             onLoad={() => {
-              window.onmessage = function handleResetDates(e) {
+              // onLoad *should* only fire after everything in the iframe has finished its own load events.
```
cdeery, line 242: What happened to HandleResetDates here?

mikix: That was Nick code - before Matthew joined us even. And he had converted some "date resetting" message passing to be more generic, but seems to have missed this one - but this code doesn't even need a name - it's just a lambda function.

cdeery (approved):
Looks good. Thanks for the walk-through!

PR #797, opened 2022-01-11
feat: AA-1138: Adds in Weekly Goal Celebration Modal

Dillon-Dumesnil:
The logic to show the modal is controlled by the backend.
Displays the modal only in courseware the first time the learner
hits their weekly learning goal. After viewing the goal, the
database row is updated to not show the modal again.

https://openedx.atlassian.net/browse/AA-1138

Related PR: https://github.com/edx/edx-platform/pull/29747


Dillon-Dumesnil:
Desktop: image
Mobile: 
image

src/courseware/course/Course.jsx
```diff
@@ -50,6 +51,10 @@ function Course({
   const celebrationOpen = shouldCelebrateOnSectionLoad(
     courseId, sequenceId, unitId, celebrateFirstSection, dispatch, celebrations,
   );
+  // If streakLengthToCelebrate is populated, that modal takes precedence. Wait til the next load to display
+  // the weekly goal celebration modal.
+  const weeklyGoalCelebrationOpen = celebrations && !celebrations.streakLengthToCelebrate && celebrations.weeklyGoal;
```
mikix: I know we talked about this on slack - this is going to be a problem eventually  :-/  Some sort of Modal priority list or something should be made...

src/courseware/course/Course.jsx
```diff
@@ -50,6 +51,10 @@ function Course({
...
+  const daysPerWeek = courseGoals && courseGoals.selected_goal && courseGoals.selected_goal.days_per_week;
```
mikix: I'm so excited to be able to use `const daysPerWeek = courseGoals?.selected_goal?.days_per_week` -- heck we might be able to shortly/now(?) since we no longer support IE11.

src/courseware/course/celebration/WeeklyGoalCelebrationModal.jsx
```diff
@@ -0,0 +1,71 @@
...
+  const { org } = useModel('coursewareMeta', courseId);
+
+  useEffect(() => {
+    if (open) {
+      recordWeeklyGoalCelebration(org, courseId);
+    }
+  }, [open]);
+
+  return (
+    <Modal
```
mikix: Does the other modal have a test? This feels like it would fall into the "developers probably won't see this normally" case where it might be nice. But if it's crazy to test it, that's fine.

mikix (approved):

src/courseware/course/Course.jsx
Dillon-Dumesnil: Yeah, I'm going to make a ticket for us to look into and simplify that. Whether it's by the backend or frontend, something should contain the master list or priority order. 
I also just realized I inadvertently took into account the first section celebration since this modal will be viewed upon loading courseware and first section won't be seen until you finish your first section so by virtue of that, goal modal will always take precedence over first section, even if both show up in the same session.

FYI: https://openedx.atlassian.net/browse/AA-1176

src/courseware/course/Course.jsx
Dillon-Dumesnil: Wat! I didn't know that was a new feature. That will be _very_ nice to use.

mikix: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

mikix: Also... check out this crazy kid: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

I don't know which JS spec version those are in, but they seem widely supported in our target browsers. So they could probably be used (and/or babel has a transform for them).

Do... you wanna try the optional chaining here and see if it works?

mikix: https://caniuse.com/mdn-javascript_operators_optional_chaining
https://caniuse.com/mdn-javascript_operators_nullish_coalescing

mikix: Haha, OK optional chaining may be too much power:
`let result = someInterface.customMethod?.();`

To optionally call that method if it's defined.

mikix: Wait, unrelatedly - why are the variables here not camel cased? Did something not get normalized in api.js?

Dillon-Dumesnil: So it looks like I can use the `?.` if we feel comfortable with it.
As for the not camel cased, it looks like the courseware only looked for truthy and never dug into the object so they just never camel cased it. I've updated that. Course Home already camel cased it properly /shrug

Dillon-Dumesnil: Tested in Firefox, Chrome, and Safari

Dillon-Dumesnil:
Updated UI for First Section Complete modal (Note: This was done as the old modal component being used has been deprecated and this is using the currently supported component):
desktop: 
image
mobile: 
image

mikix (approved):

PR #803, opened 2022-01-14
feat: stop calling course blocks rest API and assume LS exists

mikix:
- Assume that Learning Sequences is available (waffle has been
  removed)
- Stop calling course blocks API, which provided mostly duplicated
  information now.
- Refactor a bit to avoid needing to globally know which units
  exist in sequences. That is now provided just-in-time for only
  the current sequence.
- Add /first and /last URLs that you can use instead of unit IDs
  in URL paths, in service of the above point.

AA-1040
AA-1153


src/courseware/CoursewareContainer.jsx
```diff
@@ -49,11 +50,25 @@ const checkSectionToSequenceRedirect = memoize((courseStatus, courseId, sequence
   }
 });
 
-const checkUnitToSequenceUnitRedirect = memoize((courseStatus, courseId, sequenceStatus, unit) => {
```
mikix, line 52: This method was refactored to account for us no longer having a global map of units - we only know the units for the current sequence with this PR.

So instead, I added a custom call to the course blocks for just this case - which I don't believe any internal code actually does. It's mostly used for odd ball cases, like invalid sequence ids but there is a unit or similar.

src/courseware/CoursewareContainer.jsx
```diff
@@ -126,14 +172,26 @@ class CoursewareContainer extends Component {
...
+
+    // Coerce the route ids into null here because they can be undefined, but the redux ids would be null instead.
+    if (courseId !== (routeCourseId || null) || sequenceId !== (routeSequenceId || null)) {
+      // The non-route ids are pulled from redux state - they are changed at the same time as the status variables.
+      // But the route ids are pulled directly from the route. So if the route changes, and we start a fetch above,
+      // there's a race condition where the route ids are for one course, but the status and the other ids are for a
+      // different course. Since all the logic below depends on the status variables and the route unit id, we'll wait
+      // until the ids match and thus the redux states got updated. So just bail for now.
+      return;
+    }
```
mikix, line 194: The lack of this whole thing was just a big old bug as far as I can tell! A racy issue between the fetch calls and the redirect-check calls below. Didn't matter mostly, as the kind of calls that need redirects are not typical. (like sequence-only urls). But I started using them more in this PR, since we don't often have unit information anymore.

src/courseware/CoursewareContainer.test.jsx
```diff
@@ -295,13 +310,20 @@ describe('CoursewareContainer', () => {
           await loadContainer();
           expect(global.location.href).toEqual(`http://localhost/course/${courseId}`);
         });
+      });
+    });
 
-        it('should ignore the section and unit IDs and instead to the course root', async () => {
```
mikix, line 304: This test! OK, it only barely managed to work before, because:
1. It passed in `unitTree[0][0][0]` instead of the correct `unitTree[0][0][0].id` - which meant it was requesting `/courseId/sectionId/[object object]`
2. It was relying on the buggy race condition above.

So what would happen is:
1. Incoming request for `/course/section/[object]`
2. We'd see it's a request for a known section (not the expected sequence) with a provided unit
3. So redirect to `/course/[object]` and do a lookup for the unit (in this case, a garbage unitId, but the code doesn't know that yet)
4. Because of the racy code I fixed above, the section information for the actual correct section ID is still hanging around, and we think that it's the data for `[object]`
5. Which means that the code thinks we were asked for `/course/section` (rather than the actual `/course/[object]`)
6. Since that section is empty, we redirect to `/course`

I moved this test out of this "empty section" zone. And gave it an intentional "bad unit id" of `foobar`. And that code path is now handled by knowing that `foobar` fails the `sequenceMightBeUnit` check - and we redirect to toplevel course still. But more sanely.

src/courseware/course/JumpNavMenuItem.jsx
```diff
@@ -32,22 +28,14 @@ export default function JumpNavMenuItem({
...
-    const destinationString = sequences.forEach(sequence => sequence.unitIds.forEach(unitId => {
-      const complete = dispatch(checkBlockCompletion(
-        courseId,
-        sequence.id, unitId,
-      ))
-        .then(value => value);
-      if (!complete) { return `/course/${courseId}/${sequence.id}/${unitId}`; }
-    }));
-    return destinationString || `/course/${courseId}/${sequences[0].id}/${sequences[0].unitIds[0]}`;
+    return `/course/${courseId}/${sequences[0].id}`;
```
mikix, line 35: The above is MADNESS. `forEach` always returns `undefined`. So all of that block completion stuff was pointless and we always sent users to the first sequence/unit.

I think the intent though was a poor man's "go back to the last place we were" - which is ALREADY A FEATURE. If you just hit the sequence URL directly, without a unit, we look up what the last active index was and go to that.

So this change is a slight change of feature in two ways:
- Factually, we are changing from "first unit" to "last active unit" which I think is an improvement
- Theoretically with what this code was trying to do, we're changing from "earliest incompleted unit" to "last active unit" which I think is also an improvement.

src/courseware/course/JumpNavMenuItem.test.jsx
```diff
@@ -59,11 +28,8 @@ describe('JumpNavMenuItem', () => {
       {...mockData}
     />,
   );
-  it('renders menu Item as expected with button and Text and handles clicks', () => {
+  it('renders menu Item as expected with button and Text', () => {
     expect(screen.queryAllByRole('button')).toHaveLength(1);
     expect(screen.getByText('Demo Menu Item'));
-    const navButton = screen.queryAllByRole('button')[0];
-    fireEvent.click(navButton);
-    expect(mockCheckBlock).toBeCalledTimes(14); // number of units to check on load.
```
mikix, line 67: This test is also madness. It didn't check where we sent the user. Or that we even sent them anywhere. It just checked that 14 dispatch events were sent off. Doesn't even care what sort of events they were.

Anyway. I stripped the click / event checking code here. I _could've_ replaced them with fancier checks that we actually went to the right URL. But this PR was already getting sprawling. And my change is not making the tests any worse.

src/courseware/course/celebration/utils.jsx
```diff
@@ -54,10 +53,9 @@ function shouldCelebrateOnSectionLoad(courseId, sequenceId, unitId, celebrateFir
   const {
     prevSequenceId,
     nextSequenceId,
-    nextUnitId,
   } = celebrationIds;
-  const onTargetUnit = sequenceId === nextSequenceId && (!nextUnitId || unitId === nextUnitId);
-  let shouldCelebrate = onTargetUnit && celebrateFirstSection;
+  const onTargetSequence = sequenceId === nextSequenceId;
+  let shouldCelebrate = onTargetSequence && celebrateFirstSection;
```
mikix, line 58: As far as I can tell / remember, the unit ID checking was just an additional safety measure - not actually critical to this celebration code working. What it does is, as you click through Next to the next section - it just wants to make sure that you ended up at the right place before showing the celebration. And sequence should be good enough. I don't know when you would not end up at the right unit as well.

src/courseware/course/sequence/Sequence.jsx
```diff
@@ -132,7 +133,10 @@ function Sequence({
     }
   }, [(unit || {}).id]);
 
-  if (sequenceStatus === 'loading') {
+  // If sequence might be a unit, we want to keep showing a spinner - the courseware container will redirect us when
+  // it knows which sequence to actually go to.
+  const loading = sequenceStatus === 'loading' || (sequenceStatus === 'failed' && sequenceMightBeUnit);
```
mikix, line 138: This was to work around the fact that sometimes if we just have a sequenceId that is actually a unit, there's a bit of a delay before we know what to do with them. (This is for the new call to `getSequenceForUnitDeprecated`)

src/courseware/CoursewareContainer.jsx
```diff
@@ -49,11 +50,30 @@ const checkSectionToSequenceRedirect = memoize((courseStatus, courseId, sequence
...
 
-const checkUnitToSequenceUnitRedirect = memoize((courseStatus, courseId, sequenceStatus, unit) => {
-  if (courseStatus === 'loaded' && sequenceStatus === 'failed' && unit) {
-    // If the sequence failed to load as a sequence, but it *did* load as a unit, then
-    // insert the unit's parent sequenceId into the URL.
-    history.replace(`/course/${courseId}/${unit.sequenceId}/${unit.id}`);
+const checkUnitToSequenceUnitRedirect = memoize((
+  courseStatus, courseId, sequenceStatus, sequenceMightBeUnit, sequenceId, section, routeUnitId,
+) => {
+  if (courseStatus === 'loaded' && sequenceStatus === 'failed' && !section && !routeUnitId) {
```
Dillon-Dumesnil, line 60: Maybe a docstring to better explain the pattern here? Like We are trying to catch hitting the `/course/sequence`, but sequence is not a valid sequence key, but is a valid block id.

That's verbose (maybe?) and could maybe be more clear, but y'know something like that.

EDIT: I see now that that is contained below. So maybe a comment around here of "See docstring below where this is called" or something?

src/courseware/CoursewareContainer.jsx
```diff
@@ -49,11 +50,30 @@ const checkSectionToSequenceRedirect = memoize((courseStatus, courseId, sequence
...
+    if (sequenceMightBeUnit) {
+      // If the sequence failed to load as a sequence, but it is marked as a possible unit, then we need to look up the
+      // correct parent sequence for it, and redirect there.
+      const unitId = sequenceId; // just for clarity during the rest of this method
```
Dillon-Dumesnil, line 64: 👏

src/courseware/CoursewareContainer.jsx
```diff
@@ -49,11 +50,30 @@ const checkSectionToSequenceRedirect = memoize((courseStatus, courseId, sequence
...
+          }
+        },
+        () => { // error case
+          history.replace(`/course/${courseId}`);
+        },
+      );
+    } else {
+      // Invalid sequence that isn't a unit either. Redirect up to main course.
+      history.replace(`/course/${courseId}`);
+    }
```
Dillon-Dumesnil, line 80: Is it possible to combine all of these?

src/courseware/CoursewareContainer.jsx
```diff
@@ -67,6 +87,32 @@ const checkSequenceToSequenceUnitRedirect = memoize((courseId, sequenceStatus, s
   }
 });
 
+const checkSequenceUnitMarkerToSequenceUnitRedirect = memoize((courseId, sequenceStatus, sequence, unitId) => {
```
Dillon-Dumesnil, line 96: Similarly here, including a docstring for what use case is covered here.

EDIT: I see now that that is contained below. So maybe a comment around here of "See docstring below where this is called" or something?

src/courseware/CoursewareContainer.jsx
```diff
@@ -67,6 +87,32 @@ const checkSequenceToSequenceUnitRedirect = memoize((courseId, sequenceStatus, s
...
+  if (sequenceStatus !== 'loaded' || !sequence.id) {
+    return;
+  }
+
+  const hasUnits = sequence.unitIds !== undefined && sequence.unitIds.length > 0;
```
Dillon-Dumesnil: Can you do that `?` syntax thing here? Like `sequence.?unitIds.length > 0;`

src/courseware/CoursewareContainer.jsx
```diff
@@ -67,6 +87,32 @@ const checkSequenceToSequenceUnitRedirect = memoize((courseId, sequenceStatus, s
...
+
+  if (unitId === 'first') {
+    if (hasUnits) {
+      const firstUnitId = sequence.unitIds[0];
+      history.replace(`/course/${courseId}/${sequence.id}/${firstUnitId}`);
+    } else {
+      // No units... go to general sequence page
+      history.replace(`/course/${courseId}/${sequence.id}`);
+    }
```
Dillon-Dumesnil, line 110: Similar to above if we are able to combine this with the `else` below.

src/courseware/CoursewareContainer.jsx
```diff
@@ -126,14 +172,26 @@ class CoursewareContainer extends Component {
...
+    // Check if we should save our sequence position.  Only do this when the route unit ID changes.
+    this.checkSaveSequencePosition(routeUnitId);
+
+    // Coerce the route ids into null here because they can be undefined, but the redux ids would be null instead.
+    if (courseId !== (routeCourseId || null) || sequenceId !== (routeSequenceId || null)) {
+      // The non-route ids are pulled from redux state - they are changed at the same time as the status variables.
+      // But the route ids are pulled directly from the route. So if the route changes, and we start a fetch above,
+      // there's a race condition where the route ids are for one course, but the status and the other ids are for a
+      // different course. Since all the logic below depends on the status variables and the route unit id, we'll wait
+      // until the ids match and thus the redux states got updated. So just bail for now.
```
Dillon-Dumesnil, line 192: 👏

src/courseware/course/course-exit/CourseExit.test.jsx
```diff
@@ -27,11 +28,10 @@ describe('Course Exit Pages', () => {
     user_has_passing_grade: true,
     end: '2014-02-05T05:00:00Z',
   });
-  const defaultCourseBlocks = buildSimpleCourseBlocks(defaultMetadata.id, defaultMetadata.name);
+  const { courseBlocks: defaultCourseBlocks } = buildSimpleCourseBlocks(defaultMetadata.id, defaultMetadata.name);
```
Dillon-Dumesnil, line 31: What is this change doing?

src/courseware/course/sequence/Unit.test.jsx
```diff
@@ -18,7 +18,7 @@ describe('Unit', () => {
   const unitBlocks = [
     Factory.build(
       'block',
-      { type: 'problem', graded: 'true' },
+      { type: 'vertical', graded: 'true' },
       { courseId: courseMetadata.id },
     ), Factory.build(
```
Dillon-Dumesnil, line 23: Don't love this old behavior. Making "units" that are problems. C'mon

src/courseware/course/sequence/sequence-navigation/SequenceNavigation.test.jsx
```diff
@@ -70,7 +70,7 @@ describe('Sequence Navigation', () => {
     expect(testData.onNavigate).not.toHaveBeenCalled();
     // TODO: Not sure if this is working as expected, because the `contentType="lock"` will be overridden by the value
     //  from Redux. To make this provide a `fa-icon` lock we could introduce something like `overriddenContentType`.
-    expect(unitButton.firstChild).toHaveClass('fa-edit');
+    expect(unitButton.firstChild).toHaveClass('fa-tasks');
```
Dillon-Dumesnil, line 73: I thought we don't use fontawesome anymore. I understand not what you're after in this PR, but was surprised to see this change.

src/courseware/course/sequence/sequence-navigation/hooks.js
```diff
@@ -8,9 +8,10 @@ export function useSequenceNavigationMetadata(currentSequenceId, currentUnitId)
   const sequenceIds = useSelector(sequenceIdsSelector);
   const sequence = useModel('sequences', currentSequenceId);
   const courseStatus = useSelector(state => state.courseware.courseStatus);
+  const sequenceStatus = useSelector(state => state.courseware.sequenceStatus);
 
   // If we don't know the sequence and unit yet, then assume no.
-  if (courseStatus !== 'loaded' || !currentSequenceId || !currentUnitId) {
+  if (courseStatus !== 'loaded' || sequenceStatus !== 'loaded' || !currentSequenceId || !currentUnitId) {
```
Dillon-Dumesnil: Why this change? It seemed like before it was using `!currentSequenceId` as a way of checking this situation. Do we still need both checks?

src/courseware/data/api.js
```diff
@@ -1,96 +1,8 @@
 import { getConfig, camelCaseObject } from '@edx/frontend-platform';
 import { getAuthenticatedHttpClient, getAuthenticatedUser } from '@edx/frontend-platform/auth';
-import { logInfo } from '@edx/frontend-platform/logging';
 import { getTimeOffsetMillis } from '../../course-home/data/api';
 import { appendBrowserTimezoneToUrl } from '../../utils';
 
-export function normalizeBlocks(courseId, blocks) {
```
Dillon-Dumesnil, line 7: Gotta feel good.

src/courseware/data/api.js
```diff
@@ -144,39 +64,24 @@ export function normalizeLearningSequencesData(learningSequencesData) {
...
   url.searchParams.append('course_id', courseId);
   url.searchParams.append('username', authenticatedUser ? authenticatedUser.username : '');
   url.searchParams.append('depth', 3);
-  url.searchParams.append('requested_fields', 'children,effort_activities,effort_time,show_gated_sections,graded,special_exam_info,has_scheduled_content');
+  url.searchParams.append('requested_fields', 'children');
 
   const { data } = await getAuthenticatedHttpClient().get(url.href, {});
-  return normalizeBlocks(courseId, data.blocks);
+  const parent = Object.values(data.blocks).find(block => block.type === 'sequential' && block.children.includes(unitId));
+  return parent?.id;
```
Dillon-Dumesnil, line 78: :chefkiss:

src/courseware/data/api.js
```diff
@@ -144,39 +64,24 @@ export function normalizeLearningSequencesData(learningSequencesData) {
...
-        logInfo('Learning Sequences API not enabled for this user.');
-      } else {
-        logInfo(`Unexpected error calling Learning Sequences API (${error.response.status}). Ignoring.`);
-      }
-      return null;
-    }
-    throw error;
-  }
+  const { data } = await getAuthenticatedHttpClient().get(outlineUrl.href, {});
+  return normalizeLearningSequencesData(data);
```
Dillon-Dumesnil, line 84: Are errors automatically thrown without the `catch`? Just curious if you only need the catch if you plan to do error handling.

src/courseware/data/api.js
```diff
@@ -244,7 +149,7 @@ export async function getCourseMetadata(courseId) {
   return normalizeMetadata(metadata);
 }
 
-function normalizeSequenceMetadata(sequence) {
+function normalizeSequenceMetadata(courseId, sequence) {
```
Dillon-Dumesnil: Why pass this in? It doesn't look like it is used anywhere.

src/courseware/data/thunks.js
```diff
@@ -203,11 +92,11 @@ export function fetchCourse(courseId) {
   };
 }
 
-export function fetchSequence(sequenceId) {
+export function fetchSequence(courseId, sequenceId) {
   return async (dispatch) => {
     dispatch(fetchSequenceRequest({ sequenceId }));
     try {
-      const { sequence, units } = await getSequenceMetadata(sequenceId);
+      const { sequence, units } = await getSequenceMetadata(courseId, sequenceId);
```
Dillon-Dumesnil: I think I wrote this somewhere else too, but I don't think the courseId is ever used here and can be dropped.

src/instructor-toolbar/InstructorToolbar.test.jsx
```diff
@@ -15,24 +15,27 @@ jest.mock('@edx/frontend-platform', () => ({
...
-    const store = await initializeTestStore({ excludeFetchSequence: true });
-    const { courseware, models } = store.getState();
-    mockData = {
-      courseId: courseware.courseId,
-      unitId: Object.values(models.units)[0].id,
-      canViewLegacyCourseware: true,
-    };
+    const store = await initializeTestStore();
+    courseware = store.getState().courseware;
+    models = store.getState().models;
```
Dillon-Dumesnil, line 27: Why this change? Does the deconstruction not work here?

src/setupTest.js
```diff
@@ -149,15 +150,13 @@ export async function initializeTestStore(options = {}, overrideStore = true) {
     courseBlocks, sequenceBlocks, courseMetadata, sequenceMetadata,
   } = buildSimpleCourseAndSequenceMetadata(options);
 
-  let forbiddenCourseUrl = `${getConfig().LMS_BASE_URL}/api/courseware/course/${courseMetadata.id}`;
-  forbiddenCourseUrl = appendBrowserTimezoneToUrl(forbiddenCourseUrl);
```
Dillon-Dumesnil, line 155: Any idea why this was called forbidden?

src/courseware/data/__factories__/learningSequencesOutline.factory.js
```diff
@@ -40,14 +40,14 @@ export function buildOutlineFromBlocks(courseBlocks) {
     } else if (block.type === 'chapter') {
       sections[block.id] = {
```
Dillon-Dumesnil, line 45: Is this function intended to stay around long term do you think? Or would we want to one day refactor this to not pretend to build from courseBlocks anymore? If the latter, might be worth a comment in this file and the course blocks factory to call that out.

Dillon-Dumesnil (commented):
Not approving just because of all of the questions, but generally, I do not think I really had any blocking feedback. Nice job. Lots of complex stuff built into this

src/courseware/course/sequence/sequence-navigation/SequenceNavigation.test.jsx
mikix: This is coming from older (initial MFE v1) code in UnitIcon.jsx in the sequence navigation bar. It looks at the kind of content in a unit - and since I was switching them above, it changed the icon here.

But you are right that ideally it would use fontawesome - we could probably switch to paragon icons at some point. We should do a sweep of the MFE for this stuff.

src/courseware/CoursewareContainer.jsx
mikix: Oddly, I didn't think so? So two of these are from different callbacks given to resolve the promise from `getSequenceForUnitDeprecated`. So they can't really share code. And the last one is outside of that promise, so can't share code. I agree it looks redundant, but :shrug:

I guess I could move the "replace('/course')" bit into a function like "goToTopOfCourse" or something. But not sure if that's any more readable?

src/courseware/CoursewareContainer.jsx
mikix: I struggled with this too. It's nested if's - and I don't want to change anything if neither of the outer if conditions hit. So.. I dunno. I guess I could set a variable here, and then at end of function check that like `if (goToSequence) history.replace...`? But like the previous example, I favored the "simple" but verbose way. Open to refactoring slightly though, if you think that would be more clear.

src/courseware/course/course-exit/CourseExit.test.jsx
mikix: GREAT QUESTION

So `buildSimpleCourseBlocks` returns lots of different things - `courseBlocks` (the expected return for the course blocks endpoint) is just ONE of them. Look at other test files where we call it - they all do something like the change I made here.

So before.... we were returning completely non-normal (i.e. "garbage") data from the course blocks endpoint in these tests. IN FACT... I even took out the mocks for these endpoints and while I see a bunch of failed attempts to access them in the logs, all the tests in this file pass still....

But if I remove the line in the tests that fetches those endpoints, tests fail... So these tests want to make the query, but don't seem to care what the result is? I dunno. I didn't dig into it further. I just fixed this line to at least make some sort of sense. Just a drive-by fix, unrelated to the rest of the PR.

src/courseware/course/sequence/sequence-navigation/hooks.js
mikix: `!currentSequenceId` is different - that's asking "is there a sequence ID in the url?" vs checking the sequence status - which is asking "did the sequence metadata come through?"

The change was made because now the unit ids for the sequence come through the sequence metadata call exclusively, so we need to make sure that call succeeded now.

src/courseware/data/api.js
mikix: Errors will bubble up and the promise for the endpoint will return as a failure, then thunks.js will notice that and flag the fetch as failed and notify newrelic. (That's the more standard approach - catching inside of api.js is mostly for catching errors we expect and ignoring them, or re-throwing to the promise)

src/courseware/data/api.js
mikix: Ugh, great point. At an earlier version of this PR I did use it. Will drop.

src/instructor-toolbar/InstructorToolbar.test.jsx
mikix: I wanted to pull them out to use them in other methods (the `beforeEach` call below). So I added a `let` definition above, and I don't know how to set a `let` definition from a deconstruction? Not sure you can.

As for why I was moving the `mockData` definition... I don't remember if it actually had any relation to my changes. But I did note that some tests edit `mockData`! But it was being defined just once for all the tests. So the tests were leaving side effects. Wasn't affecting us yet, but I was doing some drive-by cleanup. Maybe driven by my other changes, I don't remember.

src/setupTest.js
mikix: No, that didn't make any sense to me. I suspect a copy/paste issue from another test that calls it forbidden because it returned a 403 or something. But doesn't make sense in this method...

src/courseware/data/__factories__/learningSequencesOutline.factory.js
mikix: Can do

mikix:
OK, updated with your recommendations (mostly adding comments and dropping the unnecessary `courseId` argument to `fetchSequences`).

If I were a better person, I would have done them as separate commits, but my muscle memory fingers instead added them to the previous commit. Sorry

src/courseware/CoursewareContainer.jsx
Dillon-Dumesnil: Yeah, no, I don't think making a function for it would help. I guess I was thinking of it being some sort of if the success case, if `parentId` doesn't exist, you exit the promise and ignore the error case completely and exit the promise so then they all end up outside of the promise and doing the one `history.replace` there. I guess that would also require needing to know if you already changed pages or not. This also might not make any sense because of how promises work, but this was my rough thinking and you can let me know if it is crazy:
```
    if (sequenceMightBeUnit) {
      // If the sequence failed to load as a sequence, but it is marked as a possible unit, then we need to look up the
      // correct parent sequence for it, and redirect there.
      const unitId = sequenceId; // just for clarity during the rest of this method
      getSequenceForUnitDeprecated(courseId, unitId).then(
        parentId => {
          if (parentId) {
            history.replace(`/course/${courseId}/${parentId}/${unitId}`);
            return;  // ideally this would return ll the way out of the parent as well and not sure that's possible, but this general idea
          }
        },
      );
    } 
    // Invalid sequence that isn't a unit either. Redirect up to main course.
    history.replace(`/course/${courseId}`);
```
EDIT: Also saw your comment below about this way being simple, but more verbose. And I do completely agree with that. It definitely is easy to follow and that is appreciated so I am fully happy if you decide to keep as is.

src/courseware/CoursewareContainer.jsx
```diff
@@ -197,32 +267,19 @@ class CoursewareContainer extends Component {
...
     if (nextSequence !== null) {
-      let nextUnitId = null;
-      if (nextSequence.unitIds.length > 0) {
-        [nextUnitId] = nextSequence.unitIds;
-        history.push(`/course/${courseId}/${nextSequence.id}/${nextUnitId}`);
-      } else {
-        // Some sequences have no units.  This will show a blank page with prev/next buttons.
-        history.push(`/course/${courseId}/${nextSequence.id}`);
-      }
+      history.push(`/course/${courseId}/${nextSequence.id}/first`);
```
Dillon-Dumesnil, line 270: Out of curiosity, will this code properly handle the case where this is put directly into a URL? I would think yes, because it would just match `.../:unitId` with `.../first`.

Separate consideration, you are doing a `history.push` here. IIRC, that adds it so if you hit back in your browser, this is considered one of the pages you have visited. Do we want that to be the case or should this be dropped from browser history since we are treating it as more of a redirect?

Dillon-Dumesnil (commented):
Open question about the /first and /last endpoints, but I think the rest looks good to me

Dillon-Dumesnil (approved):

src/courseware/CoursewareContainer.jsx
mikix: Tested - yes, you can manually enter `/first` and the back/forward browser actions work as expected, because while we push `/first`, it replaces itself with the final answer.

PR #808, opened 2022-01-21
fix: [AA-1018] api refactor

cdeery:
This is the first step toward clearing out the redundant metadata from the
coursewareMetadata and getting it from a common source - the courseHomeMetadata.

remove username from coursewareMetadata
Remove courseAccess from coursewareMetadata.
courseware now uses courseHomeMetadata for username and courseAccess.

Fix all unit tests
Modify classes that use metadataModel to use courseHomeMetadata for common data.
metadataModel still exists as a mechanism to distinguish if a component is under
courseware or courseHome, and it will be renamed or removed in a later refactor.


src/courseware/course/sequence/lock-paywall/LockPaywall.test.jsx
```diff
@@ -11,7 +11,7 @@ jest.mock('@edx/frontend-platform/analytics');
 
 describe('Lock Paywall', () => {
   let store;
-  const mockData = { metadataModel: 'coursewareMeta' };
+  const mockData = { notificationTrayVisible: false };
```
cdeery, line 14: metadataModel was not needed, but the test was complaining about notificationTrayVisible, hence the switch

src/courseware/course/sequence/sequence-navigation/hooks.js
```diff
@@ -2,15 +2,16 @@
...
 
 export function useSequenceNavigationMetadata(currentSequenceId, currentUnitId) {
   const sequenceIds = useSelector(sequenceIdsSelector);
   const sequence = useModel('sequences', currentSequenceId);
   const courseStatus = useSelector(state => state.courseware.courseStatus);
+  const sequenceStatus = useSelector(state => state.courseware.sequenceStatus);
 
   // If we don't know the sequence and unit yet, then assume no.
-  if (courseStatus !== 'loaded' || !currentSequenceId || !currentUnitId) {
+  if (courseStatus !== 'loaded' || sequenceStatus !== 'loaded' || !currentSequenceId || !currentUnitId) {
```
cdeery, line 14: Recommended change by @mikix to match a similar change he made elsewhere, and to resolve a testing issue.

src/courseware/data/thunks.js
```diff
@@ -130,14 +131,34 @@ export function fetchCourse(courseId) {
       getCourseMetadata(courseId),
       getCourseBlocks(courseId),
       getLearningSequencesOutline(courseId),
-    ]).then(([courseMetadataResult, courseBlocksResult, learningSequencesOutlineResult]) => {
+      getCourseHomeCourseMetadata(courseId),
```
cdeery, line 134: This is the meat of the changes - get the courseHomeMetadata even when loading courseware in order to be able to access the common metadata.

src/setupTest.js
```diff
@@ -27,9 +27,11 @@ import { appendBrowserTimezoneToUrl, executeThunk } from './utils';
 import buildSimpleCourseAndSequenceMetadata from './courseware/data/__factories__/sequenceMetadata.factory';
 
 class MockLoggingService {
-  logInfo = jest.fn();
+  // eslint-disable-next-line no-console
```
cdeery, line 30: Mocking the logger causes the tests to swallow any errors that are logged in the run. Is this reasonable to leave in?

src/shared/data/__factories__/courseMetadataBase.factory.js
```diff
@@ -5,9 +5,9 @@ import { Factory } from 'rosie'; // eslint-disable-line import/no-extraneous-dep
 import './tab.factory';
 
 export default new Factory()
-  .sequence('id', (i) => `course-v1:edX+DemoX+Demo_Course_${i}`)
   .option('host')
   .attrs({
+    id: 'course-v1:edX+DemoX+Demo_Course_1',
```
cdeery: Since coursewareMetadata and courseHomeMetadata both need to be created in the same tests, autogenerated ids require passing in ids, and complicated resolutions. There were only two tests that depended upon this functionality.

src/courseware/data/api.js
```diff
@@ -210,7 +210,6 @@ function normalizeMetadata(metadata) {
     start: data.start,
     enrollmentMode: data.enrollment.mode,
     isEnrolled: data.enrollment.is_active,
-    courseAccess: camelCaseObject(data.course_access),
```
cdeery, line 213: courseAccess and username have been removed from coursewareMetadata, and can be removed from the courseware API

src/courseware/CoursewareContainer.test.jsx
```diff
@@ -101,6 +102,7 @@ describe('CoursewareContainer', () => {
     // If we weren't given course blocks or metadata, use the defaults.
     const courseBlocks = options.courseBlocks || defaultCourseBlocks;
     const courseMetadata = options.courseMetadata || defaultCourseMetadata;
+    const courseHomeMetadata = options.courseHomeMetadata || defaultCourseHomeMetadata;
```
mikix, line 105: A thought tied to this naming, just thrown on this random line:

Historically, the MFE was divided into two parts: the courseware tab and "all the other tabs" which we distinguished as "courseware" vs "course home". As we tear down that dividing wall, the name "courseware" still makes sense for that one tab. But "course home" will lose most of its meaning and will start just being another jargon word for "the whole learning MFE" that might confuse folks down the line (and/or folks that remember it being a distinct thing).

Something we may want to consider as we clean up. No need to change anything here yet - just noting the mental discomfort I feel seeing "course home" inside the courseware tab.

Separately, seeing two different "metadatas" together is also a little pause-giving. I know this is step 1 of a larger refactor, and that kind of thing is on your brain already. Hmm, also the code folder layout will start to feel weird... Anyway! Not this PR's problem.

src/courseware/course/sequence/honor-code/HonorCode.jsx
```diff
@@ -12,7 +12,7 @@ import messages from './messages';
 
 function HonorCode({ intl, courseId }) {
   const dispatch = useDispatch();
-  const coursewareMetaData = useModel('coursewareMeta', courseId);
+  const coursewareMetaData = useModel('courseHomeMeta', courseId);
```
mikix: nit: might be worth changing the variable name too - knowing what bits of metadata are coming from where is already confusing enough.

src/courseware/course/sequence/honor-code/HonorCode.test.jsx
```diff
@@ -1,14 +1,15 @@
...
+import MockAdapter from 'axios-mock-adapter';
 import { Factory } from 'rosie';
 
 import {
-  authenticatedUser, fireEvent, initializeTestStore, render, screen, waitFor,
+  authenticatedUser, fireEvent, initializeMockApp, initializeTestStore, render, screen, waitFor,
 } from '../../../../setupTest';
 import HonorCode from './HonorCode';
 
+initializeMockApp();
```
mikix, line 12: Do you remember why this was needed? Not questioning it - it's a line a lot of tests have. Just trying to wrap my head around the tendrils of this PR's changes.

src/courseware/data/thunks.js
```diff
@@ -130,11 +131,29 @@ export function fetchCourse(courseId) {
...
+      learningSequencesOutlineResult,
+      courseHomeMetadataResult]) => {
       if (courseMetadataResult.status === 'fulfilled') {
         dispatch(addModel({
           modelType: 'coursewareMeta',
-          model: courseMetadataResult.value,
+          model: {
+            id: courseId,
+            ...courseMetadataResult.value,
+          },
```
mikix: Why this change? It has the `id` already yeah?

src/courseware/data/thunks.js
```diff
@@ -185,8 +205,8 @@ export function fetchCourse(courseId) {
         logError(courseMetadataResult.reason);
```
mikix, line 202: You probably also want a similar `logError` call for the new course home metadata result - this will send a ping to new relic (or whatever your installation uses for logging) to record the error.

src/courseware/data/thunks.js
```diff
@@ -130,11 +131,29 @@ export function fetchCourse(courseId) {
...
           modelType: 'coursewareMeta',
-          model: courseMetadataResult.value,
+          model: {
+            id: courseId,
+            ...courseMetadataResult.value,
+          },
+        }));
+      }
+
+      if (courseHomeMetadataResult.status === 'fulfilled') {
```
mikix, line 147: I don't know why the `fetchedXXX` variables aren't higher in this method - seems like they could be used here. You don't need to fix, but just odd.

src/index.jsx
```diff
@@ -66,7 +66,7 @@ subscribe(APP_READY, () => {
                 )}
               />
               <PageRoute path="/course/:courseId/course-end">
-                <TabContainer tab="courseware" fetch={fetchCourse} slice="courseware">
+                <TabContainer tab="courseware" fetch={fetchCourse} slice="courseHome">
```
mikix: This was the only one still using `courseware` yeah? Might be better just to remove the argument and switch the `TabContainer` to just hardcode `courseHomeMetadata`.

src/tab-page/LoadedTabPage.jsx
```diff
@@ -30,7 +30,7 @@ function LoadedTabPage({
     tabs,
     title,
     verifiedMode,
-  } = useModel(metadataModel, courseId);
+  } = useModel('courseHomeMeta', courseId);
```
mikix, line 33: Elsewhere in this PR, I see references to how `courseAccess` and `username` have moved. Like removing them from the factories etc. But here, we are also stopping using a lot more fields too - unless, are there other places we pull `tabs`, `title`, etc from the courseware metadata call still?

src/tab-page/LoadedTabPage.test.jsx
```diff
@@ -25,7 +25,20 @@ describe('Loaded Tab Page', () => {
 
   it('shows Instructor Toolbar if original user is staff', async () => {
     const courseMetadata = Factory.build('courseMetadata', { original_user_is_staff: true });
-    const testStore = await initializeTestStore({ courseMetadata, excludeFetchSequence: true }, false);
+    const courseHomeMetadata = Factory.build('courseHomeMetadata', {
+      courseId: courseMetadata.id,
+      // need to synchronize the id with the courseMetadata because it is autoincremented by courseMetadataBase
+      id: courseMetadata.id,
```
mikix: OK, this caused me to dig into what's happening here - and I'm confused. Which might have also caused you some confusion. Hence code like this where both values are set, neither of which matters, I think(!).

So these factory calls build a raw dictionary that is used as the mock response for the backend endpoint. Thus, its values should be snake case and match what the endpoint actually sends us. This raw dictionary will then be passed through normalization and thunks and all that.

So normally, that would mean you'd want to set `course_id` here, not `id` or `courseId`. Because that's what the endpoint actually gives. (Though I don't know who wins a race if both `course_id` and `courseId` are already set when camel casing / normalizing). The fact that the course home endpoint uses `course_id` instead of `id` (like courseware does) is a quirk that I am not sure was well thought out, but :shrug:.

HOWEVER. When looking at the code, I see that the courseHomeMetadataFactory inherits the incrementing `id` value from `courseMetadataBase` and does not set `course_id` at all - which is wrong because that is now an inaccurate mock - that's not what the LMS gives us. It gives `course_id`.

HOWEVER, that kind of doesn't matter in practice? Because in `course_home/thunks.js` (and in your new thunks code above), we just set `id: courseId` on the redux model, where `courseId` comes from the URL, not the backend. And then the model ends up with both `id: courseIdFromUrl` (from thunks) and `courseId: courseIdFromBackend` (from normalizing backend response). And then all the code probably just references `id` because that's what is the shared model value between courseware and course home sides of the code.

So ultimately, the factory value for `courseHomeMetadataFactory`'s incrementing `id` value is ignored. And setting it here doesn't matter at all, I don't think? But you said you did hit some issue with the id value incrementing - but if my theory is correct, it's not used at all. So maybe I'm wrong and something does pull it...

What probably should happen is that the `normalizeCourseHomeMetadata` call should convert `course_id` into `id` like the rest of the code expects. That's the whole point of the normalizing step - to take the hodgepodge of backend phrasing and switch it to unified MFE phrasing. (Which is a bit ironic in this case, since we made both these backends for the MFE - but normalizing should still be normalizing, even if we're just getting in our own way sometimes).

Then `thunks` can drop its custom setting of `id: courseId` and the value from the factory would actually matter? And then the factory would also need to be updated to not use `id` but rather `course_id` - because again, it's a mock and should ideally be a faithful one.

As for what you should do here... I'd say test my above theory / trace code yourself to feel confident about it. Then... maybe just let sleeping dogs lie for now and we can bite off that cleanup separately. Unless you're feeling bullish. If you do let it lie for now, you can probably drop all these factory id overrides - and also don't need to stop the incrementing maybe? Because none of this ends up getting through to the redux model, right? Or am I just going crazy and have talked myself into a metadata pretzel.

src/shared/streak-celebration/StreakCelebrationModal.test.jsx
```diff
@@ -23,6 +23,7 @@ describe('Loaded Tab Page', () => {
   let axiosMock;
   const calculateUrl = `${getConfig().ECOMMERCE_BASE_URL}/api/v2/baskets/calculate/?code=ZGY11119949&sku=8CF08E5&username=testuser`;
   const courseMetadata = Factory.build('courseMetadata', { celebrations: { streak_length_to_celebrate: 3 } });
+  const courseHomeMetadata = Factory.build('courseHomeMetadata', { id: courseMetadata.id, courseId: courseMetadata.courseId });
```
mikix: Again, I think you can just pass in `course_id: courseMetadata.id` instead of both those. In fact, `courseMetadata.courseId` is just `undefined` here, I think?

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -155,7 +155,7 @@ function StreakModal({
       title={title}
       onClose={() => {
         closeStreakCelebration();
-        recordModalClosing(metadataModel, celebrations, org, courseId, dispatch);
+        recordModalClosing('courseHomeMeta', celebrations, org, courseId, dispatch);
```
mikix: Can we just change `recordModalClosing` to not take the argument at all and hardcode `courseHomeMetadata`?

src/shared/data/__factories__/courseMetadataBase.factory.js
mikix: I am still lukewarm on this change - I kind of like the extra protections against side effects or mismatched metadata from incrementing the id. Plus, it seems like you ended up manually synchronizing the metadata factory calls anyway, by passing in `course_id` for the course home metadata. And that was the advantage of this change I thought - less busywork writing tests. But not a hill I care about defending. It's a mild change.

Though - you might as well drop the `_1` at the end of the id. Feels out of place without any incrementing happening.

src/product-tours/ProductTours.test.jsx
```diff
@@ -275,9 +281,8 @@ describe('Courseware Tour', () => {
         axiosMock.onGet(proctoredExamApiUrl).reply(200, { exam: {}, active_attempt: {} });
       });
 
-      axiosMock.onPost(`${courseId}/xblock/${defaultSequenceBlock.id}/handler/get_completion`).reply(200, {
-        complete: true,
-      });
+      const blockUrlRegExp = new RegExp(`${getConfig().LMS_BASE_URL}/courses/${courseId}/xblock/*`);
+      axiosMock.onPost(blockUrlRegExp).reply(200, { complete: true });
```
mikix, line 287: This change is surprising - both that it exists and that it doesn't break anything. Is it from a bad merge? Or is it actually related to the metadata cleanup?

src/setupTest.js
mikix: Sure? Seems reasonable.

src/setupTest.js
```diff
@@ -27,9 +27,11 @@ import { appendBrowserTimezoneToUrl, executeThunk } from './utils';
...
+  logInfo = jest.fn(infoString => console.log(infoString));
 
-  logError = jest.fn();
+  // eslint-disable-next-line no-console
+  logError = jest.fn(infoString => console.log(infoString));
```
mikix: tiniest of nits: maybe `errorString`?

src/setupTest.js
```diff
@@ -146,16 +148,19 @@ export async function initializeTestStore(options = {}, overrideStore = true) {
...
-    courseBlocks, sequenceBlocks, courseMetadata, sequenceMetadata,
+    courseBlocks, sequenceBlocks, courseMetadata, sequenceMetadata, courseHomeMetadata,
   } = buildSimpleCourseAndSequenceMetadata(options);
 
   let forbiddenCourseUrl = `${getConfig().LMS_BASE_URL}/api/courseware/course/${courseMetadata.id}`;
   forbiddenCourseUrl = appendBrowserTimezoneToUrl(forbiddenCourseUrl);
 
   const courseBlocksUrlRegExp = new RegExp(`${getConfig().LMS_BASE_URL}/api/courses/v2/blocks/*`);
   const learningSequencesUrlRegExp = new RegExp(`${getConfig().LMS_BASE_URL}/api/learning_sequences/v1/course_outline/*`);
+  let courseHomeMetadataUrl = `${getConfig().LMS_BASE_URL}/api/course_home/course_metadata/${courseMetadata.id}`;
```
mikix, line 159: Another place where we are making it not matter what the course home metadata factory's `id`/`courseId`/`course_id` value is, since we pull this `id` from the courseware metadata.

src/courseware/CoursewareContainer.test.jsx
cdeery: Excellent point. I wanted to keep the names during the initial cleanup to avoid further confusion, but we should definitely work on what would be good meaningful names.

src/courseware/course/sequence/honor-code/HonorCode.test.jsx
cdeery: I think I wanted to get the MockLoggingService so I could see some error messages that were being hidden

src/courseware/data/thunks.js
cdeery: I wanted to ask you about that. course-home/thunks specifically sets the id as the courseId, but courseware didn't.
I think I made this change when I was trying to reconcile the id/courseId disparity.
in production, I think we always set id and courseId to the same value, but in testing, we use the same courseId, and sometimes no courseId, and use different id's to distinguish test courses.
Also, I don't see id in either of the APIs. Is it really a synthetic value just for testing?

src/courseware/course/sequence/honor-code/HonorCode.jsx
cdeery: Whoops - will do. Don't need to add more confusion!

src/product-tours/ProductTours.test.jsx
cdeery: There were a bunch of calls just 404-ing in the log. They were previously invisible because the log was stubbed out with a mock. I was trying to clean the log to get noise out so I could debug my actual tests problems. You can see that the original mock just returns a minimal response. I admit it's funky. Let me see if I can find a better solution.

mikix: That's probably fine then. :shrug:

src/index.jsx
cdeery: There is a comment in TabContainer that looks like it was written by you:
```
  // The courseId from the store is the course we HAVE loaded.  If the URL changes,
  // we don't want the application to adjust to it until it has actually loaded the new data.
  const {
    courseId,
    courseStatus,
  } = useSelector(state => state[slice]);
```
That made me think it would be a  BadThing™ if I just pulled it from the store. Is that no longer the case?

src/tab-page/LoadedTabPage.jsx
cdeery: Yes. The values will now be pulled from course-home, but I have not removed the fields from the data model first. It's a stepwise change. I will be pulling them soon.

src/tab-page/LoadedTabPage.test.jsx
cdeery: It's messy. There is also the problem that the coursemetadataBase.factory uses the id in the tabs, which is common to both APIs. So the id has to be in the metadata at factory generation time.

src/shared/streak-celebration/StreakCelebrationModal.test.jsx
cdeery: both attributes are now extraneous, and are going away.

src/shared/streak-celebration/StreakCelebrationModal.jsx
cdeery: Why yes, yes we can.

src/index.jsx
mikix: I'm just talking about replacing all instances of `slice` with `"courseHome"` - not changing any functionality.

All instances of `<TabContainer>` use `slice="courseHome"`, afaict. So I just figured why not drop the complexity while we're here and remove the prop altogether.

cdeery: Yes, I see. Done.

src/courseware/data/thunks.js
cdeery: Put that thing back where I found it. Thanks for helping me.

src/courseware/data/__factories__/courseMetadata.factory.js
```diff
@@ -63,5 +63,4 @@ Factory.define('courseMetadata')
     related_programs: null,
     user_needs_integrity_signature: false,
     recommendations: null,
-    username: 'testuser',
```
Dillon-Dumesnil: Should course_access be removed above as well?

src/courseware/data/redux.test.js
```diff
@@ -106,12 +116,13 @@ describe('Data layer integration tests', () => {
       expect(state.courseware.sequenceId).toEqual(null);
 
       // check that at least one key camel cased, thus course data normalized
-      expect(state.models.coursewareMeta[courseId].courseAccess).not.toBeUndefined();
+      expect(state.models.coursewareMeta[courseId].verifiedMode).not.toBeUndefined();
```
Dillon-Dumesnil, line 119: Such an interesting test line. Lol

src/courseware/data/thunks.js
```diff
@@ -1,13 +1,14 @@
...
 import {
   getBlockCompletion,
-  postSequencePosition,
-  getCourseMetadata,
   getCourseBlocks,
+  getCourseMetadata,
   getLearningSequencesOutline,
   getSequenceMetadata,
   postIntegritySignature,
+  postSequencePosition,
```
Dillon-Dumesnil, line 9: 👏

src/tab-page/LoadedTabPage.test.jsx
```diff
@@ -25,7 +25,15 @@ describe('Loaded Tab Page', () => {
 
   it('shows Instructor Toolbar if original user is staff', async () => {
     const courseMetadata = Factory.build('courseMetadata', { original_user_is_staff: true });
```
Dillon-Dumesnil: Should `original_user_is_staff` be removed from this call now that it is included in courseHomeMetadata?

src/course-home/data/__factories__/courseHomeMetadata.factory.js
```diff
@@ -10,6 +10,7 @@ Factory.define('courseHomeMetadata')
     is_self_paced: false,
     is_enrolled: false,
     can_load_courseware: false,
+    celebrations: null,
```
Dillon-Dumesnil, line 13: This is coming into (or is already in?) the CourseHomeMetadata call? I thought this information was more tab specific. 🤷‍♂️

Dillon-Dumesnil (commented):
Just a few minor comments. Nice work

src/courseware/data/__factories__/courseMetadata.factory.js
cdeery: indeed!

src/course-home/data/__factories__/courseHomeMetadata.factory.js
cdeery: It was in both API's, and came along as part of getting the metadata from courseHomeMeta. just another redundant API field, I believe.

cdeery: I think the celebrations are course-specific. The same data appears to be in the coursewareMetadata and the courseHomeMetadata. Adding a null default here just avoids an issue with accessing an undefined field

src/courseware/course/sequence/sequence-navigation/UnitNavigationEffortEstimate.jsx
```diff
@@ -24,7 +24,7 @@ function UnitNavigationEffortEstimate({
   const sequence = useModel('sequences', sequenceId);
   const nextSequence = useModel('sequences', nextSequenceId);
 
-  if (!sequence || !nextSequence) {
+  if (!sequence || Object.keys(sequence).length === 0 || !nextSequence || Object.keys(nextSequence).length === 0) {
```
cdeery, line 27: This code was counting on a bug in useModel that returned undefined if the model existed, but the ID didn't.
Fixing the bug broke the component.

src/generic/model-store/hooks.js
```diff
@@ -1,16 +1,19 @@
 import { useSelector, shallowEqual } from 'react-redux';
 
+/*
```
cdeery, line 3: This code was returning undefined if the id was nonexistent, and an empty object if the model didn't exist, and there was code that depended on each behavior. Added a comment to set expectations about the behavior, and normalized the code.

Dillon-Dumesnil (approved):
Looks good to me. I'd say to just let MT take a review before merging in as well just to get another set of eyes on this to be safe

src/courseware/course/sequence/sequence-navigation/UnitNavigationEffortEstimate.jsx
mikix: That was me! It might make more sense now (and be similar to how other components do this), to change this file to look more like the following:
```
  const {
    effortActivities,
    unitIds,
  } = useModel('sequences', sequenceId);
  const {
    effortActivities: nextEffortActivities,
    unitIds: nextUnitIds,
  } = useModel('sequences', nextSequenceId);
```

Then maybe move up the effortActivities check, which would now be redundant with the check you modified. But also, leaving it as you have it is fine.

mikix: Ugh, maybe not worth it - just noticed that we pass the entire `nextSequence` down to `EffortEstimate`... so you'd want to pull that whole thing anyway.  :shrug:  Your code is perfect.

src/index.jsx
```diff
@@ -66,7 +66,7 @@ subscribe(APP_READY, () => {
                 )}
               />
               <PageRoute path="/course/:courseId/course-end">
-                <TabContainer tab="courseware" fetch={fetchCourse} slice="courseware">
+                <TabContainer tab="courseware" fetch={fetchCourse}>
```
mikix: This you might want to revert? Probably from an earlier version that dropped that prop, but I think we went back to using it again?

cdeery: Yes, great catch! Done

mikix (commented):
Love it

PR #812, opened 2022-01-31
fix: assume that dates & outline legacy tabs are removed

mikix:
It's harmless, but does mean a pointless redirect for the user.

[AA-799](https://openedx.atlassian.net/browse/AA-799)

### DIFF
--- a/src/course-home/data/api.js
+++ b/src/course-home/data/api.js
@@ -208,10 +208,6 @@ export async function getDatesTabData(courseId) {
     return camelCaseObject(data);
   } catch (error) {
     const { httpErrorStatus } = error && error.customAttributes;
-    if (httpErrorStatus === 404) {
-      global.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/dates`);
-      return {};
-    }
     if (httpErrorStatus === 401) {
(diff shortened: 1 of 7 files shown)


src/course-home/data/api.js
```diff
@@ -201,10 +201,6 @@ export async function getDatesTabData(courseId) {
     return camelCaseObject(data);
   } catch (error) {
     const { httpErrorStatus } = error && error.customAttributes;
-    if (httpErrorStatus === 404) {
```
cdeery, line 211: What was this doing? If the API 404s, then bail on getting the data and just go to the legacy dates page?

mikix: This was the error code the backend would give if the legacy view was enabled. So it would redirect there. But now, not needed!

cdeery: Ah! the ever delightful error code as message. Got it, thanks.

cdeery (approved):
Looks good.

mikix:
@cdeery I updated this to remove traces of **both** the legacy dates and outline tabs (was previously just the dates tab). Might be worth a quick re-review.

This should only land after both https://github.com/openedx/edx-platform/pull/29846 and https://github.com/openedx/edx-platform/pull/30064 land.

PR #814, opened 2022-02-04
chore: bump frontend-lib-special-exams to 1.15.2

bseverino:
https://github.com/edx/frontend-lib-special-exams/pull/57

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4060,9 +4060,9 @@
       }
     },
     "@edx/frontend-lib-special-exams": {
-      "version": "1.15.1",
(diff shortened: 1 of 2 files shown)


ashultz0 (approved):
change is a bump and looks good but there is that weird behavior from codecov

Dillon-Dumesnil:
Hi Bianca, I'm not sure what's going on with the codecov, but I'm going to try and rerun the job just to see what happens. If that doesn't do much, perhaps we can then try and have you rebase your changes and try again. If both of those don't work... well we can look deeper then.

bseverino:
Thanks for looking into it! Rebasing didn't work, not sure what was going on but I just nuked it and started over.

PR #827, opened 2022-02-16
asadiqbal08/Added REQUIRE_AUTHENTICATED_USER configuration flag

asadiqbal08:
Signing into MITx Online doesn't automatically sign you in to edX. MitxOnline is utilizing the Learning MFE layer in between.

After looking around the Learning MFE code. It seems, In order to fix this,

We need to accommodate the toggle for [requireAuthenticatedUser](https://github.com/openedx/frontend-platform/blob/master/src/initialize.js#L194-L196) from the environment configuration, which states that:

>> If true, turns on automatic login redirection for unauthenticated users. Defaults to false, meaning that by default the application will allow anonymous/unauthenticated sessions.


openedx-webhooks:
Thanks for the pull request, @asadiqbal08! I've created [OSPR-6463](https://openedx.atlassian.net/browse/OSPR-6463) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


Please let us know once your PR is ready for our review and all tests are green.

natabene:
@asadiqbal08 Thank you for your contribution. Is this ready for our review?

asadiqbal08:
Yes.

davidjoy:
@mikix (or team) Would you be able to sanity check this?  My understanding is that there are public/unauthenticated pages in this MFE, which is why we don't have that flag turned on.  To me it feels inappropriate to give the build process the option of turning it on and off if it's going to mean certain parts of the MFE just start breaking/won't be available.

mikix:
> My understanding is that there are public/unauthenticated pages in this MFE, which is why we don't have that flag turned on.

Yes exactly. All courses have the outline page as public in some form. Most of the time (for "private" courses), it's just a page that says "hey, go login or register to see more". But some courses can opt into a public outline page that shows the course outline (very rare, but possible). Or even a fully public course, where the courseware and everything is public for unauthenticated users (also rare, but we had a covid course like that recently).

> To me it feels inappropriate to give the build process the option of turning it on and off if it's going to mean certain parts of the MFE just start breaking/won't be available.

Yeah, I'd be leery of this - it's basically disabling bits of the learning MFE user experience.

asadiqbal08:
So we do not have any options to make it configurable in MFE ?

mikix:
> So we do not have any options to make it configurable in MFE ?

Not as written/expected today. Running in this configuration means that (A) you are silently (and maybe unexpectedly?) removing `public-outline` and `public` course support and (B) the MFE currently has a nice little "unauthenticated user" landing page on the course outline, which I guess you could safely skip by redirecting to the login page instead, but it's just not the expected UX.

I may not fully understand the original issue that caused this solution. There might be a different solution that doesn't change how the MFE expects to operate.

Can you expand on the problem you're trying to solve of `Signing into MITx Online doesn't automatically sign you in to edX`. Does the login button at the top of the MFE not send you to the login page just like `REQUIRE_AUTHENTICATED_USER` does?

asadiqbal08:
>> Can you expand on the problem you're trying to solve of Signing into MITx Online doesn't automatically sign you in to edX. 

Let me explain, Sorry for not pointing the issue in the description earlier, We are trying to accommodate this [problem](https://github.com/mitodl/mitxonline/issues/393) and found out that, If we Configured the Open edX to support of `OAuth2 authentication` from MITx Online as described [HERE](https://github.com/mitodl/mitxonline/blob/main/docs/source/configuration/open_edx.rst#configure-open-edx-to-support-oauth2-authentication-from-mitx-online) and with auto redirection in Learning MFE to sign-in should follow the `Oauth2 authentication` mechanism of OpenEdx.

>> Does the login button at the top of the MFE not send you to the login page.

It sends but auto redirecting is not happening with default value of `REQUIRE_AUTHENTICATED_USER: false`. User has to click on the `sign in / register button` for further redirection and we are thinking, with OpenEdx `oauth2` auto redirection flow, user will be get signed in without following any extra steps.

mikix:
Just thought of another issue with requiring authentication - there is a "unsubscribe from goal reminder emails" landing page in the MFE that we don't want to require authentication for. Certainly from a UX pov, but maybe also from a legal spam pov (not sure on the regulations around unsubscribing)?

I think a change that would make me less concerned would be just affecting the pages you care about (the course tabs), rather than all of the MFE. This would avoid issues with intentionally-unauthenticated pages like the above unsubscribe landing, and future issues with new such pages. Something like swapping in [AuthenticatedPageRoute](https://github.com/openedx/frontend-platform/blob/master/src/react/AuthenticatedPageRoute.jsx) when configured to do so (or using similar logic as it does).

But anyway - doing this redirect is ONE way to solve the issue. But like, HOW is the redirect fixing it? The browser hits the LMS and it knows that you're logged into MITx Online somehow? (Does it query the MITx Online servers? There must be some token sharing happening somewhere - can we just do that more smartly ahead of time?) Can we fix the jwt cookie to also add the learning MFE domain to it? Sorry, our auth framework is not something I'm fully fluent in.

This feels like possibly a hammer where we want a scalpel.

natabene:
Closing this based on comments above.

openedx-webhooks:
@asadiqbal08 Even though your pull request wasn’t merged, please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/frontend-app-learning&entry.867055334=https://github.com/openedx/frontend-app-learning/pull/827&entry.1484655318=https://github.com/asadiqbal08&entry.752974735=2022-02-16+10:24&entry.1917517419=2022-04-01+19:20&entry.2133058324=No) so we can improve your experience in the future.

.env.development
```diff
@@ -41,3 +41,4 @@ TWITTER_HASHTAG='myedxjourney'
 TWITTER_URL='https://twitter.com/edXOnline'
 USER_INFO_COOKIE_NAME='edx-user-info'
 SESSION_COOKIE_DOMAIN='localhost'
+REQUIRE_AUTHENTICATED_USER='true'
```
asadiqbal08, line 44: It should be `false`

PR #829, opened 2022-02-17
fix: AA-1194: Update the breadcrumb link

Dillon-Dumesnil:
Before, for sections, it would link you to the chapter which would
cause a JS error since the sequence endpoint expects sequentials.
This updates to now link to the first subsection within a section when
you hit the section breadcrumb

### DIFF
--- a/src/courseware/course/CourseBreadcrumbs.jsx
+++ b/src/courseware/course/CourseBreadcrumbs.jsx
@@ -15,8 +15,7 @@ import JumpNavMenuItem from './JumpNavMenuItem';
 function CourseBreadcrumb({
   content, withSeparator, courseId, sequenceId, unitId, isStaff,
 }) {
-  const defaultContent = content.filter(destination => destination.default)[0] || { id: courseId, label: '' };
-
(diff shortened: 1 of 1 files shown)


src/courseware/course/CourseBreadcrumbs.jsx
```diff
@@ -33,7 +32,9 @@ function CourseBreadcrumb({
           ? (
             <Link
               className="text-primary-500"
-              to={`/course/${courseId}/${defaultContent.id}`}
+              to={defaultContent.sequences.length
+                ? `/course/${courseId}/${defaultContent.sequences[0].id}/${defaultContent.sequences[0].unitIds[0]}`
+                : `/course/${courseId}/${defaultContent.id}`}
```
mikix: Hmm, I know I told you to do it this way with adding the unit id - but can you test with just the sequence id? I believe that will send you to wherever you were last looking - which might be more correct for this use case. If that works fine, that might be nicer.

Dillon-Dumesnil: Ah based on `checkSequenceToSequenceUnitRedirect`, I think you are correct. I will check and update if it feels better.

mikix (approved):

PR #828, opened 2022-02-16
fix: deny proctored exam access to audit and honor enrollment tracks

schenedx:
### Description
[MST-1273](https://openedx.atlassian.net/browse/MST-1273)
This change would prevent audit and honor enrollment mode learners from accessing any exam type **except** timed exams.
Will not merge this PR until https://github.com/openedx/edx-platform/pull/29929 is deployed to production

@openedx/masters-devs-cosmonauts Please review.

PR #830, opened 2022-02-17
fix: revert "fix: deny proctored exam access to audit and honor enrollment tracks (#828)

schenedx:
### Description
This reverts commit 45d5141769bfa57ead8f9ce147cc51ee434d4141.

Found a bug in [the PR](https://github.com/openedx/frontend-app-learning/pull/828). This PR reverts that change.

@openedx/masters-devs-cosmonauts Please review quickly.

The bug was that, the [code at frontend-lib-special-exam](https://github.com/edx/frontend-lib-special-exams/blob/main/src/exam/Exam.jsx#L61), which is part of the version 1.15.4, is excluding too much. When the content type is not of an exam type, the variable examType has the value `""`. I need to only check against an array of proctored exam types.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3966,9 +3966,9 @@
       }
     },
     "@edx/frontend-lib-special-exams": {
-      "version": "1.15.4",
-      "resolved": "https://registry.npmjs.org/@edx/frontend-lib-special-exams/-/frontend-lib-special-exams-1.15.4.tgz",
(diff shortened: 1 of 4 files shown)


alangsto (approved):

michaelroytman (approved):

Dillon-Dumesnil (approved):
Sure thing

PR #833, opened 2022-02-18
Paragon updates

mikix:
A series of commits that gets us up to a pretty recent paragon release.

Where possible, I tried to remove old ways of doing things in favor of fancy new paragon ways.

- Removed our disparate uses of react-responsive and react-break in favor of paragon's reponsive support.
- Adapted to `Card` and `Spinner` changes.
- Removed our custom `Tour` component in favor of paragon's drop-in replacement `ProductTour`.

You'll note that I only get us to paragon 19.3.0, instead of 19.5.0, the most recent version. That's because starting with 19.4.0, paragon uses some css that isn't supported in IE11. And we still declare support for that browser in webpack, so we get some bogus warnings if we use newer versions of paragon. The fix for that is in https://github.com/openedx/frontend-app-learning/pull/728 - but we can't land that until we also update our version of node because node 12 doesn't have some old polyfills that we'd need as soon as we drop existing polyfills. I don't fully understand the issue, but Adam S does. The node upgrade for that is starting in https://github.com/openedx/frontend-app-learning/pull/813


Dillon-Dumesnil (approved):
Sure

PR #836, opened 2022-02-24
feat: Add Value Prop past expiration messaging for gated content

julianajlk:
[REV-2500](https://openedx.atlassian.net/browse/REV-2500).
When the expiration date passes for a learner in a course, the Value Prop messaging should display that the date has passed and they should enroll in a new course run. 

I'm dividing the above ticket into 2 PRs. This PR is for the Gated Content. Course home and courseware changes can be found [here](https://github.com/openedx/frontend-app-learning/pull/853).

Screen Shot 2022-03-03 at 6 46 37 PM

**This PR:**
- Adds the past expiration messaging for gated content banner
- Adds event tracking for link click
- Updates gated content tests

**To manually test:**
- Checkout my branch `julianajlk/REV-2500/value-prop-past-upgrade` in learning MFE directory
- Run `npm ci build` and `npm start`
- Go to Course Home http://localhost:2000/course/course-v1:edX+DemoX+Demo_Course/home
- Click on "start course" to go to Courseware page
- Go to an unit that has gated content http://localhost:2000/course/course-v1:edX+DemoX+Demo_Course/block-v1:edX+DemoX+Demo_Course+type@sequential+block@edx_introduction/block-v1:edX+DemoX+Demo_Course+type@vertical+block@vertical_0270f6de40fc
- Make sure you view as "Audit" learner

**A note on unit tests:**
I'm using the `<Hyperlink>` component for the "View Course Details" link with the icon because it has the exact styling and icon we want. However, when the tests for `LockPaywall.jsx` are run locally, it adds the alt text of "... in a new tab" but when the build is run here, it adds the alt text "... Open in a new tab". 

This means that when I use the link selector with a specific name, it will either fail locally or in prod because the name does not have the exact alt text. Example:
`screen.getByRole('link', { name: 'View Course Details Opens in a new tab' })` vs. ``screen.getByRole('link', { name: 'View Course Details in a new tab' })``

Keeping this test passing the build here but failing locally is not ideal because future engineers will have a failing test if they run `npm run test` locally that is not related to their work. 

**Possible solutions:**
1) Fix the issue at the Paragon level. This is outside of this ticket's scope and will require more time. Edit: #paragon-working-group has been notified and a PAR ticket will be created. 
2) Remove the use of the `<Hyperlink>` Paragon component and use plain HTML anchor tag and exported SVG for the logo instead.
3) The actual Paragon link is already tested, so since I only need to assert that the click event fired the intended action (and not that the alt text/role is correct), use the `screen.getByText()` selector instead.
_**^ chose solution 3.**_

### DIFF
--- a/src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
+++ b/src/courseware/course/sequence/lock-paywall/LockPaywall.jsx
@@ -65,6 +65,16 @@ function LockPaywall({
     });
   };
 
+  const logClickPastExpiration = () => {
+    sendTrackEvent('edx.bi.ecommerce.gated_content.past_expiration.link_clicked', {
+      ...eventProperties,
+      linkCategory: 'gated_content',
+      linkName: 'course_details',
+      linkType: 'link',
+      pageName: 'in_course',
+    });
+  };
+
   return (
     <Alert variant="light" aria-live="off" icon={Locked} className="lock-paywall-container">
       <div className="row">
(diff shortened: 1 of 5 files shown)


inventhouse (approved):

dianekaplan (approved):
Great work! No questions :)

PR #837, opened 2022-02-28
chore: update transifex scripts to api v3

mikix:
Transifex is sunsetting anything below v3 later this year. Let's get ahead of our scripts unexpectedly failing.

- Switch 'reactifex' dev dep to our maintained fork at   '@edx/reactifex' (it has an older released version number, but it's still maintained unlike the original)
- Update Makefile commands for new v3 transifex API

See https://github.com/openedx/frontend-app-support-tools/commit/4510b63dcecc711e347596e27f06f1ed4bec89b0 for a similar conversion - support-tools is only other repo using this version of the scripts.

The new script requires `TRANSIFEX_AUTH_TOKEN`, whereas the old scripts required `SECRET_USER` and `SECRET_PWD` to be set. I can't see where we set either in our edx.org scripting, so I'll assume that both versions are set somewhere, because support-tools is working fine. But this is the sort of thing that can only really be tested live, since I don't have an auth token locally to test this change with.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,9 @@
-transifex_resource = frontend-app-learning
+export TRANSIFEX_RESOURCE=frontend-app-learning
 transifex_langs = "ar,fr,es_419,zh_CN"
 
 transifex_utils = ./node_modules/.bin/transifex-utils.js
 i18n = ./src/i18n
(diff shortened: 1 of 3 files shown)


Dillon-Dumesnil (approved):
Go for it

PR #838, opened 2022-02-28
fix: [AA-1206] resolve access APIs

cdeery:
As part of eliminating redundant fields course_access was removed from
 the Courseware metadata API. There are some differences between the
 two APIs - courseware returned an error if the course had flags
 preventing it from being loaded in the MFE, and courseHome had a
 second field, can_load_courseware, that returned a boolean.

 This fix unifies the handling of the access fields to behave consistently.

This is the frontend change to https://github.com/openedx/edx-platform/pull/29920

### DIFF
--- a/src/course-home/data/__factories__/courseHomeMetadata.factory.js
+++ b/src/course-home/data/__factories__/courseHomeMetadata.factory.js
@@ -9,7 +9,7 @@ Factory.define('courseHomeMetadata')
     title: 'Demonstration Course',
     is_self_paced: false,
     is_enrolled: false,
-    can_load_courseware: false,
+    can_load_courseware: true,
     celebrations: null,
     course_access: {
       additional_context_user_message: null,
--- a/src/courseware/data/thunks.js
+++ b/src/courseware/data/thunks.js
@@ -93,7 +93,9 @@ export function fetchCourse(courseId) {
         logError(courseHomeMetadataResult.reason);
       }
       if (fetchedMetadata && fetchedCourseHomeMetadata) {
-        if (courseHomeMetadataResult.value.courseAccess.hasAccess && fetchedOutline) {
+        if (courseHomeMetadataResult.value.courseAccess.hasAccess
+          && courseHomeMetadataResult.value.canLoadCourseware
+          && fetchedOutline) {
           // User has access
           dispatch(fetchCourseSuccess({ courseId }));
           return;
--- a/src/shared/access.js
+++ b/src/shared/access.js
@@ -4,7 +4,7 @@ import { getLocale } from '@edx/frontend-platform/i18n';
 // This function inspects an access denied error and provides a redirect url (looks like a /redirect/... path),
 // which then renders a nice little message while the browser loads the next page.
 // This is basically a frontend version of check_course_access_with_redirect in the backend.
-export function getAccessDeniedRedirectUrl(courseId, activeTabSlug, courseAccess, start, unitId) {
+export function getAccessDeniedRedirectUrl(courseId, activeTabSlug, canLoadCourseware, courseAccess, start, unitId) {
   let url = null;
   switch (courseAccess.errorCode) {
     case 'audit_expired':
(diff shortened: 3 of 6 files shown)


src/shared/access.js
```diff
@@ -33,7 +33,13 @@ export function getAccessDeniedRedirectUrl(courseId, activeTabSlug, courseAccess
     case 'authentication_required':
     case 'enrollment_required':
     default:
-      if (activeTabSlug !== 'outline') {
+      // if the learner has access to the course, but it is not enabled in the mfe, there is no
+      // error message, but canLoadCourseware will be false
+      if (canLoadCourseware === false) {
+        if (unitId) {
+          url = `/redirect/courseware/${courseId}/unit/${unitId}`;
+        }
```
Dillon-Dumesnil: Am I correct in reading this as "If you are loading the Course Outline tab in the MFE, but do not have access to the Courseware for this course in the MFE, attempt to redirect to the legacy view?"

Should this maybe be rewritten as 
```
if (canLoadCourseware === false) {
        if (unitId) {
          url = `/redirect/courseware/${courseId}/unit/${unitId}`;
        }
}
if ((activeTabSlug !== 'outline') {
        url = `/redirect/course-home/${courseId}`;
      }
```
I believe this gets at the issue you want without co-opting the behavior the course home has already created. With this approach you should also delete lines 24-32 in this file.

Alternatively, you might be able to just switch the case of `'microfrontend_disabled'` to `canLoadCourseware === false` and then leave the rest the same.

Let me know what you think.

src/courseware/data/thunks.js
```diff
@@ -205,7 +205,9 @@ export function fetchCourse(courseId) {
         logError(courseHomeMetadataResult.reason);
       }
       if (fetchedMetadata && fetchedCourseHomeMetadata) {
-        if (courseHomeMetadataResult.value.courseAccess.hasAccess && fetchedBlocks) {
+        if (courseHomeMetadataResult.value.courseAccess.hasAccess
+          && courseHomeMetadataResult.value.canLoadCourseware
```
Dillon-Dumesnil, line 97: 👍

src/course-home/data/__factories__/courseHomeMetadata.factory.js
```diff
@@ -9,7 +9,7 @@ Factory.define('courseHomeMetadata')
     title: 'Demonstration Course',
     is_self_paced: false,
     is_enrolled: false,
-    can_load_courseware: false,
+    can_load_courseware: true,
```
Dillon-Dumesnil, line 12: Why this switch? Feels like it could be odd given some of the other defaults we use in this file already (for example, `is_enrolled: false,`)

Dillon-Dumesnil (commented):
Couple of questions about this approach.

src/course-home/data/__factories__/courseHomeMetadata.factory.js
cdeery: One exception to false values in the factory is that course_access.has_access is true by default. Adding in the additional check for canLoadCourseware breaks a lot of tests. It seemed reasonable to have it as the default.

src/shared/access.js
cdeery: yes, you are correct. The "else" hides a code path.
I can take out the case for 'microfrontend_disabled', but I didn't think I could put the case to be something that wasn't the result of the expression in the switch statement.

src/shared/access.js
```diff
@@ -21,19 +21,17 @@ export function getAccessDeniedRedirectUrl(courseId, activeTabSlug, courseAccess
...
-      break;
     case 'authentication_required':
     case 'enrollment_required':
     default:
-      if (activeTabSlug !== 'outline') {
+      // if the learner has access to the course, but it is not enabled in the mfe, there is no
+      // error message, canLoadCourseware will be false.
+      // This is only used for courseware
+      if (activeTabSlug === 'courseware' && canLoadCourseware === false) {
+        if (unitId) {
```
Dillon-Dumesnil: Nit: just add unitId into the above check since all of those must be true to enter the loop.

src/shared/access.js
```diff
@@ -21,19 +21,17 @@ export function getAccessDeniedRedirectUrl(courseId, activeTabSlug, courseAccess
...
-        url = `/redirect/courseware/${courseId}/unit/${unitId}`;
-      }
-      break;
     case 'authentication_required':
     case 'enrollment_required':
     default:
-      if (activeTabSlug !== 'outline') {
+      // if the learner has access to the course, but it is not enabled in the mfe, there is no
+      // error message, canLoadCourseware will be false.
+      // This is only used for courseware
```
Dillon-Dumesnil: Nit: remove this comment. I think it is implied by the activeTabSlug check

Dillon-Dumesnil (approved):
Small nits and then other things to still patch up in the backend, but looks good to me

PR #839, opened 2022-03-01
fix: resize course handout iframe as contents change

mikix:
This can happen if a handout has e.g. expanding list elements.

[AA-1215](https://openedx.atlassian.net/browse/AA-1215)


Dillon-Dumesnil (approved):
Sure

PR #840, opened 2022-03-02
feat: AA-1205: Enable Entrance Exam support for Learning MFE

Dillon-Dumesnil:
Adds an alert to the courseware if the section is an Entrance Exam. Also
adds a listener to reload the page upon receiving a message from the LMS
indicating the user has now passed the exam.

Commit also contains misc. clean up for i18n messages switching to variable names.

Related to https://github.com/openedx/edx-platform/pull/30001

### DIFF
--- a/src/alerts/sequence-alerts/hooks.js
+++ b/src/alerts/sequence-alerts/hooks.js
@@ -1,0 +1,56 @@
+import { useSelector } from 'react-redux';
+
+import { useModel } from '../../generic/model-store';
+import { ALERT_TYPES, useAlert } from '../../generic/user-messages';
+
+import messages from './messages';
+
+function useSequenceBannerTextAlert(sequenceId) {
+  const sequence = useModel('sequences', sequenceId);
+  const sequenceStatus = useSelector(state => state.courseware.sequenceStatus);
+
+  // Show Alert that comes along with the sequence
+  useAlert(sequenceStatus === 'loaded' && sequence.bannerText, {
+    code: null,
+    dismissible: false,
+    text: sequence.bannerText,
+    type: ALERT_TYPES.INFO,
+    topic: 'sequence',
+  });
+}
+
+function useSequenceEntranceExamAlert(courseId, sequenceId, intl) {
+  const course = useModel('coursewareMeta', courseId);
+  const sequence = useModel('sequences', sequenceId);
+  const sequenceStatus = useSelector(state => state.courseware.sequenceStatus);
+  const {
+    entranceExamCurrentScore,
+    entranceExamEnabled,
+    entranceExamId,
+    entranceExamMinimumScorePct,
+    entranceExamPassed,
+  } = course.entranceExamData || {};
+  const entranceExamAlertVisible = sequenceStatus === 'loaded' && entranceExamEnabled && entranceExamId === sequence.sectionId;
+  let entranceExamText;
+
+  if (entranceExamPassed) {
+    entranceExamText = intl.formatMessage(
+      messages.entranceExamTextPassed, { entranceExamCurrentScore: entranceExamCurrentScore * 100 },
+    );
+  } else {
+    entranceExamText = intl.formatMessage(messages.entranceExamTextNotPassing, {
+      entranceExamCurrentScore: entranceExamCurrentScore * 100,
+      entranceExamMinimumScorePct: entranceExamMinimumScorePct * 100,
+    });
+  }
+
+  useAlert(entranceExamAlertVisible, {
+    code: null,
+    dismissible: false,
+    text: entranceExamText,
+    type: ALERT_TYPES.INFO,
+    topic: 'sequence',
+  });
+}
+
+export { useSequenceBannerTextAlert, useSequenceEntranceExamAlert };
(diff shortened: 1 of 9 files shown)


src/courseware/course/sequence/Sequence.test.jsx
```diff
@@ -116,6 +116,43 @@ describe('Sequence', () => {
     expect(screen.queryAllByRole('button').length).toEqual(0);
   });
 
+  it('renders correctly with Entrance Exam content and passing score', async () => {
```
Dillon-Dumesnil: I have no idea how to make this test work for alerts so if anyone knows, would appreciate the help.

src/courseware/data/__factories__/courseMetadata.factory.js
```diff
@@ -49,6 +49,7 @@ Factory.define('courseMetadata')
     course_exit_page_is_active: true,
     user_has_passing_grade: false,
     certificate_data: null,
+    entrance_exam_data: null,
```
mikix: That's not realistic though, yeah? The backend always provides it, just with some default values in all the fields?

src/courseware/course/sequence/Sequence.jsx
```diff
@@ -97,26 +97,53 @@ function Sequence({
...
+    });
+  }
+
+  useAlert(entranceExamAlertVisible, {
+    code: null,
+    dismissible: false,
+    text: entranceExamText,
+    type: ALERT_TYPES.INFO,
+    topic: 'sequence',
+  });
```
mikix: This hunk of entrance exam alert-condition-checking code might be a convient thing to throw into a `useEntranceExamAlert` hook in some hooks.js file, just to isolate it?

src/courseware/course/sequence/Sequence.jsx
```diff
@@ -97,26 +97,53 @@ function Sequence({
...
-          code: null,
-          dismissible: false,
-          text: sequence.bannerText,
-          type: ALERT_TYPES.INFO,
-          topic: 'sequence',
-        });
+    function receiveMessage(event) {
+      const { type } = event.data;
+      if (type === 'entranceExamComplete') {
+        global.location.reload();
```
mikix, line 109: This is a dramatic action - might be worth a comment on this side as to what is happening / why.

src/courseware/course/sequence/Sequence.jsx
```diff
@@ -97,26 +97,53 @@ function Sequence({
...
-        id = add({
-          code: null,
-          dismissible: false,
-          text: sequence.bannerText,
-          type: ALERT_TYPES.INFO,
-          topic: 'sequence',
-        });
+    function receiveMessage(event) {
+      const { type } = event.data;
+      if (type === 'entranceExamComplete') {
```
mikix: Have we been using camel case or snake case or what for these?

I think TNL used `plugin.resize` style names. I just recently followed their lead a bit with `lmshtmlfragment.resize`. This is all a bit of a wild west. Namespacing is unclear, etc. I don't care what form you end up using (or if you use this one), but just wanted to call it out.

mikix (approved):

src/courseware/course/sequence/Sequence.jsx
Dillon-Dumesnil: Seems like the wild west. I thought camelCase made sense just because the sender is in JS, but could see how a little bit of namespacing via `.` could be helpful. Hmm. Not sure what namespace this should even fall under. `sequence.entranceExamComplete`? `entranceExam.complete`?

src/alerts/sequence-alerts/hooks.js
```diff
@@ -0,0 +1,56 @@
...
+    text: sequence.bannerText,
+    type: ALERT_TYPES.INFO,
+    topic: 'sequence',
+  });
+}
+
+function useSequenceEntranceExamAlert(courseId, sequenceId, intl) {
+  const course = useModel('coursewareMeta', courseId);
+  const sequence = useModel('sequences', sequenceId);
+  const sequenceStatus = useSelector(state => state.courseware.sequenceStatus);
```
Dillon-Dumesnil, line 25: Any thoughts on this method vs passing in the course, sequence, and sequenceStatus directly?

src/courseware/course/sequence/Sequence.jsx
Dillon-Dumesnil: Let me know what you think of this message

mikix: Sure!

src/alerts/sequence-alerts/hooks.js
mikix: No strong opinions - I know I should, but eh, we aren't very consistent in the MFE and I think my system-programming-object-oriented brain has the wrong biases from how React native people think about the tradeoffs of passing everything in vs looking everything up from react global state.

mikix (approved):

src/courseware/course/sequence/Sequence.jsx
Dillon-Dumesnil: Went with `entranceExam.passed`. Complete is an overloaded term here and it's more about passed than complete anyway. And adds in a little namespace of entranceExam in case that helps us in the future. Luckily fairly easy to change for us in the future should we desire

src/alerts/sequence-alerts/hooks.js
muselesscreator: I think where only needed for the hook, it is reasonable to fetch the data directly to the hook instead of the component.  Makes them more re-usable, which is a huge part of the point of hooks anyway.

PR #841, opened 2022-03-03
fix: guard access to a react ref before using it

mikix:
This was causing some errors ever since we started using the ref in an event handler - I'm not entirely sure why the ref stops being valid, unless it's a lifecycle thing. But anyway, this seems to stop the error in testing.


Dillon-Dumesnil (approved):
Maybe related to us having multiple iframes on this same page? So the `current` changes?

mikix:
> Maybe related to us having multiple iframes on this same page? So the current changes?

Each instance would have its own wrapping `LmsHtmlFragment`, which is where this ref is kept. I'm not sure.

PR #843, opened 2022-03-03
chore: update frontend-build, paragon, header, and footer

mikix:
A collection of edx-owned npm updates. These required an actual
code change of using our own svg rather than directly loading an
svg from paragon, because paragon has its own svgo config that
can potentially conflict with our version of svgo - as it does
when we update frontend-build.

And with the latest versions of frontend-build, we can now use
the latest versions of paragon.

Header and footer updates thrown in for free.

### DIFF
--- a/src/course-home/outline-tab/widgets/LearningGoalButton.jsx
+++ b/src/course-home/outline-tab/widgets/LearningGoalButton.jsx
@@ -2,7 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
-import { ReactComponent as FlagIntenseIcon } from '@edx/paragon/icons/svg/flag.svg';
+// These flag svgs are derivatives of the Flag icon from paragon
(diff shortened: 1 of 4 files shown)


src/course-home/outline-tab/widgets/LearningGoalButton.jsx
```diff
@@ -2,7 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
-import { ReactComponent as FlagIntenseIcon } from '@edx/paragon/icons/svg/flag.svg';
+// These flag svgs are derivatives of the Flag icon from paragon
+import { ReactComponent as FlagIntenseIcon } from './flag_black.svg';
```
mikix, line 6: Doing this the previous way is actually an anti-pattern in my mind. You want these icons to update in lockstep, and not to allow paragon's version to drift from ours.

Dillon-Dumesnil (approved):

PR #852, opened 2022-03-07
Update testing-library/react to 12.x

mikix:
From 10.x. Along the way, also drop a lot of other unnecessary dev dependencies.


src/courseware/course/NotificationTrigger.test.jsx
```diff
@@ -53,18 +53,6 @@ describe('Notification Trigger', () => {
...
-    const { container } = render(<NotificationTrigger {...mockData} notificationStatus="active" />);
-    expect(container).toBeInTheDocument();
-    expect(screen.getByTestId('notification-dot')).toBeInTheDocument();
-    jest.useFakeTimers();
-    setTimeout(() => {
-      expect(localStorage.setItem).toHaveBeenCalledTimes(2);
-      expect(screen.queryByRole('notification-dot')).not.toBeInTheDocument();
-    }, 3000);
-    jest.runAllTimers();
-  });
```
mikix: This test was wholly broken.

1. It was doing a `queryByRole` instead of `queryByTestId` so it would never find anything anyway, even without a timeout. 
2. This component (`NotificationTrigger`) no longer even holds the 3-second logic. I guess at some point we moved that to NotificationTray.

src/courseware/course/NotificationTray.test.jsx
```diff
@@ -80,6 +80,18 @@ describe('NotificationTray', () => {
     expect(screen.queryByText('You have no new notifications at this time.')).toBeInTheDocument();
   });
 
+  it('marks notification as seen 3 seconds later', async () => {
```
mikix: See `NotificationTrigger.jsx` comment for why this was moved here & fixed up.

package.json
```diff
@@ -34,7 +34,6 @@
     "@edx/brand": "npm:@edx/brand-openedx@1.1.0",
     "@edx/frontend-component-footer": "10.2.0",
     "@edx/frontend-component-header": "2.4.3",
-    "@edx/frontend-enterprise-utils": "1.1.1",
```
mikix, line 37: Unused these days - a leftover of when we had our own custom header implementation and pulled a link from the config in this package.

Dillon-Dumesnil (approved):

PR #861, opened 2022-03-07
fix: [AA-1207] unify source of tabs

cdeery:
courseware and courseHome both provide tabs to the mfe.
This PR unifies the calls so that tab descriptions are only fetched from courseHome metadata

These are the frontend changes to these backend changes: https://github.com/openedx/edx-platform/pull/30023


src/course-home/data/api.js
```diff
@@ -90,14 +90,22 @@ function normalizeAssignmentPolicies(assignmentPolicies, sectionScores) {
...
   return {
     ...data,
     tabs: data.tabs.map(tab => ({
-      // The API uses "courseware" as a slug for both courseware and the outline tab. We switch it to "outline" here for
+      // The API uses "courseware" as a slug for both courseware and the outline tab.
+      // If needed, we switch it to "outline" here for
       // use within the MFE to differentiate between course home and courseware.
-      slug: tab.tabId === 'courseware' ? 'outline' : tab.tabId,
+      slug: tab.tabId === 'courseware' ? rootSlug : tab.tabId,
+      id: tab.tabId === 'courseware' ? rootSlug : tab.tabId,
```
mikix: This is good and fine and we talked about doing it this way. But would another way to do this same thing be to fix this not at the data layer but at the render layer? Like, have the outline version of `LoadedTabPage` modify the `activeTabSlug` it passes to `CourseTabsNavigation`, or even have `CourseTabsNavigation` be directly aware that an `outline` tab slug is the same as `courseware`? Not sure where we want this edge case, and maybe it is more appropriate to do that where we "normalize" this data indeed. Anyway, just rambling. This is good.

src/course-home/data/api.js
```diff
@@ -90,14 +90,22 @@ function normalizeAssignmentPolicies(assignmentPolicies, sectionScores) {
   });
 }
 
-function normalizeCourseHomeCourseMetadata(metadata) {
+/**
+ * Tweak the metadata for consistency
+ * @param metadata the data to normalize
+ * @param rootSlug either 'courseware' or 'outline' depending on the context
+ * @returns {{isMasquerading: (*|boolean), tabs}}
```
mikix: This is slightly misleading. It also returns all the other data too. And `tabs` was in the incoming data - we just modify it. The only addition is `isMasquerading`.

src/course-home/outline-tab/widgets/StartOrResumeCourseCard.jsx
```diff
@@ -56,7 +56,7 @@ function StartOrResumeCourseCard({ intl }) {
         )}
       />
       {/* Footer is needed for internal vertical spacing to work out. If you can remove, be my guest */}
-      <Card.Footer />
+      <Card.Footer><></></Card.Footer>
```
mikix, line 59: Lol, did I mess this up? You saw some rendering issue?

src/course-home/data/api.js
```diff
@@ -90,14 +90,22 @@ function normalizeAssignmentPolicies(assignmentPolicies, sectionScores) {
...
   const data = camelCaseObject(metadata);
   return {
     ...data,
     tabs: data.tabs.map(tab => ({
-      // The API uses "courseware" as a slug for both courseware and the outline tab. We switch it to "outline" here for
+      // The API uses "courseware" as a slug for both courseware and the outline tab.
+      // If needed, we switch it to "outline" here for
       // use within the MFE to differentiate between course home and courseware.
-      slug: tab.tabId === 'courseware' ? 'outline' : tab.tabId,
+      slug: tab.tabId === 'courseware' ? rootSlug : tab.tabId,
```
mikix, line 107: What caused the addition of `id`? Courseware is using it? Feels odd to have two identical fields - that's the sort of thing that's normally normalized out, not in!  :) Can we change courseware to use `slug`? Or the rest of MFE to use `id`?

src/courseware/course/course-exit/CourseRecommendations.jsx
```diff
@@ -106,8 +106,8 @@ function CourseCard({
           <Card.ImageCap src={image.src} />
           <Card.Header title={truncate(title, 70, { reserveLastWord: -1 })} subtitle={subtitle} size="sm" />
           {/* Section is needed for internal vertical spacing to work out. If you can remove, be my guest */}
-          <Card.Section />
-          <Card.Footer textElement={intl.formatMessage(messages.recommendationsCourseFooter)} />
+          <Card.Section> <></> </Card.Section>
+          <Card.Footer textElement={intl.formatMessage(messages.recommendationsCourseFooter)}><></></Card.Footer>
```
mikix, line 110: Is this something your IDE is doing? Or real issues?

src/courseware/course/course-exit/CourseNonPassing.jsx
```diff
@@ -16,7 +16,8 @@ import { logClick, logVisit } from './utils';
 
 function CourseNonPassing({ intl }) {
   const { courseId } = useSelector(state => state.courseware);
-  const { org, tabs, title } = useModel('coursewareMeta', courseId);
+  const { org, title } = useModel('coursewareMeta', courseId);
+  const { tabs } = useModel('courseHomeMeta', courseId);
```
mikix, line 20: Not necessary to do now, but you could also grab `org` and `title` from `courseHomeMeta` while you're here. (and in `CourseInProgress.jsx`)

src/generic/tabs/Tabs.jsx
```diff
@@ -37,7 +37,7 @@ export default function Tabs({ children, className, ...attrs }) {
         key="overflow"
       >
         <Dropdown className="h-100">
-          <Dropdown.Toggle variant="link" className="nav-link h-100">
+          <Dropdown.Toggle variant="link" className="nav-link h-100" id="learn.course.tabs.navigation.overflow.menu">
```
mikix, line 40: lol really making sure you don't conflict :)

But seriously, nice silencing of that warning.

src/setupTest.js
```diff
@@ -2,7 +2,6 @@ import 'core-js/stable';
 import 'regenerator-runtime/runtime';
 import '@testing-library/jest-dom';
 import '@testing-library/jest-dom/extend-expect';
-import 'jest-chain';
```
mikix, line 5: Love this change - you should also drop `jest-chain` from package.json and run `npm install` to update package-lock.json

Dillon-Dumesnil (commented):
Echoing a lot of what Mike Terry said. No major pieces outside of that.

src/course-home/data/api.js
cdeery: I'll leave it here unless we figure out that it needs to be handled elsewhere. This seemed like the change closest to the point where we know what context we are loading.

src/course-home/data/api.js
cdeery: Good point. This was autogenerated. I will clean it up.

src/course-home/outline-tab/widgets/StartOrResumeCourseCard.jsx
cdeery: Yes, there are a handful of errors that I am seeing in tests and in the browser where the new version of Paragon throws errors if required params are missing. This one failed for lack of children, and this is the minimal child object that I could pass.

src/course-home/data/api.js
cdeery: React complains that the list doesn't have an id. We do that a lot in our code - just ignore the console errors. I was trying to reduce the noise.

src/courseware/course/course-exit/CourseRecommendations.jsx
cdeery: Real issue. Paragon appears to have more strict requirements where children are required.

src/courseware/course/course-exit/CourseNonPassing.jsx
cdeery: I will be doing cleanup on the remaining redundant fields once this PR is done. I'm trying to be methodical so that it doesn't spin out too much.

src/generic/tabs/Tabs.jsx
cdeery: So much noise

src/setupTest.js
cdeery: Brilliant. Will do.

src/courseware/course/course-exit/CourseRecommendations.jsx
cdeery: I'm seeing all sorts of errors in the console with issues about required props, but they are not showing up in prod. I can't tell if these errors are a phantom caused by my local dev environment, or if they are legit.

src/course-home/data/api.js
mikix: Ah OK, a list-key warning. Like "each child in a list needs a unique key field" or something? But that's normally for the `key` field, not the `id` field?

But let's assume it does need the id field instead, in some generated component list. This fix feels like we're fixing things in the wrong layer (the data rather than the rendering side, which has some id requirement for updating the dom - stuff that the data layer knows nothing about). Like, I would expect the data to be one thing and then see the following in the rendering layer:
```
{tabs.map((tab) => <Tab id={tab.slug} {...tab} />)}
```

But I'm expecting that we do something like the following, where we don't explicitly set what react needs and just throw the data blob at react. And thus adding an id field in the data fixes the warning.
```
{tabs.map((tab) => <Tab {...tab} />)}
```

All that said though, it's not a big deal. Just raises my code-smell alarms when I see us normalizing two identical fields like this. Something somewhere is doing the wrong thing if this is the fix.

cdeery: No, you are right. it's solving the wrong problem in the wrong place. Backing it out.

mikix (approved):

PR #865, opened 2022-03-11
fix: trust the course grade the LMS gives us for progress page

mikix:
Rather than recompute it ourselves. Now that the backend is fixed to consider only visible grades for the course grade, we don't need to try to work around its logic (which is more accurate/consistent in general).

Related to https://github.com/openedx/edx-platform/pull/30043

(This is essentially a partial revert of https://github.com/openedx/frontend-app-learning/pull/480 which was a fix for [AA-845](https://openedx.atlassian.net/browse/AA-845))

[AA-1217](https://openedx.atlassian.net/browse/AA-1217)

### DIFF
--- a/src/course-home/progress-tab/ProgressTab.test.jsx
+++ b/src/course-home/progress-tab/ProgressTab.test.jsx
@@ -133,6 +133,8 @@ describe('Progress Tab', () => {
       });
       await fetchAndRender();
       expect(screen.queryByRole('button', { name: 'Grade range tooltip' })).not.toBeInTheDocument();
+      expect(screen.getByTestId('currentGradeTooltipContent').innerHTML).toEqual('50%');
+      expect(screen.getByTestId('gradeSummaryFooterTotalWeightedGrade').innerHTML).toEqual('50%');
       expect(screen.getByText('A weighted grade of 75% is required to pass in this course')).toBeInTheDocument();
     });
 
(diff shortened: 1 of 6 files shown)


src/course-home/progress-tab/ProgressTab.test.jsx
```diff
@@ -442,9 +448,8 @@ describe('Progress Tab', () => {
       expect(screen.queryAllByTestId('blocked-icon')).toHaveLength(4);
     });
 
-    it('renders correct current grade tooltip when showGrades is false', async () => {
-      // The learner has a 50% on the first assignment and a 100% on the second, making their grade a 75%
-      // The second assignment has showGrades set to false, so the grade reflected to the learner should be 50%.
+    it('does not render subsections for which showGrades is false', async () => {
```
Dillon-Dumesnil, line 451: Is this a change in behavior? Looks like before it would be showing up and just have the difference in grade computed vs. shown? Should we still check that the grade returned is correct?

Dillon-Dumesnil (commented):
One question about a test, but looks like nice cleanup

src/course-home/progress-tab/ProgressTab.test.jsx
mikix: No, not a change in behavior. Just that I realized we didn't actually have a test for that!

The situation before my PRs:
1. LMS sends over true course grade (including even invisible grades)
2. LMS also sends all subsection grades (each marked whether they are visible or not)
3. MFE only shows the subsection grades it is supposed to
4. MFE ignores sent course grade and recalculates it based on visible subsections

Now with my PRs:
1. LMS sends the "user visible" course grade (i.e. the result from step 4 above, but more accurate because it includes all the LMS grader logic)
2. same
3. same
4. MFE just uses the given course grade from the LMS now

Dillon-Dumesnil (approved):

PR #875, opened 2022-03-17
fix: Improve Guard for Iframe resize

cdeery:
Fix line that was causing JS errors when it was called in a context without a valid document

### DIFF
--- a/src/course-home/outline-tab/LmsHtmlFragment.jsx
+++ b/src/course-home/outline-tab/LmsHtmlFragment.jsx
@@ -28,7 +28,7 @@ export default function LmsHtmlFragment({
 
   const iframe = useRef(null);
   function resetIframeHeight() {
-    if (iframe.current) {
+    if (iframe?.current?.contentWindow?.document?.body) {
       iframe.current.height = iframe.current.contentWindow.document.body.scrollHeight;
     }
   }


src/course-home/outline-tab/LmsHtmlFragment.jsx
```diff
@@ -10,7 +10,7 @@ export default function LmsHtmlFragment({
   ...rest
 }) {
   const wholePage = `
-    <html>
+    <html >
```
mikix: typo?

src/course-home/outline-tab/LmsHtmlFragment.jsx
```diff
@@ -28,7 +28,7 @@ export default function LmsHtmlFragment({
 
   const iframe = useRef(null);
   function resetIframeHeight() {
-    if (iframe.current) {
+    if (iframe?.current?.contentWindow?.document?.body?.scrollHeight) {
```
mikix: I said this in slack too - but I think you want to chop off the `?.scrollHeight` at the end there - it's valid for the iframe content to size to zero, and we'd want to follow along in that case.

src/course-home/outline-tab/LmsHtmlFragment.jsx
cdeery: yes

src/course-home/outline-tab/LmsHtmlFragment.jsx
cdeery: Yes you did! Fixed

mikix (approved):
:chef_kiss:

PR #878, opened 2022-03-22
fix: [AA-1219] Crash when locked content rendered without unit ID

cdeery:
Allows SequenceNavigation to complete rendering even if unitId is not provided.

Currently, the page crashes when trying to show that a page is locked content. This change aligns with other work to allow the UnitButton to render even if it does not have a unitId.

### DIFF
--- a/src/courseware/course/sequence/sequence-navigation/UnitButton.jsx
+++ b/src/courseware/course/sequence/sequence-navigation/UnitButton.jsx
@@ -69,8 +69,13 @@ UnitButton.defaultProps = {
   showCompletion: true,
 };
 
-const mapStateToProps = (state, props) => ({
-  ...state.models.units[props.unitId],
-});
+const mapStateToProps = (state, props) => {
+  if (props.unitId) {
+    return {
+      ...state.models.units[props.unitId],
+    };
+  }
+  return {};
+};
 
 export default connect(mapStateToProps)(UnitButton);


src/courseware/course/sequence/sequence-navigation/UnitButton.jsx
```diff
@@ -69,8 +69,13 @@ UnitButton.defaultProps = {
...
-const mapStateToProps = (state, props) => ({
-  ...state.models.units[props.unitId],
-});
+const mapStateToProps = (state, props) => {
+  if (props.unitId) {
+    return {
+      ...state.models.units[props.unitId],
+    };
+  }
+  return {};
```
mikix, line 78: Should this be `return props;`?

mikix (approved):

src/courseware/course/sequence/sequence-navigation/UnitButton.jsx
cdeery: If I understand this correctly, it returns additional state values that are added to props. is that right?

mikix: Right you are!

From https://react-redux.js.org/using-react-redux/connect-mapstate:
```You do not need to include values from ownProps in the object returned from mapStateToProps. connect will automatically merge those different prop sources into a final set of props.```

PR #885, opened 2022-03-25
fix: stop stripping most inaccessible sequences from navigation

mikix:
I had previously made a "fix" to strip all inaccessible sequences from the learning sequence outline hierarachy, as a way to filter out unreleased sequences. See commit d1f19a9.

But that was too big a hammer and stripped a lot of released-but-inaccessible sequences too (e.g. prerequisites).

So now, we adopt a more nuanced approach and explicitly just filter out sequences that are both inaccessible AND unreleased.

[AA-1219](https://openedx.atlassian.net/browse/AA-1219)


cdeery (approved):
Looks good. Thanks for adding the comments, that will help when we are looking at this again 6 months from now!

PR #908, opened 2022-04-04
fix: [MICROBA-1769] Fix nonPassing states

Tj-Tracy:
When a learner is in the non-passing state for a certificate, we don't
want to give them messaging about certificate status IF the certificate
is not released yet for the course.

Before this change, learners who were were in the nonpassing state would
be informed before finishing the course or before the certificates were
released. After this change, learners who are nonpassing will be in a "not passing not
available" state (similar to "earned but not available" state) until the
course issues its certificate.

nonPassing course-end when certificates are not viewable:
image

Passing when certificates are not viewable:
image

nonPassing when certs are viewable:
image

passing when certs are viewable are unchanged for the course end screen.

certificate alert sidebar for passing and nonpassing before certs are visible:
image

Course home remains unchanged. information about certificates is only found there if a user has been awarded a cert.

### DIFF
--- a/src/course-home/progress-tab/ProgressTab.test.jsx
+++ b/src/course-home/progress-tab/ProgressTab.test.jsx
@@ -31,7 +31,9 @@ describe('Progress Tab', () => {
   courseMetadataUrl = appendBrowserTimezoneToUrl(courseMetadataUrl);
   const progressUrl = new RegExp(`${getConfig().LMS_BASE_URL}/api/course_home/progress/*`);
   const masqueradeUrl = `${getConfig().LMS_BASE_URL}/courses/${courseId}/masquerade`;
-
+  const now = new Date();
+  const tomorrow = new Date(now.getFullYear(), now.getMonth(), now.getDate() + 1);
+  const overmorrow = new Date(now.getFullYear(), now.getMonth(), now.getDate() + 2);
   function setMetadata(attributes, options) {
     const courseMetadata = Factory.build('courseHomeMetadata', attributes, options);
     axiosMock.onGet(courseMetadataUrl).reply(200, courseMetadata);
@@ -806,9 +808,52 @@ describe('Progress Tab', () => {
         sendTrackEvent.mockClear();
       });
 
-      it('Displays text for nonPassing case when learner does not have a passing grade', async () => {
+      it('Displays text for nonPassing case when learner does not have a passing grade and certificates are viewable', async () => {
         await fetchAndRender();
         expect(screen.getByText('In order to qualify for a certificate, you must have a passing grade.')).toBeInTheDocument();
+      });
+
+      it('Displays text for notAvailableNonPassing case when a learner does not have a passing grade and certificates are not viewable', async () => {
+        setMetadata({
+          can_view_certificate: false,
+          is_enrolled: true,
+        });
+        setTabData({
+          end: tomorrow.toISOString(),
+        });
+        await fetchAndRender();
+        expect(screen.getByText(`This course ends on ${tomorrow.toLocaleDateString('en-us', {
+          year: 'numeric',
+          month: 'long',
+          day: 'numeric',
+        })}, final grades and any earned certificates are scheduled to be available after ${tomorrow.toLocaleDateString('en-us', {
+          year: 'numeric',
+          month: 'long',
+          day: 'numeric',
+        })}.`)).toBeInTheDocument();
+      });
+
+      it('Handles certificate available dates for notAvailableNotPassing', async () => {
+        setMetadata({
+          can_view_certificate: false,
+          is_enrolled: true,
+        });
+        setTabData({
+          end: tomorrow.toISOString(),
+          certificate_data: {
+            certificate_available_date: overmorrow.toISOString(),
+          },
+        });
+        await fetchAndRender();
+        expect(screen.getByText(`This course ends on ${tomorrow.toLocaleDateString('en-us', {
+          year: 'numeric',
+          month: 'long',
+          day: 'numeric',
+        })}, final grades and any earned certificates are scheduled to be available after ${overmorrow.toLocaleDateString('en-us', {
+          year: 'numeric',
+          month: 'long',
+          day: 'numeric',
+        })}.`)).toBeInTheDocument();
       });
 
       it('sends event when visiting progress tab when learner is not passing', async () => {
--- a/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
+++ b/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
@@ -22,6 +22,8 @@ function CertificateStatus({ intl }) {
   const {
     isEnrolled,
     org,
+    isSelfPaced,
+    canViewCertificate,
   } = useModel('courseHomeMeta', courseId);
 
   const {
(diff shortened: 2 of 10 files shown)


src/course-home/data/__factories__/progressTabData.factory.js
```diff
@@ -6,6 +6,7 @@ Factory.define('progressTabData')
   .attrs({
     access_expiration: null,
     end: '3027-03-31T00:00:00Z',
+    can_view_certificate: true,
```
cdeery: Common fields should be in courseHomeMeta. It is loaded for tabs and for courseware.

cdeery (commented):
The engage team has been working to get fields that are duplicated across APIs consolidated. Can you add canViewCertificate to courseHomeMetadata instead of adding it to both the progress tab and courseware? Course home metadata will have been loaded for either of those cases.

src/course-home/data/__factories__/progressTabData.factory.js
Tj-Tracy: Sure thing!

cdeery (approved):
Everything looks good. Thanks for consolidating the API calls.
Now all I need is an opportunity to casually work "overmorrow" into a conversation somewhere!

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
```diff
@@ -22,6 +22,8 @@ function CertificateStatus({ intl }) {
   const {
     isEnrolled,
     org,
+    isSelfPaced,
+    canViewCertificate,
```
oliviaruizknott, line 26: Do we need both of these? Does the `canViewCertificate` not already take into account whether or not the course `isSelfPaced`?

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
```diff
@@ -90,9 +92,25 @@ function CertificateStatus({ intl }) {
...
+        day: 'numeric',
+      });
+      certAvailabilityDate = intl.formatDate(certificateAvailableDate || end, {
+        year: 'numeric',
+        month: 'long',
+        day: 'numeric',
+      });
+    }
+    //  These values just pass through if they are not needed by the message
+    body = intl.formatMessage(messages[`${certCase}Body`], { endDate, certAvailabilityDate });
```
oliviaruizknott, line 113: When the certificate is not available, passing and non-passing statuses render the same thing in the UI, right? That means this turns into the same things as the case starting on `171` below? Any way we could consolidate these two sections?

It is not clear form reading the code that these two states will result in the same thing. That would be nice.

Especially since these two places render the same text, at least having a single string referenced that is translated once in one place (and can be updated once in one place) would be good.

Also—for my own curiosity—when/why do we use the `intl.formatMessage` syntax vs the inline `FormattedMessage` component?

src/course-home/progress-tab/ProgressTab.test.jsx
```diff
@@ -806,11 +808,54 @@ describe('Progress Tab', () => {
...
+          month: 'long',
+          day: 'numeric',
+        })}, final grades and any earned certificates are scheduled to be available after ${overmorrow.toLocaleDateString('en-us', {
+          year: 'numeric',
+          month: 'long',
+          day: 'numeric',
+        })}.`)).toBeInTheDocument();
+      });
+
       it('sends event when visiting progress tab when learner is not passing', async () => {
```
oliviaruizknott, line 859: These tests for sent events make me wonder if we should add an event for "when visiting progress tab when certificate not available".

See tests for "earned but unavailable" case (search for that string in this file).

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -248,6 +249,29 @@ function CourseCelebration({ intl }) {
       }
       break;
     default:
+      // To properly handle messaging in the case where a course has not ended yet
+      // and certificateData is null, we want to make sure we serve a generic message
```
oliviaruizknott, line 253: Flexible code like this in the `CourseCelebration` component is exactly what I was thinking!—but I don’t think your current logic would ever get here.

src/courseware/course/course-exit/CourseNonPassing.jsx
```diff
@@ -38,21 +56,79 @@ function CourseNonPassing({ intl }) {
...
+                        alt={`${intl.formatMessage(messages.certificateImage)}`}
+                        className="w-100"
+                        style={{ maxWidth: '13rem' }}
+                      />
+                    </div>
+                  </div>
+                </Alert>
+              </div>
+            )
+        }
```
oliviaruizknott, line 131: Oooooh ... I do not like that you are having to recreate this UI here ...

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
oliviaruizknott: (Okay I wrote how I think the redundancy should be addressed in my overall review comment.)

oliviaruizknott (changes requested):
**Blocking**
- Send event on Progress page
  - Looking at the other tests in `ProgressTab.test.jsx`, it looks like we send some events when users visit this page in various cases. See the case covered by the test starting on line 859, for example. Can you make sure we are sending the equivalent event for our new case? And if that’s already magically taken care of, can you add a test that covers it?

**Non-blocking**

Okay, you solved this differently than I expected, and of course 💁‍♀️ , I think my way is better. 😛 I’m going to present my case here, but this is non-blocking since:
1. Your solution works!
2. You’ve already put more time than you wanted into this
3. My suggestions are more invasive
4. Probably I am just attached to my own ideas

Here goes:

**TL;DR:** Add a new COURSE_EXIT_MODE instead: use that to render the correct thing once instead of manually re-writing  existing UI in all the places it’s needed.

I was expecting a change to the general "course exit mode" logic:
- A new [`COURSE_EXIT_MODE`](https://github.com/openedx/frontend-app-learning/blob/af837fcac8b64b3007ff6f5cc371329a02565ad8/src/courseware/course/course-exit/utils.js#L7-L12) (perhaps `certNotAvailable`)
- Updates to the logic in the [`getCourseExitMode` utility function.](https://github.com/openedx/frontend-app-learning/blob/af837fcac8b64b3007ff6f5cc371329a02565ad8/src/courseware/course/course-exit/utils.js#L28)

I suggest that `getCourseExitMode` receive that new `canViewCertificate` data as an argument, and conditionally return a new `COURSE_EXIT_MODE` called something like `certNotAvailable`.

A check like this:

```
if (!canViewCertificate) {
  return COURSE_EXIT_MODES.certificateNotAvailable
}
```

—could happen [after the check for `hasScheduledContent` and before any checks about passing. ](https://github.com/openedx/frontend-app-learning/blob/af837fcac8b64b3007ff6f5cc371329a02565ad8/src/courseware/course/course-exit/utils.js#L55-L60)

In the case that `!canViewCertificate`, we want to render the one thing, without a care for whether or not the user is passing. Skip checking passing stuff.

Both the `CertificateStatus` and `CourseExit` components call this method to determine what UI to display. 

**CertStatus.jsx**
In `CertStatus.jsx`, you can add a single case to handle:
```
else if (mode === COURSE_EXIT_MODES.certNotAvailable) { ...
```
—maybe around line 93.5. This would eliminate the current redundancy where you’re having to tell it to render the same thing in both the not passing and passing cases.

**CourseExit.jsx + CourseCelebration.jsx**
Similarly, in `CourseExit.jsx`, you can add a single case to handle `mode === COURSE_EXIT_MODES.certNotAvailable`. In this case, it would render the `CourseCelebration` component.

You’ve already set up `CourseCelebration` to know whether or not the learner `canViewCertificate`, so you can add a check for that that will set the variables properly before the `switch` starts on line 85. It should look like the `earned_but_not_available` case currently on 125; but that case could now be removed.*

This would allow you to remove the re-writing of the celebration screen in `CourseNonPassing.jsx`. I don’t think you’d need any changes in that component, actually.

*(Any references to / uses of `earned_but_not_available` as a `CELEBRATION_STATUS` can be removed entirely, wherever it shows up; that should not be conceptually distinct from "not earned but not available").

Tj-Tracy:
Closing out this PR because the Scope of the ticket has changed. I'll be making a new branch to refactor these changes in a more appropriate way.

PR #918, opened 2022-04-08
fix: [MICROBA-1769] Cert status before course end

Tj-Tracy:
Right now, learners who are nonpassing are able to view information
about thier certificates early at the course end screen and progress
pages. This is because we show messaging around the nonpassing state in
some cases before a course ends and certificates are available. This can
also lead to cases where grades are not finalized and students who may
be passing see a scary nonpassing message instead.

This change makes it so during the course exit, a student who finishes a
course before the course is over will see the celebration screen
regardless of passing status. Once the course is over (or if
certificates are available immediately), and they are
still not passing, they will see the nonpassing messaging. The same
change was made for the certificate status alert in the progress tab.

### DIFF
--- a/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
+++ b/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
@@ -22,6 +22,8 @@ function CertificateStatus({ intl }) {
   const {
     isEnrolled,
     org,
+    canViewCertificate,
+    isSelfPaced,
   } = useModel('courseHomeMeta', courseId);
 
   const {
@@ -45,6 +47,8 @@ function CertificateStatus({ intl }) {
     hasScheduledContent,
     isEnrolled,
     userHasPassingGrade,
+    null, // CourseExitPageIsActive
+    canViewCertificate || isSelfPaced,
   );
 
   const eventProperties = {
@@ -156,7 +160,7 @@ function CertificateStatus({ intl }) {
         certAvailabilityDate = <FormattedDate value={certificateAvailableDate} day="numeric" month="long" year="numeric" />;
         body = (
           <FormattedMessage
-            id="courseCelebration.certificateBody.notAvailable.endDate"
+            id="courseCelebration.certificateBody.notAvailable.certificateAvailableDate"
             defaultMessage="This course ends on {endDate}. Final grades and any earned certificates are
             scheduled to be available after {certAvailabilityDate}."
             description="This shown for leaner when they are eligible for certifcate but it't not available yet, it could because leaners just finished the course quickly!"
@@ -178,10 +182,22 @@ function CertificateStatus({ intl }) {
         }
         break;
 
-      // This code shouldn't be hit but coding defensively since switch expects a default statement
       default:
-        certCase = null;
-        certEventName = 'no_certificate_status';
+        // if user completes a course before certificates are available, treat it as notAvailable
+        // regardless of passing or nonpassing status
+        if (!canViewCertificate && !isSelfPaced) {
+          certCase = 'notAvailable';
+          endDate = intl.formatDate(end, {
+            year: 'numeric',
+            month: 'long',
+            day: 'numeric',
+          });
+          body = intl.formatMessage(messages.notAvailableEndDateBody, { endDate });
+        } else {
+          // This code shouldn't be hit but coding defensively since switch expects a default statement
+          certCase = null;
+          certEventName = 'no_certificate_status';
+        }
         break;
     }
   }
--- a/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
+++ b/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
@@ -23,6 +23,7 @@ function CertificateStatus({ intl }) {
     isEnrolled,
     org,
     canViewCertificate,
+    userTimezone,
   } = useModel('courseHomeMeta', courseId);
 
   const {
@@ -61,6 +62,7 @@ function CertificateStatus({ intl }) {
   let certStatus;
   let certWebViewUrl;
   let downloadUrl;
+  const timezoneFormatArgs = userTimezone ? { timeZone: userTimezone } : {};
 
   if (certificateData) {
     certStatus = certificateData.certStatus;
@@ -190,6 +192,8 @@ function CertificateStatus({ intl }) {
             year: 'numeric',
             month: 'long',
             day: 'numeric',
+          }, {
+            ...timezoneFormatArgs,
           });
           body = intl.formatMessage(messages.notAvailableEndDateBody, { endDate });
         }
--- a/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
+++ b/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
@@ -192,7 +192,6 @@ function CertificateStatus({ intl }) {
             year: 'numeric',
             month: 'long',
             day: 'numeric',
-          }, {
             ...timezoneFormatArgs,
           });
           body = intl.formatMessage(messages.notAvailableEndDateBody, { endDate });
--- a/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
+++ b/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
@@ -195,6 +195,9 @@ function CertificateStatus({ intl }) {
             ...timezoneFormatArgs,
           });
           body = intl.formatMessage(messages.notAvailableEndDateBody, { endDate });
+        } else {
+          certCase = null;
+          certEventName = 'no_certificate_status';
         }
         break;
     }
--- a/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
+++ b/src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
@@ -23,7 +23,6 @@ function CertificateStatus({ intl }) {
     isEnrolled,
     org,
     canViewCertificate,
-    isSelfPaced,
   } = useModel('courseHomeMeta', courseId);
 
   const {
@@ -48,7 +47,7 @@ function CertificateStatus({ intl }) {
     isEnrolled,
     userHasPassingGrade,
     null, // CourseExitPageIsActive
-    canViewCertificate || isSelfPaced,
+    canViewCertificate,
   );
 
   const eventProperties = {
@@ -160,7 +159,7 @@ function CertificateStatus({ intl }) {
         certAvailabilityDate = <FormattedDate value={certificateAvailableDate} day="numeric" month="long" year="numeric" />;
         body = (
           <FormattedMessage
-            id="courseCelebration.certificateBody.notAvailable.certificateAvailableDate"
+            id="courseCelebration.certificateBody.notAvailable.endDate"
             defaultMessage="This course ends on {endDate}. Final grades and any earned certificates are
             scheduled to be available after {certAvailabilityDate}."
             description="This shown for leaner when they are eligible for certifcate but it't not available yet, it could because leaners just finished the course quickly!"
@@ -185,7 +184,7 @@ function CertificateStatus({ intl }) {
       default:
         // if user completes a course before certificates are available, treat it as notAvailable
         // regardless of passing or nonpassing status
-        if (!canViewCertificate && !isSelfPaced) {
+        if (!canViewCertificate) {
           certCase = 'notAvailable';
           endDate = intl.formatDate(end, {
             year: 'numeric',
@@ -193,10 +192,6 @@ function CertificateStatus({ intl }) {
             day: 'numeric',
           });
           body = intl.formatMessage(messages.notAvailableEndDateBody, { endDate });
-        } else {
-          // This code shouldn't be hit but coding defensively since switch expects a default statement
-          certCase = null;
-          certEventName = 'no_certificate_status';
         }
         break;
     }
--- a/src/courseware/course/course-exit/CourseCelebration.jsx
+++ b/src/courseware/course/course-exit/CourseCelebration.jsx
@@ -133,7 +133,7 @@ function CourseCelebration({ intl }) {
         <>
           <p>
             <FormattedMessage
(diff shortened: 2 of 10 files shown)


src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -248,6 +251,25 @@ function CourseCelebration({ intl }) {
       }
       break;
     default:
+      if (!canViewCertificate && !isSelfPaced) {
+        visitEvent = 'celebration_course_not_ended';
```
oliviaruizknott: I’m not sold on this name since it also handles cases where the course may have ended but the cert available date hasn’t lapsed. Can we reuse the `celebration_with_unavailable_cert` event name from the `earned_but_not_available` case?

I wonder who uses this data and how important distinguishing these cases is. If it is important, then I imagine the way that this catches 3 cases isn’t great either ... which is why I’d advocate for just using the `celebration_with_unavailable_cert` for all until / unless we can get this whole mess more generally sorted out.

oliviaruizknott (commented):
I am satisfied—though not overjoyed—with this approach. Overjoyment was too high an expectation for this area of the code. 😛

Will add proper review whenever this is more finalized.

src/courseware/course/course-exit/CourseCelebration.jsx
Tj-Tracy: Yeah unfortunately we cannot remove the distinction between the new default case and `earned_not_available` at the moment. I'm fine with using this as the event name for all 3 cases. which are:
- Not passing, course not ended
- Passing, course not ended
- Passing, course ended, certificate available date not past

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
```diff
@@ -156,7 +160,7 @@ function CertificateStatus({ intl }) {
         certAvailabilityDate = <FormattedDate value={certificateAvailableDate} day="numeric" month="long" year="numeric" />;
         body = (
           <FormattedMessage
-            id="courseCelebration.certificateBody.notAvailable.endDate"
+            id="courseCelebration.certificateBody.notAvailable.certificateAvailableDate"
```
mikix: This is going to erase all translations for this string and instead have the new string you added with this id be wrongly translated. You may not love the id name, but you should keep it and use something else for the new string you add.

I would also be super cautious about trying to reuse an id across files like `notAvailable.certificateAvailableDate` here. If one file drifts the defaultMessage from the other, I dunno if our transifex tooling would yell at us or silently pick one message to use. I would probably prefer to just have duplicated translations than risk incorrect ones.

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
```diff
@@ -45,6 +47,8 @@ function CertificateStatus({ intl }) {
     hasScheduledContent,
     isEnrolled,
     userHasPassingGrade,
+    null, // CourseExitPageIsActive
+    canViewCertificate || isSelfPaced,
```
mikix: That feels like an odd pairing of variables. Knowing that self-paced courses can view the cert immediately is the sort of business logic that I'd prefer live in the backend, which would ideally just report `can_view_certificate = true` in that case.

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
```diff
@@ -178,10 +182,22 @@ function CertificateStatus({ intl }) {
...
+        if (!canViewCertificate && !isSelfPaced) {
+          certCase = 'notAvailable';
+          endDate = intl.formatDate(end, {
+            year: 'numeric',
+            month: 'long',
+            day: 'numeric',
+          });
+          body = intl.formatMessage(messages.notAvailableEndDateBody, { endDate });
+        } else {
+          // This code shouldn't be hit but coding defensively since switch expects a default statement
```
mikix: You can probably drop the `since switch expects a default statement` bit from this comment now. It also sounds like this code branch was being hit? So maybe the whole comment needs a rethink.

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -248,6 +250,28 @@ function CourseCelebration({ intl }) {
       }
       break;
     default:
+      if (!canViewCertificate && !isSelfPaced) {
+        //  We reuse the cert event here. Since this default state is so
+        //  Similar to the earned_not_available state, this event name should be fine
+        //  to cover the same cases.
+        visitEvent = 'celebration_with_unavailable_cert';
+        certHeader = intl.formatMessage(messages.certificateHeaderNotAvailable);
+        const endDate = intl.formatDate(end, {
```
mikix, line 260: You'll want to pass timezone arguments here. Search elsewhere in the MFE for timezoneFormatArgs for examples.

src/courseware/course/course-exit/utils.js
```diff
@@ -73,12 +74,14 @@ function getCourseExitNavigation(courseId, intl) {
...
     courseExitPageIsActive,
   } = useModel('coursewareMeta', courseId);
+  const { canViewCertificate, isSelfPaced } = useModel('courseHomeMeta', courseId);
   const exitMode = getCourseExitMode(
     certificateData,
     hasScheduledContent,
     isEnrolled,
     userHasPassingGrade,
     courseExitPageIsActive,
+    canViewCertificate || !isSelfPaced,
```
mikix: This logic is different than the other times this is called. Intentional or typo?

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
Tj-Tracy: I changed this back to the original id. I noticed I also had an id that was duplicated as you said, and fixed it. Is that what you had meant by the second part of you comment? I'm not very familiar with how transifex works.

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
Tj-Tracy: I modified the API on the platform end and removed `isSelfPaced` from the code I added.

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
Tj-Tracy: removed the `else` portion of this if statement.

src/courseware/course/course-exit/CourseCelebration.jsx
Tj-Tracy: user timezone added 👍

src/courseware/course/course-exit/utils.js
Tj-Tracy: This was a typo 😬 But Ive since removed the need to distinguish self-paced courses.

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
mikix: Yeah that's what I meant, thanks.

Quick digression on react-intl (which is what is ultimately providing all the `formatMessage` stuff) and how it uses ids, just because it's often a confusing topic:

The id is the "lookup value" for `react-intl`. Contrast this with `gettext` used on the backend. In `gettext`-land, the english string is the "lookup value". So if I have python code like:

`gettext('Hello Mike')`

And want to change that to:

`gettext('Hello, Mike')`

I have just wiped away all existing translations for `Hello Mike` because gettext can no longer find translations when it looks up `Hello, Mike` -- they were only defined for the lookup value of `Hello Mike`. So when it can't find a translation, it will show the default English message (which is the same as the id lookup for `gettext` of "Hello, Mike".

This is not as bad as it sounds, since most translation tooling (like transifex) will show translators similar strings & past translations etc. But it **will** usually require a manual step by a translator to get that translated again.

Compare with the `react-intl` approach of using an id for the lookup value. This gives some flexibility to make minor changes to the string without wiping away translations. Now if I have a formatted message like:

`formatMessage({ id: 'hello', defaultMessage: 'Hello Mike'})`

I'm saying `I will look up translations with the id key "hello", and if I can't find it, I'll use the provided default English message "Hello Mike"`.

If I change that default English message to be `Hello, Mike`, I'm only changing that default English value -- all existing translations are left alone. But future translators will see `Hello, Mike` in the translation UI and will incorporate the comma.

But if we ever change the id, `react-intl` will no longer be able to find the previous translations, so we effectively wipe them out until translators get a chance to translate the string again. This is often an appropriate thing! What if I wanted to change the English version to `Bye, Mike` -- that feels like I've actually changed the meaning and we **want** translators to re-translate it. So I might also change the id to something like `bye`.

Now this sounds nice! Engineers now have the ability to decide for themselves whether their string modifications are changing the meaning of the phrase or not (i.e. to wipe out existing translations or not).

**HOWEVER** this has pros and cons. Full disclosure, I am more a fan of the `gettext` approach, but both are fine as long as engineers are educated about it (hence why I'm spending this time).

- Engineers may not understand all that nuance above and may change strings without changing the id value, creating obviously bad translations.
- Engineers may not be the best judge of what sorts of string modifications would require a change in translation, especially considering the wide world of languages out there.
- There are a lot of edge cases that feel odd for an engineer to be adjudicating (like what if I changed it to `Hello Mike!`...? Is that "worth" retranslating? What's the value placed upon Spanish users having the correct emphasis there?)

So `gettext` is very conservative and just says "hey folks, you make _any_ change and I'm gonna let the translators have a pass at this" -- preferring correctness over the convenience of being able to edit typos or the value of close-enough translations. That's where I generally fall myself, but I can see the appeal of either side.

Also, what I've described is my best understanding of the system, there may be bits I've got wrong or tooling that is fancier than I think (for example, maybe transifex itself notices english changes and wipes out translations like `gettext` would - though in that case, it's just transifex being fancy - another tool might not do that).

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
mikix: Nice. Though I just went and looked at the backend part of this change, and doesn't `certificates_viewable_for_course` already take `self_paced` into account? It calls `_should_certificate_be_visible` and that looks like it returns true for a self paced course.

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
```diff
@@ -178,10 +183,20 @@ function CertificateStatus({ intl }) {
...
+        // regardless of passing or nonpassing status
+        if (!canViewCertificate) {
+          certCase = 'notAvailable';
+          endDate = intl.formatDate(end, {
+            year: 'numeric',
+            month: 'long',
+            day: 'numeric',
+          }, {
+            ...timezoneFormatArgs,
+          });
```
mikix: `formatDate` takes a third argument? I'd expect this to look like:

```
         intl.formatDate(end, {
            year: 'numeric',
            month: 'long',
            day: 'numeric',
            ...timezoneFormatArgs,
          });
```

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
mikix: I think the `else` portion had some value - it did set `certEventName` for the call to `sendTrackEvent` - that might still be good.

src/courseware/course/course-exit/CourseCelebration.jsx
```diff
@@ -130,7 +133,7 @@ function CourseCelebration({ intl }) {
         <>
           <p>
             <FormattedMessage
-              id="courseCelebration.certificateBody.notAvailable.endDate.v2"
+              id="courseCelebration.certificateBody.notAvailable.certificateAvailableDate"
```
mikix: This is one of those things where it might be better to just leave the uglier id - you'd wipe out all translations if you change this.

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
Tj-Tracy: hmm, maybe I misunderstood the code example I looked at. I changed it to model your suggestion. Is there an easy way to test this? just change my browser timezone?

src/courseware/course/course-exit/CourseCelebration.jsx
Tj-Tracy: Changed this one back as well.

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
Tj-Tracy: This is a great explanation, thanks for writing this out. I was not really aware of the nuances of the system as Ive not really edited a translation string outside platform. I think Ill take some of this explanation over for documentation on the MFEs we own.

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
Tj-Tracy: re-added without the comment 👍

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
Tj-Tracy: Hmm, it does. Ive confirmed its working locally after removing the extra piece from the platform PR. No change is needed here though.

src/course-home/progress-tab/certificate-status/CertificateStatus.jsx
mikix: Change your timezone setting in your account page (without these args, it would use your browser timezone, but sometimes users set that explicitly in their edx preferences).

But as for misunderstanding, I'm _guessing_ this is what happened. You saw the **component** version of `formatMessage`: `<FormattedMessage>` that looked like this:

```
    <FormattedRelative
      key="timeRemaining"
      value={endDate}
      {...timezoneFormatArgs}
    />
```

And at quick glance, it looks like we're adding a copied object as a new argument to the end of that. But in reality, the parenthesis are jsx boundaries for returning to javascript code, and we are just multiple new args to the component, not one new object at the end.

mikix (approved):

PR #919, opened 2022-04-08
fix: downgrade react and upgrade some other deps to align

mikix:
Now that we are using node 16, peer dependencies are much more strict about aligning between all of our dependencies.

This PR downgrades react from 17 to 16 (no changes) and upgrades paragon and frontend-lib-special-exams to all be on the same page about what peer dependency ranges are valid.


src/setupTest.js
```diff
@@ -5,14 +5,13 @@ import '@testing-library/jest-dom/extend-expect';
 import './courseware/data/__factories__';
 import './course-home/data/__factories__';
 import { getConfig, mergeConfig } from '@edx/frontend-platform';
-import { configure as configureI18n } from '@edx/frontend-platform/i18n';
+import { configure as configureI18n, IntlProvider } from '@edx/frontend-platform/i18n';
```
mikix, line 8: This was because we don't directly depend on react-intl, so npm was complaining about the below direct use of it. Instead, grab the pass-through component from frontend-platform.

cdeery (approved):
That looks like a pretty low-impact change. Nice that we got a bump in the paragon version out of it.

PR #921, opened 2022-04-11
fix: add back es-check & fsevents for now to fix build

mikix:
A previous commit (7f37575) dropped es-check, which dropped fsevents, which caused our build system (which is still using npm@6) to fail with an error like `Unsupported platform for fsevents` when trying to install fsevents through a dependency (e.g. when installing npm aliases).

I am reintroducing all the package-lock changes from that commit to get back fsevents in a state where that error does not occur.

I think a longer-term fix would be to instead upgrade our build system to node16 / npm6. But this is an easy fix for now to unblock the builds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "@testing-library/user-event": "13.5.0",
     "axios-mock-adapter": "1.20.0",
     "codecov": "3.8.3",
+    "es-check": "6.2.1",
     "husky": "7.0.4",
     "jest": "27.5.1",
     "rosie": "2.1.0"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -70,6 +70,7 @@
     "@testing-library/user-event": "13.5.0",
     "axios-mock-adapter": "1.20.0",
     "codecov": "3.8.3",
+    "es-check": "6.2.1",
```
mikix, line 73: A comment here would be _lovely_, but json insanely does not allow comments.

jmyatt (approved):

PR #929, opened 2022-04-19
fix: remove support for the legacy courseware pages

mikix:
Access to learners for these pages [has been removed](https://github.com/openedx/edx-platform/pull/30238), so we don't need to keep any support for it around. Simplifies some code paths.


cdeery (approved):
This does make the front end simpler. Nice.

PR #944, opened 2022-05-11
refactor: Replace PDF course certificate view code

ruzniaievdm:


openedx-webhooks:
Although this pull request is already closed, I've created [OSPR-6672](https://openedx.atlassian.net/browse/OSPR-6672) so that we can track it in Jira.

PR #956, opened 2022-07-14
feat: [VAN-1000] - Set weekly goal through welcome email via query param

shafqatfarhan:
**Description**
Set weekly goal via query param and delete param to avoid setting the goal multiple times.

**Ticket**
https://2u-internal.atlassian.net/browse/VAN-1000

### DIFF
--- a/src/course-home/outline-tab/widgets/WeeklyLearningGoalCard.jsx
+++ b/src/course-home/outline-tab/widgets/WeeklyLearningGoalCard.jsx
@@ -1,7 +1,9 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
+import { useLocation } from 'react-router-dom';
 import PropTypes from 'prop-types';
 
 import { Form, Card, Icon } from '@edx/paragon';
+import { history } from '@edx/frontend-platform';
 import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 import { getAuthenticatedUser } from '@edx/frontend-platform/auth';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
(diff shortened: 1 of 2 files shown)


zainab-amir (approved):

src/course-home/outline-tab/widgets/WeeklyLearningGoalCard.jsx
```diff
@@ -65,6 +69,21 @@ function WeeklyLearningGoalCard({
     }
   }
 
+  useEffect(() => {
+    const currentParams = new URLSearchParams(location.search);
+    const weeklyGoal = Number(currentParams.get('weekly_goal'));
+    if (weeklyGoal) {
+      handleSelect(weeklyGoal, true);
```
cdeery, line 76: I think the input needs to be checked for correct values before it is consumed.

src/course-home/outline-tab/widgets/WeeklyLearningGoalCard.jsx
```diff
@@ -32,8 +34,9 @@ function WeeklyLearningGoalCard({
   const [daysPerWeekGoal, setDaysPerWeekGoal] = useState(daysPerWeek);
   // eslint-disable-next-line react/prop-types
   const [isGetReminderSelected, setGetReminderSelected] = useState(subscribedToReminders);
+  const location = useLocation();
 
-  function handleSelect(days) {
+  function handleSelect(days, triggeredFromEmail = false) {
     // Set the subscription button if this is the first time selecting a goal
     const selectReminders = daysPerWeekGoal === null ? true : isGetReminderSelected;
     setGetReminderSelected(selectReminders);
```
cdeery, line 42: Am I correct that the expectation here is that if the learner does not want to be subscribed to email reminders then they have the opportunity to do that when they go to the page? Does the focus need to be set to the unsubscribe button? How would users with assistive technology know they could unsubscribe?

cdeery (changes requested):
A couple of comments

src/course-home/outline-tab/widgets/WeeklyLearningGoalCard.jsx
shafqatfarhan: done

src/course-home/outline-tab/widgets/WeeklyLearningGoalCard.jsx
shafqatfarhan: Yes, learners have an option to unsubscribe to reminder emails. Currently, the focus is not set to the unsubscribe button even when the user manually sets the goal from the UI.

To support assistive technology, we are already setting the `aria-checked` attribute to indicate the currently selected weekly goal button in their group.

shafqatfarhan:
@cdeery PR is ready for review again. Thanks.

cdeery (approved):
Thanks for the changes. Looks good

PR #958, opened 2022-07-19
feat: Added support to redirect to DSC if required.

HammadAhmadWaqas:
https://2u-internal.atlassian.net/browse/ENT-5992


src/shared/access.js
```diff
@@ -18,6 +18,9 @@ export function getAccessDeniedRedirectUrl(courseId, activeTabSlug, courseAccess
     case 'survey_required':
       url = `/redirect/survey/${courseId}`;
       break;
+    case 'data_sharing_access_required':
+      url = `/redirect/consent?consentPath=${encodeURIComponent(courseAccess.developerMessage)}`;
```
sameenfatima78, line 22: Are we certain that we'll always receive `developerMessage` and it can never be empty? Would suggest handling the case if not and adding a test case for that as well.

src/courseware/CoursewareRedirectLandingPage.jsx
```diff
@@ -32,6 +33,13 @@ export default () => {
             global.location.assign(`${getConfig().LMS_BASE_URL}/dashboard${location.search}`);
           }}
         />
+        <PageRoute
+          path={`${path}/consent/`}
+          render={({ location }) => {
```
sameenfatima78, line 38: Doesn't the render prop function requires to return a JSX element?

HammadAhmadWaqas: We are redirecting to a new page, so we don't need to return anything, You can see above <PageRoute> elements. All work in the same way.

src/shared/access.js
HammadAhmadWaqas: Yes, we are 100% certain that we'all always receive `developerMessage` because this line is under `case 'data_sharing_access_required'` and I am sending this developerMessage from [here](https://github.com/openedx/edx-platform/pull/30741/files#diff-c53baebd2502a4f0aaefa2553d6d427370f049aaaeed8f560b8a532e54a5303eR234) for this the error `data_sharing_access_required `

irfanuddinahmad (approved):

zamanafzal (approved):

PR #990, opened 2022-10-18
fix: update transifex flag for tx cli 1.4.0

leangseu-edx:
The Transifex cli started requiring the -t or --translations flag in the pull command in order to fetch translations.

https://github.com/transifex/cli/pull/96


nsprenkle (approved):
:shipit:

PR #997, opened 2022-11-02
expired onboarding exam messaging fix

ashultz0:
Currently expiring soon is displayed 28 days before expiration and forever afterwards. Adds an actual expired state for after.

Also clarifies the expring soon message which assumed other course, that was not necessarily true.

Also updates the take action lines when you do not have valid onboarding to make sure they appear for everything not currently valid or in process, and updates the submitted process lines to not appear for expired statuses.

This also requires expiration time updates provided in https://github.com/openedx/edx-proctoring/pull/1088 to make sure that expiration time properly comes along with more exam states.

### DIFF
--- a/src/course-home/outline-tab/widgets/ProctoringInfoPanel.jsx
+++ b/src/course-home/outline-tab/widgets/ProctoringInfoPanel.jsx
@@ -35,6 +35,7 @@ function ProctoringInfoPanel({ intl }) {
     error: 'error',
     otherCourseApproved: 'otherCourseApproved',
     expiringSoon: 'expiringSoon',
+    expired: 'expired',
   };
 
   function getReadableStatusClass(examStatus) {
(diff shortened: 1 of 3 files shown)


src/course-home/outline-tab/widgets/ProctoringInfoPanel.jsx
```diff
@@ -77,11 +83,19 @@ function ProctoringInfoPanel({ intl }) {
     return borderClass;
   }
 
+  function isExpired(dateString) {
+    // Returns true if the expiration date is within 28 days
```
alangsto: Should update this comment for expired, not expiring soon

schenedx (approved):
👍

alangsto (approved):
One small comment fix, but otherwise the logic looks good to me

mattcarter (approved):
LGTM

- [x] code review
- [ ] code execution

PR #1014, opened 2022-12-13
build: major paragon version upgrade

Mashal-m:
Major version upgrade of `edx/paragon`.


Mashal-m:
Closing because the paragon version is upgraded now.

PR #1017, opened 2022-12-17
chore: upgrade `@edx/frontend-build` from v9 -> v12

adamstankiewicz:
* Upgrades `@edx/frontend-build` from v9 -> v12
* Addresses some new ESLint errors.
* Disables majority of new ESLint errors in `.eslintrc.js` file (some of these rules should really be turned back on and issues fixed).


.eslintrc.js
```diff
@@ -1,11 +1,19 @@
...
-    rules: {
-      'import/named': 'off',
-      'import/no-extraneous-dependencies': 'off',
-    },
-  }],
+const config = createConfig('eslint', {
+  rules: {
+    // TODO: all these rules should be renabled/addressed. temporarily turned off to unblock a release.
+    'react-hooks/rules-of-hooks': 'off',
+    'react-hooks/exhaustive-deps': 'off',
```
adamstankiewicz, line 7: Of these disabled ESLint rules, these 2 are the most important as they largely the catch instances where we're not following how React is intended to be used, notably the ["Rules of Hooks"](https://reactjs.org/docs/hooks-rules.html).

From the React docs:

> That’s what allows React to correctly preserve the state of Hooks between multiple useState and useEffect calls. (If you’re curious, we’ll explain this in depth [below](https://reactjs.org/docs/hooks-rules.html#explanation).)

leangseu-edx (approved):
LGTM 👍 . For the coverage, I will refactor the unit tests.

PR #1045, opened 2023-01-13
chore: install react share and implement social share demo

leangseu-edx:
Hard code to use `window.expSocialShare` from optimizely.

image


muselesscreator (approved):
LGTM, aside from the linting issues git is calling out

PR #1048, opened 2023-01-20
fix: disable share feature on component unmount

muselesscreator:
Teensy fix to disable share button on page change.


leangseu-edx (approved):
LGTM 👍

PR #1056, opened 2023-01-24
fix: update iframe static stylesheet

leangseu-edx:
Move css to static location.

Screen Shot 2023-01-24 at 3 14 30 PM

PR #1057, opened 2023-01-25
feat: copy static to dist folder for prod

leangseu-edx:
TLDR; assets didn’t get copy for build so no file exist and react router fail silently

From previous change https://github.com/openedx/frontend-app-learning/pull/1056, I moved `css` to a static folder because that was incorrect place. However, that didn't solve the problem because `webpack` fail to copy the file over and react-router was failing silently. React router couldn't find the file, it just return empty `html`.

### DIFF
--- a/webpack.prod.config.js
+++ b/webpack.prod.config.js
@@ -8,7 +8,6 @@ config.plugins.push(
   new CopyPlugin({
     patterns: [
       {
-
         from: path.resolve(__dirname, './public/static'),
         to: path.resolve(__dirname, './dist/static'),
       },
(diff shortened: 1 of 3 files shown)


webpack.prod.config.js
```diff
@@ -0,0 +1,19 @@
...
+const { createConfig } = require('@edx/frontend-build');
+const CopyPlugin = require('copy-webpack-plugin');
+
+const config = createConfig('webpack-prod');
+
+config.plugins.push(
+  new CopyPlugin({
+    patterns: [
+      {
+
```
muselesscreator:
```suggestion
```

muselesscreator (approved):

PR #1064, opened 2023-02-15
fix: exams with no due date now display exam type

ilee2u:
Previously, exams would not display the exam type (Timed, Midterm, etc) if the due date value was null:
image

The logic in the component for subsections on the learner end has been modified to account for when the due date value is null:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,3 @@ module.config.js
 
 # Local environment overrides
 .env.private
-
-# Installed Packages
-packages
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ module.config.js
 
 # Local environment overrides
 .env.private
+
+# Installed Packages
+packages
--- a/src/course-home/outline-tab/SequenceLink.jsx
+++ b/src/course-home/outline-tab/SequenceLink.jsx
@@ -39,7 +39,7 @@ const SequenceLink = ({
   const coursewareUrl = <Link to={`/course/${courseId}/${id}`}>{title}</Link>;
   const displayTitle = showLink ? coursewareUrl : title;
 
-  const dueDateString = due ? '{description} due {assignmentDue}' : '{description} (no due date)';
+  const dueDateString = due ? '{description} due {assignmentDue}' : '{description}';
 
   return (
     <li>
--- a/src/course-home/outline-tab/SequenceLink.jsx
+++ b/src/course-home/outline-tab/SequenceLink.jsx
@@ -41,7 +41,7 @@ const SequenceLink = ({
 
   const dueDateMessage = (
     <FormattedMessage
-      id="learning.outline.sequence-due"
+      id="learning.outline.sequence-due-date-set"
       defaultMessage="{description} due {assignmentDue}"
       description="Used below an assignment title"
       values={{
@@ -63,7 +63,7 @@ const SequenceLink = ({
 
   const noDueDateMessage = (
     <FormattedMessage
-      id="learning.outline.sequence-not-due"
+      id="learning.outline.sequence-due-date-not-set"
       defaultMessage="{description}"
       description="Used below an assignment title"
       values={{
--- a/src/course-home/outline-tab/SequenceLink.jsx
+++ b/src/course-home/outline-tab/SequenceLink.jsx
@@ -39,6 +39,8 @@ const SequenceLink = ({
   const coursewareUrl = <Link to={`/course/${courseId}/${id}`}>{title}</Link>;
   const displayTitle = showLink ? coursewareUrl : title;
 
+  const dueDateString = due ? '{description} due {assignmentDue}' : '{description} (no due date)';
+
   return (
     <li>
       <div className={classNames('', { 'mt-2 pt-2 border-top border-light': !first })}>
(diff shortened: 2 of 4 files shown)


michaelroytman (approved):
Nice job!

src/course-home/outline-tab/SequenceLink.jsx
```diff
@@ -70,31 +72,29 @@ const SequenceLink = ({
...
+              description="Used below an assignment title"
+              values={{
+                assignmentDue: (
+                  <FormattedTime
+                    key={`${id}-due`}
+                    day="numeric"
+                    month="short"
+                    year="numeric"
+                    timeZoneName="short"
+                    value={due}
```
alangsto: how does this message render if due date is null? Will the `FormattedTime` component error at all?

alangsto: actually never mind, saw your screenshot!

src/course-home/outline-tab/SequenceLink.jsx
```diff
@@ -39,6 +39,8 @@ const SequenceLink = ({
   const coursewareUrl = <Link to={`/course/${courseId}/${id}`}>{title}</Link>;
   const displayTitle = showLink ? coursewareUrl : title;
 
+  const dueDateString = due ? '{description} due {assignmentDue}' : '{description} (no due date)';
```
alangsto: I might not specify that there is no due date, just because of personalized learning schedules (failing to find documentation on that right now). I know those weren't accounted for prior to this change, but it might be cleaner to just have the description

alangsto:
Is there any way to add a unit test for this change?

src/course-home/outline-tab/SequenceLink.jsx
ilee2u: To clarify, are there ways for individual learners to set their own deadlines for stuff like exams, which could conflict with the subtitle here saying "no due date"? 

As a general example:
"Timed Exam (no due date), self set due date is 02/24/2023"

alangsto: Learners in self-paced courses have personalized learning schedules (PLS). They aren't able to choose their own deadlines, but the backend automatically creates a schedule with due dates for them. If those due dates pass without the learner completing the work, the learner has the option to shift those due dates to the future. This behavior can be repeated over and over again. 

So that's why I'm suggesting for exams with no due date to just include the description with no other information

ilee2u: Got it. I just pushed a new commit with the respective change.

ilee2u:
> Is there any way to add a unit test for this change?

I'm not sure how to create a unit test for this. Is the Jest file ```src/course-home/outline-tab/OutlineTab.test.jsx``` the right place to do so?

alangsto:
@ilee2u I think that's a good place to add it!

ilee2u:
Ok, quick context for this latest commit:

I decided to work on getting past the validation error I received in my last commits, which stated that "Messages must be statically evaluate-able for extraction" in react-intl's FormattedMessage components. I did a quick fix as a proof of concept based on this Stack Overflow thread: https://stackoverflow.com/questions/57788375/react-setting-up-dynamic-formattedmessage-error-messages-must-be-statically

I encapsulated two separate FormattedMessage components into two separate const variables, each with a different defaultMessage attribute for when a due date is present or not. In the return() statement, the respective FormattedMessage component is rendered based on whether or not the due date ("due" in the code) is null.

This is definitely a dirty fix, which the Stack Overflow thread I linked mentions. I can refactor this further to be cleaner if you'd like.

EDIT: Also, I figured out that I can validate my work before committing by running ```make validate``` :^)

.gitignore
```diff
@@ -23,3 +23,6 @@ module.config.js
 
 # Local environment overrides
 .env.private
+
+# Installed Packages
+packages
```
alangsto: I don't think this should be added, I'm not seeing it in other MFE .gitignore files

ilee2u: Ah right. I have that for now since I have the frontend-lib-special-exams package installed in a directory called "packages/", which I installed to get the app working. I tried to remove the packages directory itself, but that caused the app to throw an error along the lines of "frontend-lib-special-exams not found", which I didn't know how to resolve.

I can either figure out how to uninstall the package, or remove both this line of code and the packages directory once this PR is ready to merge.

ilee2u:
Ok! I've begun implementation of a unit test. I used the docs for [testing-library](https://testing-library.com/docs/queries/about/). I decided to use ```screen.getByText()``` to see if the correct elements would appear based on whether or not an exam has a set due date:

```
    // TODO: insert unit test for if due date is null or not, what the subtitle
    // If due date set, make sure subtitle says description AND due date
    it('exam subsection description and due date display correctly', async () => {
      const { courseBlocks } = await buildMinimalCourseBlocks(courseId, 'Title', { resumeBlock: true });
      setTabData({
        course_blocks: { blocks: courseBlocks.blocks },
      });
      await fetchAndRender();

      // Look for a substring that says "(exam type) Exam due (datetime)""
      expect(screen.getByText(/Exam due/)).toBeInDocument();
    });

    // If due date is NOT set, ONLY display description
    it('exam subsection description and due date display correctly', async () => {
      const { courseBlocks } = await buildMinimalCourseBlocks(courseId, 'Title', { resumeBlock: true });
      setTabData({
        course_blocks: { blocks: courseBlocks.blocks },
      });
      await fetchAndRender();
      // Look for a substring that only says "(exam type) Exam"
      expect(screen.getByText(/Exam$/)).toBeInDocument();
    });
```
    
I'm not sure how to modify whatever version of this MFE test is working on, such that I can test both cases where a due date is set/not set for an exam subsection like so: 
image

TL;DR: I tried my best to make a unit test based off the docs, but it definitely won't work since I don't actually know how to set up the mock version of the page that gets tested (amongst other gaps in my knowledge here). I'd appreciate any help for setting that up.

ilee2u:
Attempting to implement a unit test is beyond the scope of this PR. Ready for final review.

zacharis278 (approved):
🎉

PR #1078, opened 2023-03-08
feat: update special-exams library

zacharis278:
Adds extra parameter for attempt id when sending messages to external proctoring software

details: https://github.com/openedx/frontend-lib-special-exams/pull/86


varshamenon4 (approved):
Looks good!

PR #1080, opened 2023-03-09
feat: update special exams version

alangsto:
## [MST-1643](https://2u-internal.atlassian.net/browse/MST-1643)

Update the special exams lib, and ensure that `EXAMS_BASE_URL` is read from the environment. It should be nulled because it's being used to gate new exams behavior.


varshamenon4 (approved):
Looks good!

PR #1081, opened 2023-03-10
fix: discussion sidebar loads very slow

muhammadadeeltajamul:
Discussion was loading very slow because the iframe gets loaded first and then the call was made to load discussions MFE in sidebar. This process was taking more than 10 sec. 
After this change sidebar will be loaded by default but will be hidden. When user will click on sidebar button, sidebar will open instantly reducing the load time for user

Ticket Link: [inf-766](https://2u-internal.atlassian.net/browse/INF-766)


awais-ansari (approved):

ayesha-waris (approved):
LGTM

PR #1083, opened 2023-03-13
feat: fetch exam access token

varshamenon4:
When required (if Unit is an exam and if using the exams IDA), this code fetches the exam access token. If the exam access token is not granted the iframe does not load. The exam access token is added to the iframe url.

Here are the associated tickets:
[MST-1599](https://2u-internal.atlassian.net/browse/MST-1599)
[MST-1781](https://2u-internal.atlassian.net/browse/MST-1781)

### DIFF
--- a/src/courseware/course/sequence/Unit.jsx
+++ b/src/courseware/course/sequence/Unit.jsx
@@ -1,6 +1,8 @@
 import { getConfig } from '@edx/frontend-platform';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import { logError } from '@edx/frontend-platform/logging';
 import { AppContext, ErrorPage } from '@edx/frontend-platform/react';
+import { getExamAccess, fetchExamAccess, isExam } from '@edx/frontend-lib-special-exams';
 import { Modal } from '@edx/paragon';
 import PropTypes from 'prop-types';
 import React, {
@@ -65,6 +67,16 @@ function useLoadBearingHook(id) {
   }, [id]);
 }
 
+function addExamAccessToIframeUrl(accessToken, iframeUrl) {
+  let url = iframeUrl;
+  if (isExam()) {
+    if (accessToken) {
+      url += `&exam_access=${accessToken}`;
+    }
+  }
+  return url;
+}
+
 export function sendUrlHashToFrame(frame) {
   const { hash } = window.location;
   if (hash) {
@@ -83,6 +95,7 @@ const Unit = ({
 }) => {
   const { authenticatedUser } = useContext(AppContext);
   const view = authenticatedUser ? 'student_view' : 'public_view';
+
   let iframeUrl = `${getConfig().LMS_BASE_URL}/xblock/${id}?show_title=0&show_bookmark_button=0&recheck_access=1&view=${view}`;
   if (format) {
     iframeUrl += `&format=${format}`;
@@ -93,6 +106,8 @@ const Unit = ({
   const [showError, setShowError] = useState(false);
   const [modalOptions, setModalOptions] = useState({ open: false });
   const [shouldDisplayHonorCode, setShouldDisplayHonorCode] = useState(false);
+  const [examAccessToken, setExamAccessToken] = useState('');
+  const [blockExamAccess, setBlockExamAccess] = useState(isExam());
 
   const unit = useModel('units', id);
   const course = useModel('coursewareMeta', courseId);
@@ -140,6 +155,16 @@ const Unit = ({
     sendUrlHashToFrame(document.getElementById('unit-iframe'));
   }, [id, setIframeHeight, hasLoaded, iframeHeight, setHasLoaded, onLoaded]);
 
+  useEffect(() => {
+    if (isExam()) {
+      fetchExamAccess().finally(() => {
+        const examAccess = getExamAccess();
+        setExamAccessToken(examAccess);
+        setBlockExamAccess(false);
+      }).catch((error) => logError(error));
+    }
+  }, [id]);
+
   return (
     <div className="unit">
       <h1 className="mb-0 h3">{unit.title}</h1>
@@ -175,7 +200,7 @@ const Unit = ({
           <HonorCode courseId={courseId} />
         </Suspense>
       )}
-      {!shouldDisplayHonorCode && !hasLoaded && !showError && (
+      {(!shouldDisplayHonorCode || blockExamAccess) && !hasLoaded && !showError && (
         <PageLoading
           srMessage={intl.formatMessage(messages.loadingSequence)}
         />
@@ -208,12 +233,12 @@ const Unit = ({
           dialogClassName="modal-lti"
         />
       )}
-      {!shouldDisplayHonorCode && (
+      {!shouldDisplayHonorCode && !blockExamAccess && (
         <div className="unit-iframe-wrapper">
           <iframe
             id="unit-iframe"
             title={unit.title}
-            src={iframeUrl}
+            src={addExamAccessToIframeUrl(examAccessToken, iframeUrl)}
             allow={IFRAME_FEATURE_POLICY}
             allowFullScreen
             height={iframeHeight}
--- a/src/courseware/course/sequence/Unit.test.jsx
+++ b/src/courseware/course/sequence/Unit.test.jsx
@@ -1,9 +1,24 @@
 import React from 'react';
 import { Factory } from 'rosie';
+import { fetchExamAccess, getExamAccess, isExam } from '@edx/frontend-lib-special-exams';
 import {
   initializeTestStore, loadUnit, messageEvent, render, screen, waitFor,
 } from '../../../setupTest';
 import Unit, { sendUrlHashToFrame } from './Unit';
+
+const originalIsExam = jest.requireActual('@edx/frontend-lib-special-exams').isExam();
+const originalFetchExamAccess = jest.requireActual('@edx/frontend-lib-special-exams').fetchExamAccess();
+const originalGetExamAccess = jest.requireActual('@edx/frontend-lib-special-exams').getExamAccess();
+jest.mock('@edx/frontend-lib-special-exams', () => ({
+  ...jest.requireActual('@edx/frontend-lib-special-exams'),
+  isExam: jest.fn(),
+  fetchExamAccess: jest.fn(),
+  getExamAccess: jest.fn(),
+}));
+isExam.mockImplementation(() => originalIsExam).mockReturnValue(false);
+fetchExamAccess.mockImplementation(() => originalFetchExamAccess).mockResolvedValue();
+const examAccessToken = 'EXAMACCESSTOKEN';
+getExamAccess.mockImplementation(() => originalGetExamAccess).mockReturnValue(examAccessToken);
 
 describe('Unit', () => {
   let mockData;
@@ -173,5 +188,32 @@ describe('Unit', () => {
     render(<Unit {...mockData} />);
     expect(React.useEffect).toHaveBeenCalled();
     expect(mockHashCheck).toHaveBeenCalled();
+  });
+
+  it('updates url if exam and exam access granted', async () => {
+    isExam.mockReturnValue(true);
+
+    render(<Unit {...mockData} />);
+    expect(isExam).toHaveBeenCalled();
+    await waitFor(() => expect(screen.getByTitle(unit.display_name)).toHaveAttribute('src', `http://localhost:18000/xblock/${mockData.id}?show_title=0&show_bookmark_button=0&recheck_access=1&view=student_view&format=${mockData.format}&exam_access=${examAccessToken}`));
+  });
+
+  it('does not update url if exam and exam access not granted', async () => {
+    isExam.mockReturnValue(true);
+    fetchExamAccess.mockRejectedValue();
+    getExamAccess.mockReturnValue('');
+
+    render(<Unit {...mockData} />);
+    expect(isExam).toHaveBeenCalled();
+    await waitFor(() => expect(screen.getByTitle(unit.display_name)).toHaveAttribute('src', `http://localhost:18000/xblock/${mockData.id}?show_title=0&show_bookmark_button=0&recheck_access=1&view=student_view&format=${mockData.format}`));
+  });
+
+  it('does not update url if not exam', async () => {
+    isExam.mockReturnValue(false);
+
+    render(<Unit {...mockData} />);
+    expect(isExam).toHaveBeenCalled();
+    const renderedUnit = screen.getByTitle(unit.display_name);
+    expect(renderedUnit).toHaveAttribute('src', `http://localhost:18000/xblock/${mockData.id}?show_title=0&show_bookmark_button=0&recheck_access=1&view=student_view&format=${mockData.format}`);
   });
 });
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "@edx/brand": "npm:@edx/brand-openedx@1.2.0",
         "@edx/frontend-component-footer": "11.6.3",
         "@edx/frontend-component-header": "3.6.4",
-        "@edx/frontend-lib-special-exams": "2.10.0",
+        "@edx/frontend-lib-special-exams": "2.12.0",
         "@edx/frontend-platform": "3.4.1",
         "@edx/paragon": "20.28.4",
         "@fortawesome/fontawesome-svg-core": "1.3.0",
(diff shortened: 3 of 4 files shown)


src/courseware/course/sequence/Unit.jsx
```diff
@@ -83,11 +84,19 @@ const Unit = ({
...
+  if (exam.id) {
+    checkExamAccessToken();
+  }
+
   let iframeUrl = `${getConfig().LMS_BASE_URL}/xblock/${id}?show_title=0&show_bookmark_button=0&recheck_access=1&view=${view}`;
   if (format) {
     iframeUrl += `&format=${format}`;
   }
-
+  if (examAccessToken) {
```
zacharis278: I think there might be some async issues to consider here. What happens if the api call behind `checkExamAccessToken` isn't done yet?

src/courseware/course/sequence/Unit.test.jsx
```diff
@@ -174,4 +179,28 @@ describe('Unit', () => {
     expect(React.useEffect).toHaveBeenCalled();
     expect(mockHashCheck).toHaveBeenCalled();
   });
+
+  it('calls fetchExamAccess if isExam upon block id', () => {
```
varshamenon4: Is this the right approach to testing this? That is, mocking the functions and making sure they are called? Have some additional tests which still need to be implemented below but could use a look to make sure it's in the right direction.

src/index.jsx
```diff
@@ -123,6 +123,7 @@ initialize({
         ENTERPRISE_LEARNER_PORTAL_HOSTNAME: process.env.ENTERPRISE_LEARNER_PORTAL_HOSTNAME || null,
         ENABLE_JUMPNAV: process.env.ENABLE_JUMPNAV || null,
         ENABLE_NOTICES: process.env.ENABLE_NOTICES || null,
+        EXAMS_BASE_URL: process.env.EXAMS_BASE_URL || null,
```
varshamenon4: Not sure why this dupe key error is happening? I tried to add the merge config in tests but wondering if this has to do with it being in both frontend-lib-special-exams and frontend-app-learning as env variables. And if this is the case, I'm not sure if it's better to completely get rid of in one place or another.

src/courseware/course/sequence/Unit.jsx
```diff
@@ -1,6 +1,7 @@
 import { getConfig } from '@edx/frontend-platform';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import { AppContext, ErrorPage } from '@edx/frontend-platform/react';
+import { getExamAccess, fetchExamAccess, isExam } from '@edx/frontend-lib-special-exams';
```
varshamenon4, line 5: When I run the tests locally on the IDE, I get an error that these functions cannot be found. I'm wondering if it has to do with this export/import being reflected in my branch of frontend-lib-special-exams and not the main branch?

varshamenon4 (commented):
Some clarifying questions for review!

src/courseware/course/sequence/Unit.test.jsx
zacharis278: yeah this looks correct to me

src/index.jsx
zacharis278: why is EXAMS_BASE_URL needed here?

src/courseware/course/sequence/Unit.jsx
zacharis278: I think that's it, likely has to do with installing via module.config vs. what's in package.json

varshamenon4: ahhh okay.

src/index.jsx
varshamenon4: I believe because we want to merge config in frontend-app-learning with these other settings--though it's not called in frontend-app-learning. Since frontend-app-learning already imports the edx frontend platform dependency.

zacharis278:
@varshamenon4 not sure if you already plan to have this covered but I think we're missing some tests around how the blockExamAccess flag works.

src/courseware/course/sequence/Unit.test.jsx
```diff
@@ -174,4 +175,42 @@ describe('Unit', () => {
...
+    const mockFetchExamAccess = jest.fn(() => fetchExamAccess());
+
+    const effectSpy = jest.spyOn(React, 'useEffect');
+    effectSpy.mockImplementation(() => {
+      if (mockIsExam()) {
+        mockFetchExamAccess();
+      }
+    });
+
+    render(<Unit {...mockData} />);
```
varshamenon4, line 196: @Zacharis278 still getting used to jest testing, but is this the correct way to test the isExam and fetchExamAccess calls? And yes, like you mentioned I still need to test if blockExamAccess is working correctly. I'm guessing this would happen my mocking those other functions and their return values and making sure that the blockExamAccess is set correctly? And then also testing that the iframe url either has or doesn't have the exam access token?

src/courseware/course/Course.test.jsx
```diff
@@ -16,6 +17,13 @@ import * as thunks from '../data/thunks';
 
 jest.mock('@edx/frontend-platform/analytics');
 
+// const originalIsExam = jest.requireActual('@edx/frontend-lib-special-exams').isExam();
```
varshamenon4: I tried mocking some of the pieces of the Unit that were added to hopefully resolve the tests in Course and Sequence but that didn't seem to work. I am looking into the errors now. It's showing an unexpected error but I'm also seeing console errors about incorrect return type for Course and Sequence so going to dig into that.

src/courseware/course/sequence/Unit.jsx
```diff
@@ -180,6 +203,11 @@ const Unit = ({
           srMessage={intl.formatMessage(messages.loadingSequence)}
         />
       )}
+      {!showError && !hasLoaded && blockExamAccess && (
+        <PageLoading
+          srMessage={intl.formatMessage(messages.loadingSequence)}
```
varshamenon4, line 205: This just continues to render the loading sequence like above but I'm wondering if there should be an error like the code below or if this is the right expected behavior for when exam access is blocked.

src/courseware/course/sequence/Unit.test.jsx
```diff
@@ -174,4 +189,37 @@ describe('Unit', () => {
...
+    expect(React.useEffect).toHaveBeenCalled();
+    expect(isExam).toHaveBeenCalled();
+    const renderedUnit = screen.getByTitle(unit.display_name);
+    expect(renderedUnit).toHaveAttribute('src', `http://localhost:18000/xblock/${mockData.id}?show_title=0&show_bookmark_button=0&recheck_access=1&view=student_view&format=${mockData.format}&exam_access=${examAccessToken}`);
+  });
+
+  it('updates url with token if exam and exam access granted', () => {
+    isExam.mockReturnValue(true);
+    // todo: should fetch
```
varshamenon4: Not sure how to mock when exam access is not granted. I'm thinking of returning a rejected Promise, but I think I'll need to update the mockFetchExamAccess to handle both resolved/rejected cases.

src/courseware/course/sequence/Unit.test.jsx
```diff
@@ -174,4 +189,37 @@ describe('Unit', () => {
...
     expect(mockHashCheck).toHaveBeenCalled();
   });
+
+  it('updates url if exam and exam access granted', () => {
+    isExam.mockReturnValue(true);
+
+    render(<Unit {...mockData} />);
+    expect(React.useEffect).toHaveBeenCalled();
+    expect(isExam).toHaveBeenCalled();
+    const renderedUnit = screen.getByTitle(unit.display_name);
```
varshamenon4: Getting an error here as it's just leading to the loading sequence. When I was debugging I could see that the block exam access was successfully set to false when exam access is granted, but there seemed to be some async issue happening. I'm wondering if I'm missing something on mocking the Promise.

varshamenon4 (commented):
Had some questions about testing that I could use some eyes on!

src/courseware/course/sequence/Unit.jsx
zacharis278: I see what you mean, I think fetchExamAccess in our useEffect hook should always do `setBlockExamAccess(false)` after the promise returns. That way the user gets an error and not an infinite spinner. You could accomplish that by simple changing `fetchExamAccess().then(() => {` to `fetchExamAccess().finally(() => {`

src/courseware/course/Course.test.jsx
zacharis278: ok I see what's happening here. We could mock out the the api into special-exams library but I think that's the wrong way to go about things because it would force ever place we import this to do the same. What's happening is a simple null check problem is frontend-lib-special-exams because exam state isn't initialized. IMO because api.js is an external interface it should be resilient to that.

I think a simple null check in the isExam function would solve this but to avoid issues in the future id recommend:
if no exam state
- isExam - return false
- getExamAccess - return empty string + log an error
- fetchExamAccess- return failed promise + log an error

src/courseware/course/sequence/Unit.test.jsx
zacharis278: Yup your render here causes a promise to fire (fetchExamAccess) inside useEffect because isExam is true. So, your test needs to handle that by either wrapping the render in something like act() or having screen.getByTitle wait until the iframe loads with the token.

Right now your test can't find the iframe because on the first draw of the Unit it's still in the loading state

src/courseware/course/Course.test.jsx
varshamenon4: Ah I see @Zacharis278. And to clarify, this change would be in isExam itself/frontend-lib-special-exams. And you're suggesting doing the bullet points for each api function to be safe, so isExam, getExamAccess, and fetchExamAccess respectively?

src/courseware/course/sequence/Unit.jsx
varshamenon4: Okay so I'll update this and also update to render the error instead of page loading?

zacharis278: I don't even think you _need_ to do anything in the UI here since the xblock itself is going to render its own error. It might be nicer but in the interest of getting things across the line the default xblock error should be enough.

src/courseware/course/Course.test.jsx
zacharis278: yeah it would have to be in the frontend-lib-special-exams

src/courseware/course/sequence/Unit.test.jsx
```diff
@@ -174,4 +189,35 @@ describe('Unit', () => {
...
+  it('updates url if exam and exam access granted', async () => {
+    isExam.mockReturnValue(true);
+
+    render(<Unit {...mockData} />);
+    expect(React.useEffect).toHaveBeenCalled();
+    expect(isExam).toHaveBeenCalled();
+    await waitFor(() => expect(screen.getByTitle(unit.display_name)).toHaveAttribute('src', `http://localhost:18000/xblock/${mockData.id}?show_title=0&show_bookmark_button=0&recheck_access=1&view=student_view&format=${mockData.format}&exam_access=${examAccessToken}`));
+  });
+
+  it('does not update url if exam and exam access not granted', async () => {
```
varshamenon4, line 201: Could you take a look at this test? Not sure how to go about testing this case. I tried adding a catch to the useEffect, but I'm not sure that that is correct, as it seems to resolve and set the exam access token when I test this.

src/courseware/course/Course.test.jsx
varshamenon4: Added to frontend-lib-special-exams in this PR: https://github.com/openedx/frontend-lib-special-exams/pull/97

src/courseware/course/sequence/Unit.test.jsx
varshamenon4: Fixed.

src/courseware/course/sequence/Unit.jsx
```diff
@@ -140,6 +155,17 @@ const Unit = ({
     sendUrlHashToFrame(document.getElementById('unit-iframe'));
   }, [id, setIframeHeight, hasLoaded, iframeHeight, setHasLoaded, onLoaded]);
 
+  useEffect(() => {
+    if (isExam()) {
+      fetchExamAccess().finally(() => {
+        setExamAccessToken(getExamAccess());
```
varshamenon4: Found some bugginess with this switch in logic (below, adding the check for exam access token before setting block exam access), that this is now not loading because the exam access token isn't set (but I believe this is some type of syncing issue).

varshamenon4: Basically, the exam access token is eventually set, but when I check for it in line 61, it's showing up as blank. As I had it before (when it was working), the blockExamAccess was always set to false when an exam. But since now with the finally after fetch exam access, it should always get to blockExamAccess set to false. I'm going to try hacking the error case to see if it never gets to the finally in this case.

varshamenon4: I now don't think this is an issue. Because when I test that fetchExamAccess returns a rejected value, then it shouldn't get to the finally block. So I don't need the logic altogether to check the examAccessToken within the useEffect block, in the finally(). I just need to test/hack the case where the exam access token is not granted. But lemme know if you think I should investigate.

src/courseware/course/sequence/Unit.jsx
```diff
@@ -140,6 +155,16 @@ const Unit = ({
     sendUrlHashToFrame(document.getElementById('unit-iframe'));
   }, [id, setIframeHeight, hasLoaded, iframeHeight, setHasLoaded, onLoaded]);
 
+  useEffect(() => {
+    if (isExam()) {
+      fetchExamAccess().finally(() => {
+        const examAccess = getExamAccess();
+        setExamAccessToken(examAccess);
+        setBlockExamAccess(false);
+      }).catch((error) => logError(error));
```
varshamenon4, line 164: Need to catch an error in case of a rejected promise returned. But lemme know if there is a better approach.

src/courseware/course/sequence/Unit.jsx
```diff
@@ -180,6 +205,11 @@ const Unit = ({
           srMessage={intl.formatMessage(messages.loadingSequence)}
         />
       )}
+      {!showError && !hasLoaded && blockExamAccess && (
```
varshamenon4: Just wanted to confirm that these are the conditions we want met to have loading sequence?

varshamenon4 (commented):
Ready for re-review.

src/courseware/course/sequence/Unit.jsx
zacharis278: this is a 'we should never hit this line' type error right? In which case, this looks fine.

src/courseware/course/sequence/Unit.jsx
zacharis278: I think this logic leaves a gap if blockExamAccess is false (which renders the iframe) but the iframe content isn't loaded yet (so hasLoaded==False). We might get two spinners in that case because the statement above also creates one if !hasLoaded.

So maybe `!showError && (!hasLoaded || blockExamAccess)`, or better yet combining this logic with the above avoid the issue of something unintentionally rendering PageLoading twice entirely.

zacharis278 (changes requested):

varshamenon4:
Yup, you're right on this creating 2 spinners. I noticed that when I had the logic to block exam access and it was creating two spinners. I'll update. @Zacharis278 

https://github.com/openedx/frontend-app-learning/pull/1083#discussion_r1165875958

src/courseware/course/sequence/Unit.jsx
varshamenon4: Yup, you're right on this creating 2 spinners. I noticed that when I had the logic to block exam access and it was creating two spinners. I'll update. @Zacharis278

src/courseware/course/sequence/Unit.jsx
varshamenon4: Yeah we shouldn't. I was just getting some warnings when I tried to test having fetchExamAccess return an rejected promise.

src/courseware/course/sequence/Unit.jsx
varshamenon4: Updated the logic!

src/courseware/course/sequence/Unit.test.jsx
```diff
@@ -125,7 +140,7 @@ describe('Unit', () => {
     render(<Unit {...mockData} />);
     window.postMessage(testMessageWithOffset, '*');
 
-    await expect(waitFor(() => expect(window.scrollTo()).toHaveBeenCalled()));
+    expect(waitFor(() => expect(window.scrollTo()).toHaveBeenCalled()));
```
zacharis278: is this change intentional?

varshamenon4: Ah I was testing something and had commented it out but I forgot to add back the await. Fixing now

varshamenon4: Fixed!

zacharis278 (approved):

PR #1090, opened 2023-03-23
feat: create component to decode params

leangseu-edx:
TICKET: https://2u-internal.atlassian.net/browse/AU-1076

What changed:
- redirect encoded params

How to test:
visit:
  - http://localhost:2000/course/course-v1:edX+DemoX+Demo_Course/home
  - http://localhost:2000/course%2Fcourse-v1%3AedX%2BDemoX%2BDemo_Course/home
  - http://localhost:2000course-v1%2525253AedX%2525252BDemoX%2525252BDemo_Course/home

All should land you at  http://localhost:2000/course/course-v1:edX+DemoX+Demo_Course/home

### DIFF
--- a/src/courseware/CoursewareRedirectLandingPage.jsx
+++ b/src/courseware/CoursewareRedirectLandingPage.jsx
@@ -7,6 +7,8 @@ import { PageRoute } from '@edx/frontend-platform/react';
 import queryString from 'query-string';
 import PageLoading from '../generic/PageLoading';
 
+import DecodePageRoute from '../decode-page-route';
+
 const CoursewareRedirectLandingPage = () => {
   const { path } = useRouteMatch();
   return (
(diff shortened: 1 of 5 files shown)


schenedx (approved):
👍 Nice unit tests.
This is a risky change. I'd like to run some tests on stage. Please work with your team to manual test this on stage.

nsprenkle:
Have we run this past frontend working group? I feel like we can't be the only ones dealing with this issue and it feels like we should make sure this is a pattern that aligns with our frontend wants.

leangseu-edx:
@nsprenkle what do you mean by align with our frontend?
Do you think this is a feature that worth sharing in a common library? Or do you think this change going to break other code?

muselesscreator (approved):
LGTM.  We may want a non-rewriting-path option in the future if this winds up being over-zealous, but for now it seems a reasonable enough solution to avoid having to do larger overrides to the PageRoute component logic

PR #1097, opened 2023-04-06
feat: upgrade special exams version and add required config values

alangsto:
Upgrade to the latest version of the special exams lib, which includes these changes: https://github.com/openedx/frontend-lib-special-exams/compare/feb32dd...a37b9fa

The new version requires two new config variables to be set, so those have been added as part of this PR.


package.json
```diff
@@ -32,7 +32,7 @@
     "@edx/brand": "npm:@edx/brand-openedx@1.2.0",
     "@edx/frontend-component-footer": "11.6.3",
     "@edx/frontend-component-header": "3.6.4",
-    "@edx/frontend-lib-special-exams": "~2.4.0",
+    "@edx/frontend-lib-special-exams": "2.8.0",
```
zacharis278: maybe we should just keep this constrained throughout the project so we control what gets pulled in?

zacharis278 (approved):

PR #1111, opened 2023-05-09
fix: gracefully handle 403 responses in tab loading

jansenk:
From https://github.com/openedx/edx-platform/pull/32210 :
> The issue:
>
> When a learner navigates to a Learning MFE course overview page, we hit (the course outline) API.(the course outline) API uses get_course_with_access to check whether or not the learner can access this course. If not, get_course_with_access raises a CourseAccessError, which causes the request to return a 302 and a URL to redirect to.
>
> This is useful in cases where the user is requesting a web page - they will be automatically redirected to another page depending on their state. However, this is being requested as an API request. When the error is raised, we then attempt to make an API request to various other sites. In this case, it's the learner home, which then requests the auth service, and causes the CORS. The CORS is an error, but it's not really the issue here. We shouldn't be making an ajax request to the redirect URL at all. We should be redirecting the user's web browser to the redirect URL.

The above PR is the change to the outline API to prevent the auto redirect and to instead return a 403.

There is an existing mechanism in the TabPage component to handle this redirect. Redirect information is passed via the  course metadata endpoint, and all we need to do is not raise an exception when loading the outline. We have the same behavior in the progress and outline views as well, so I also made this change there, to expect and ignore the 403. 

Testing instructions in the edx-platform PR

### DIFF
--- a/src/course-home/data/api.js
+++ b/src/course-home/data/api.js
@@ -204,10 +204,16 @@ export async function getDatesTabData(courseId) {
     const { data } = await getAuthenticatedHttpClient().get(url);
     return camelCaseObject(data);
   } catch (error) {
-    const { httpErrorStatus } = error && error.customAttributes;
+    const httpErrorStatus = error?.response?.status;
     if (httpErrorStatus === 401) {
       // The backend sends this for unenrolled and unauthenticated learners, but we handle those cases by examining
       // courseAccess in the metadata call, so just ignore this status for now.
+      return {};
+    }
+    if (httpErrorStatus === 403) {
+      // The backend sends this if there is a course access error and the user should be redirected. The redirect
+      // info is included in the course metadata request and will be handled there as long as this call returns
+      // without an error
       return {};
     }
     throw error;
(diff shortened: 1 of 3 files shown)


src/course-home/data/api.js
```diff
@@ -330,6 +330,11 @@ export async function getOutlineTabData(courseId) {
     headers,
   } = tabData;
 
+  if (data.course_access_redirect) {
+    logInfo(`Redirected away from course outline: ${data.error_code} - ${data.developer_message}`);
+    window.location.replace(data.url);
```
schenedx: #NIT can you add in code comment to spell out why you are doing this `windows.location.replace` call?
It is clear from your PR description. It's not clear just reading the code.

schenedx (approved):
👍

muselesscreator (approved):
LGTM

nsprenkle (dismissed):
This seems semantically like the wrong thing to do. If I'm asking for access to a course and I don't have it, it seems like I would like a `401 - Unauthorized`, no? Not a `302`. And then it should be the responsibility of the MFE to redirect, not the API.

leangseu-edx:
> This seems semantically like the wrong thing to do. If I'm asking for access to a course and I don't have it, it seems like I would like a `401 - Unauthorized`, no? Not a `302`. And then it should be the responsibility of the MFE to redirect, not the API.

302 just mean redirect. It make sense if you go to `unauth.com` then to redirect you to `unauth.com/login`. However 302 as an api doesn't really make much sense.

I agree with @nsprenkle that the backend should raise `Unauthorized` then we should wrap `await getAuthenticatedHttpClient().get(url);` in try/catch for the frontend.

schenedx (approved):
👍

mattcarter (approved):
approved

leangseu-edx (approved):

PR #1112, opened 2023-05-15
fix: logging an error when unit iframe fails to load

davidjoy:
Right now we log nothing to the logging service when a unit iframe fails to load.  The ErrorPage that’s shown isn’t using the ErrorBoundary, so we had no indication that something went wrong.  This solves the problem closer to the source where the error originates.

This PR also bumps frontend-platform to 4.3.0 to pick up the new feature that automatically logs the userId to the logging service whenever an error occurs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "@edx/frontend-component-footer": "12.0.0",
         "@edx/frontend-component-header": "4.0.0",
         "@edx/frontend-lib-special-exams": "2.16.2",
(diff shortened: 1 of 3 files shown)


robrap (commented):
Nice. Do we need a test, or too tricky if this isn't yet tested?

adzuci (approved):

davidjoy:
Bummer, but I don't think I can write a test for this with the current code structure, especially given that it's in the onLoad of an iframe loading in a unit test.  There are no other tests for that code.

PR #1131, opened 2023-06-22
fix: modifies sidebar state such that it remains open

ayesha-waris:
[INF-915](https://2u-internal.atlassian.net/browse/INF-915)

- Removed state tracking and such that the sidebar always opens when a user navigates to a unit.
- The close button should still allow users to close the sidebar, but it should re-open when user reloads the page or navigates to another unit that has discussions enabled.

### DIFF
--- a/src/courseware/course/sidebar/SidebarContextProvider.jsx
+++ b/src/courseware/course/sidebar/SidebarContextProvider.jsx
@@ -40,7 +40,7 @@ const SidebarProvider = ({
       setCurrentSidebar(initialSidebar);
     }
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [initialSidebar, verifiedMode]);
+  }, [initialSidebar, verifiedMode, unitId]);
 
   const onNotificationSeen = useCallback(() => {
     setNotificationStatus('inactive');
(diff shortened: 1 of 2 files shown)


src/courseware/course/sidebar/SidebarContextProvider.jsx
```diff
@@ -40,7 +40,7 @@ const SidebarProvider = ({
       setCurrentSidebar(initialSidebar);
```
sundasnoreen12: No need of line 26 const showDiscussionSidebar = localStorage.getItem('showDiscussionSidebar') !== 'false';

src/courseware/course/sidebar/SidebarContextProvider.jsx
```diff
@@ -40,7 +40,7 @@ const SidebarProvider = ({
...
     }
```
sundasnoreen12, line 32: for line 30-32 use this
const initialSidebar = SIDEBARS.DISCUSSIONS.ID.

src/courseware/course/sidebar/SidebarContextProvider.jsx
sundasnoreen12: remove check for line 23

sundasnoreen12 (changes requested):

sundasnoreen12 (approved):
PR LGTM

awais-ansari (approved):

muhammadadeeltajamul (approved):

PR #1148, opened 2023-07-28
feat: bump to frontend-build alpha

muselesscreator:
Proof-of-concept consumption of the alpha branch of `frontend-build` without actually using or locally configuring typescript.

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -7,4 +7,14 @@ on:
 
 jobs:
   commitlint:
-    uses: openedx/.github/.github/workflows/commitlint.yml@master
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: remove tsconfig.json # see issue https://github.com/conventional-changelog/commitlint/issues/3256
+        run: |
+          rm -f tsconfig.json
+      - name: Check commits
+        uses: wagoid/commitlint-github-action@v5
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -77,5 +77,5 @@ validate:
 
 .PHONY: validate.ci
 validate.ci:
-	npm ci
+	npm ci --legacy-peer-deps
```
leangseu-edx, line 80: I disagree with doing this. We should resolve the dependency.

muselesscreator: this is a proof-of-concept and that is not feasible right now.  This is the currently-agreed-on workaround and is only necessary until we merge alpha into master.

.github/workflows/commitlint.yml
```diff
@@ -7,4 +7,14 @@ on:
...
 jobs:
   commitlint:
-    uses: openedx/.github/.github/workflows/commitlint.yml@master
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: remove tsconfig.json # see issue https://github.com/conventional-changelog/commitlint/issues/3256
```
marlonkeating, line 16: Do we need this step before introducing a tsconfig?

muselesscreator: no, just leaving it here for when we do add one.  hence the `-f`

bradenmacdonald:
I believe this PR is no longer useful, so closing it out. We are on a new version of frontend-build now, with lots of nice features including optional typescript.

PR #1163, opened 2023-08-17
feat: add config value for chat response url

alangsto:
## [MST-2041](https://2u-internal.atlassian.net/browse/MST-2041)

Expose config value so that it can be consumed by `frontend-lib-learning-assistant`, which will be installed in the learning MFE.

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -137,6 +137,7 @@ initialize({
         EXAMS_BASE_URL: process.env.EXAMS_BASE_URL || null,
         PROCTORED_EXAM_FAQ_URL: process.env.PROCTORED_EXAM_FAQ_URL || null,
         PROCTORED_EXAM_RULES_URL: process.env.PROCTORED_EXAM_RULES_URL || null,
+        CHAT_RESPONSE_URL: process.env.CHAT_RESPONSE_URL || null,
       }, 'LearnerAppConfig');
     },
   },


src/index.jsx
```diff
@@ -137,6 +137,7 @@ initialize({
         EXAMS_BASE_URL: process.env.EXAMS_BASE_URL || null,
         PROCTORED_EXAM_FAQ_URL: process.env.PROCTORED_EXAM_FAQ_URL || null,
         PROCTORED_EXAM_RULES_URL: process.env.PROCTORED_EXAM_RULES_URL || null,
+        CHAT_RESPONSE_URL: process.env.CHAT_RESPONSE_URL || null,
```
michaelroytman, line 140: We'll need to add this to [setUpTest.js](https://github.com/openedx/frontend-app-learning/blob/5ee61904d5a99b58e4d749d7612bdb88d519f98b/src/setupTest.js#L54) too, right? I can also do that as part of my work, if you'd prefer.

alangsto: The value should only be added there if it's being used in testing. There's nothing using it right now which is why I opted out!

michaelroytman: When I render the component during tests, I'll need that to be set, so I can set it in my pull request, then - just thinking ahead!

michaelroytman (approved):

PR #1165, opened 2023-08-21
fix: should not fetch courseTopics for lti provider

awais-ansari:
### [INF-1024](https://2u-internal.atlassian.net/browse/INF-1024)

- [course-topics](api/discussion/v1/courses/course-id) This API call raises an error when the discussion provider is set to LTI. e.g Yellowdig. This API call should only call when the discussion provider is edx.
- Error - "Discussion is disabled for the course."

### DIFF
--- a/src/courseware/course/sidebar/sidebars/discussions/DiscussionsTrigger.jsx
+++ b/src/courseware/course/sidebar/sidebars/discussions/DiscussionsTrigger.jsx
@@ -3,7 +3,7 @@ import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import { Icon } from '@edx/paragon';
 import { QuestionAnswer } from '@edx/paragon/icons';
 import PropTypes from 'prop-types';
-import React, { useContext, useEffect } from 'react';
+import React, { useContext, useEffect, useMemo } from 'react';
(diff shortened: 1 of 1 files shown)


sundasnoreen12 (approved):
PR LGTM

src/courseware/course/sidebar/sidebars/discussions/DiscussionsTrigger.jsx
```diff
@@ -23,12 +23,17 @@ const DiscussionsTrigger = ({
...
   const baseUrl = getConfig().DISCUSSIONS_MFE_BASE_URL;
+  const ltiProvider = useMemo(
+    () => tabs?.find(tab => tab.slug === 'lti_discussion'),
+    [tabs],
+  );
 
   useEffect(() => {
     // Only fetch the topic data if the MFE is configured.
-    if (baseUrl) {
+    if (baseUrl && !ltiProvider) {
```
AhtishamShahid: We should check if this tab exists instead of checking lit tab
```
{
    "tab_id": "discussion",
    "title": "Discussion",
    "url": "https://discussions.stage.edx.org/course-v1:edx+Incontext01+run02/"
}
```

AhtishamShahid (approved):

PR #1166, opened 2023-08-22
Install Xpert chatbot from frontend-lib-learning-assistant.

michaelroytman:
### Description

This commit installs the Xpert chatbot feature from the [frontend-lib-learning-assistant](https://github.com/edx/frontend-lib-learning-assistant) repository into the [frontend-lib-learning](https://github.com/openedx/frontend-app-learning) application.

This component is rendered by the Course component. The component is only rendered when a few conditions are satisfied.

**JIRA**: [MST-2034](https://2u-internal.atlassian.net/browse/MST-2034) (private)

### Demo

**NOTE**: Please ignore the line over the video camera icon. This was fixed in https://github.com/edx/frontend-lib-learning-assistant/pull/12.

https://github.com/openedx/frontend-app-learning/assets/11871801/99d3e0c0-eafa-40a9-ad8f-840f9b514180


src/courseware/course/Course.jsx
```diff
@@ -93,10 +93,10 @@ const Course = ({
         />
         {shouldDisplayTriggers && (
           <>
-            <ChatTrigger
+            <Chat
+              enabled={course.learningAssistantEnabled}
```
michaelroytman, line 97: I'd prefer to read these values from the Redux store, but it'll make testing more difficult because I'll have to set up that part of the Redux store, so I've chosen to use props here in the interest of getting this out more quickly.

src/courseware/course/chat/Chat.jsx
```diff
@@ -0,0 +1,55 @@
...
+  enabled: PropTypes.bool.isRequired,
+  enrollmentMode: PropTypes.string,
+  courseId: PropTypes.string.isRequired,
+};
+
+Chat.defaultProps = {
+  enrollmentMode: null,
+};
+
+export default injectIntl(Chat);
```
michaelroytman, line 55: I'm not sure if this is really necessary, since there's no text to internationalize right now.

src/courseware/course/chat/Chat.test.jsx
```diff
@@ -0,0 +1,157 @@
...
+
+import Chat from './Chat';
+
+jest.mock('@edx/frontend-platform/analytics');
+
+initializeMockApp();
+
+const courseId = 'course-v1:edX+DemoX+Demo_Course';
+let testCases = [];
+let enabledTestCases = [];
```
michaelroytman, line 17: I tried to refactor the test to generate the test cases instead of declaring them outright. This is because the conditions under which the Xpert chatbot should appear have grown more complex, so it's easier to generate the objects. Let me know if this makes it more confusing to read.

src/courseware/course/chat/Chat.test.jsx
```diff
@@ -0,0 +1,157 @@
...
+              enrollmentMode={test.enrollmentMode}
+              isStaff={false}
+              enabled
+              courseId={courseId}
+            />
+          </BrowserRouter>,
+          { store },
+        );
+
+        // This is rather brittle and depends on the implementation of the Xpert component, but there
```
michaelroytman: Any other ideas here?

src/setupTest.js
```diff
@@ -116,6 +117,7 @@ export async function initializeTestStore(options = {}, overrideStore = true) {
       models: modelsReducer,
       courseware: coursewareReducer,
       courseHome: courseHomeReducer,
+      learningAssistant: learningAssistantReducer,
```
michaelroytman, line 120: I'm a little confused why the `learningAssistantReducer` is necessary, but the `tours` or `recommendations` reducers aren't. Without the `learningAssistantReducer`, other tests were breaking, so I added it.

src/courseware/course/chat/Chat.jsx
```diff
@@ -0,0 +1,55 @@
...
+  );
+
+  const shouldDisplayChat = (
+    enabled
+    && (isVerifiedEnrollmentMode || isStaff) // display only to non-audit or staff
+  );
+
+  return (
+    <>
+      {/* Use a portal to ensure that component overlay does not compete with learning MFE styles. */}
```
alangsto, line 35: Thank you adding this comment!

src/courseware/course/chat/Chat.test.jsx
alangsto: Is there an ID that we can query by instead? I'm hesitant to query by button name because that could change

michaelroytman: We could add a `data-testid` property to the button that we could query for. I'm a little torn, because I think we _should_ be querying by roles like button, because it's a more representative and accessible way of testing. And I wonder if these tests _should_ break if we change the text of the button. But, on the other hand, these tests aren't meant to test how the toggle button looks and behaves.

Actually, now that I think of it, since I'm only rendering the `Xpert` in the test, there should only be a single button in the DOM, so I could get away with this.

```const chat = screen.queryByRole('button');```

What do you think of that? I feel like it's more reasonable to assume there is a button in the DOM than a button with a particular "name". If you prefer the test ID, I can work on adding that.

alangsto: If it's only rendering the Xpert Component in the test, I like the idea of querying by the button role alone, as opposed to a button with a particular name. That seems like an appropriate solution to me that is definitely less brittle.

michaelroytman:
@alangsto I addressed your feedback. Let me know what you think!

alangsto (approved):
Looks good to me!

arbrandes:
Hey there!  This just came to my attention.  Is there an OEP, ADR, or other documentation (including any community discussions) that describe the intention behind this new functionality?

PR #1180, opened 2023-09-07
feat: grant audit learners access

alangsto:
## [MST-2130](https://2u-internal.atlassian.net/browse/MST-2130)

The learning assistant should be visible to audit learners in courses for which the assistant has been enabled.

### DIFF
--- a/src/courseware/course/chat/Chat.jsx
+++ b/src/courseware/course/chat/Chat.jsx
@@ -18,17 +18,26 @@ const Chat = ({
     'credit',
     'masters',
     'executive-education',
+    'paid-executive-education',
+    'paid-bootcamp',
   ];
 
-  const isVerifiedEnrollmentMode = (
+  const AUDIT_MODES = [
+    'audit',
+    'honor',
+    'unpaid-executive-education',
+    'unpaid-bootcamp',
+  ];
+
+  const isEnrolled = (
     enrollmentMode !== null
     && enrollmentMode !== undefined
-    && VERIFIED_MODES.some(mode => mode === enrollmentMode)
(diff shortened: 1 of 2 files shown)


src/courseware/course/chat/Chat.jsx
```diff
@@ -20,15 +20,22 @@ const Chat = ({
     'executive-education',
   ];
 
-  const isVerifiedEnrollmentMode = (
+  const AUDIT_MODES = [
```
michaelroytman, line 25: Nit. My only comment is there may be a mismatch between these modes and those on the backend.

I noticed that a `paid-bootcamp` mode is missing from `VERIFIED_MODES`. I also noticed that there are three Executive Education modes: `paid-executive-education`, `unpaid-executive-education`, and `executive-education`, and we're missing the `paid-executive-education` mode from `VERIFIED_MODES`. I don't know if `paid-executive-education` and `executive-education` resolve to the same mode at some point (e.g. `executive-education`), but it's something I noticed.

This isn't an issue for our release right now because of the modes available in the alpha courses, and nothing catastrophic will happen as a result of the mismatch, but I think it's worth taking a quick look at.

michaelroytman (approved):

src/courseware/course/chat/Chat.jsx
alangsto: Let me take a closer look to make sure it matches up

alangsto: Just updated `VERIFIED_MODES` to include both `paid-bootcamp` and `paid-executive-education`

PR #1199, opened 2023-10-04
Added waffle flag state for Courseware Search

rijuma:
## Description
Ticket: [KBK-36](https://2u-internal.atlassian.net/browse/KBK-36) (internal)
This relates to an `edx-platform` PR which adds the endpoint to the LMS (https://github.com/openedx/edx-platform/pull/33372)

There's a project to add `edx-search` into the course home. Right now I'm creating a placeholder button which will respond to the Course Waffle Flag toggle of `courseware.mfe_courseware_search`.

## UI changes
search-placeholder

If the flag is enabled, a search button will be shown on the course's home navigation. I'm using the `float` property to position itself, which might sound controversial. This was the easiest way to achieve it since the `CourseTabsNavigation` component has a wrapping logic whenever the navigation items don't fit. I've tried to move the navigation to a flex container but it just messes up with this logic. Luckily, the button is small enough so it does not interfere with the remaining space after folding.

search-plaheholder-2

## Testing Instructions:

Navigate to the LMS Admin and toggle `courseware.mfe_courseware_search` flag via:
- Waffle Flags: `http://(...)/admin/waffle/flag/`
- Course Waffle Flags: `http://(...)/admin/waffle_utils/waffleflagorgoverridemodel/`

## Results:

You should be able to see the search placeholder or not (see screenshot above) depending on the enabled state for the Waffle Flag.

### DIFF
--- a/src/course-home/courseware-search/CoursewareSearch.test.jsx
+++ b/src/course-home/courseware-search/CoursewareSearch.test.jsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import {
+  act,
+  initializeMockApp,
+  render,
+  screen,
+  waitFor,
+} from '../../setupTest';
+import { fetchCoursewareSearchSettings } from '../data/thunks';
+import { CoursewareSearch } from './index';
+
+jest.mock('../data/thunks');
+
+function renderComponent() {
+  const { container } = render(<CoursewareSearch />);
+  return container;
+}
+
+describe('CoursewareSearch', () => {
+  beforeAll(async () => {
+    initializeMockApp();
+  });
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('Should not render when the waffle flag is disabled', async () => {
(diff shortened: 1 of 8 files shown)


src/course-home/courseware-search/CoursewareSearch.test.jsx
```diff
@@ -0,0 +1,46 @@
...
+    });
+  });
+
+  it('Should render when the waffle flag is enabled', async () => {
+    fetchCoursewareSearchSettings.mockImplementation(() => Promise.resolve({ enabled: false }));
+    await act(async () => renderComponent());
+    await waitFor(() => {
+      expect(fetchCoursewareSearchSettings).toHaveBeenCalledTimes(1);
+      expect(screen.queryByTestId('courseware-search-button')).not.toBeInTheDocument();
+    });
```
rijuma: This should be the `true` case. Updating test.

schenedx (approved):
👍

PR #1212, opened 2023-10-16
Added Courseware Search container popover

rijuma:
## Description
Ticket: [KBK-42 🔒](https://2u-internal.atlassian.net/browse/KBK-42)

We need a [modal 🔒](https://2u-internal.atlassian.net/wiki/spaces/Kubrick/pages/614432780/Component+breakdown+based+on+Figma#Full-Screen-Modal) to show the search components and the results.

> [!IMPORTANT]
> I've added `@testing-library/react-hooks@^8.0.1` dependency. This allows testing standalone hooks rather than creating a testing component to use the hook. It makes it much cleaner to test. 

### instructor view

| 480px wide  | 1024px wide |
| --- | --- |
| search-w480-instructor | search-w1024-instructor |

### Student view

| 480px wide  | 1024px wide |
| --- | --- |
| search-w480-student | search-w1024-student |

The Design is a bit peculiar, and I couldn't match the UI using Paragon's [FullscreenModal](https://paragon-openedx.netlify.app/components/modal/fullscreen-modal/), [ModalLayer](https://paragon-openedx.netlify.app/components/modal/modal-layer/) nor [ModalDialog](https://paragon-openedx.netlify.app/components/modal/modal-dialog/).

The UI has a modal-like overlay without a backdrop that anchors just over the Course Home tabs and extends to the bottom of the page.

I've managed to create some hooks to solve the UI. Maybe this can be revisited and replaced by some of Paragon's previously mentioned components in the future.

The overlay has some placeholder text to test the inner scrolling. When the modal pops up, it scrolls to the top of the page and freeze the `<body>` scrolling.

## Notes
I've added a prop to the slice `course-home/data/slice.js`. I don't see there's any unit tests on slices in the repo, maybe because they are small and simple or maybe they can be tested indirectly. I could add a `slice.test.js` for that extra prop if needed.

## Testing Instructions:

In order to see the search button and overlay:
- Follow the instructions to enable the waffle flags as indicated in [this PR](https://github.com/openedx/frontend-app-learning/pull/1199).
- You should be able to click the search icon to popup the overlay.

## Results:

You should be able to click and open the search modal when the waffle flag is enabled.


src/course-home/courseware-search/hooks.js
```diff
@@ -0,0 +1,64 @@
...
+  }, [courseId]);
+
+  return enabled;
+}
+
+export function useCoursewareSearchState() {
+  const enabled = useCoursewareSearchFeatureFlag();
+  const show = useSelector(state => state.courseHome.showSearch);
+
+  return { show: enabled && show };
```
schenedx, line 24: Why should this return a Json, not a direct boolean? Is it because hooks has to be in JSON format?

src/course-home/courseware-search/hooks.js
```diff
@@ -0,0 +1,64 @@
...
+      global.removeEventListener('scroll', recalculate);
+    };
+  }, []); // Empty array ensures that effect is only run on mount
+
+  return elementInfo;
+}
+
+export function useLockScroll() {
+  useEffect(() => {
+    window.scrollTo(0, 0);
```
schenedx, line 64: I am curious to see how unit test can cover this statement.

schenedx (commented):
See my questions and comments inline. I understand your tech decisions here. I think your choices here are sensible.

src/course-home/courseware-search/hooks.js
aethant: The tests are running in `jsdom` so there's access to the `window` object. You would then set `scrollTo` with `jest.fn` and listen for the method to be called with `0, 0` params.  
  
psuedo-code:  
\`\`\`  
const spyScrollTo = jest.fn()  
Object.defineProperty(window, 'scrollTo', { value: spyScrollTo });  
expect(spyScrollTo).toHaveBeenCalledWith({top: 0, left: 0});  
\`\`\`

src/course-home/courseware-search/hooks.js
aethant: Hooks do not require JSON to be returned, or anything for that matter. Rather, he is likely returning a JSON object here as it is being used to set state in the reducer.

src/course-home/courseware-search/hooks.js
```diff
@@ -0,0 +1,64 @@
...
+  const enabled = useCoursewareSearchFeatureFlag();
+  const show = useSelector(state => state.courseHome.showSearch);
+
+  return { show: enabled && show };
+}
+
+export function useElementBoundingBox(elementId) {
+  const [elementInfo, setElementInfo] = useState(undefined);
+
+  useLayoutEffect(() => {
```
aethant, line 37: I can't recall - does `useLayoutEffect` automatically throttle/debounce calls? If not, would it be useful to add a slight debounce here?

src/course-tabs/CourseTabsNavigation.jsx
```diff
@@ -5,33 +5,41 @@ import classNames from 'classnames';
...
+              href={url}
+            >
+              {title}
+            </a>
+          ))}
+        </Tabs>
+      </div>
+      {show ? (
+        <CoursewareSearch />
+      ) : undefined}
```
aethant: I believe it would be better to return `null` here rather than `undefined` - conversely, you could also render this as:

`{show && <CoursewareSearch />}` and avoid the issue altogether.

src/course-home/courseware-search/hooks.js
rijuma: Usually I like to return a single prop object rather than a value for context, it kind of forces you to take the prop name into account when handling the response. I can change it to a direct value if you disagree.

schenedx: I am trying to learn if there are advantage of doing it this way. Both you and @aethant provided good reasons. I am satisfied with the explanations.

src/course-home/courseware-search/hooks.js
rijuma: Indeed. I'm working on the tests atm.

For context, I've added the scroll top because I lock the page scrolling when the user opens the Search panel. This is just a fail safe to prevent some weird edge case of someone opening the search UI somehow while scrolled because it might get the close button hidden from the viewport.

src/course-home/courseware-search/hooks.js
rijuma: I thought about debouncing it, the problem is that it might create visual artifacts when the UI updates while scrolling, but I guess that's not important since it's scrolling to the top automatically. I'll debounce it with 200-300ms just to make it lighter to the browser when resizing. `useLayout()` doesn't seem to include debouncing by itself.

src/course-tabs/CourseTabsNavigation.jsx
rijuma: True. It should return `null` instead. In fact, it's probably outputting a warning to the console.

The thing about `{show && <CoursewareSearch />}` is that you need to be sure there's actually `false` or `null` in show and not something _falsy_ like `0` that will be rendered. In this case we can probably feel safe that it's a boolean, but I think it's just a good practice to make sure to return `null`.

src/course-home/courseware-search/hooks.js
rijuma: Besides that, I intentionally named the method `useCoursewareSearchState()` in case we need to store more state props to use on search, not only the show or hide.

src/course-tabs/CourseTabsNavigation.jsx
aethant: That's a good point, forgot about the `0`

src/course-home/courseware-search/hooks.js
rijuma: This was tested with a simple `jest.spyOn(window, 'scrollTo')`. (see `src/course-home/courseware-search/hooks.test.jsx`)

src/course-tabs/CourseTabsNavigation.jsx
rijuma: Fixed. Thanks for the keen sight.

src/course-home/courseware-search/hooks.js
rijuma: Added a debounce of 100ms.

src/course-home/courseware-search/CoursewareSearch.test.jsx
```diff
@@ -21,26 +30,55 @@ describe('CoursewareSearch', () => {
...
+      renderComponent();
+    });
+
+    it('Should use useElementBoundingBox() and useLockScroll() hooks', () => {
+      expect(useElementBoundingBox).toBeCalledTimes(1);
+      expect(useLockScroll).toBeCalledTimes(1);
+    });
+
+    it('Should have a "--modal-top-position" CSS variable matching the CourseTabsNavigation top position', () => {
+      renderComponent();
```
schenedx: Why do you need to call `renderComponent` again here, when `beforeAll` already called that?

src/course-home/courseware-search/CoursewareSearch.test.jsx
```diff
@@ -21,26 +30,55 @@ describe('CoursewareSearch', () => {
...
+
+      const section = screen.getByTestId('courseware-search-section');
+      expect(section.style.getPropertyValue('--modal-top-position')).toBe(`${tabsTopPosition}px`);
+    });
+
+    it('Should dispatch setShowSearch(true) when clicking the close button', () => {
+      renderComponent();
```
schenedx: Same question as why this is called again when `beforeAll` has called this function already

src/course-home/courseware-search/hooks.test.jsx
```diff
@@ -0,0 +1,187 @@
...
+
+  describe('useCoursewareSearchFeatureFlag', () => {
+    const renderTestHook = async (enabled = true) => {
+      useParams.mockImplementation(() => ({ courseId: enabled ? 123 : 456 }));
+      let hook;
+      await act(async () => { (hook = renderHook(() => useCoursewareSearchFeatureFlag())); });
+      return hook;
+    };
+
+    test('should return true if feature is enabled', async () => {
```
schenedx: Just want to learn here: What is the difference between `it` and `test` in unit tests like these?

schenedx (approved):
Minor feedback and a learning question. Good job 👍

src/course-home/courseware-search/CoursewareSearch.test.jsx
rijuma: Whoops!

src/course-home/courseware-search/CoursewareSearch.test.jsx
rijuma: Whoopsie!

src/course-home/courseware-search/hooks.test.jsx
rijuma: `it` is an alias of `test` actually, but to clean it up I renamed it all to `it`.

PR #1214, opened 2023-10-20
Add CoursewareSearchForm component

davidnuon:
## Description

Ticket: [KBK-41](https://2u-internal.atlassian.net/browse/KBK-41)

Adds a search bar for the Courseware Search feature. 

Depends on #1212

### DIFF
--- a/src/course-home/courseware-search/CoursewareSearchBar.jsx
+++ b/src/course-home/courseware-search/CoursewareSearchBar.jsx
@@ -1,0 +1,32 @@
+import React from 'react';
+import { SearchField } from '@edx/paragon';
+import PropTypes from 'prop-types';
+
+const CoursewareSearchBar = ({
+  onSubmit,
+  onChange,
+  placeholder,
+}) => (
+  <SearchField.Advanced
+    onSubmit={onSubmit}
+    onChange={onChange}
+    submitButtonLocation="external"
+    className="courseware-search-bar"
+    data-testid="courseware-search-bar"
+  >
+    <div className="pgn__searchfield_wrapper">
+      <SearchField.Label />
+      <SearchField.Input placeholder={placeholder} />
+      <SearchField.ClearButton />
+    </div>
+    <SearchField.SubmitButton submitButtonLocation="external" />
+  </SearchField.Advanced>
+);
+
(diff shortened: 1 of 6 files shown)


src/course-home/courseware-search/CoursewareSearchBarTest.jsx
aethant, line 1: It looks as if your test file is misnamed, hence the coverage issues. Per the repo, seems it should be `.test.jsx`

That or its inceptioning and wants tests for tests.

src/course-home/courseware-search/CoursewareSearchBar.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { SearchField } from '@edx/paragon';
+import PropTypes from 'prop-types';
+
+const CoursewareSearchBar = ({
```
rijuma: I find this name a bit ambiguous, since you can't tell by the name what the component does. I insist to something like `CoursewareSearchForm` since it's more evident. The filters are not to be included in this component so it is actually just the form. This is only a suggestion, you can ignore this if you disagree.

src/course-home/courseware-search/messages.js
```diff
@@ -16,6 +16,11 @@ const messages = defineMessages({
     defaultMessage: 'Search this course',
     description: 'Title for the Courseware Search module.',
   },
+  searchBarPlaceholderText: {
+    id: 'learn.coursewareSerch.searchBarPlaceholderText',
+    defaultMessage: 'Search',
+    description: 'Placeholder text courseware for the search bar.',
```
rijuma: I think this is meant to be a context for the translator. I'd rather put something more descriptive as `Placeholder text for the Courseware Search input control`.

src/course-home/courseware-search/CoursewareSearchBar.jsx
davidnuon: 👍

aethant:
@davidnuon I think it has to be "fix:" rather than "lint: Fix" for the check to pass

rijuma (approved):
LGTM!

aethant (approved):

PR #1224, opened 2023-10-25
feat: Added CoursewareSearchResults UI component

rijuma:
## Description
Ticket: [KBK-44 🔒](https://2u-internal.atlassian.net/browse/KBK-44)

We need a [Search Results UI component 🔒](https://2u-internal.atlassian.net/wiki/spaces/Kubrick/pages/614432780/Component+breakdown+based+on+Figma#Search-Results) to show the search components and the results.

### Showing results

| 320px wide  | 1024px wide |
| --- | --- |
| image | image |

### No results

| 320px wide  | 1024px wide |
| --- | --- |
| image | image |

## Testing Instructions:

This shows mocked results, so in order to see the variations:
- Follow the instructions to enable the waffle flags as indicated in https://github.com/openedx/frontend-app-learning/pull/1199.
- Searching for `lorem ipsum` will show the "no results" variant.
- Searching for anything other than `lorem ipsum` will show a fixed number of results.
- Performing an empty search will clear the results.

### DIFF
--- a/src/course-home/courseware-search/CoursewareSearchResults.jsx
+++ b/src/course-home/courseware-search/CoursewareSearchResults.jsx
@@ -1,0 +1,78 @@
+import React from 'react';
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import {
+  Folder, TextFields, VideoCamera, Article,
+} from '@edx/paragon/icons';
+import { Icon } from '@edx/paragon';
+import PropTypes from 'prop-types';
+import messages from './messages';
+
+const iconTypeMapping = {
+  document: Folder,
+  text: TextFields,
+  video: VideoCamera,
+};
+const defaultIcon = Article;
+
+const CoursewareSearchResults = ({ intl, results }) => {
+  if (!results.length) {
+    return (
+      <div className="courseware-search-results">
+        <p className="courseware-search-results__empty" data-testid="no-results">{intl.formatMessage(messages.searchResultsNone)}</p>
+      </div>
+    );
+  }
+
+  return (
+    <div className="courseware-search-results" data-testid="search-results">
+      {results.map(({
+        title, href, type, breadcrumbs, contentMatches, isExternal,
+      }) => {
+        const key = type.toLowerCase();
+        const icon = iconTypeMapping[key] || defaultIcon;
+
+        const linkProps = { href };
+
+        if (isExternal) {
+          linkProps.target = '_blank';
+          linkProps.rel = 'nofollow';
+        }
+
+        return (
+          <a className="courseware-search-results__item" {...linkProps}>
+            <div className="courseware-search-results__icon"><Icon src={icon} /></div>
+            <div className="courseware-search-results__info">
+              <div className="courseware-search-results__title">
+                <span>{title}</span>
+                {contentMatches ? (<em>{contentMatches}</em>) : null }
+              </div>
+              {breadcrumbs?.length ? (
+                <ul className="courseware-search-results__breadcrumbs">
+                  {breadcrumbs.map(bc => (<li><div>{bc}</div></li>))}
+                </ul>
+              ) : null}
+            </div>
+          </a>
+        );
+      })}
+    </div>
+  );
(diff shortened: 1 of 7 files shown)


rijuma:
This is a UI component with mocked behavior for testing. I don't think this needs to have coverage atm. It will be wrapped up in an implementation ticket.

src/course-home/courseware-search/CoursewareSearchResults.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+const iconTypeMapping = {
+  document: Folder,
+  text: TextFields,
+  video: VideoCamera,
+};
+const defaultIcon = Article;
+
+const CoursewareSearchResults = ({ intl, results }) => {
+  if (!results.length) {
```
aethant: Should `results` have  a default value and/or a check for undefined here? i.e. `if (!results?.length)` , `if (!results || !results.length) {` or `({intl, results = [] })` ? Reading the `propTypes` it doesn't seem like `results` is required.

Reason: if `results` === undefined, `.length` will throw.

rijuma: Whoops, I thought it had the conditional. Nice catch.

src/course-home/courseware-search/CoursewareSearchResults.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+  return (
+    <div className="courseware-search-results" data-testid="search-results">
+      {results.map(({
+        title, href, type, breadcrumbs, contentMatches, isExternal,
+      }) => {
+        const key = type.toLowerCase();
+        const icon = iconTypeMapping[key] || defaultIcon;
+
+        const linkProps = { href };
```
aethant: Maybe its me not liking mutating an object, but this could be expressed as such:

  
`const linkProps = {`   
 `...href,`   
 `...(isExternal`   
 `? {`   
 `target: "_blank",`   
 `rel: "nofollow" }`   
 `: {})`  
 `}`

src/course-home/courseware-search/CoursewareSearchResults.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+        return (
+          <a className="courseware-search-results__item" {...linkProps}>
+            <div className="courseware-search-results__icon"><Icon src={icon} /></div>
+            <div className="courseware-search-results__info">
+              <div className="courseware-search-results__title">
+                <span>{title}</span>
+                {contentMatches ? (<em>{contentMatches}</em>) : null }
+              </div>
+              {breadcrumbs?.length ? (
```
aethant, line 48: Similar to above - is `breadcrumbs` being checked to exist/have at least an empty array value `[]`?

src/course-home/courseware-search/CoursewareSearchResults.jsx
rijuma: I don't like mutations either. But since everything is inside the same scope, I'm standing between a small mutation, and a hacky code just to prevent it. I really hope JS comes up with some syntactic sugar for conditional object properties. Let me think about the lesser evil here..

src/course-home/courseware-search/CoursewareSearchResults.jsx
rijuma: This does have an optional chaining, so anything other than an array (or a sneaky object with a `length` property will be falsy.

rijuma: Besides, I'm using `PropTypes` on the component for the proper format. I think that a lot of boilerplate for not using typescript is not worth it IMHO.

src/course-home/courseware-search/CoursewareSearchResults.jsx
rijuma: I've made an alternative without mutations, but it's clearer to read. Let me know what you think.

aethant (approved):

schenedx (approved):
👍

PR #1231, opened 2023-10-31
[KBK-51] Clear search results when search field is cleared

aethant:
[KBK-51](https://2u-internal.atlassian.net/browse/KBK-51)

As a user, I expect that when I clear the search input, the filter will disappear and no content will be presented.


schenedx:
This person no longer works for 2U. Closing the PR.

PR #1234, opened 2023-11-08
chore: make notification display by default

leangseu-edx:
Make some change to have notification correctly.

The bug was introduced on this pr https://github.com/openedx/frontend-app-learning/pull/1139


mattcarter (approved):
Approved but you need to fix the spaces causing the test failure.

PR #1250, opened 2023-12-05
feat: remove segment event

alangsto:
## [COSMO-9](https://2u-internal.atlassian.net/browse/COSMO-9)

With the end of our first experiment related to the learning assistant experience, the Cosmonauts team would also like to remove a segment event logged for each render of the chat component. This segment was added for both experimental and diagnostic purposes, so with the end of the experiment, this is no longer needed.


varshamenon4 (approved):
lgtm!

PR #1251, opened 2023-12-11
feat: Add redirect to enterprise learner dashboard 

pwnage101:
When a learner tries to load the B2C course page for a course that starts in the future (error_code="course_not_started"), normally the learning MFE automatically redirects to the B2C dashboard.  This change supports an alternate error_code "course_not_started_enterprise_learner" to trigger an alternative redirect to the B2B (enterprise) learner dashboard.

This does two main things:

1. When the course metadata API response indicates course_access.error_code = "course_not_started_enterprise_learner" then redirect to "/redirect/enterprise-learner-dashboard".
2. When the top-level router matches path "/redirect/enterprise-learner-dashboard" then redirec to to the value of config `ENTERPRISE_LEARNER_PORTAL_URL`.

ENT-8078

Companion PR to create the new `course_not_started_enterprise_learner` error code: https://github.com/openedx/edx-platform/pull/33913/

### DIFF
--- a/global-setup.js
+++ b/global-setup.js
@@ -1,0 +1,4 @@
+// Force all tests to run in UTC to prevent tests from being sensitive to host timezone.
+module.exports = async () => {
+  process.env.TZ = 'UTC';
+};
--- a/jest.config.js
+++ b/jest.config.js
@@ -16,5 +16,6 @@ module.exports = createConfig('jest', {
     'react-markdown': '<rootDir>/node_modules/react-markdown/react-markdown.min.js',
   },
   testTimeout: 30000,
-  testEnvironment: 'jsdom'
+  testEnvironment: 'jsdom',
+  globalSetup: "./global-setup.js"
 });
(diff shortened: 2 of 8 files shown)


global-setup.js
```diff
@@ -0,0 +1,4 @@
+// Force all tests to run in UTC to prevent tests from being sensitive to host timezone.
+module.exports = async () => {
+  process.env.TZ = 'UTC';
+};
```
pwnage101, line 4: Spent 25 minutes scratching my head on a failing js test until I realized it was sensitive to the time zone of the host machine, and running `TZ=UTC npm run test` fixed it.  This addition causes all tests to always run in UTC, regardless of host TZ.

adamstankiewicz: Out of curiosity, does setting `process.env.TZ` in the existing `setupTest.js` file work (in an attempt to avoid needing to setup the `globalSetup` config).

adamstankiewicz (approved):
These changes LGTM (with 1 non-blocking question). Defer to any of Jansen's feedback, if any; otherwise, ✅

global-setup.js
pwnage101: yes, it would have!  I figured just do it for the whole project but happy to revert back to just doing it for one test class if you feel like it's too heavy-handed.  This was the failing test (date off by one day):

https://github.com/openedx/frontend-app-learning/pull/1251/files#diff-729c9dcf69224055c50e2364be6ba2387aada65cd0929bf928271c8b3e1c561aR511

PR #1253, opened 2023-12-14
UI search UI a11y changes

rijuma:
## Description
Tickets: [KBK-74 🔒](https://2u-internal.atlassian.net/browse/KBK-74), [KBK-77 🔒](https://2u-internal.atlassian.net/browse/KBK-77)

These changes improves the Accessibility for the UI on Courseware Search. Some of the update requests for accessibility are not possible from this implementation, since we are using Paragon components. A request to update these accessibility issues on Paragon will be performed.

# Ticket KBK-74

Done.

> Change the `<h2>Search this course</h2>` with `<h2 aria-level="1">Search this course</h2>` or `<h1 class="h2">Search this course</h1>`. The main heading should be an `H1` semantically.

Done.

> The light gray text is too light (3.23:1, minimum is 4.5:1). The Paragon design system has standard values for hint text.
> The main underline for the tab-list (the lighter gray) should have minimum 3:1 contrast vs. background.

Color contrast updates:
- Navigation Tab border changed from #e9e6e4 (Paragon default) to gray-300 (#8F8F8F) Contrast change: [from 1.21:1 to 3.23:1](https://webaim.org/resources/contrastchecker/?fcolor=8F8F8F&bcolor=FFFFFF).
- Active tab border bottom increased from `2px` to `4px` to make more evident that the tab is currently active. This is to complement the darkening of the tab border from the previous change.
- All light gray text changed from gray-400 (#8F8F8F) to gray-500 (#707070) - Contrast change: [from 3.23:1 to 4.95](https://webaim.org/resources/contrastchecker/?fcolor=707070&bcolor=FFFFFF).

> For the Search button <button type="submit" class="pgn__searchfield__button btn btn-primary">Search<span class="sr-only">submit search</span></button>. Remove the sr-only span, otherwise it will read as "Search submit search".

An empty string was provided to the parameter that adds the span with the `sr-only` class. This results in an empty span with a `sr-only`. I'll assume the span will be ignored, otherwise we should request an update on Paragon to remove the element when there's no text for that label.

> For <div class="courseware-search__results-summary"> add aria-live="polite" aria-relevant="all" aria-atomic="true".

Done.

> In narrow viewport, if there is a More... button, the focusable <li> should have role="tab". (Spin up a new ticket for a Paragon accessibility update if this is not possible on our side).

This is a Paragon issue. A request to update this will be created.

> The logic seems to be off on narrow viewports. Where it shows the More button when there is room for Sequence for example. And when Sequence is selected it doesn’t appear to be selected (though this particular filter result might be an edge case).

This is a Paragon issue. A request to update this will be created.

> Change <label for="pgn-searchfield-input-1" class="m-0"><span class="sr-only">search</span></label> to <label for="pgn-searchfield-input-1" class="m-0"><span class="sr-only">search box</span></label> Better to have more distinct labels, for voice input programs.

The search component in Paragon has a property to set both the label and the button text (i18n), but share the same property. So currently is not possible to have a different text between the label and the search submit button. This is a Paragon issue. A request to update this will be created.

# Ticket KBK-77

Currently, the result count from the search endpoint is not equivalent to the result item count. We've decided to remove this count from the results label.

# UI Changes

These are screenshots showing the results in the UI updates.

## New
search-after

## Old

search-before


src/index.scss
```diff
@@ -38,24 +38,6 @@
...
-  }
-
-  .nav a {
-    &:not(.active):hover {
-      background-color: $light-400;
-      border-bottom: none;
-    }
-  }
-}
-
```
rijuma, line 58: This was moved to its own css file for the `CourseTabsNavigation` component.

src/course-home/courseware-search/CoursewareSearch.test.jsx
```diff
@@ -246,24 +246,14 @@ describe('CoursewareSearch', () => {
       expect(screen.queryByTestId('courseware-search-summary').textContent).toBe('No results found.');
     });
 
-    it('should show a summary for a single result', () => {
+    it('should show a summary for the results', () => {
```
schenedx, line 249: Why did you remove the test case for multiple results?

src/course-tabs/course-tabs-navigation.scss
```diff
@@ -0,0 +1,24 @@
...
+      border-bottom: none;
+    }
+  }
+
+  &__search-toggle {
+    position: absolute;
+    top: .05rem;
+    right: 0;
+  }
+}
```
schenedx: #NIT newline please?

schenedx (approved):
A question on the removal of a test case.
No blocking comments. LGTM 👍

src/course-home/courseware-search/CoursewareSearch.test.jsx
rijuma: It used to have two different messages, one for a single result and another for multiple/zero results, like:
- 1 match found for ${search}
- 2 matches found for ${search}

This required two i18n entries for different languages as well.

Since we are not considering the count any more, it will always be the same copy.

PR #1257, opened 2023-12-19
fix: prevent error when there are no results for a given filter

germanolleunlp:
https://2u-internal.atlassian.net/browse/KBK-79

### DIFF
--- a/src/course-home/courseware-search/CoursewareSearchResults.jsx
+++ b/src/course-home/courseware-search/CoursewareSearchResults.jsx
@@ -15,7 +15,7 @@ const iconTypeMapping = {
 
 const defaultIcon = Article;
 
-const CoursewareSearchResults = ({ results }) => {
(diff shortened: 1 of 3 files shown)


src/course-home/courseware-search/CoursewareSearchResults.jsx
```diff
@@ -15,7 +15,7 @@ const iconTypeMapping = {
 
 const defaultIcon = Article;
 
-const CoursewareSearchResults = ({ results }) => {
+const CoursewareSearchResults = ({ results = [] }) => {
```
schenedx, line 18: Could we develop a test case for this?

schenedx (approved):
Either way, I understand why we need this change.

src/course-home/courseware-search/CoursewareSearchResults.jsx
germanolleunlp: added ✅

PR #1259, opened 2023-12-21
fix: ignore failed outline API call when learner has no access

pwnage101:
If the learner doesn't even have access to the course (e.g. because the course starts in the future), don't worry about a 404 fetching the course outline since we're not planning to use it anyway.

ENT-8078

matrix
===

* [current behavior] if the learner has access to the course and tab data is available: stay on course home page.
* [current behavior] if the learner has access to the course but the tab data is not available, display error page.
* [current behavior] if the learner does not have access to the course, but the tab data is available, dispatch to enterprise redirect logic.
* [NEW BEHAVIOR] if the learner does not have access to the course, but the tab data is not available, dispatch to enterprise redirect logic (WAS: display error page).

A combination of factors contribute to this specific scenario with new behavior, but the vast majority of hits on this page won't trigger this scenario.  It just so happens that it's likely to happen while testing on stage, which gets in the way of testing other features which rely on dispatching to fetchTabDenied.

### DIFF
--- a/src/course-home/data/thunks.js
+++ b/src/course-home/data/thunks.js
@@ -38,28 +38,41 @@ export function fetchTab(courseId, tab, getTabData, targetUserId) {
   return async (dispatch) => {
     dispatch(fetchTabRequest({ courseId }));
     try {
-      const courseHomeCourseMetadata = await getCourseHomeCourseMetadata(courseId, 'outline');
-      dispatch(addModel({
-        modelType: 'courseHomeMeta',
-        model: {
-          id: courseId,
-          ...courseHomeCourseMetadata,
-        },
-      }));
-      const tabDataResult = getTabData && await getTabData(courseId, targetUserId);
-      if (tabDataResult) {
+      const promisesToFulfill = [getCourseHomeCourseMetadata(courseId, 'outline')];
+      if (getTabData) {
+        promisesToFulfill.push(getTabData(courseId, targetUserId));
+      }
+      const [
+        courseHomeCourseMetadataResult,
+        tabDataResult,
+      ] = await Promise.allSettled(promisesToFulfill);
+      if (courseHomeCourseMetadataResult.status === 'fulfilled') {
+        dispatch(addModel({
+          modelType: 'courseHomeMeta',
+          model: {
+            id: courseId,
+            ...courseHomeCourseMetadataResult.value,
+          },
+        }));
+      }
+      if (tabDataResult?.status === 'fulfilled') {
         dispatch(addModel({
           modelType: tab,
           model: {
             id: courseId,
-            ...tabDataResult,
+            ...tabDataResult.value,
           },
         }));
       }
-      // Disable the access-denied path for now - it caused a regression
-      if (!courseHomeCourseMetadata.courseAccess.hasAccess) {
+      if (courseHomeCourseMetadataResult.status === 'rejected') {
+        throw courseHomeCourseMetadataResult.reason;
+      } else if (!courseHomeCourseMetadataResult.value.courseAccess.hasAccess) {
+        // If the learner does not have access to the course, short cut to dispatch to a denied response regardless of
+        // the tabDataResult.
         dispatch(fetchTabDenied({ courseId }));
-      } else if (tabDataResult || !getTabData) {
+      } else if (tabDataResult?.status === 'rejected') {
+        throw tabDataResult.reason;
+      } else {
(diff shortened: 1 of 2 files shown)


src/course-home/data/thunks.js
```diff
@@ -38,28 +38,39 @@ export function fetchTab(courseId, tab, getTabData, targetUserId) {
...
+        throw courseHomeCourseMetadataResult.reason;
+      } else if (!courseHomeCourseMetadataResult.value.courseAccess.hasAccess) {
         dispatch(fetchTabDenied({ courseId }));
-      } else if (tabDataResult || !getTabData) {
+      } else if (tabDataResult && tabDataResult.status === 'rejected') {
+        throw tabDataResult.reason;
+      } else {
         dispatch(fetchTabSuccess({
           courseId,
           targetUserId,
```
pwnage101: Notable changes:
* There's now a default case; one of [fetchTabDenied, fetchTabSuccess, fetchTabFailure] is always dispatched.
* A non-accessible course is now handled using fetchTabDenied regardless of the tab data query results.

src/course-home/data/thunks.js
```diff
@@ -38,28 +38,39 @@ export function fetchTab(courseId, tab, getTabData, targetUserId) {
...
-      // Disable the access-denied path for now - it caused a regression
-      if (!courseHomeCourseMetadata.courseAccess.hasAccess) {
+      if (courseHomeCourseMetadataResult.status === 'rejected') {
+        throw courseHomeCourseMetadataResult.reason;
+      } else if (!courseHomeCourseMetadataResult.value.courseAccess.hasAccess) {
         dispatch(fetchTabDenied({ courseId }));
-      } else if (tabDataResult || !getTabData) {
+      } else if (tabDataResult && tabDataResult.status === 'rejected') {
+        throw tabDataResult.reason;
+      } else {
```
adamstankiewicz: nit: I wonder if it'd make sense to co-locate the conditions that result in throwing an exception?

```jsx
if (courseHomeCourseMetadataResult.status === 'rejected') {
  throw courseHomeCourseMetadataResult.reason;
} else if (tabDataResult?.status === 'rejected') {
  throw tabDataResult.reason;
} else if (!courseHomeCourseMetadataResult.value.courseAccess.hasAccess) {
 dispatch(fetchTabDenied({ courseId }));
} else {
  dispatch(fetchTabSuccess({
    courseId,
    targetUserId,
  }));
}
```

src/course-home/data/thunks.js
```diff
@@ -38,28 +38,39 @@ export function fetchTab(courseId, tab, getTabData, targetUserId) {
...
+      if (courseHomeCourseMetadataResult.status === 'fulfilled') {
+        dispatch(addModel({
+          modelType: 'courseHomeMeta',
+          model: {
+            id: courseId,
+            ...courseHomeCourseMetadataResult.value,
+          },
+        }));
+      }
+      if (tabDataResult && tabDataResult.status === 'fulfilled') {
```
adamstankiewicz: nit:

```suggestion
      if (tabDataResult?.status === 'fulfilled') {
```

src/course-home/data/thunks.js
```diff
@@ -38,28 +38,39 @@ export function fetchTab(courseId, tab, getTabData, targetUserId) {
...
         }));
       }
-      // Disable the access-denied path for now - it caused a regression
-      if (!courseHomeCourseMetadata.courseAccess.hasAccess) {
+      if (courseHomeCourseMetadataResult.status === 'rejected') {
+        throw courseHomeCourseMetadataResult.reason;
+      } else if (!courseHomeCourseMetadataResult.value.courseAccess.hasAccess) {
         dispatch(fetchTabDenied({ courseId }));
-      } else if (tabDataResult || !getTabData) {
+      } else if (tabDataResult && tabDataResult.status === 'rejected') {
```
adamstankiewicz: nit:

```suggestion
      } else if (tabDataResult?.status === 'rejected') {
```

adamstankiewicz (approved):
LGTM (with a couple nits), deferring to any owning squad's review feedback.

src/course-home/data/thunks.js
pwnage101: Hm, I'm pretty sure I want to dispatch fetchTabDenied as soon as possible, which is why the conditional is ordered like that.  I added a comment to clarify this.

nsprenkle (approved):

PR #1282, opened 2024-02-05
feat: gate visibility of chat component by active attempt

alangsto:
## [COSMO-86](https://2u-internal.atlassian.net/browse/COSMO-86)

The chat component should not be visible if a learner has an active attempt. The special exams library has been refactored so that the exam state is accessible outside of components defined in the exams library. This PR makes use of the newly available exams reducer to gave chat visibility.

### DIFF
--- a/src/courseware/course/chat/Chat.jsx
+++ b/src/courseware/course/chat/Chat.jsx
@@ -1,4 +1,5 @@
 import { createPortal } from 'react-dom';
+import { useSelector } from 'react-redux';
 import PropTypes from 'prop-types';
 
 import { Xpert } from '@edx/frontend-lib-learning-assistant';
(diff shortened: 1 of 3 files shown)


src/courseware/course/chat/Chat.jsx
```diff
@@ -37,10 +42,13 @@ const Chat = ({
     return endDate ? utcDate > endDate : false; // evaluate if end date has passed only if course has end date
   };
 
+  const hasActiveAttemptId = !!activeAttemptId;
+
   const shouldDisplayChat = (
     enabled
     && (hasVerifiedEnrollment || isStaff) // display only to verified learners or staff
     && !endDatePassed()
+    && !hasActiveAttemptId
```
michaelroytman: Nit: I think you can just use `!activeAttemptId`, right? That'll cast `activeAttemptId` to a boolean and then negate it. `hasActiveAttemptId` is unnecessary, because you're negating the value `activeAttemptId` twice.

michaelroytman (commented):
This looks good to me so far. Looking forward to testing this out once the feature branch is merged!

michaelroytman (approved):

PR #1283, opened 2024-02-06
chore: upgrade frontend-component-header version

mubbsharanwar:
upgrade frontend-component-header version to enable show/hide username from header based on the flag

VAN-1804


bradenmacdonald:
This PR seems to be obsolete. Feel free to re-open if I'm mistaken and it's still relevant.

PR #1284, opened 2024-02-06
Refactor use of frontend-lib-special-exams public API to use hooks.

michaelroytman:
### Description:

**Jira**: [COSMO-175](https://2u-internal.atlassian.net/browse/COSMO-175)

**NOTE**: The corresponding pull request in `frontend-lib-special-exams` is a prerequisite to these changes: https://github.com/openedx/frontend-lib-special-exams/pull/134. Those changes must be merged and deployed and the version of the library updated here before these changes can be merged.

This commit refactors the use of the `frontend-lib-special-exams` public API to use hooks.

This commit also imports the root reducer from the `frontend-lib-special-exams` library. This root reducer is used for the specialExams slice when configuring the store for this application.


src/courseware/course/sequence/Unit/hooks/useExamAccess.js
```diff
@@ -12,14 +12,23 @@ export const stateKeys = StrictDict({
 const useExamAccess = ({
   id,
 }) => {
-  const [accessToken, setAccessToken] = useKeyedState(stateKeys.accessToken, '');
```
michaelroytman, line 15: We no longer need to maintain this as state, since the `useExamAccessToken` hook takes care of getting the updated values from the `specialExams` store.

src/courseware/course/sequence/Unit/hooks/useExamAccess.test.js
```diff
@@ -1,98 +0,0 @@
-import React from 'react';
```
michaelroytman, line 1: The diff is showing up this way, because I had to rename this from a `js` file to a `jsx` file.

src/courseware/course/sequence/Unit/hooks/useExamAccess.test.jsx
```diff
@@ -0,0 +1,88 @@
+import { logError } from '@edx/frontend-platform/logging';
```
michaelroytman, line 1: I used the article [How to test custom React hooks](https://kentcdodds.com/blog/how-to-test-custom-react-hooks) to decide how to test this hook. I decided to focus on testing the return values instead of the internals of the hook, since that may change (as it did in this pull request). I think these tests will be more stables this way.

alangsto (approved):

PR #1286, opened 2024-02-07
feat: show chat for desktop only

varshamenon4:
### Context

The Chatbot UI is conditionally displayed based on the screen width of a browser. The current breakpoint allows for mobile browsers in landscape mode to be able to view the chat toggle, although the tool itself has not been tested for mobile use yet. 

The breakpoint should be changed to prevent mobile users from seeing the chat toggle. The updated breakpoint should only apply to the Chat toggle, not to any other triggers in the Learning MFE.

### Changes
- Added new breakpoint specific to chat
- Set to min medium based on available breakpoints here: https://paragon-openedx.netlify.app/foundations/responsive/#available-breakpoints

### Screenshots

Screenshot 2024-02-07 at 3 17 09 PM
Screenshot 2024-02-07 at 3 17 23 PM

### DIFF
--- a/src/courseware/course/Course.jsx
+++ b/src/courseware/course/Course.jsx
@@ -54,6 +54,7 @@ const Course = ({
     celebrations && !celebrations.streakLengthToCelebrate && celebrations.weeklyGoal,
   );
   const shouldDisplayTriggers = windowWidth >= breakpoints.small.minWidth;
+  const shouldDisplayChat = windowWidth >= breakpoints.medium.minWidth;
   const daysPerWeek = course?.courseGoals?.selectedGoal?.daysPerWeek;
 
   useEffect(() => {
@@ -82,7 +83,7 @@ const Course = ({
           isStaff={isStaff}
           unitId={unitId}
         />
-        {shouldDisplayTriggers && (
+        {shouldDisplayChat && (
           <>
             <Chat
               enabled={course.learningAssistantEnabled}
@@ -93,6 +94,10 @@ const Course = ({
               unitId={unitId}
               endDate={course.end ? course.end : ''}
             />
+          </>
+        )}
+        {shouldDisplayTriggers && (
+          <>
             {enableNewSidebar === 'true' ? <NewSidebarTriggers /> : <SidebarTriggers /> }
           </>
         )}
--- a/src/courseware/course/Course.test.jsx
+++ b/src/courseware/course/Course.test.jsx
@@ -17,6 +17,14 @@ jest.mock('@edx/frontend-lib-special-exams/dist/data/thunks.js', () => ({
   ...jest.requireActual('@edx/frontend-lib-special-exams/dist/data/thunks.js'),
   checkExamEntry: () => jest.fn(),
 }));
+const mockChatTestId = 'fake-chat';
+jest.mock(
+  './chat/Chat',
+  // eslint-disable-next-line react/prop-types
+  () => function ({ courseId }) {
+    return <div className="fake-chat" data-testid={mockChatTestId}>Chat contents {courseId} </div>;
+  },
+);
 
 const recordFirstSectionCelebration = jest.fn();
 // eslint-disable-next-line no-import-assign
@@ -316,5 +324,42 @@ describe('Course', () => {
       render(<Course {...testData} />, { store: testStore, wrapWithRouter: true });
       await waitFor(() => expect(screen.getByText('To access course materials, you must score 70% or higher on this exam. Your current score is 30%.')).toBeInTheDocument());
     });
+  });
+
+  it('displays chat when screen is wide enough (browser)', async () => {
+    const courseMetadata = Factory.build('courseMetadata', {
+      learning_assistant_enabled: true,
+      enrollment: { mode: 'verified' },
+    });
+    const testStore = await initializeTestStore({ courseMetadata }, false);
+    const { courseware } = testStore.getState();
+    const { courseId, sequenceId } = courseware;
+    const testData = {
+      ...mockData,
+      courseId,
+      sequenceId,
+    };
+    render(<Course {...testData} />, { store: testStore, wrapWithRouter: true });
+    const chat = screen.queryByTestId(mockChatTestId);
+    await expect(chat).toBeInTheDocument();
+  });
+
+  it('does not display chat when screen is too narrow (mobile)', async () => {
+    global.innerWidth = breakpoints.extraSmall.minWidth;
+    const courseMetadata = Factory.build('courseMetadata', {
+      learning_assistant_enabled: true,
+      enrollment: { mode: 'verified' },
+    });
+    const testStore = await initializeTestStore({ courseMetadata }, false);
+    const { courseware } = testStore.getState();
+    const { courseId, sequenceId } = courseware;
+    const testData = {
+      ...mockData,
+      courseId,
+      sequenceId,
+    };
+    render(<Course {...testData} />, { store: testStore, wrapWithRouter: true });
+    const chat = screen.queryByTestId(mockChatTestId);
+    await expect(chat).not.toBeInTheDocument();
   });
 });


src/courseware/course/Course.jsx
```diff
@@ -54,6 +54,7 @@ const Course = ({
     celebrations && !celebrations.streakLengthToCelebrate && celebrations.weeklyGoal,
   );
   const shouldDisplayTriggers = windowWidth >= breakpoints.small.minWidth;
+  const shouldDisplayChat = windowWidth >= breakpoints.medium.minWidth;
```
varshamenon4, line 57: I set the breakpoint for chat to medium min width (based on https://paragon-openedx.netlify.app/foundations/responsive/#available-breakpoints) figuring that this will account for most mobile landscape pixel widths, but wondering if there is a specific pixel width to use here. Tried looking at other usages of breakpoints in frontend-app-learning for ideas. Wondering if I should reach out to fedx slack channel for ideas too?

src/courseware/course/Course.jsx
```diff
@@ -82,7 +83,7 @@ const Course = ({
           isStaff={isStaff}
           unitId={unitId}
         />
-        {shouldDisplayTriggers && (
+        { shouldDisplayChat && (
```
varshamenon4: Should I test this? I didn't see existing testing in Course.test.jsx for this so I wasn't sure if this is something that needs to have a unit test. If so, would I just test that the chat ui appears/doesn't appear depending on screen width?

src/courseware/course/Course.jsx
varshamenon4: Reached out to the fedx Slack channel and got some good ideas:

> Hmm I don't know of any solution that comes to mind. There is this userAgentData.mobile property from the Navigator object but it's still considered experimental and doesn't have a ton of browser support. The other recommendation I've seen is to just use userAgent but it's not recommended as the value can be easily modified.
I think you have a good solution (and correct screen width). The only downside I can spot right away is if someone on a desktop decides to look at the page on a narrow width, they wouldn't see the chatbot 

> There is a small npm package you can use that is currently in use in [one repository](https://github.com/search?q=org%3Aopenedx+%2Freact-device-detect%2F&type=code) to determine if a user is in the browser or a mobile device.
https://www.npmjs.com/package/react-device-detect

After doing some research, it seems like the current approach is best. For the npm package, I saw this: "If you just want your React app to behave differently or look different on mobiles in general, CSS @media queries and matchMedia are probably what you want." And the user agent data/user agent approach seems to have some downsides too.

Did also get confirmation from the team and from the channel that this width would be the correct one.

src/courseware/course/Course.jsx
zacharis278: For the sake of having coverage I think we should if it's easy. It looks like a really simple test since windowWidth is just passed as a prop. (if we were testing CSS I'd say don't bother)

zacharis278 (approved):
I'll leave adding the test up to you, approved either way 👍

src/courseware/course/Course.jsx
varshamenon4: This is a good point. I can add a couple of tests I'm just running into some trouble with testing this. Should I be mocking the chat basically? Would this be most similar to testing the sidebar? Adding the test to Course.test.jsx.

src/courseware/course/Course.test.jsx
```diff
@@ -317,4 +317,52 @@ describe('Course', () => {
       await waitFor(() => expect(screen.getByText('To access course materials, you must score 70% or higher on this exam. Your current score is 30%.')).toBeInTheDocument());
     });
   });
+
+  it('displays chat when screen is wide enough (browser)', async () => {
+    const courseMetadata = Factory.build('courseMetadata', {
+      learning_assistant_enabled: true,
+      enrollment: { mode: 'verified' },
+    });
+    const specialExams = {
```
varshamenon4: Here is where I tried to create specialExams and add to testStore, but I'm thinking that this isn't the right way to add. Getting a bunch of console errors along the lines of:
`Warning: Failed prop type: The prop `unitId` is marked as required in `BookmarkButton`, but its value is `undefined`.
        at isBookmarked (/edx/app/frontend-app-learning/src/courseware/course/bookmark/BookmarkButton.jsx:27:3)
        at courseId (/edx/app/frontend-app-learning/src/courseware/course/sequence/Unit/index.jsx:18:3)
        at gated (/edx/app/frontend-app-learning/src/courseware/course/sequence/SequenceContent.jsx:13:3)
        at _classCallCheck (/edx/app/frontend-app-learning/node_modules/@edx/src/i18n/injectIntlWithShim.jsx:13:24)
        at injectIntl(ShimmedIntlComponent)
        at div
        at div
        at div
        at div
        at isGated (/edx/app/frontend-app-learning/node_modules/@edx/frontend-lib-special-exams/src/exam/Exam.jsx:24:3)
        at _classCallCheck (/edx/app/frontend-app-learning/node_modules/@edx/src/i18n/injectIntlWithShim.jsx:13:24)
        at injectIntl(ShimmedIntlComponent)
        at children (/edx/app/frontend-app-learning/node_modules/@edx/frontend-lib-special-exams/src/exam/ExamWrapper.jsx:15:24)
        at SequenceExamWrapper
        at div
        at unitId (/edx/app/frontend-app-learning/src/courseware/course/sequence/Sequence.jsx:32:3)
        at courseId (/edx/app/frontend-app-learning/src/courseware/course/sidebar/SidebarContextProvider.jsx:14:3)
        at courseId (/edx/app/frontend-app-learning/src/courseware/course/Course.jsx:24:3)
        at CourseWrapper (/edx/app/frontend-app-learning/src/courseware/course/Course.jsx:153:36)
        at children (/edx/app/frontend-app-learning/src/generic/user-messages/UserMessagesProvider.jsx:32:33)
        at basenameProp (/edx/app/frontend-app-learning/node_modules/react-router/lib/components.tsx:378:13)
        at basename (/edx/app/frontend-app-learning/node_modules/react-router-dom/index.tsx:354:3)
        at Provider (/edx/app/frontend-app-learning/node_modules/react-redux/lib/components/Provider.js:21:20)
        at store (/edx/app/frontend-app-learning/node_modules/@edx/src/react/OptionalReduxProvider.jsx:9:49)
        at _classCallCheck (/edx/app/frontend-app-learning/node_modules/@edx/src/react/ErrorBoundary.jsx:15:22)
        at IntlProvider (/edx/app/frontend-app-learning/node_modules/react-intl/src/components/provider.js:83:47)
        at store (/edx/app/frontend-app-learning/node_modules/@edx/src/react/AppProvider.jsx:47:39)
        at IntlProvider (/edx/app/frontend-app-learning/node_modules/react-intl/src/components/provider.js:83:47)
        at children (/edx/app/frontend-app-learning/src/setupTest.js:178:22)
`
The part that is most interesting is:
`at isGated (/edx/app/frontend-app-learning/node_modules/@edx/frontend-lib-special-exams/src/exam/Exam.jsx:24:3)
        at _classCallCheck (/edx/app/frontend-app-learning/node_modules/@edx/src/i18n/injectIntlWithShim.jsx:13:24)
        at injectIntl(ShimmedIntlComponent)
        at children (/edx/app/frontend-app-learning/node_modules/@edx/frontend-lib-special-exams/src/exam/ExamWrapper.jsx:15:24)
        at SequenceExamWrapper
`
which is what makes me think this has to do with the special exams piece not being mocked correctly in my tests.

michaelroytman: I tend to find all of those console errors to be noise. I think they're related to the fact that a lot of component props correspond to pieces of Redux state, and before the API calls to hydrate the Redux store complete, these props are `undefined`, which causes prop type check failures.

Is the issue that this test is failing? I can see that in the Github actions. I'm not seeing a larger stack trace corresponding to that test, though, so I'm assuming it's just a matter of the assertion failing. Is there something I'm missing?

I believe that this is happening because of the changes we've made to `Xpert`. `Xpert` will now make an API call before it renders to fetch the data it needs to know whether it should render (i.e. whether it's enabled). Because of that, the `Xpert` component is not going to be in the DOM when you try to render the `Course`, because the API function has likely not completed. That's why there's no `toggle-button` in the DOM.

Ordinarily, you'd use the async functions of the React Testing Library to wait for the component to appear in your test. I'm not sure what's happening to that API call. It's not being mocked out in the test, so perhaps it's being made and floating out in the ether. Or maybe it's failing silently. Regardless, the `Xpert` component will likely never show up, even if you wait for it.

We actually ran into this same problem in `Chat.test.jsx`. I think your two options are:

1. Mock out the API call that `Xpert` is making. This is actually difficult, because the function that's making the API call is defined in another library (`frontend-lib-special-exams`). Based on some initial research, I think [msw](https://www.npmjs.com/package/msw) (Mock Service Worker) could work here, because you can specify the URL to mock and not the Javascript function that makes the API call, but I don't see it used in this repository.

2. Mock out the `Chat` component. This might feel a little odd. My rationale is that what your unit test is intending to test is the breakpoint - whether _anything_ appears when the screen size is large enough, right? We don't actually care, for the purposes of the unit test, what appears. We've already tested every other part of the `Xpert` flow, so we don't need to actually test that what's appearing is `Xpert`. Take a look at how we've mocked out Xpert in the [Chat test file](https://github.com/openedx/frontend-app-learning/blob/master/src/courseware/course/chat/Chat.test.jsx#L13-L28). Notice how we mocked out the component with a dummy one. You could do something similar here. I think the choice is up to you whether to mock out `Xpert` or `Chat`. `Chat` is a pretty dumb component and doesn't do much, so I think it would be fine to mock it out, but maybe there is code in there that you still want to run as part of your test?

I hope that helps! Let me know if I've missed an important detail!

varshamenon4: This is so very helpful thank you! I think the instinct is right to just mock out the Chat component, especially since I just want to see that the screen size behavior is working correctly. I'll take a look at the Chat file for how to mock out a component.

PR #1287, opened 2024-02-08
CoursewareSearch - Show filters if needed only

rijuma:
Description
Ticket: [KBK-108 🔒](https://2u-internal.atlassian.net/browse/KBK-108)

These changes hide the filter options with no results (no use). It also hides the filter entirely if the results consists of only one result type. In addition to this, I'm fixing some other minor issues that prevented the search results to show properly.

UI changes
Full results
all-result-types

Partial filters results
two-result-types

Only one result type (no need for filtering)
one-result-type-only

Removed double message on empty results (also improved the UI a bit)
no-results

### DIFF
--- a/src/course-home/courseware-search/CoursewareResultsFilter.jsx
+++ b/src/course-home/courseware-search/CoursewareResultsFilter.jsx
@@ -22,9 +22,13 @@ export const CoursewareSearchResultsFilter = ({ intl }) => {
 
   const { results: data = [] } = lastSearch;
 
-  if (!data.length) { return null; }
+  // If there's no data, we show an empty result.
+  if (!data.length) { return <CoursewareSearchResults />; }
 
   const results = useMemo(() => {
+    // This reducer distributes the data into different groups to make it easy to
+    // use on the filters.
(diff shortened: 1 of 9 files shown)


src/course-home/courseware-search/CoursewareResultsFilter.jsx
```diff
@@ -27,23 +22,34 @@ export const CoursewareSearchResultsFilter = ({ intl }) => {
 
   const { results: data = [] } = lastSearch;
 
-  const results = useMemo(() => data.reduce((acc, { type, ...rest }) => {
-    acc[allFilterKey] = [...(acc[allFilterKey] || []), { type: allFilterKey, ...rest }];
-    if (type === allFilterKey) { return acc; }
+  if (!data.length) { return null; }
+
+  const results = useMemo(() => {
```
schenedx, line 28: #NIT can you add a paragraph of inline comment here to explain what you are doing?
The reduce function doesn't explain it well.

schenedx (approved):
👍

PR #1297, opened 2024-02-22
feat: update chat visibility

alangsto:
The backend for the learning assistant will not return 200 responses for courses that don't pass certain date conditions. We should replicate that behavior on the frontend to ensure that users do not see 400 errors.

### DIFF
--- a/src/courseware/course/Course.jsx
+++ b/src/courseware/course/Course.jsx
@@ -70,6 +70,23 @@ const Course = ({
 
   const SidebarProviderComponent = enableNewSidebar === 'true' ? NewSidebarProvider : SidebarProvider;
 
+  const chatValidDates = () => {
+    const date = new Date();
+    const utcDate = date.toISOString();
+
+    const enrollmentStartDate = course.enrollmentStart || utcDate;
+    const startDate = course.start || enrollmentStartDate;
+    const enrollmentEndDate = course.enrollmentEnd || utcDate;
+    const endDate = course.end || enrollmentEndDate;
+
+    return (
+      startDate <= enrollmentStartDate
+      && enrollmentStartDate <= utcDate
+      && utcDate <= enrollmentEndDate
+      && enrollmentEndDate <= endDate
+    );
+  };
+
   return (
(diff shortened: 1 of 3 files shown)


src/courseware/course/Course.jsx
```diff
@@ -70,6 +70,23 @@ const Course = ({
 
   const SidebarProviderComponent = enableNewSidebar === 'true' ? NewSidebarProvider : SidebarProvider;
 
+  const chatValidDates = () => {
```
michaelroytman, line 73: Since `course` comes from the Redux store, I think we could read in the `course` with the `useModel` hook and compute this data within the `Chat` component to avoid having `Chat` specific logic in the `Course` component. What do you think?

src/courseware/course/chat/Chat.test.jsx
```diff
@@ -182,7 +181,7 @@ describe('Chat', () => {
           enabled
           courseId={courseId}
           contentToolsEnabled={false}
-          endDate={new Date(currentTime.getTime() - 10 * 60000).toISOString()}
+          validDates={false}
```
michaelroytman, line 184: Nit: I'd rename this test, because it's testing something less specific now.

src/courseware/course/Course.jsx
michaelroytman: Also, technically, this bit of code is untested. If you move this into `Chat`, you'll have to add this data to the mock Redux store for it to work properly in your test. That would get you better test coverage, but would be more work.

If you want to get this out and fast-follow with the refactor/testing changes, I'm good with that. But let me know if you don't think this is necessary.

michaelroytman (approved):
This looks good to me. I have two comments about where you're computing whether the date is valid and testing that function. I approved this because I'm open to not addressing my comments, but let me know what you think is best.

src/courseware/course/Course.jsx
alangsto: I think I'd like to get this out and then work on updating the tests/refactor afterwards. Once I merge this I'll start that work.

PR #1298, opened 2024-02-22
feat: refactor date visibility

alangsto:
Date related code for the Chat component should be moved to the Chat component itself.

### DIFF
--- a/src/courseware/course/chat/Chat.test.jsx
+++ b/src/courseware/course/chat/Chat.test.jsx
@@ -1,5 +1,6 @@
 import { BrowserRouter } from 'react-router-dom';
 import React from 'react';
+import { Factory } from 'rosie';
 
 import {
   initializeMockApp,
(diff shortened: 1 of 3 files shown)


src/courseware/course/chat/Chat.test.jsx
```diff
@@ -173,6 +171,18 @@ describe('Chat', () => {
...
   it('if course end date has passed, component should not be visible', async () => {
+    store = await initializeTestStore({
+      specialExams: {
+        activeAttempt: {
+          attempt_id: 1,
+        },
+      },
+      courseMetadata: Factory.build('courseMetadata', {
+        start: '2014-02-03T05:00:00Z',
+        end: '2014-02-05T05:00:00Z',
```
alangsto, line 182: I'm just testing course end date here, because if the enrollment start and end are null, they will both default to the respective course start or course end date. I can add additional tests for each case if there is a need for it.

michaelroytman (approved):

PR #1305, opened 2024-02-28
course translation waffle flag

leangseu-edx:
Disable showing course translation on the flag disabled

Ticket: https://2u-internal.atlassian.net/browse/AU-1858

### DIFF
--- a/src/courseware/data/api.js
+++ b/src/courseware/data/api.js
@@ -110,6 +110,8 @@ function normalizeMetadata(metadata) {
     showCalculator: data.show_calculator,
     notes: camelCaseObject(data.notes),
     marketingUrl: data.marketing_url,
+    language: data.language,
+    wholeCourseTranslationEnabled: data.whole_course_translation_enabled,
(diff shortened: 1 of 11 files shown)


src/courseware/data/api.js
```diff
@@ -110,6 +110,8 @@ function normalizeMetadata(metadata) {
     showCalculator: data.show_calculator,
     notes: camelCaseObject(data.notes),
     marketingUrl: data.marketing_url,
+    language: data.language,
```
BrandonHBodine, line 113: Is this referencing the source/original language for the course?

BrandonHBodine (approved):
Left a question but otherwise lgtm.

src/courseware/data/api.js
leangseu-edx: It is just available field for the course overview. TBH, I am not even sure if this would work. Currently it has locale code instead of language code.

BrandonHBodine: Locale like `en-us` or locale as in `us`?

PR #1312, opened 2024-03-04
fix: Fixes expandAll button functionality

### LINKED ISSUE #1311: Expand All button is not working correctly
"The 'Expand All' button functionality appears to be malfunctioning. Despite expanding all sections, the button label remains as 'Expand All'. Upon clicking the 'Expand All' button, the label changes to 'Collapse All', indicating the intended action. Similar case for vice versa (Collapsing all sections still shows 'Collapse all' text on button instead of 'Expand All'.  As in the below image, all sections are collapsed, the button should show 'Expand All' text rather than 'Collapse All'. 

Screenshot 2024-03-04 at 4 07 53 PM

hinakhadim:
On collapsing all the sections, the button now automatically show 'Expand All'. Similar in case of vice versa. Fixes #1311 

Before: 
Screenshot 2024-03-04 at 4 07 53 PM

After:
image

### DIFF
--- a/src/course-home/outline-tab/Section.jsx
+++ b/src/course-home/outline-tab/Section.jsx
@@ -33,13 +33,17 @@ const Section = ({
 
   const [open, setOpen] = useState(defaultOpen);
 
+  const updateExpandedSection = (status) => {
+    setOpen(status);
+    setExpandedSections((prevObj) => ({ ...prevObj, [section.id]: status }));
+  };
+
   useEffect(() => {
     setOpen(expand);
   }, [expand]);
 
   useEffect(() => {
-    setOpen(defaultOpen);
-    setExpandedSections((prevObj) => ({ ...prevObj, [section.id]: defaultOpen }));
+    updateExpandedSection(defaultOpen);
   // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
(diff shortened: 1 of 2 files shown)


src/course-home/outline-tab/Section.jsx
```diff
@@ -78,20 +80,20 @@ const Section = ({
         styling="card-lg"
         title={sectionTitle}
         open={open}
-        onToggle={() => { setOpen(!open); }}
+        onToggle={() => { setOpen(!open); setExpandedSections((prevObj) => ({ ...prevObj, [section.id]: !open })); }}
```
arbrandes: I feel like at this point the `on` event is getting verbose enough that a proper handling function would be better.  There's also a lot in common with the next two changes, so it's probably going to be worth it for DRYness.

arbrandes (changes requested):

src/course-home/outline-tab/Section.jsx
hinakhadim: updated it

arbrandes:
Since this changes user-facing behavior, we'll need product confirmation on whether it's actually a bug or not.  (If it's a bug you don't necessarily need to go through the [Product Review Process](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review).  @jmakowski1123, what's your take?

PR #1314, opened 2024-03-04
lk/add translation src dest

leangseu-edx:
- chore: add src and dest translation

### DIFF
--- a/src/courseware/data/api.js
+++ b/src/courseware/data/api.js
@@ -110,8 +110,6 @@ function normalizeMetadata(metadata) {
     showCalculator: data.show_calculator,
     notes: camelCaseObject(data.notes),
     marketingUrl: data.marketing_url,
-    language: data.language,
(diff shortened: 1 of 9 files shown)


src/courseware/data/api.js
```diff
@@ -110,8 +110,6 @@ function normalizeMetadata(metadata) {
     showCalculator: data.show_calculator,
     notes: camelCaseObject(data.notes),
     marketingUrl: data.marketing_url,
-    language: data.language,
-    wholeCourseTranslationEnabled: data.whole_course_translation_enabled,
```
nsprenkle, line 114: Don't you still want this "enabled" field?

nsprenkle (approved):
LGTM, with one open question.

src/courseware/data/api.js
leangseu-edx: This place was incorrect. I realize that after connect it to the backend.

PR #1328, opened 2024-03-18
fetch config and move feedback widget for the plugin

leangseu-edx:
- follow the contract to get waffle flag and available languages from the api
- move feedback widget
- update tests

ticket: https://2u-internal.atlassian.net/browse/AU-1915


nsprenkle (changes requested):
The data values don't match contract. Please update so they match the contract or, if it makes more sense to change the contract, let's do that, but they must match

Rodra (approved):
Looks great !

plugins/UnitTranslationPlugin/translation-selection/index.jsx
```diff
@@ -63,18 +71,28 @@ const TranslationSelection = ({
         courseId={courseId}
         selectedLanguage={selectedLanguage}
         setSelectedLanguage={setSelectedLanguage}
+        availableLanguages={availableLanguages}
         id={id}
       />
+      <FeedbackWidget
```
Rodra, line 77: Why are we rendering the widget here ?

leangseu-edx: This is just a place holder. It create the virtual dom here. But it get moved with ref.

Rodra: Ohh I see... COOL !

Rodra (approved):

nsprenkle (approved):
LGTM but before you merge, can you update PR title? This is no longer just language config, it is also feedback, etc.

PR #1330, opened 2024-03-20
feat: whole course translations

leangseu-edx:
What changed:
- implementation of whole course translation
- use frontend plugin framework's pattern
- NOTE: most of the existing code in `src/` doesn't get much changed. The reason there were a lot of files was because I tried to fix the warning message during tests. It used to print way too many error and warning messages. It was too tedious to debug/write more tests. This doesn't remove all of the warning but it should take care of the common and unnecessary one.


env.config.jsx
```diff
@@ -0,0 +1,22 @@
+import UnitTranslationPlugin from '@plugins/UnitTranslationPlugin';
```
BrandonHBodine: Isn't this file in `.gitignore`?

package.json
```diff
@@ -32,12 +32,12 @@
...
     "@edx/frontend-component-footer": "12.7.1",
     "@edx/frontend-component-header": "4.11.1",
-    "@edx/frontend-lib-special-exams": "2.29.1",
     "@edx/frontend-lib-learning-assistant": "^1.24.1",
+    "@edx/frontend-lib-special-exams": "2.29.1",
     "@edx/frontend-platform": "5.6.1",
     "@edx/openedx-atlas": "^0.6.0",
-    "@openedx/paragon": "21.11.3",
-    "@edx/react-unit-test-utils": "npm:@edx/react-unit-test-utils@1.7.1",
+    "@edx/paragon": "20.46.0",
```
BrandonHBodine: Isn't this package deprecated now in favor of `@openedx/paragon` https://www.npmjs.com/package/@edx/paragon mentions it

BrandonHBodine (commented):
Left a couple questions. Since it's top of mind for me; I noticed there's a conflict between paragon versions @edx/paragon and @openedx/paragon.

jsnwesson:
Solely inputting on the Plugin framework implementation: this looks good! The implementation looks good, including the `env.config.jsx` example (assuming that the actual file was intended to be .gitignored). If the limitation for gitignoring it is because the default content is living in `env.config`, would I be able to recommend that you try the `@openedx/frontend-plugin-framework` package? 

The main change you'd see is that `defaultContents` wouldn't be defined in `env.config.jsx`, but instead you can leave the default component _inside_ the PluginSlot wrapper ([see README for setup](https://github.com/openedx/frontend-plugin-framework?tab=readme-ov-file#host-micro-frontend-mfe)). 

What ends up taking the place of `defaultContents` is a boolean `keepDefault` that you can use to determine if you'd want to use the default component in your deploy.  

Note: there is an intention to make the default content more easily customizable when several components are involved (as was the intended feature with defining the content in `env.config`) but that implementation shouldn't require a drastic change in the setup you have now.

env.config.jsx
leangseu-edx: No, we need to deploy this part of the MFE for now. There isn't a good way to deploy it separately yet.

leangseu-edx: Sorry, I thought you were talking about the line. I am removing it. This file got through from rebasing.

leangseu-edx:
> Solely inputting on the Plugin framework implementation: this looks good! The implementation looks good, including the `env.config.jsx` example (assuming that the actual file was intended to be .gitignored). If the limitation for gitignoring it is because the default content is living in `env.config`, would I be able to recommend that you try the `@openedx/frontend-plugin-framework` package?
> 
> The main change you'd see is that `defaultContents` wouldn't be defined in `env.config.jsx`, but instead you can leave the default component _inside_ the PluginSlot wrapper ([see README for setup](https://github.com/openedx/frontend-plugin-framework?tab=readme-ov-file#host-micro-frontend-mfe)).
> 
> What ends up taking the place of `defaultContents` is a boolean `keepDefault` that you can use to determine if you'd want to use the default component in your deploy.
> 
> Note: there is an intention to make the default content more easily customizable when several components are involved (as was the intended feature with defining the content in `env.config`) but that implementation shouldn't require a drastic change in the setup you have now.

So I added a pr to make the package build correctly https://github.com/openedx/frontend-plugin-framework/pull/48. However, I still run into issue of unavailable context. When I used the transpiled code, the react component that replaced the slot lost the application context. I can't access application data.

src/courseware/course/sequence/Unit/index.jsx
```diff
@@ -28,16 +30,27 @@ const Unit = ({
...
+  }));
+
+  const iframeUrl = getUrl();
 
   return (
     <div className="unit">
-      <h1 className="mb-0 h3">{unit.title}</h1>
+      <div className="mb-0">
+        <h3 className="h3">{unit.title}</h3>
+        {/* <PluginSlot
```
Rodra: Should this be commented ?

Rodra (approved):
Just one question, but apart from that this looks AWESOME

src/courseware/course/sequence/Unit/index.jsx
leangseu-edx: No but the package `@openedx/frontend-plugin-framework` doesn't work as it should be. I have commented for now while working for solution.

Rodra (approved):

src/courseware/course/sequence/Sequence.jsx
```diff
@@ -144,57 +148,60 @@ const Sequence = ({
...
+              }}
+            />
+            )}
+          </div>
         </div>
+        {enableNewSidebar === 'true' ? <NewSidebar /> : <Sidebar />}
       </div>
-      {enableNewSidebar === 'true' ? <NewSidebar /> : <Sidebar />}
-    </div>
+      <div id="whole-course-translation-feedback-widget" />
```
arbrandes, line 203: We will likely have to convert `sequence-container` to a plugin slot so you can insert this div as part of the plugin.  Thoughts?

plugins/UnitTranslationPlugin/index.jsx
```diff
@@ -0,0 +1,43 @@
...
+  );
+};
+
+UnitTranslationPlugin.propTypes = {
+  id: PropTypes.string.isRequired,
+  courseId: PropTypes.string.isRequired,
+  unitId: PropTypes.string.isRequired,
+};
+
+export default UnitTranslationPlugin;
```
arbrandes, line 43: We're going to have to move this plugin out of this repository in the next couple of weeks, before `redwood.master` is created.

example.env.config.jsx
```diff
@@ -0,0 +1,24 @@
+import UnitTranslationPlugin from '@plugins/UnitTranslationPlugin';
```
arbrandes, line 1: It's great for documentation to have an example plugin defined, but it's unfortunate that we won't be able to leave it pointing to this particular plugin.

.env
```diff
@@ -4,6 +4,7 @@
 NODE_ENV='production'
 
 ACCESS_TOKEN_COOKIE_NAME=''
+AI_TRANSLATIONS_URL=''
```
arbrandes, line 7: We won't be able to leave this in `master` for much longer.  See other comments.

arbrandes (commented):
Can't thank you enough for adopting FPF and using it for AI translations, @leangseu-edx.  Otherwise, beyond it being unfortunate that there are a lot of unrelated changes in this PR (I'd've preferred them in a separate one), there is a big stickler here: we can't have the plugin live in the repository itself.  It'll have to be removed from master this week so that it doesn't land in Redwood.

There's also a specifc place (`sequence-container`) that will need to be converted to a slot.  See comment.

src/courseware/course/sequence/Sequence.jsx
leangseu-edx: I was thinking of doing that as well. It require an extra fetch to the server for the same response. If that is a prefer way, I don't mind.

plugins/UnitTranslationPlugin/index.jsx
leangseu-edx: Should I start the process for make a monorepo or do we have one already?

.env
leangseu-edx: This can be move to `env.config.js`

example.env.config.jsx
leangseu-edx: I thought we would leave this as an example.

plugins/UnitTranslationPlugin/index.jsx
arbrandes: @leangseu-edx, this plugin can live anywhere at 2U's discretion, with one exception: it can't be in the openedx org.  Otherwise, it's up to you folks whether it is even public.

example.env.config.jsx
arbrandes: Sure, but let's at least make it clearly fictitious.  Or even better, maybe make a skeleton sample plugin that lives in `plugins`, and point to that.

This is not the most urgent bit, though.

.env
arbrandes: To 2U's own version of it, yes, but preferably not the example one that lives in this repository.

src/courseware/course/sequence/Sequence.jsx
arbrandes: If a plugin needs to make extra requests, that is not only fine, but expected.  I would avoid introducing extra requests to the default experience, though.

To be clear, all I mean to suggest is for `sequence-container` to either be wrapped in a `<PluginSlot />` or contain one or more of them, not that it should become a plugin itself.  In other words, what exists today should at most be "default content".

PR #1341, opened 2024-03-29
fix: modal iframe height to 100%

ilee2u:
Proposed Solution for this CR: https://2u-internal.atlassian.net/browse/COSMO-227

The iframe within a modal in the learning MFE was set to a default height of 100vh. It should actually be 100% in order to match the height of the modal it's in, not the view height itself


schenedx (approved):
👍

michaelroytman (approved):

leangseu-edx (approved):
Seem like a reasonable change.

PR #1345, opened 2024-04-05
feat(AU-1949): Restrict WCT to verified track learners only

Rodra:
https://2u-internal.atlassian.net/browse/AU-1949

### DIFF
--- a/plugins/UnitTranslationPlugin/index.test.jsx
+++ b/plugins/UnitTranslationPlugin/index.test.jsx
@@ -24,7 +24,9 @@ describe('<UnitTranslationPlugin />', () => {
     useState.mockReturnValue([{ enabled, availableLanguages }, jest.fn()]);
   };
   it('render empty when translation is not enabled', () => {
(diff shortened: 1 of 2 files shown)


plugins/UnitTranslationPlugin/index.test.jsx
```diff
@@ -24,15 +24,19 @@ describe('<UnitTranslationPlugin />', () => {
     useState.mockReturnValue([{ enabled, availableLanguages }, jest.fn()]);
   };
   it('render empty when translation is not enabled', () => {
-    useModel.mockReturnValue({ language: 'en' });
+    useModel
+      .mockReturnValueOnce({ language: 'en' })
+      .mockReturnValueOnce({ verifiedMode: { accessExpirationDate: null } });
```
leangseu-edx, line 33: Try to use `jest-when` to mock instead. This rely on the order of the useModel get call in `index.jsx` file.

leangseu-edx (approved):
LGTM 👍

PR #1346, opened 2024-04-08
fix: fixed zindex and width issue

sundasnoreen12:
[INF-1348](https://2u-internal.atlassian.net/browse/INF-1348)

**Description**

When accessing the following module link, the Discussion sidebar to the right appears but is not accessible while using a Chrome browser.

[Course | edX](https://learning.edx.org/course/course-v1:MITx+6.431x+1T2024/block-v1:MITx+6.431x+1T2024+type@sequential+block@Lec__5_Probability_mass_functions_and_expectations/block-v1:MITx+6.431x+1T2024+type@vertical+block@ch6-s2-tab2)

The error is not replicated when using Firefox.

**Solution:** 

The problem has been resolved by adjusting the width for larger screens.

### DIFF
--- a/src/courseware/course/new-sidebar/common/SidebarBase.jsx
+++ b/src/courseware/course/new-sidebar/common/SidebarBase.jsx
@@ -41,7 +41,7 @@ const SidebarBase = ({
 
   return (
     <section
-      className={classNames('ml-0 ml-lg-4 h-auto align-top', {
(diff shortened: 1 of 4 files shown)


leangseu-edx (approved):
LGMT 👍

awais-ansari (approved):

PR #1347, opened 2024-04-08
feat: add plugin slot for fbe lock paywall

alangsto:
## [COSMO-240](https://2u-internal.atlassian.net/browse/COSMO-240)

This PR creates a plugin slot for FBE lock paywall messaging. The existing lock paywall messaging is set as the default, so there is no change to the messaging unless otherwise configured.

### DIFF
--- a/src/courseware/course/sequence/Unit/UnitSuspense.jsx
+++ b/src/courseware/course/sequence/Unit/UnitSuspense.jsx
@@ -2,6 +2,7 @@ import React, { Suspense } from 'react';
 import PropTypes from 'prop-types';
 
 import { useIntl } from '@edx/frontend-platform/i18n';
+import { PluginSlot } from '@openedx/frontend-plugin-framework';
 
 import { useModel } from '@src/generic/model-store';
 import PageLoading from '@src/generic/PageLoading';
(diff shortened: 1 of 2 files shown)


schenedx (approved):
👍

michaelroytman (approved):

src/courseware/course/sequence/Unit/UnitSuspense.jsx
```diff
@@ -24,19 +25,24 @@ const UnitSuspense = ({
...
       {shouldDisplayContentGating && (
-        suspenseComponent(messages.loadingLockedContent, LockPaywall)
+        <Suspense fallback={<PageLoading srMessage={formatMessage(messages.loadingLockedContent)} />}>
+          <PluginSlot
+            id="fbe_message_plugin"
+            pluginProps={{
+              courseId,
+            }}
+          >
+            <LockPaywall courseId={courseId} />
```
arbrandes, line 38: Putting things behind a PluginSlot is a good first step, but the point is to make it so the default behavior _doesn't_ include unwanted features.  So, as a next step, we're going to have to remove `LockPaywall` altogether.

PR #1350, opened 2024-04-09
chore: patch frontend-build deployment

leangseu-edx:
This pr is similar to https://github.com/openedx/frontend-app-learner-record/pull/302/

The can be remove once https://github.com/openedx/frontend-build/pull/515 is merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "lint": "fedx-scripts eslint --ext .js --ext .jsx .",
     "lint:fix": "fedx-scripts eslint --fix --ext .js --ext .jsx .",
     "prepare": "husky install",
+    "postinstall": "patch-package",
     "snapshot": "fedx-scripts jest --updateSnapshot",
     "start": "fedx-scripts webpack-dev-server --progress",
     "test": "fedx-scripts jest --coverage --passWithNoTests"
@@ -84,6 +85,7 @@
     "jest": "^26.6.3",
     "jest-console-group-reporter": "^1.0.1",
     "jest-when": "^3.6.0",
+    "patch-package": "^8.0.0",
     "postcss-loader": "^8.1.1",
     "rosie": "2.1.1",
     "sass": "^1.72.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -84,6 +85,7 @@
     "jest": "^26.6.3",
     "jest-console-group-reporter": "^1.0.1",
     "jest-when": "^3.6.0",
+    "patch-package": "^8.0.0",
```
BrandonHBodine, line 88: What is the intent of using this package for?

leangseu-edx: `patch-package` will try to read `@openedx+frontend-build+13.0.30.patch` file after install. If you look at the file `@openedx+frontend-build+13.0.30.patch`, it is just telling `node_module/@openedx/frontend-build` to update specific file. Essentially the change for the that was just so `webpack` would pick up `JS_CONFIG_FILEPATH` environment variable

BrandonHBodine: Interesting; this does seem pretty complex though. What's the need for injecting this patch code here instead of updating the https://github.com/openedx/frontend-build package?

BrandonHBodine: NVM, I see that this is a temporary fix for that pr in the frontend-build like you mention in the PR description,

BrandonHBodine (approved):

arbrandes:
@leangseu-edx, we're probably going to have to remove this unless a compelling reason is found for this `.env` variable to exist in the first place.  See https://github.com/openedx/frontend-build/pull/515#issuecomment-2079430708.

bradenmacdonald:
@leangseu-edx @arbrandes Is this patch still doing anything? I'd like to remove it, but it's unclear if it's still useful or not.

bradenmacdonald:
@jsnwesson Do you know if this patch can be removed?

arbrandes:
@bradenmacdonald, we should create a PR to remove it and give 2U a couple of days to review it.  Are you up for it, or should I?

bradenmacdonald:
@arbrandes Sure! I've opened it: https://github.com/openedx/frontend-app-learning/pull/1621 - but could you please help me get it reviewed?

PR #1351, opened 2024-04-11
build: Update to the latest codecov and add a repo token.

feanil:
Codecov is failing more often without this token so update to the latest
version and add a token to the github actions secrets for this repo that
can be used to fix coverage uploading.


jesperhodge:
I notified Aurora since they are owners and I don't want to overstep by approving in their repo. But to me this looks completely fine.

brian-smith-tcril (approved):
:shipit:

PR #1356, opened 2024-04-16
fix: fixed expended view unit bar issue

sundasnoreen12:
[INF-1348](https://2u-internal.atlassian.net/browse/INF-1348)

**Description**
Fixed expended view unit bar issue on XL screen.

**Screenshots**
**Before**
Screenshot 2024-04-16 at 3 05 05 PM

**After**
Screenshot 2024-04-16 at 3 05 21 PM

### DIFF
--- a/src/courseware/course/sequence/sequence-navigation/SequenceNavigationTabs.jsx
+++ b/src/courseware/course/sequence/sequence-navigation/SequenceNavigationTabs.jsx
@@ -30,7 +30,6 @@ const SequenceNavigationTabs = ({
         ref={containerRef}
         className={classNames('sequence-navigation-tabs-container', {
           'navigation-tab-width': isOnXLDesktop && currentSidebar,
-          'expanded-navigation-tab-width ': isOnXLDesktop && !currentSidebar,
         })}
       >
         <div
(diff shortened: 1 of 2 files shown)


src/courseware/course/sequence/sequence-navigation/SequenceNavigationTabs.jsx
```diff
@@ -23,7 +29,8 @@ const SequenceNavigationTabs = ({
       <div
         ref={containerRef}
         className={classNames('sequence-navigation-tabs-container', {
-          'navigation-tab-width': isOnXLDesktop,
+          'navigation-tab-width': isOnXLDesktop && currentSidebar,
+          'expanded-navigation-tab-width ': isOnXLDesktop && !currentSidebar,
```
awais-ansari: You don't need to add width to an expanded navigation tab. It should take the default width. Just add `'navigation-tab-width': isOnXLDesktop && currentSidebar,` when a sidebar is open and screensize is XL.

awais-ansari (changes requested):

awais-ansari (approved):

PR #1360, opened 2024-04-16
feat(AU-1894): send track event when requesting translation

Rodra:
https://2u-internal.atlassian.net/browse/AU-1894

### DIFF
--- a/plugins/UnitTranslationPlugin/translation-selection/index.jsx
+++ b/plugins/UnitTranslationPlugin/translation-selection/index.jsx
@@ -1,6 +1,7 @@
 import React, { useContext, useEffect } from 'react';
 import PropTypes from 'prop-types';
 
+import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 import { AppContext } from '@edx/frontend-platform/react';
 import { IconButton, Icon, ProductTour } from '@openedx/paragon';
 import { Language } from '@openedx/paragon/icons';
@@ -29,6 +30,20 @@ const TranslationSelection = ({
     courseId,
     language,
   });
+
+  useEffect(() => {
+    if (courseId && language && selectedLanguage && unitId && userId) {
+      const eventName = 'edx.whole_course_translations.translation_requested';
+      const eventProperties = {
+        courseId,
+        sourceLanguage: language,
+        targetLanguage: selectedLanguage,
+        unitId,
+        userId,
+      };
+      sendTrackEvent(eventName, eventProperties);
+    }
+  }, [courseId, language, selectedLanguage, unitId, userId]);
 
   useEffect(() => {
     dispatch(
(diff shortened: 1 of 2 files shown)


plugins/UnitTranslationPlugin/translation-selection/index.jsx
```diff
@@ -1,6 +1,7 @@
 import React, { useContext, useEffect } from 'react';
 import PropTypes from 'prop-types';
 
+import { sendTrackEvent } from '@edx/frontend-platform/analytics';
```
BrandonHBodine, line 4: What's the destination of this? Does it go to a specific platform?

Rodra: My understanding is that `sendTrackEvent` will send the event to Segment. These [docs](https://openedx.atlassian.net/wiki/spaces/AN/pages/13205895/Event+Design+and+Review+Process) may be out of date, but it’s what I’m using to differentiate the two. We use `sendTrackEvent` in the Open Courses Xpert Learning Assistant, for what it’s worth, and that’s logged to Segment and then made available in Snowflake.

plugins/UnitTranslationPlugin/translation-selection/index.jsx
```diff
@@ -30,6 +31,20 @@ const TranslationSelection = ({
     language,
   });
 
+  useEffect(() => {
+    if (courseId && language && selectedLanguage && unitId && userId) {
+      const eventName = 'edx.whole_course_translations.translation_requested';
+      const eventProperties = {
```
leangseu-edx, line 37: I am not sure if we should just send the event when `sourceLanguage !== targetLanguage`. I am pretty sure we can identify all of the users that didn't use the feature with page view.

plugins/UnitTranslationPlugin/translation-selection/index.jsx
```diff
@@ -30,6 +31,20 @@ const TranslationSelection = ({
...
+        courseId,
+        sourceLanguage: language,
+        targetLanguage: selectedLanguage,
+        unitId,
+        userId,
```
leangseu-edx, line 42: Remove `userId` it is PII.

plugins/UnitTranslationPlugin/translation-selection/index.jsx
Rodra: Yes I think adding that condition is OK, didn't understand last sentence. What's page view ?

plugins/UnitTranslationPlugin/translation-selection/index.jsx
Rodra: I had it as needed in the AC, could we still calculate engagement/completion accurately later ?

leangseu-edx (approved):

PR #1366, opened 2024-04-18
feat: Added PluginSlot wrapping UpgradeNotification components

rijuma:
# Description
Ticket: [COSMO-246 🔒](https://2u-internal.atlassian.net/browse/COSMO-246)

Adds a plugin slot wrapping the UpgradeNotification component in the sidebar notification panel and the sidebar on the course outline.

> [!IMPORTANT]  
> I've added a new `PluginSlot` component to wrap in the tests. The idea is that now supports children and the use of `testId`. We needed the update to be able to wrap a component to be updated by the Plugin config.

The parameters from the original UpgradeNotification component are passed to the PluginSlot as:
```{ upgradeNotificationProps: (...original props) }```


zacharis278 (changes requested):
looks good, my only comment is the naming convention for these slots is required to be generic and not just for upgrade widgets

rijuma:
> looks good, my only comment is the naming convention for these slots is required to be generic and not just for upgrade widgets

I've updated the slot ids to `notification_widget` and `outline_tab`.

zacharis278 (approved):

PR #1367, opened 2024-04-22
feat: Added Plugin Slot wrapping UpgradeNotification in NotificationTray

rijuma:
# Description
Ticket: [COSMO-246 🔒](https://2u-internal.atlassian.net/browse/COSMO-246)

In my previous PR https://github.com/openedx/frontend-app-learning/pull/1366 I forgot to add an extra slot in the `NotificationsTry` wrapping the `UpgradeNotifications` component.

### DIFF
--- a/src/courseware/course/sidebar/sidebars/notifications/NotificationTray.jsx
+++ b/src/courseware/course/sidebar/sidebars/notifications/NotificationTray.jsx
@@ -2,6 +2,7 @@ import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import classNames from 'classnames';
 import React, { useContext, useEffect, useMemo } from 'react';
 import { sendTrackEvent } from '@edx/frontend-platform/analytics';
+import { PluginSlot } from '@openedx/frontend-plugin-framework';
 import { useModel } from '../../../../../generic/model-store';
 import UpgradeNotification from '../../../../../generic/upgrade-notification/UpgradeNotification';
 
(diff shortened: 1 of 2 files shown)


zacharis278 (approved):

src/courseware/course/sidebar/sidebars/notifications/NotificationTray.jsx
```diff
@@ -75,21 +92,13 @@ const NotificationTray = ({ intl }) => {
...
-            org={org}
-            upgradeNotificationCurrentState={upgradeNotificationCurrentState}
-            setupgradeNotificationCurrentState={setUpgradeNotificationCurrentState}
-          />
+          <PluginSlot
+            id="notification_tray"
+            pluginProps={{ upgradeNotificationProps }}
+            testId="notification-tray-slot"
+          >
+            <UpgradeNotification {...upgradeNotificationProps} />
```
arbrandes, line 100: A few things:

* The `verifiedMode` check should be removed, and instead delegated to `pluginProps`
* The `UpgradeNotification` component should not be present by default
* If there are no other uses of `UpgradeNotification`, the component itself should be removed from the codebase
* The props are far too feature-specific; we should probably rely on global context in the future

arbrandes (commented):
This is a good step in the right direction, but there's more we'll have to do.

src/courseware/course/sidebar/sidebars/notifications/NotificationTray.jsx
zacharis278: @arbrandes can we move this conversation to open PR into main here since this was just a subset of the work? https://github.com/openedx/frontend-app-learning/pull/1368

arbrandes: Of course!  I didn't realize this was a PR into a feature branch.  :)

PR #1377, opened 2024-05-01
chore: make sidebar less intrusive for mobile

leangseu-edx:
Ticket: https://2u-internal.atlassian.net/browse/AU-1930

What changed?
- Mobile user will only need to need to close the notifications once.


BrandonHBodine (approved):
lgtm

PR #1379, opened 2024-05-02
feat: generic sidebar notification plugin

zacharis278:
~This is just a POC to prove out the API of params we're passing I didn't touch the new sidebar or write any tests.~

Removes upgrade component specific properties from sidebar/notification plugin slots.

Since we're merging into the feature branch this would be the resulting diff against main: https://github.com/openedx/frontend-app-learning/compare/master...zhancock/generic-notification-plugin

**Risks/Concerns**
This component is still highly coupled with the state of the learning MFE that detail is just hidden into the component now. If the redux store changes structure it could break our plugin. Once this is working it would be good to evaluate the component code we've copied over, there's probably a lot of conditions we can carve out if the upgrade component is 2U specific.

Pairs with https://github.com/edx/frontend-plugin-advertisements/pull/10
^^ in case this isn't publicly visible the approach is to useModel and get all of these fields rather than passing them over a pluggable API.

eg
```
  const {
    accessExpiration,
    marketingUrl,
    offer,
    timeOffsetMillis,
  } = useModel('coursewareMeta', courseId);
```

### DIFF
--- a/src/course-home/outline-tab/OutlineTab.jsx
+++ b/src/course-home/outline-tab/OutlineTab.jsx
@@ -196,7 +196,7 @@ const OutlineTab = ({ intl }) => {
             )}
             <CourseTools />
             <PluginSlot
-              id="outline_tab_notifications"
+              id="outline_tab_notifications_plugin"
               pluginProps={{ courseId }}
             >
               <UpgradeNotification
--- a/src/course-home/outline-tab/OutlineTab.jsx
+++ b/src/course-home/outline-tab/OutlineTab.jsx
@@ -124,20 +124,6 @@ const OutlineTab = ({ intl }) => {
     }
   }, [location.search]);
 
-  const upgradeNotificationProps = {
-    offer,
-    verifiedMode,
-    accessExpiration,
-    contentTypeGatingEnabled: datesBannerInfo.contentTypeGatingEnabled,
-    marketingUrl,
-    upsellPageName: 'course_home',
-    userTimezone,
-    timeOffsetMillis,
-    courseId,
-    org,
-    shouldDisplayBorder: true,
-  };
-
   return (
     <>
       <div data-learner-type={learnerType} className="row w-100 mx-0 my-3 justify-content-between">
@@ -210,11 +196,22 @@ const OutlineTab = ({ intl }) => {
             )}
             <CourseTools />
             <PluginSlot
-              id="outline_tab"
-              pluginProps={upgradeNotificationProps}
-              testId="outline-tab-slot"
+              id="outline_tab_notifications"
+              pluginProps={{ courseId }}
             >
-              <UpgradeNotification {...upgradeNotificationProps} />
+              <UpgradeNotification
+                offer={offer}
+                verifiedMode={verifiedMode}
+                accessExpiration={accessExpiration}
+                contentTypeGatingEnabled={datesBannerInfo.contentTypeGatingEnabled}
+                marketingUrl={marketingUrl}
+                upsellPageName="course_home"
+                userTimezone={userTimezone}
+                shouldDisplayBorder
+                timeOffsetMillis={timeOffsetMillis}
+                courseId={courseId}
+                org={org}
+              />
             </PluginSlot>
             <CourseDates />
             <CourseHandouts />
--- a/src/course-home/outline-tab/OutlineTab.test.jsx
+++ b/src/course-home/outline-tab/OutlineTab.test.jsx
@@ -132,19 +132,14 @@ describe('Outline Tab', () => {
       expect(expandedSectionNode).toHaveAttribute('aria-expanded', 'true');
     });
 
-    it('renders the Notification wrapper', async () => {
+    it('includes outline_tab_notifications_plugin slot', async () => {
       const { courseBlocks } = await buildMinimalCourseBlocks(courseId, 'Title', { resumeBlock: true });
       setTabData({
         course_blocks: { blocks: courseBlocks.blocks },
       });
       await fetchAndRender();
 
-      const pluginSlot = screen.getByTestId('outline-tab-slot');
-      expect(pluginSlot).toBeInTheDocument();
-
-      // The Upgrade Notification should be inside the PluginSlot.
-      const UpgradeNotification = pluginSlot.querySelector('.upgrade-notification');
-      expect(UpgradeNotification).toBeInTheDocument();
+      expect(screen.getByTestId('outline_tab_notifications_plugin')).toBeInTheDocument();
     });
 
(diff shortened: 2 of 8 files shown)


src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -210,11 +196,22 @@ const OutlineTab = ({ intl }) => {
...
             <CourseTools />
             <PluginSlot
-              id="outline_tab"
-              pluginProps={upgradeNotificationProps}
-              testId="outline-tab-slot"
+              id="outline_tab_notifications"
+              pluginProps={{ courseId }}
             >
-              <UpgradeNotification {...upgradeNotificationProps} />
+              <UpgradeNotification
```
zacharis278, line 202: revert back to what's in main today, no change in behavior if the plugin is off.

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -210,11 +196,22 @@ const OutlineTab = ({ intl }) => {
             )}
             <CourseTools />
             <PluginSlot
-              id="outline_tab"
-              pluginProps={upgradeNotificationProps}
-              testId="outline-tab-slot"
+              id="outline_tab_notifications"
```
zacharis278: I renamed this one so it's a bit more clear this is just plugging into the notification area

src/courseware/course/sidebar/sidebars/notifications/NotificationTray.jsx
```diff
@@ -94,10 +78,27 @@ const NotificationTray = ({ intl }) => {
         ? (
           <PluginSlot
             id="notification_tray"
-            pluginProps={upgradeNotificationProps}
-            testId="notification-tray-slot"
+            pluginProps={{
+              courseId,
+              notificationCurrentState: upgradeNotificationCurrentState,
```
zacharis278, line 83: This drives the 'red dot' notification icon. I've removed the 'upgrade' from how this is passed to the plugin to keep it general. On the learning MFE side all the variables relating to this are still called 'upgrade...' we should rename all of that eventually. Probably at the same time we remove all the upgrade code from the learning MFE.

I think these would belong on the API for this plugin slot since any plugin here may want to drive that icon. That said, the code for the red dot is tightly coupled to upgrade right now but you could use it creatively for other purposes. It would probably take further consideration on how we handle notifications from multiple sources (if you put more than 1 plugin here) when the need arises. My thinking is the existence of a getter/setter wouldn't need to change so what I have here should be a stable interface.

arbrandes:
> the approach is to useModel and get all of these fields rather than passing them over a pluggable API.

For better or worse, I think that will be inevitable in some of the first cases (like this one).  This will still help guide how to build out an API that we can consider stable, though.

src/tests/MockedPluginSlot.jsx
```diff
@@ -1,11 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const MockedPluginSlot = ({ children, testId }) => {
-  if (!testId) { return children ?? 'PluginSlot'; } // Return its content if PluginSlot slot is wrapping any.
-
-  return <div data-testid={testId}>{children}</div>;
-};
+const MockedPluginSlot = ({ children, id }) => (
```
zacharis278, line 4: I had to modify how this works a bit because we get react errors if `testId` is passed to the PluginSlot component. We can only pass id, pluginProps, and children. I opted to use `id` instead here since that would be required to have a functional plugin slot.

src/courseware/course/new-sidebar/sidebars/discussions-notifications/notifications/NotificationsWidget.test.jsx
```diff
@@ -80,7 +80,7 @@ describe('NotificationsWidget', () => {
     });
   });
 
-  it('renders upgrade card', async () => {
+  it('includes notification_widget_plugin slot', async () => {
```
zacharis278, line 83: for this and other components I split testing the plugin slot and behavior of the default content w/o a plugin since we expect the latter to be removed.

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -210,11 +196,22 @@ const OutlineTab = ({ intl }) => {
             )}
             <CourseTools />
             <PluginSlot
-              id="outline_tab"
-              pluginProps={upgradeNotificationProps}
-              testId="outline-tab-slot"
+              id="outline_tab_notifications_plugin"
```
zacharis278, line 199: changed the name on these to align convention with existing plugins `xxxxx_plugin`

src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -132,19 +132,14 @@ describe('Outline Tab', () => {
...
       });
       await fetchAndRender();
 
-      const pluginSlot = screen.getByTestId('outline-tab-slot');
-      expect(pluginSlot).toBeInTheDocument();
-
-      // The Upgrade Notification should be inside the PluginSlot.
-      const UpgradeNotification = pluginSlot.querySelector('.upgrade-notification');
-      expect(UpgradeNotification).toBeInTheDocument();
+      expect(screen.getByTestId('outline_tab_notifications_plugin')).toBeInTheDocument();
```
rijuma, line 142: Why removing the check for `UpgradeNotification`? If the idea is to make the test case simpler by checking the slot presence only I still think that we should check that `UpgradeNotification` is rendered in another test.

varshamenon4 (approved):
lgtm!

src/course-home/outline-tab/OutlineTab.test.jsx
rijuma: I just saw below you mentioned that the behavior was split into separate tests, but I'll leave the comment here since I'm not seeing an added test for this file. This test case was previously used to check the existence of the `UpgradeNotifications` before we added the plugin slot, that's why I think it's removing that case. Feel free to ignore my comment if I'm mistaken.

rijuma (approved):

src/course-home/outline-tab/OutlineTab.test.jsx
zacharis278: There's a whole set of tests for the existing upgrade notification further down the file: https://github.com/openedx/frontend-app-learning/blob/eba6583a7093bccd3518e0c8df6c83dbd8ee7328/src/course-home/outline-tab/OutlineTab.test.jsx#L1179

rijuma: You're right. I thought this test was modified when the Plugin Slot was added but it was a new test case. My mistake.

PR #1382, opened 2024-05-08
feat: update courseware search button

varshamenon4:
Update Courseware Search Toggle to be easier to find and more descriptive.

**Small screen:**
Screenshot 2024-05-17 at 9 03 20 AM

**Medium screen:**
Screenshot 2024-05-17 at 9 03 31 AM

**Large screen:**
Screenshot 2024-05-17 at 9 03 43 AM

### DIFF
--- a/src/course-home/courseware-search/CoursewareSearchToggle.jsx
+++ b/src/course-home/courseware-search/CoursewareSearchToggle.jsx
@@ -1,6 +1,6 @@
 import React, { useEffect } from 'react';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
-import { Button, Icon } from '@openedx/paragon';
+import { Button } from '@openedx/paragon';
 import { Search } from '@openedx/paragon/icons';
 import { useDispatch } from 'react-redux';
 import messages from './messages';
@@ -25,16 +25,17 @@ const CoursewareSearchToggle = ({
   if (!enabled) { return null; }
 
   return (
-    <div className="courseware-searc-toggle">
+    <div className="courseware-search-toggle">
       <Button
-        variant="tertiary"
+        variant="outline-primary"
         size="sm"
         className="p-1 mt-2 mr-2 rounded-lg"
         aria-label={intl.formatMessage(messages.searchOpenAction)}
         onClick={handleSearchOpenClick}
         data-testid="courseware-search-open-button"
+        iconAfter={Search}
       >
-        <Icon src={Search} />
+        {messages.contentSearchButton}
       </Button>
     </div>
   );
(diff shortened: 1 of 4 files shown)


src/course-home/courseware-search/CoursewareSearchToggle.jsx
```diff
@@ -25,16 +25,17 @@ const CoursewareSearchToggle = ({
...
+    <div className="courseware-search-toggle">
       <Button
-        variant="tertiary"
+        variant="outline-primary"
         size="sm"
         className="p-1 mt-2 mr-2 rounded-lg"
         aria-label={intl.formatMessage(messages.searchOpenAction)}
         onClick={handleSearchOpenClick}
         data-testid="courseware-search-open-button"
+        iconAfter={Search}
```
varshamenon4: I tried looking for the icon that matched the one in the figma here: https://paragon-openedx.netlify.app/foundations/icons/ but couldn't figure out how to find. Is there a way to find within the figma?

src/course-home/courseware-search/messages.js
```diff
@@ -2,79 +2,84 @@ import { defineMessages } from '@edx/frontend-platform/i18n';
 
 const messages = defineMessages({
   searchOpenAction: {
-    id: 'learn.coursewareSerch.openAction',
+    id: 'learn.coursewareSearch.openAction',
```
varshamenon4, line 5: Since I'm already updating this file to add contentSearchButton, I'm just fixing the typo in coursewareSearch.

varshamenon4 (commented):
Just a couple of notes for review. Thanks!

zacharis278 (dismissed):

src/course-home/courseware-search/CoursewareSearchToggle.jsx
varshamenon4: Updated to ManageSearch

src/course-tabs/CourseTabsNavigation.jsx
```diff
@@ -16,23 +16,27 @@ const CourseTabsNavigation = ({
   return (
     <div id="courseTabsNavigation" className={classNames('course-tabs-navigation', className)}>
       <div className="container-xl">
```
varshamenon4, line 18: I added the CoursewareSearchToggle as a sibling component to the Tabs. I labeled the divs as rows/columns to avoid collision between the tabs and the search (this does work better as you can see in the screenshare/screenshot). I still need to figure out how to move the search toggle to the right as it was before. Maybe I need to update the size of the column so it is smaller? 2 main questions:

1. Is there a way to move the search to the right?
2. Are there potential problems/corner cases I could run into with this approach?

varshamenon4 (commented):
Figured out an approach to avoid collisions (see screenshots/screenshares), but the placement of the search toggle is now off so investigating that. Would appreciate any thoughts on this approach!

src/course-home/courseware-search/messages.js
rijuma: Whoopsie!

rijuma (approved):
LGTM

schenedx (approved):
👍

zacharis278 (approved):

zacharis278:
@varshamenon4 we chose to merge this and test on stage. This isn't enabled outside of edX courses yet so if there's any UX issues we can just follow up with another PR

PR #1394, opened 2024-05-23
style: use brand style for search button border

zacharis278:
The default theme uses rounded boarders but the edx.org theme does not. Not setting anything here will use the appropriate border styling based on the site theme.


alangsto (approved):

schenedx (approved):
👍

PR #1407, opened 2024-06-03
feat(AU-2073): Trigger track event when iframe fails on load

Rodra:


src/courseware/course/sequence/Unit/hooks/useIFrameBehavior.js
```diff
@@ -94,6 +96,11 @@ const useIFrameBehavior = ({
   const handleIFrameLoad = () => {
     if (!hasLoaded) {
       setShowError(true);
+      sendTrackEvent('edx.bi.error.learning.iframe_load_failed', {
+        iframeUrl,
+        unitId: id,
+        courseId,
```
nsprenkle: Strictly speaking, you don't need the `courseId` here since course ID is part of the unit ID:

`xblock-v1:{course_id}{block_id}`

nsprenkle (approved):

PR #1412, opened 2024-06-24
feat: Add PluginSlot wrapping SequenceNavigation next button

KristinAoki:
JIRA Ticket: [TNL-11662 🔒](https://2u-internal.atlassian.net/browse/TNL-11662)

Add new plugin slot that wraps the next button in `SequenceNavigation`.  The plugin will allow for content to appear after the next button is clicked before moving on the next subsection.

### DIFF
--- a/src/courseware/course/sequence/sequence-navigation/SequenceNavigation.jsx
+++ b/src/courseware/course/sequence/sequence-navigation/SequenceNavigation.jsx
@@ -10,8 +10,9 @@ import {
   isRtl,
   getLocale,
 } from '@edx/frontend-platform/i18n';
-
+import { PluginSlot } from '@openedx/frontend-plugin-framework';
 import { useSelector } from 'react-redux';
+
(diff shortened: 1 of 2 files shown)


src/courseware/course/sequence/sequence-navigation/SequenceNavigation.jsx
```diff
@@ -95,17 +101,37 @@ const SequenceNavigation = ({
...
-      <Button
-        variant="link"
-        className="next-btn"
-        onClick={nextHandler}
-        disabled={disabled}
-        iconAfter={nextArrow}
-        as={disabled ? undefined : Link}
-        to={disabled ? undefined : nextLink}
+      <PluginSlot
+        id="next_button_slot"
```
BrandonHBodine, line 105: Is there a naming convention that was establish for these plugin IDs?

BrandonHBodine (approved):
Left a question but lgtm.

Issue #1316, opened 2024-03-07
Node 20 upgrade


arbrandes:
### Description

See the parent ticket for the general description and motivation, https://github.com/openedx/public-engineering/issues/267.

```[tasklist]
- [ ] https://github.com/openedx/frontend-app-learning/pull/1445
- [ ] https://github.com/openedx/frontend-app-learning/pull/1443
- [ ] https://github.com/openedx/frontend-app-learning/pull/1454
```

PR #1445, opened 2024-08-23
test: Add Node 20 to CI matrix

BilalQamar95:
### Description

As a first step in the upgrade to Node 20, add it to the CI matrix as a non-blocking test.

See [the tracking issue](https://github.com/openedx/frontend-app-learning/issues/1316) for further information.

### DIFF
--- a/src/courseware/course/content-tools/calculator/Calculator.jsx
+++ b/src/courseware/course/content-tools/calculator/Calculator.jsx
@@ -341,7 +341,6 @@ class Calculator extends Component {
                         <li>arccsch(4x+y)</li>
                       </ul>
                     </td>
-                    { /* eslint-disable-next-line jsx-a11y/control-has-associated-label */ }
                     <td dir="auto" />
                   </tr>
                   <tr>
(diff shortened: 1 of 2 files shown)


src/courseware/course/content-tools/calculator/Calculator.jsx
```diff
@@ -341,6 +341,7 @@ class Calculator extends Component {
                         <li>arccsch(4x+y)</li>
                       </ul>
                     </td>
+                    { /* eslint-disable-next-line jsx-a11y/control-has-associated-label */ }
```
UsamaSadiq: Let's mention why this change is needed.

BilalQamar95: The error `(A control must be associated with a text label)` and the corresponding ESLint ignore might seem self-explanatory as the rule `jsx-a11y/control-has-associated-label` typically applies to interactive controls like buttons, inputs, or other form elements that require labels for accessibility. In this case, `<td>` is a table cell element, which doesn’t function as an interactive control and therefore doesn’t require a label. The `eslint-disable` comment was added to prevent unnecessary warnings for this specific use case.

UsamaSadiq (approved):

brian-smith-tcril (approved):

PR #1452, opened 2024-09-06
fix: show studio button if user has access

KristinAoki:
## Description

This PR adds `studio_access` attribute to a courses metadata for LMS requests. This field is used when determining is the "View in Studio" button is showed in the Instructors Toolbar. Currently the button appears for all staff members regardless of their studio permissions. However, the "Limited Staff" role does not have access to Studio, but is shown the Studio button and sees a 403 error when trying to load the Studio page. This change impacts Authors.

## Supporting information

JIRA Ticket: [AU-2055 🔒](https://2u-internal.atlassian.net/browse/AU-2055)

> Expected behavior
>
> The Studio button should not be visible since this role doesn’t have permission to view content in Studio.
> 
> Studio will not load because this role doesn’t grant access to it. 
> 
> Actual behavior
> The Studio button is visible even though this role doesn’t have permission to view content in Studio.
> 
> Studio does not load. Instead, an error message displays:
> 
>  > 403 Forbidden

## Testing instructions

Using `KristinAoki/add-cms-access-check-for-lms` branch

1. Go to Instructor > Membership > “Course Team Management”.
2. Select “Limited Staff” from the dropdown. 
3. Add a test account of yours to this role.
4. Save your changes.
5. As this test account, sign in.. 
6. Go to the course you added yourself to. 
7. See the Banner bar that contains the masquerade (“View this course as:”) toolbar.
8. Confirm that the Studio button is not present.
9. Log back into instructor account
10. Confirm that the Studio button is present.

## Deadline

None

## Other information

This is dependent on openedx/edx-platform PR [#35444](https://github.com/openedx/edx-platform/pull/35444)

### DIFF
--- a/src/instructor-toolbar/InstructorToolbar.jsx
+++ b/src/instructor-toolbar/InstructorToolbar.jsx
@@ -23,10 +23,10 @@ function getInsightsUrl(courseId) {
   return urlFull;
 }
 
-function getStudioUrl(courseId, unitId) {
+function getStudioUrl(courseId, unitId, hasStudioAccess) {
   const urlBase = getConfig().STUDIO_BASE_URL;
   let urlFull;
-  if (urlBase) {
+  if (urlBase && hasStudioAccess) {
     if (unitId) {
       urlFull = `${urlBase}/container/${unitId}`;
     } else if (courseId) {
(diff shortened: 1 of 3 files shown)


BrandonHBodine (approved):
lgtm!

jansenk (approved):

PR #1486, opened 2024-09-13
Revert "fix: show studio button if user has access"

KristinAoki:
Reverts openedx/frontend-app-learning#1452

This change works as expected for superusers and users with the "Staff" and "Limited Staff" roles. However, users with the "Admin" role are not seeing the Studio button even though they should have the permission to see it. Additional check will need to be added to the backend before this can be added back.


BrandonHBodine (approved):
lgtm

PR #1494, opened 2024-10-03
fix: pass extra prop to plugin slot

mubbsharanwar:
Passing model as a prop to plugin slot for dynamic model selection 

[SONIC-717](https://2u-internal.atlassian.net/browse/SONIC-717)


aht007 (approved):
LGTM

alangsto (approved):

PR #1507, opened 2024-10-29
feat: update courseware search api name

alangsto:
## [COSMO-309](https://2u-internal.atlassian.net/browse/COSMO-309)

Courseware search is not explicitly enabled by a flag on the backend, so the function should be renamed on the frontend.


rijuma (approved):

schenedx (commented):
👍

PR #1514, opened 2024-11-07
feat: add page not found route

KristinAoki:
## Description
This PR replace the blank screen for invalid URLs with a header, footer, and page not found message. The page not found page also sends a logging error and tracking event so that redirects to invalid urls can be monitored and tracked.

### Before
Screenshot 2024-11-07 at 4 29 08 PM

### After
Screenshot 2024-11-07 at 4 19 52 PM

## Supporting Information

JIRA Ticket: [AU-1049 🔒 ](https://2u-internal.atlassian.net/browse/AU-1049)

> Due to MFEs' current client-side routing implementation, users just saw a blank page with a footer (ie, a “client-side 404”) instead of a true 404 page. The routing failures yielded a 200 OK page and are not reported to the server in any way, so no alerts are triggered. For contrast, 404s in Legacy LMS pages are logged and also share a more coherent story with users.
>
> image

Testing

1. Navigate to a course
2. Change the URL to an invalid path
3. Confirm that the `PageNotFound` component is displayed
4. Click the homepage URL
5. Confirm that it navigates back to the learner dashboard

### DIFF
--- a/src/generic/PageNotFound.jsx
+++ b/src/generic/PageNotFound.jsx
@@ -1,0 +1,43 @@
+import { getConfig } from '@edx/frontend-platform';
+import { Hyperlink } from '@openedx/paragon';
+import { useIntl } from '@edx/frontend-platform/i18n';
+import FooterSlot from '@openedx/frontend-slot-footer';
+
+import HeaderSlot from '../plugin-slots/HeaderSlot';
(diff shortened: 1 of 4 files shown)


BrandonHBodine (approved):
lgtm

PR #1543, opened 2024-11-25
feat: wrap existing sidebars in `frontend-plugin-framework` `PluginSlot`s

brian-smith-tcril:
This PR is an MVP implementation of slots for the sidebars. The sidebars have grown quite complex over time, so the goal of this PR is to  "wrap the complexity" and ensure everything sidebar related is wrapped in a slot.

The existing show/hide logic is still intact *inside* the slots, so the slots will not be shown/hidden by the existing sidebar triggers. This was an intentional decision to avoid coupling the new slots to existing sidebar logic.

My hope is that once this PR lands we'll have a good jumping off point for discussions around how to architect the MFE to support simple and reasonable built-in sidebars as well as anything site operators want to build.


src/plugin-slots/CourseOutlineMobileSidebarTriggerSlot/README.md
```diff
@@ -0,0 +1,39 @@
+# Course Outline Sidebar Trigger Slot
+
+### Slot ID: `course_outline_mobile_sidebar_trigger_slot`
+
+## Description
+
+This slot is used to replace/modify/hide the course outline sidebar mobile trigger.
```
bradenmacdonald, line 7: But what _is_ the "course outline sidebar mobile trigger"?

src/plugin-slots/CourseOutlineSidebarSlot/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Slot ID: `course_outline_sidebar_slot`
+
+## Description
+
+This slot is used to replace/modify/hide the course outline sidebar.
+
+## Example
+
+![📊 in sidebar slot](./screenshot.png)
```
bradenmacdonald: These screenshots with some 📊 in the slot are kind of helpful, but it would also be helpful to see what the default content looks like. If we had to have only one screenshot, I think I would prefer something like this:

plugin-slot

src/plugin-slots/CourseOutlineSidebarSlot/README.md
```diff
@@ -0,0 +1,39 @@
+# Course Outline Sidebar Slot
+
+### Slot ID: `course_outline_sidebar_slot`
+
+## Description
+
+This slot is used to replace/modify/hide the course outline sidebar.
```
bradenmacdonald, line 7: I think it's worth mentioning that this slot is only visible if the `courseware.enable_navigation_sidebar` waffle flag is enabled.

src/plugin-slots/NotificationsDiscussionsSidebarSlot/index.jsx
```diff
@@ -0,0 +1,30 @@
...
+        mergeProps: true,
+      }}
+    >
+      {isNewDiscussionSidebarViewEnabled ? <NewSidebar /> : <Sidebar />}
+    </PluginSlot>
+  );
+};
+
+NotificationsDiscussionsSidebarSlot.propTypes = {
+  courseId: PropTypes.string.isRequired,
```
bradenmacdonald: Can we please use `.tsx` and TypeScript types instead of `.jsx` and `propTypes` for this sort of net new component?

src/plugin-slots/NotificationsDiscussionsSidebarSlot/index.jsx
```diff
@@ -0,0 +1,30 @@
...
+};
+
+export default NotificationsDiscussionsSidebarSlot;
```
bradenmacdonald: Nit: with the latest frontend-build, you don't have to use default exports, but you can if you want to.

bradenmacdonald (commented):
Cool. The discussion [here](https://github.com/openedx/frontend-app-learning/pull/1496#discussion_r1854649011) about providing `courseId` via standardized context is very relevant here too.

xitij2000:
Just one small nit. I think we can save some file size with not quality loss by running these png files through something like [oxipng](https://github.com/shssoichiro/oxipng). I did so on my PR and it saved around 30-40% on file sizes. They are small files but over time with a lot of images it'll add up.

src/plugin-slots/CourseOutlineSidebarSlot/README.md
brian-smith-tcril: That isn't the case for this slot. In the PR description I mentioned

> The existing show/hide logic is still intact inside the slots, so the slots will not be shown/hidden by the existing sidebar triggers. This was an intentional decision to avoid coupling the new slots to existing sidebar logic.

I have also added a `course_breadcrumbs_slot` as the breadcrumbs are shown/hidden based on the same waffle flag, so in the spirit of "wrapping the complexity" that logic is also contained within a slot now.

src/plugin-slots/CourseOutlineSidebarSlot/README.md
brian-smith-tcril: I've updated all the READMEs to have 2 screenshots now. 1 showing default content and 1 showing custom content.

src/plugin-slots/NotificationsDiscussionsSidebarSlot/index.jsx
brian-smith-tcril: Good call! I've updated all the slots to be `.tsx` now!

src/plugin-slots/NotificationsDiscussionsSidebarSlot/index.jsx
brian-smith-tcril: Updated to use named exports/imports everywhere!

brian-smith-tcril:
> Just one small nit. I think we can save some file size with not quality loss by running these png files through something like [oxipng](https://github.com/shssoichiro/oxipng). I did so on my PR and it saved around 30-40% on file sizes. They are small files but over time with a lot of images it'll add up.

Great suggestion! I optimized them here https://github.com/openedx/frontend-app-learning/compare/5b453a31c244950d557c25d98b56445edaab4525..7453d94b1d5c8fdfe0b91acded9789c2465bc772

src/plugin-slots/CourseOutlineMobileSidebarTriggerSlot/README.md
brian-smith-tcril: This ties in to the "wrapping complexity" goal of this PR. The course outline sidebar trigger (the button that opens up the sidebar) is displayed in a different location (both visually and within the DOM) on mobile vs on desktop. The `CourseOutlineTrigger` component is used in both `Sequence` and `Course`. In `Course`, the component is used with `isMobileView` always set to `true`, where as in `Sequence` it is always set to `false`.

With the new slots, `CourseOutlineMobileSidebarTriggerSlot` exists only in `Course`, and `CourseOutlineSidebarTriggerSlot` exists only in `Sequence`.

I'm not sure what the best way to communicate this complexity in the README would be, any suggestions are very warmly welcome!

bradenmacdonald: I see. I do think it's more clear with the updated README screenshots showing before+after! Otherwise I don't have any suggestions :/

bradenmacdonald (approved):
Other than the ongoing discussion about how to handle context/props like `courseId`, I think this looks great. Thanks for adding those additional screenshots to the READMEs - it really helps!

arbrandes (approved):
Works for me!  👍🏼

PR #1551, opened 2024-12-09
feat: add progress certificate status plugin slot

KristinAoki:
## Description
This PR adds a plugin slot for the `<CertificateStatus />` component to allow users to change the render depending on the learner's enrollment track. This change if to help better align the experiences of the home page sidebar and the progress page sidebar.

## Supplemental Information
JIRA Ticket: [AU-2305 🔒 ](https://2u-internal.atlassian.net/browse/AU-2305)
> We want to be able to align the experience on the progress page for audit users with the experience on the home page


hajorg (approved):
LGTM

PR #1578, opened 2025-01-27
feat: update next unit button plugin for left sidebar navigation usage

KristinAoki:
This PR makes the following changes:

1. Renames the `next_button_slot` to `next_unit_top_nav_trigger_slot` to better capture its role
2. Moves the plugin to the central plugin folder to make it less hidden in the code
3. Add the plugin to `UnitNavigation` when the buttons appear at the top of the unit page. This happens when using left sidebar navigation.


src/courseware/course/sequence/sequence-navigation/UnitNavigation.jsx
```diff
@@ -38,13 +38,33 @@ const UnitNavigation = ({
     const { exitActive, exitText } = GetCourseExitNavigation(courseId, intl);
     const buttonText = (isLastUnit && exitText) ? exitText : intl.formatMessage(messages.nextButton);
     const disabled = isLastUnit && !exitActive;
+    const variant = 'outline-primary';
```
BrandonHBodine, line 41: Nit: Pulling these out into variables is a little weird to me. I prefer the text being included in the props. Is this a normal practice in this codebase?

BrandonHBodine (approved):
lgtm

src/courseware/course/sequence/sequence-navigation/UnitNavigation.jsx
KristinAoki: I put them into a variable because they are being used in two spots and want to guarantee that their values always stay in sync

PR #1579, opened 2025-01-29
feat: update proctoring info panel api call

alangsto:
## [COSMO-626](https://2u-internal.atlassian.net/browse/COSMO-626)

All exams related requests should be routed through the exams service, if the `EXAMS_BASE_URL` is truthy. This follows the same pattern of https://github.com/openedx/frontend-lib-special-exams/blob/main/src/data/api.js.

### DIFF
--- a/src/course-home/outline-tab/OutlineTab.test.jsx
+++ b/src/course-home/outline-tab/OutlineTab.test.jsx
@@ -54,7 +54,7 @@ describe('Outline Tab', () => {
   const goalUrl = `${getConfig().LMS_BASE_URL}/api/course_home/save_course_goal`;
   const masqueradeUrl = `${getConfig().LMS_BASE_URL}/courses/${courseId}/masquerade`;
   const outlineUrl = `${getConfig().LMS_BASE_URL}/api/course_home/outline/${courseId}`;
-  const proctoringInfoUrl = `${getConfig().LMS_BASE_URL}/api/edx_proctoring/v1/user_onboarding/status?is_learning_mfe=true&course_id=${encodeURIComponent(courseId)}&username=MockUser`;
+  const proctoringInfoUrl = `${getConfig().EXAMS_BASE_URL}/api/v1/student/course_id/${encodeURIComponent(courseId)}/onboarding?username=MockUser`;
 
   const store = initializeStore();
   const defaultMetadata = Factory.build('courseHomeMetadata');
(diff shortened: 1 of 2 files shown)


src/course-home/outline-tab/OutlineTab.test.jsx
```diff
@@ -54,7 +54,7 @@ describe('Outline Tab', () => {
   const goalUrl = `${getConfig().LMS_BASE_URL}/api/course_home/save_course_goal`;
   const masqueradeUrl = `${getConfig().LMS_BASE_URL}/courses/${courseId}/masquerade`;
   const outlineUrl = `${getConfig().LMS_BASE_URL}/api/course_home/outline/${courseId}`;
-  const proctoringInfoUrl = `${getConfig().LMS_BASE_URL}/api/edx_proctoring/v1/user_onboarding/status?is_learning_mfe=true&course_id=${encodeURIComponent(courseId)}&username=MockUser`;
+  const proctoringInfoUrl = `${getConfig().EXAMS_BASE_URL}/api/v1/student/course_id/${encodeURIComponent(courseId)}/onboarding?username=MockUser`;
```
alangsto, line 57: I spent a while trying to figure out how to mock out `getConfig` to try adding a test for the legacy endpoint, but wasn't able to mock it successfully. I've updated the URL in the tests to match what is being called by the `getProctoringInfoData` in the case where `EXAMS_BASE_URL` is truthy, because that setting has a truthy value for the test environment.

varshamenon4: That makes sense! I think it's sufficient/okay to not have testing for the legacy endpoint here. That being said, is it possible to test locally instead? Also okay with leaving as is, but just figured it could be a nice check.

varshamenon4 (approved):
lgtm!

src/course-home/outline-tab/OutlineTab.test.jsx
alangsto: Yes I did test locally as well!

rijuma: Just leaving a comment: Is [this example](https://github.com/edx/frontend-plugin-recommendations/blob/78f0148fade9b015d84463f3e7b2eefdd8eda6bd/src/setupTest.js#L22) useful to mock the `getConfig` values?

varshamenon4: Nice!

alangsto: @rijuma that's what I tried to do for the mock, but it wasn't actually mocking the correct thing.

PR #1580, opened 2025-01-29
fix: accessibility issues on outline and unit pages

KristinAoki:
## Description

This PR fixes the following accessibility issues:

1. Header used for screenreader only text
2. Element focus when expanding and dismissing welcome message
3. Bookmark button using wrong ARIA attributing while processing bookmark status

## Supporting Information
JIRA Tickets: [AU-2368 🔒](https://2u-internal.atlassian.net/browse/AU-2368), [AU-2369 🔒](https://2u-internal.atlassian.net/browse/AU-2369), and [AU-2370 🔒](https://2u-internal.atlassian.net/browse/AU-2370)

AU-2368
> The visually hidden heading "Level 2 headings may be created by course providers in the future." acts as a
hint and does not qualify as a heading.
>
> Text that does not have the purpose of a heading either visually or structurally must not use heading markup, otherwise this will create a confusing heading structure for screen reader users. This is a failure of [WCAG 1.3.1 Info and Relationships](https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html).

AU-2369
>  In course updates, the “Show More” and “Dismiss” buttons dynamically add or remove content from the course home. Dynamically-added content needs to be added to the focus order directly after the element that exposed the content, or keyboard focus needs to be moved to the new content.
> 
> When keyboard users select the “Show More” button, the keyboard focus moves to the "Show Less" control instead of moving to the newly added content. Please move the keyboard focus to the newly added content on activating "Show More" control.
>
> When the “Dismiss” button is pressed, the keyboard focus returns to the body/top of the page, which is not a logical order. This can make the user interface difficult to navigate and understand. Ensure that the order in which elements receive keyboard focus follows a meaningful and logical sequence. For example, if the “Dismiss” button removes content, the keyboard focus could move to the previous interactive element.

AU-2370
> The “Bookmark this page” button toggles the aria-disabled attribute very quickly, which can be confusing to screen reader users if they are given delayed information about the button’s availability. Please remove the aria-disabled attribute and use an aria-busy attribute when the button is updating to avoid confusion for the screen reader users.

## Testing

AU-2369
1. Navigate to a course outline that has a welcome message
4. Using a keyboard, click on the "Show more" button
5. Press the tab button on the keyboard, confirm that the focus is not on the "Show less" button
6. Confirm that if there is a interactive element in the welcome message, it received the focus
7. Click the "Show less" button
8. Confirm that the focus remains on the button
9. Click on the "Dismiss" button and confirm that the focus moved to the "Expand all" button

AU-2370
1. Open the first unit in course
2. Confirm that the bookmark is outlined
3. Open browser dev tools
4. Confirm that `aria-busy=false` for the bookmark button
5. Click on the bookmark
6. Confirm that `aria-busy` was updated quickly
7. Confirm that the bookmark is filled

### DIFF
--- a/src/course-home/outline-tab/widgets/WelcomeMessage.jsx
+++ b/src/course-home/outline-tab/widgets/WelcomeMessage.jsx
@@ -1,7 +1,7 @@
-import React, { useState, useMemo } from 'react';
+import { useState, useMemo, useRef } from 'react';
 import PropTypes from 'prop-types';
 
-import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+import { useIntl } from '@edx/frontend-platform/i18n';
 import { Alert, Button, TransitionReplace } from '@openedx/paragon';
 import truncate from 'truncate-html';
 
@@ -11,11 +11,13 @@ import messages from '../messages';
 import { useModel } from '../../../generic/model-store';
 import { dismissWelcomeMessage } from '../../data/thunks';
 
-const WelcomeMessage = ({ courseId, intl }) => {
+const WelcomeMessage = ({ courseId, nextElementRef }) => {
+  const intl = useIntl();
   const {
     welcomeMessageHtml,
   } = useModel('outline', courseId);
 
+  const messageBodyRef = useRef();
   const [display, setDisplay] = useState(true);
 
   // welcomeMessageHtml can contain comments or malformatted HTML which can impact the length that determines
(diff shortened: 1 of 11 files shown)


src/course-home/outline-tab/widgets/WelcomeMessage.jsx
```diff
@@ -49,46 +51,56 @@ const WelcomeMessage = ({ courseId, intl }) => {
...
-          <LmsHtmlFragment
-            className="inline-link"
-            data-testid="long-welcome-message-iframe"
-            key="full-html"
-            html={cleanedWelcomeMessageHtml}
-            title={intl.formatMessage(messages.welcomeMessage)}
-          />
-        )}
-      </TransitionReplace>
+      {/* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */}
```
BrandonHBodine: This seems to be a meaningful lint. Is there supposed to be a `role` attribute on this `div`?

KristinAoki: I looked into a bit more and realized that I was using the wrong value for `tabIndex`. I have updated it to `tabIndex="-1"` which allows the div to be programmatically focusable, but not keyboard focusable. Making it not keyboard focusable is good because it will not disrupt the natural flow of the page while using keyboard navigation, but will allow us to focus on it when we need the content inside the `div`.

BrandonHBodine: Ah gotcha, that makes sense. Thanks for fixing that!

BrandonHBodine (approved):
lgtm!

PR #1581, opened 2025-01-30
fix: Unify Xpert audit trial eligibility between backend and frontend

rijuma:
## Description
Ticket: [COSMO-612 🔒](https://2u-internal.atlassian.net/browse/COSMO-612)

These changes improve the definition of the Xpert audit trial eligibility and make consistent between backend and frontend.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "start": "fedx-scripts webpack-dev-server --progress",
     "dev": "PUBLIC_PATH=/learning/ MFE_CONFIG_API_URL='http://localhost:8000/api/mfe_config/v1' fedx-scripts webpack-dev-server --progress --host apps.local.openedx.io",
     "test": "fedx-scripts jest --coverage --passWithNoTests",
+    "test:watch": "fedx-scripts jest --watch --passWithNoTests",
     "types": "tsc --noEmit"
   },
   "author": "edX",
--- a/src/courseware/course/chat/Chat.jsx
+++ b/src/courseware/course/chat/Chat.jsx
@@ -6,7 +6,7 @@ import { Xpert } from '@edx/frontend-lib-learning-assistant';
 import { getConfig } from '@edx/frontend-platform';
 import { injectIntl } from '@edx/frontend-platform/i18n';
 
-import { AUDIT_MODES, VERIFIED_MODES } from '@src/constants';
+import { ALLOW_UPSELL_MODES, VERIFIED_MODES } from '@src/constants';
 import { useModel } from '../../../generic/model-store';
 
 const Chat = ({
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -21,6 +21,7 @@
     "start": "fedx-scripts webpack-dev-server --progress",
     "dev": "PUBLIC_PATH=/learning/ MFE_CONFIG_API_URL='http://localhost:8000/api/mfe_config/v1' fedx-scripts webpack-dev-server --progress --host apps.local.openedx.io",
     "test": "fedx-scripts jest --coverage --passWithNoTests",
+    "test:watch": "fedx-scripts jest --watch --passWithNoTests",
```
rijuma, line 24: This is super handy to work on creating/updating tests.

src/courseware/course/chat/Chat.jsx
```diff
@@ -22,67 +22,48 @@ const Chat = ({
...
-  } = course;
+  // If is disabled or taking an exam, we don't show the chat.
+  if (!enabled || activeAttempt?.attempt_id || exam?.id) { return null; }
 
-  const hasVerifiedEnrollment = (
-    enrollmentMode !== null
-    && enrollmentMode !== undefined
-    && VERIFIED_MODES.includes(enrollmentMode)
-  );
+  // If is not staff and doesn't have an entollment, we don't show the chat.
```
alangsto: nit: typo in enrollment

src/courseware/course/chat/Chat.jsx
```diff
@@ -22,67 +22,48 @@ const Chat = ({
...
-    && AUDIT_MODES.includes(enrollmentMode)
+  const verifiedMode = VERIFIED_MODES.includes(enrollmentMode); // Enrollment verified
+  const auditMode = (
+    !isStaff
+    && !verifiedMode
+    && ALLOW_UPSELL_MODES.includes(enrollmentMode) // Can upgrade course
     && getConfig().ENABLE_XPERT_AUDIT
   );
+  // If user has no access, we don't show the chat.
+  if (!isStaff && !(verifiedMode || auditMode)) { return null; }
```
alangsto, line 39: Is there a reason this check is separated from not having an enrollment mode? It seems like they might be able to be combined.

alangsto (commented):
A few nits

src/courseware/course/chat/Chat.jsx
```diff
@@ -22,67 +22,48 @@ const Chat = ({
   } = useSelector(state => state.specialExams);
   const course = useModel('coursewareMeta', courseId);
 
-  const {
-    accessExpiration,
-    start,
-    end,
-  } = course;
+  // If is disabled or taking an exam, we don't show the chat.
+  if (!enabled || activeAttempt?.attempt_id || exam?.id) { return null; }
```
varshamenon4, line 26: Was there a reason for taking these out of the consts and returning null? I wonder if there is any downside to returning null values?

src/courseware/course/chat/Chat.jsx
varshamenon4: I agree this could be combined. Specifically I would suggest grouping the conditional statements into the following constants. I think this would reduce/remove the if / null statements. I also believe that using the constants would make more readable and necessitate fewer comments, since all of the conditional statements will be grouped by the constants. This would also require a more minimal refactor of the existing code (especially since this is openedx/Axim).

- hasVerifiedEnrollment (same as current)
- hasValidAuditEnrollmentAndXpertAccess (enrollmentMode !== null && enrollmentMode !== undefined && AUDIT_MODES.includes(enrollmentMode) AND ALLOW_UPSELL_MODES.includes(enrollmentMode)AND ENABLE_XPERT_AUDIT setting is true)

varshamenon4 (commented):
Looks good so far! Just some refactoring suggestions.

src/courseware/course/chat/Chat.jsx
rijuma: The common practice for a component that does not render is returning `null`.

src/courseware/course/chat/Chat.jsx
varshamenon4, line 1: Are there any tests that need to be updated or added?

src/courseware/course/chat/Chat.jsx
rijuma: I like to have these in the same way as is usually preferred early returns in functions, instead of wrapping the next code into unnecessary if statements.

src/courseware/course/chat/Chat.jsx
rijuma: I'm more inclined for the early `null` returns when we have enough information to tell that the component doesn't need to be render. Having the condition at the end means running unnecessary code and logic.

I thought it was easier to skip checking the other conditions if there was no enrollment, since having no enrollment would affect both, and it would avoid checking it in each condition as before.

rijuma: >Note: We agreed on a call to follow up on this approach.

src/courseware/course/chat/Chat.jsx
rijuma: Nope, I just swapped `AUDIT_MODES`  with `ALLOW_UPSELL_MODES` and rearranged the code, but the logic practically stays the same besides that.

src/courseware/course/chat/Chat.jsx
rijuma: Fixed. Thanks!

alangsto (approved):

varshamenon4 (approved):

PR #1582, opened 2025-02-06
feat(AU-2375): Update breadcrumbs and outline navigation bar UI

Rodra:
### Description

Extend CourseOutlineSidebarTriggerSlot props to amplify plugin usage.

### DIFF
--- a/src/courseware/course/sequence/Sequence.jsx
+++ b/src/courseware/course/sequence/Sequence.jsx
@@ -43,6 +43,7 @@ const Sequence = ({
     originalUserIsStaff,
   } = useModel('courseHomeMeta', courseId);
   const sequence = useModel('sequences', sequenceId);
+  const section = useModel('sections', sequence ? sequence.sectionId : null);
   const unit = useModel('units', unitId);
   const sequenceStatus = useSelector(state => state.courseware.sequenceStatus);
   const sequenceMightBeUnit = useSelector(state => state.courseware.sequenceMightBeUnit);
@@ -161,7 +162,13 @@ const Sequence = ({
   const defaultContent = (
     <>
       <div className="sequence-container d-inline-flex flex-row w-100">
-        <CourseOutlineSidebarTriggerSlot />
+        <CourseOutlineSidebarTriggerSlot
+          courseId={courseId}
+          sectionId={section ? section.id : null}
+          sequenceId={sequenceId}
+          isStaff={isStaff}
+          unitId={unitId}
+        />
         <CourseOutlineSidebarSlot />
         <div className="sequence w-100">
           {!isEnabledOutlineSidebar && (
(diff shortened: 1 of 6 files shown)


Rodra:
@openedx/committers-frontend-app-learning this PR is ready for review

bradenmacdonald:
I'll try to take a look by tomorrow.

src/courseware/course/sequence/Sequence.jsx
```diff
@@ -161,7 +162,13 @@ const Sequence = ({
   const defaultContent = (
     <>
       <div className="sequence-container d-inline-flex flex-row w-100">
-        <CourseOutlineSidebarTriggerSlot />
+        <CourseOutlineSidebarTriggerSlot
+          courseId={courseId}
```
bradenmacdonald: Per a big discussion on https://github.com/openedx/frontend-app-learning/pull/1496 , I think `courseId` should not be a prop, but instead any plugins that need it can get it with the `useContextId` hook. See [the example code for `ProgressTabCourseGradeSlot`](https://github.com/openedx/frontend-app-learning/tree/master/src/plugin-slots/ProgressTabCourseGradeSlot) which shows this.

Ideally, `isStaff` would be likewise always available via a `useIsStaff` or `useUser` hook, but I don't think it's been implemented/standardized yet.

I think the other props are reasonable for this.

bradenmacdonald (changes requested):
Mostly looks good. Please just swap out `courseId` in favor of `useContextId`, and can you please update [the slot README](https://github.com/openedx/frontend-app-learning/tree/master/src/plugin-slots/CourseOutlineSidebarTriggerSlot) to list the new props?

src/courseware/course/sequence/Sequence.jsx
Rodra: Regarding your suggestion on using `useContextId` to get the `courseId` instead of passing it as a prop to my plugin, I have an issue when trying to use it from the view that's inserting that plugin. I used the hook as in here https://github.com/openedx/frontend-app-learning/pull/1496/files#diff-295513048177ffa3f1a1d414820290303766b7c74b29050d29aea1a08a59305cR19, looks like that hook is using the `courseId` saved in `courseHome` from redux store but that data seems to be available from the course home page. When using the Sequence/Unit view, the data is available in `courseware` and `courseHome` has `null/undefined` data. Is there a courseware hook that I could use or should I go back to use the courseId as prop ?

Screenshot 2025-02-24 at 12 23 48 PM
Screenshot 2025-02-24 at 12 24 12 PM

bradenmacdonald: Could you just update the `useContextId` hook, so that it falls back? Like this:

```diff
- export const useContextId = () => useSelector<RootState>(state => state.courseHome.courseId);
+ export const useContextId = () => useSelector<RootState>(state => state.courseware.courseId ?? state.courseHome.courseId);
```

bradenmacdonald (approved):
LGTM!

jristau1984:
@bradenmacdonald can you merge this if you have no update requests?  The Aurora team does not have merge rights.  Thanks!

KristinAoki:
@jristau1984 I now have merge rights so I can merge it

PR #1614, opened 2025-02-28
feat: Search a11y updated behavior.

rijuma:
## Description

These changes fix several accessibility issues with the Courseware Search panel.

These changes include:
- Moving the close panel button closer to the search button, so it doesn't get lost on large viewports.
- Moved the close button tab order to the third place. When opening the search panel, the search input is auto focused, then the focus moves to the search button and after that the close button takes focus.
- Updated browser's back navigation behavior while the search dialog is open. If the search panel is open when going back it closes the dialog instead.
- Reworked the component to use the Dialog HTML element with the modal functionality. This adds more quality of life features like closing the panel when hitting escape or clicking on the backdrop and restricts the element focus to cycle through the panel controls only (before these changes, the user was able to get focus into elements behind the search panel).

## Screenshots

Screenshot 2025-02-28 at 09 38 59

### DIFF
--- a/.husky/_/husky.sh
+++ b/.husky/_/husky.sh
@@ -1,31 +1,9 @@
-#!/bin/sh
-if [ -z "$husky_skip_init" ]; then
-  debug () {
-    if [ "$HUSKY_DEBUG" = "1" ]; then
-      echo "husky (debug) - $1"
-    fi
-  }
+echo "husky - DEPRECATED
 
-  readonly hook_name="$(basename "$0")"
-  debug "starting $hook_name..."
+Please remove the following two lines from $0:
 
-  if [ "$HUSKY" = "0" ]; then
-    debug "HUSKY env variable is set to 0, skipping hook"
-    exit 0
-  fi
+#!/usr/bin/env sh
+. \"\$(dirname -- \"\$0\")/_/husky.sh\"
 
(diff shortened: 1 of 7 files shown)


.husky/_/husky.sh
```diff
@@ -1,31 +1,9 @@
-#!/bin/sh
-if [ -z "$husky_skip_init" ]; then
-  debug () {
-    if [ "$HUSKY_DEBUG" = "1" ]; then
-      echo "husky (debug) - $1"
-    fi
-  }
+echo "husky - DEPRECATED
```
schenedx: Can you confirm this husky change is intentional?

schenedx (approved):
LGTM. I appreciate the updates!

.husky/_/husky.sh
rijuma: Thank you Simon. I don't even know what made this change.

alangsto (approved):

.husky/_/husky.sh
bradenmacdonald: See https://github.com/openedx/frontend-app-learning/pull/1608 for more information. You will likely need to run `npx husky` after pulling the latest version of this repo. Also if you see a bunch of sass warnings, they're safe to ignore and I'll fix them soon :)

PR #1617, opened 2025-02-28
feat: Update previous and next unit navigation buttons design

Rodra:
### Axim Team:

We've received feedback about the placement of these buttons, and we'd like to make the following change on the open edx platform as well: Move the previous and next buttons at the top of the page down (in line with the content header). For this reason, I think it should be OK to make this change within the MFE itself.

### Description

The top previous and next buttons should be moved to match the design pictured below and attached.

Screenshot 2025-02-28 at 4 43 03 PM
Screenshot 2025-02-28 at 4 23 21 PM
Screenshot 2025-02-28 at 4 23 30 PM
Screenshot 2025-02-28 at 4 24 06 PM

There is a requirement from support to make sure the bottom of the page previous and next buttons remain

### After
Screenshot 2025-03-03 at 3 04 46 PM
Screenshot 2025-03-03 at 3 04 55 PM

### Caveat

Since there's an already margin and padding for the unit content the arrow buttons were added within the inside content without affecting that design. The wanted changes locate the buttons as if there was no margin but we are not skipping this default configuration unless is requested.

Screenshot 2025-03-03 at 3 06 57 PM

### DIFF
--- a/src/courseware/course/sequence/Sequence.jsx
+++ b/src/courseware/course/sequence/Sequence.jsx
@@ -196,7 +196,6 @@ const Sequence = ({
           )}
 
           <div className="unit-container flex-grow-1 pt-4">
-            {isEnabledOutlineSidebar && renderUnitNavigation(true)}
             <SequenceContent
               courseId={courseId}
               gated={gated}
@@ -204,6 +203,8 @@ const Sequence = ({
               unitId={unitId}
               unitLoadedHandler={handleUnitLoaded}
               isOriginalUserStaff={originalUserIsStaff}
+              isEnabledOutlineSidebar={isEnabledOutlineSidebar}
+              renderUnitNavigation={renderUnitNavigation}
             />
             {unitHasLoaded && renderUnitNavigation(false)}
           </div>
--- a/src/courseware/course/sequence/Sequence.jsx
+++ b/src/courseware/course/sequence/Sequence.jsx
@@ -196,6 +196,7 @@ const Sequence = ({
           )}
 
           <div className="unit-container flex-grow-1 pt-4">
+            {isEnabledOutlineSidebar && renderUnitNavigation(true)}
             <SequenceContent
               courseId={courseId}
               gated={gated}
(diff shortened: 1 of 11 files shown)


KristinAoki:
@Rodra the slack discussion link that you posted is to a private thread. Can you give a summary in the description of the discussion?

src/courseware/course/sequence/sequence-navigation/UnitNavigation.jsx
```diff
@@ -23,23 +23,31 @@ const UnitNavigation = ({
...
-      variant="outline-secondary"
-      buttonLabel={intl.formatMessage(messages.previousButton)}
-      buttonStyle="previous-button justify-content-center"
-      onClick={onClickPrevious}
-      previousLink={previousLink}
-    />
-  );
+  const renderPreviousButton = () => {
+    const buttonStyle = isAtTop
+      ? 'previous-icon-button text-dark mr-4'
```
KristinAoki:
```suggestion
      ? 'previous-icon-button text-dark mr-3'
```

src/plugin-slots/NextUnitTopNavTriggerSlot/index.tsx
```diff
@@ -47,6 +49,7 @@ export const NextUnitTopNavTriggerSlot : React.FC<Props> = ({
         nextLink,
         disabled,
         buttonText,
+        isAtTop,
```
KristinAoki, line 50: The same change needs to be added to `pluginProps`

src/courseware/course/sequence/Sequence.jsx
```diff
@@ -196,6 +196,7 @@ const Sequence = ({
           )}
 
           <div className="unit-container flex-grow-1 pt-4">
+            {isEnabledOutlineSidebar && renderUnitNavigation(true)}
```
KristinAoki: The designs have the buttons inline with the bookmark icon change. However, the styling and location of the bookmark button is not currently being changed. Can you make it so that the buttons are at the same level as the unit title?

KristinAoki (changes requested):

src/plugin-slots/NextUnitTopNavTriggerSlot/index.tsx
KristinAoki, line 1: Will you remove `courseId` from the props as it is best practice to fetch inside the plugin via `useContext`?

KristinAoki (approved):

PR #1644, opened 2025-03-20
fix: useExamAccess hook does not make call to fetchExamAccessToken after isExam changes.

michaelroytman:
### Description
This commit fixes a bug that prevents the `fetchExamAccessToken` hook in the `useExamAccess` hook from running. This occurs when the value of `isExam` returned by the `useIsExam` hook changes from `false` to `true`. Because the dependency array of the `useEffect` hook within the `useExamAccess` hook was `['id']`, the `useEffect` hook would not rerun on changes to `isExam`. The fix is to add `isExam` to the dependency array.

**Jira**: [COSMO-673](https://2u-internal.atlassian.net/browse/COSMO-683) (private)

### DIFF
--- a/src/courseware/course/sequence/Unit/hooks/useExamAccess.js
+++ b/src/courseware/course/sequence/Unit/hooks/useExamAccess.js
@@ -18,9 +18,10 @@ const useExamAccess = ({
   const fetchExamAccessToken = useFetchExamAccessToken();
 
   // NOTE: We cannot use this hook in the useEffect hook below to grab the updated exam access token in the finally
-  //       block, due to the rules of hooks. Instead, we get the value of the exam access token from a call to the hook.
+  //       block, due to the rules of hooks. Instead, we get the value of the exam access token from a call to
+  //       the hook below.
   //       When the fetchExamAccessToken call completes, the useExamAccess hook will re-run
-  //       (due to a change to the Redux store, and, thus, a change to the the context), at which point the updated
+  //       (due to a change to the Redux store, and, thus, a change to the context), at which point the updated
   //       exam access token will be fetched via the useExamAccessToken hook call below.
   //       The important detail is that there should never be a return value (false, '').
   const examAccessToken = useExamAccessToken();
@@ -35,7 +36,7 @@ const useExamAccess = ({
           logError(error);
         });
     }
-  }, [id]);
+  }, [id, isExam]);
 
   return {
     blockAccess,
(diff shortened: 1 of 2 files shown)


src/courseware/course/sequence/Unit/hooks/useExamAccess.js
```diff
@@ -35,7 +36,7 @@ const useExamAccess = ({
           logError(error);
         });
     }
-  }, [id]);
+  }, [id, isExam]);
```
varshamenon4, line 39: Nice find!

src/courseware/course/sequence/Unit/hooks/useExamAccess.test.jsx
```diff
@@ -16,10 +17,27 @@ jest.mock('@edx/frontend-lib-special-exams', () => ({
...
+
 const mockFetchExamAccessToken = jest.fn().mockImplementation(() => Promise.resolve());
 useFetchExamAccessToken.mockReturnValue(mockFetchExamAccessToken);
 
-const testAccessToken = 'test-access-token';
+const mockUseIsExam = (initialState = false) => {
+  const [isExam, setIsExam] = useState(initialState);
+
+  // This setTimeout block is intended to replicate the case where a unit is an exam, but
+  // the call to fetch exam metada has not yet completed. That is the value of isExam starts
```
varshamenon4: Is this supposed to be metadata?
```suggestion
  // the call to fetch exam metadata has not yet completed. That is the value of isExam starts
```

src/courseware/course/sequence/Unit/hooks/useExamAccess.test.jsx
```diff
@@ -43,31 +61,60 @@ describe('useExamAccess hook', () => {
...
 
-      const { result } = renderHook(() => useExamAccess({ id }));
+      const { result, waitForNextUpdate } = renderHook(() => useExamAccess({ id }));
       const { accessToken, blockAccess } = result.current;
 
       expect(accessToken).toEqual('');
       expect(blockAccess).toBe(true);
       expect(mockFetchExamAccessToken).toHaveBeenCalled();
+
+      // This is to get rid of the act(...) warning.
```
varshamenon4, line 74: Nice!

varshamenon4 (approved):
Nice! Just one nit/typo!

src/courseware/course/sequence/Unit/hooks/useExamAccess.test.jsx
michaelroytman: Thanks for catching that! I'll push up a fix.

PR #1645, opened 2025-03-21
feat: Add slots to add tab links

xitij2000:
# Description

This PR adds mechanisms to add course tabs using a frontend plugin. 

Private-Ref: https://tasks.opencraft.com/browse/BB-9341


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

xitij2000:
@bradenmacdonald Would love to get your thoughts on the approach used in this PR. It's doing a few non-standard things. 

We wanted to add a new course tab, for which we need two things.

1. We need some way to add an actual tab. This is a straightforward slot. 
2. We need some place for the tab to link to. 

To add a new page that feels like part of the platform we need some way to extend the routes to allow creating a new page. I tried making the routes pluggable, but the router refuses to work with components that aren't a Route. Which led to the approach used here, which is to add a config value for defining new routes, each route gets its own slot so you can create a new page there.

bradenmacdonald:
@xitij2000 I don't think plugins can add `<Route>` objects directly, because as you said, the Router won't accept any children that aren't `<Route>`. But I think what you _can_ do is add a catch-all route going to an empty page that has a plugin slot and on that page, each plugin can add a `<Routes><Route ... /></Routes>` entry as a separate slot, and that will work fine.

This is because:
* `<Routes>` can appear anywhere within the page hierarchy
* Multiple `<Routes>` can even appear on the same page

See https://github.com/openedx/frontend-app-authoring/blob/aeefcc639f430c4769b0903cad0da7977cb40cf3/src/pages-and-resources/PagesAndResources.jsx#L88-L93 for an example of this. It's not wrapped in a plugin slot, but it would work just the same way if it were.

Also, I think we want to move away from a very config-heavy plugin approach toward something where plugins "register themselves" and require minimal config other than installation, so I think this sort of approach (adding `<Routes>` to an empty slot) is better long-term.

itsjeyd:
Hey @xitij2000, are you still planning to continue working on this PR?

xitij2000:
@itsjeyd I'm looking into the budget situation for this, will update once that is clear.

xitij2000:
@bradenmacdonald  I've switched to the mechanism you recommended. One issue is that now I can no longer have a route like `/course/:courseId/xyz` but other than that it seems to work well.

bradenmacdonald:
@xitij2000 Sorry for the slow reply. If the slot is named "Course Page Slot" then maybe you should mount the slot as a catch-all for `/course/:courseId/*`, and then plugins _could_ add new routes like `/course/:courseId/xyz`. They couldn't add new routes like `/somethingelse/:courseId/`, but that's fine - I think the former is more consistent with the idea of a "course page slot" than the latter, and you could even add a separate "Learning Page Slot" for adding routes to `/` that aren't nested under existing paths.

xitij2000:
> @xitij2000 Sorry for the slow reply. If the slot is named "Course Page Slot" then maybe you should mount the slot as a catch-all for `/course/:courseId/*`, and then plugins _could_ add new routes like `/course/:courseId/xyz`. They couldn't add new routes like `/somethingelse/:courseId/`, but that's fine - I think the former is more consistent with the idea of a "course page slot" than the latter, and you could even add a separate "Learning Page Slot" for adding routes to `/` that aren't nested under existing paths.

I tried this and unfortunately it is simply not working. It keeps getting redirected to the home page. I think this is because there is already a match for /course/:courseId/:sequenceId which is also catchall. The only way out I think would be to have `/course/:courseId/plugin/*` or similar (i..e use `p` instead of plugin or `c` or some other separator like `_`).

bradenmacdonald:
@xitij2000 Ah, I see. Well thanks for trying. What do you think makes the most sense then? I kinda like the idea of having a a slot under `/course/:courseId/plugin/*` as you're suggesting.

Though maybe we should also ask @arbrandes if there's a more official way for plugins to add new URLs, especially in frontend-base world.

xitij2000:
> @xitij2000 Ah, I see. Well thanks for trying. What do you think makes the most sense then? I kinda like the idea of having a a slot under `/course/:courseId/plugin/*` as you're suggesting.
> 
> Though maybe we should also ask @arbrandes if there's a more official way for plugins to add new URLs, especially in frontend-base world.

That is a good point. I think there is a better way with frontend-base which AFAIK would work across all MFEs.  From what I remember with frontend-base you can simply extend the list of routes via frontend-template-site. So at that point of time this work might not be needed at all. 

Is my understandin correct @arbrandes? If so we can pause this work and just temporarily maintain the drift on the client side till the better approach from frontend-base is implemented.

arbrandes:
@xitij2000,

> you can simply extend the list of routes via frontend-template-site

That is correct.  Any plugin (any "app") will be able to easily extend routes.  You can see how it works [in the "slot showcase" plugin/app](https://github.com/openedx/frontend-base/blob/main/shell/dev/slotShowcase/app.tsx#L47-L54) (usable by running `npm run dev` in frontend-base).

xitij2000:
@arbrandes Thanks for clarifying!

@bradenmacdonald I think given the above it might make sense to refactor this PR to only contain the changes relevant to adding a new tab and remove the code related to custom routes since that approach will change with the move to frontend-base. For now we can keep this bit of code drift in the client branch. 

CC: @pkulkark

bradenmacdonald:
@xitij2000 OK, makes sense.

xitij2000:
@bradenmacdonald I've removed all the routes-related code.

src/plugin-slots/CourseTabLinksSlot/README.md
```diff
@@ -0,0 +1,45 @@
...
+          widget: {
+            id: 'custom_tab',
+            type: DIRECT_PLUGIN,
+            RenderWidget: ()=> (
+              <a
+                className={classNames('nav-item flex-shrink-0 nav-link')}
+                href="#"
+              >
+                Custom Tab
+              </a>
```
bradenmacdonald: Can the example use `CourseTabLink` or is that not part of the plugin API?

bradenmacdonald (commented):
@xitij2000 This looks good to me, though @farhaanbukhsh is the maintainer now and @arbrandes knows more about plugin slots and the future direction of this MFE. Do either of you guys want to look at this or approve it for us?

src/plugin-slots/CourseTabLinksSlot/README.md
xitij2000: That was a last minute refactoring so forgot to update the docs. It makes sense to have that here. Will update.

itsjeyd:
Hey @farhaanbukhsh and @arbrandes, a friendly reminder to check out this PR. Braden already [reviewed it](https://github.com/openedx/frontend-app-learning/pull/1645#pullrequestreview-3311060259), so a quick look from one of you would probably be enough to get it over the line.

src/course-tabs/CourseTabLink.tsx
```diff
@@ -0,0 +1,20 @@
...
+interface CourseTabLinkProps {
+  slug: string;
+  activeTabSlug?: string;
+  url: string;
+  title: string;
+}
+
+export const CourseTabLink = ({
+  slug, activeTabSlug, url, title,
+}:CourseTabLinkProps) => (
```
farhaanbukhsh:
```suggestion
}: CourseTabLinkProps) => (
```

src/course-tabs/CourseTabLinksList.tsx
```diff
@@ -0,0 +1,25 @@
...
+interface CourseTabLinkListProps {
+  tabs: Array<{
+    title: string;
+    slug: string;
+    url: string;
+  }>,
+  activeTabSlug?: string;
+}
+
+export const CourseTabLinksList = ({ tabs, activeTabSlug }:CourseTabLinkListProps) => (
```
farhaanbukhsh:
```suggestion
export const CourseTabLinksList = ({ tabs, activeTabSlug }: CourseTabLinkListProps) => (
```

src/course-tabs/CourseTabsNavigation.jsx
```diff
@@ -44,19 +48,4 @@ const CourseTabsNavigation = ({
...
-    slug: PropTypes.string.isRequired,
-    url: PropTypes.string.isRequired,
-  })).isRequired,
-};
-
-CourseTabsNavigation.defaultProps = {
-  activeTabSlug: undefined,
-  className: null,
-};
-
```
farhaanbukhsh, line 61: 👍🏾

farhaanbukhsh: 👍🏾

farhaanbukhsh (changes requested):
@xitij2000 this is good work and I have few nits but overall this looks good to me

xitij2000:
@farhaanbukhsh  I've addressed the points you raised. Do have a look and merge if it feel ready.

farhaanbukhsh (approved):
👍

Good work on the slot and it does add great functionality

- ✅ I tested this on Tutor devstack
- ✅ I read through the code
- ✅  I checked for accessibility issues, it uses Paragon which has accessibility built in.
- ✅  Includes documentation

PR #1647, opened 2025-03-24
feat: Drop canShowUpgradeSock course data.

feanil:
DEPR: https://github.com/openedx/edx-platform/issues/36429

This piece of data is not being used anywhere but was still being
consumed so just drop the data so that the backend can be updated to no
longer provide the data.

The backend API is being updated in https://github.com/openedx/edx-platform/pull/36436


bradenmacdonald (approved):
Thanks!

PR #1658, opened 2025-03-26
feat: add course end dashboard plugin slots

jansenk:
Add plugin slots to allow control of course end dashboard links


src/plugin-slots/CourseExitPluginSlots/CourseExitViewCoursesPluginSlot/index.jsx
```diff
@@ -0,0 +1,37 @@
...
+  content: PropTypes.shape({
+    href: PropTypes.string.isRequired,
+  }).isRequired,
+};
+
+const CourseExitViewCoursesPluginSlot = ({ href }) => (
+  <PluginSlot id="course_exit_view_courses_slot">
+    <ViewCoursesLink content={{ href }} />
+  </PluginSlot>
+);
```
KristinAoki:
```suggestion
const CourseExitViewCoursesPluginSlot = () => {
  const href = `${getConfig().LMS_BASE_URL}/dashboard`;
  return (
    <PluginSlot id="course_exit_view_courses_slot">
      <ViewCoursesLink content={{ href }} />
    </PluginSlot>
  );
};
```

src/courseware/course/course-exit/CourseExit.jsx
```diff
@@ -63,21 +61,10 @@ const CourseExit = ({ intl }) => {
...
     <>
-      <div className="row w-100 mt-2 mb-4 justify-content-end">
-        <Button
-          variant="outline-primary"
-          href={`${getConfig().LMS_BASE_URL}/dashboard`}
-        >
-          {intl.formatMessage(messages.viewCoursesButton)}
-        </Button>
-      </div>
+      <CourseExitViewCoursesPluginSlot href={`${getConfig().LMS_BASE_URL}/dashboard`} />
```
KristinAoki:
```suggestion
      <CourseExitViewCoursesPluginSlot />
```
See comment on plugin file

src/courseware/course/course-exit/DashboardFootnote.jsx
```diff
@@ -1,30 +1,27 @@
...
+  const intl = useIntl();
   const { courseId } = useSelector(state => state.courseware);
   const { org } = useModel('courseHomeMeta', courseId);
   const { administrator } = getAuthenticatedUser();
 
   const dashboardLink = (
     <Hyperlink
       style={{ textDecoration: 'underline' }}
-      destination={`${getConfig().LMS_BASE_URL}/dashboard`}
+      destination={content?.dashboardFootnoteUrl || `${getConfig().LMS_BASE_URL}/dashboard`}
```
KristinAoki: Is this conditional necessary? In the default version, you always pass in a url so it should always be defined.

KristinAoki (changes requested):

KristinAoki (approved):

src/plugin-slots/CourseExitPluginSlots/CourseExitViewCoursesPluginSlot/index.jsx
```diff
@@ -0,0 +1,39 @@
...
+  return (
+    <PluginSlot id="course_exit_view_courses_slot">
+      <ViewCoursesLink content={{ href }} />
+    </PluginSlot>
+  );
+};
+
+CourseExitViewCoursesPluginSlot.propTypes = {};
+
+export default CourseExitViewCoursesPluginSlot;
```
bradenmacdonald: This doesn't need to block this PR if you didn't know, but: whenever creating new files in this repo or any MFE repo, please:
1. Use TypeScript `.tsx` instead of `.jsx`
2. Use TypeScript types instead of `propTypes` - [propTypes have been deprecated for 8 years!](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-proptypes-and-defaultprops)

You also don't need to use default exports anymore, but you can if you want to.

jansenk: hey @bradenmacdonald thanks for the heads up. this PR fell by the wayside after the 2u shakeup.
I'm getting some CI`react/prop-types` linting failures because of missing propTypes. I'm surprised that's the case. I tried to find some other PRs where someone makes this change, and I don't see anyone using any kind of `/* eslint-disable react/prop-types */`. Am I missing something?

jansenk: https://github.com/openedx/frontend-app-learning/pull/1632 is a good example. The check is green and there are no ignore directives. I'm not sure what I'm missing.

bradenmacdonald: I haven't checked in detail, but it's likely that the linter is still requiring propTypes in .jsx files, but not in .tsx files. Could that be it? Because it's better to have propTypes than nothing, and you can't replace propTypes with types in a jsx file (unless you use JSDoc syntax but then you might as well convert to .tsx anyways)

jansenk: I changed them to .tsx files. I'm not particularly familiar with the minutiae of typescript but it seems like I needed Params in both places in order for the linter to be okay with the change
`const ComponentName: React.FC<Params> = ({ p1, p2, p3}: Params) => {`
Not sure why both were needed, again, the linked PR seems to get by just using the angle bracket / java-style-parameterized-class syntax, but a green checkmark is a green checkmark

bradenmacdonald: Yeah you should only need it in one or the other:

```tsx
const ComponentName = ({ p1, p2, p3}: Params) => {
```
or
```tsx
const ComponentName: React.FC<Params> = ({ p1, p2, p3}) => {
```

bradenmacdonald: But I guess there's no harm in specifying both.

nsprenkle (approved):

PR #1410, opened 2024-06-20
feat: removes sidebar upgrade notification component

zacharis278:
Followup to #1368. 

~These components are specific to edx.org and will be moved to a 2U owned plugin. There are some open questions and I'd like to go back and manually test everything once the 2U plugin is updated to no longer import these. Leaving this in draft state for now.~

edx.org has migrated these components to a 2U owned plugin. This code is no longer being used and can be deprecated. While the LockPaywall is also unused we have opted not to remove it as originally intended. See note below

##### Removed by this PR

- ~LockPaywall: Upgrade ad shown in place of graded unit if content gating is enabled~ No longer included in this PR
- UpgradeNotification: Sidebar upgrade advertisement
- ~UpsellBullets: Only used by Upgrade/Paywall~ Cannot be removed since we're keeping the LockPaywall

##### Not Removed by this PR
- Other instances of the Upgrade Button

##### Open Questions
Unsure on what to do with the LockPaywall. Should this actually be removed? It puts us in a weird state where 'nothing' would be displayed if content gating is enabled and that plugin slot isn't configured. I'm thinking this should either be reworked so the `shouldDisplayContentGating` logic moves into the plugin itself or we don't touch LockPaywall at all here.

**Decision: We will not be removing the LockPaywall as part of this effort**

PR #1675, opened 2025-04-14
feat: removes Upgrade Notification as default content

arbrandes:
As a follow-up to https://github.com/openedx/frontend-app-learning/pull/1368, remove the UpgradeNotification component from the sidebar's default content.


src/courseware/course/new-sidebar/sidebars/discussions-notifications/notifications/NotificationsWidget.test.tsx
```diff
@@ -130,44 +109,6 @@ describe('NotificationsWidget', () => {
...
-
-    await act(async () => {
-      fireEvent.click(closeNotificationButton);
-    });
-
-    await waitFor(() => {
-      expect(screen.queryByTestId(testId)).not.toBeInTheDocument();
-    });
-  });
-
```
arbrandes, line 170: These tests can no longer run because they rely on a "Close" button that can now only be provided by a plugin.

(I don't approve of the way this NotificationsWidget was implemented in the first place.  It should not have relied on an optional widget for an apparently essential feature.  But it's beyond the scope of this PR to fix the architecture here.)

arbrandes:
Picking up https://github.com/openedx/frontend-app-learning/pull/1410 since the original author left the project.

@bradenmacdonald and/or @brian-smith-tcril, mind giving this a thumbs up?

brian-smith-tcril (approved):
LGTM!

Not sure what's going on with the codecov upload action, I'm seeing an "Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable" error - not worth blocking this PR over.

PR #1676, opened 2025-04-15
Remove unused <ShareButton>, replace query-string with newer APIs

bradenmacdonald:
This PR:
* removes a `<ShareButton>` component which seems to have been an experimental thing that is no longer used.
* Replaces use of the NPM `query-string` package with `useSearchParams` from react-router and standard web APIs (`URLSearchParams` / `URL`).
* Converts some files from .js to .ts

Private ref: MNG-4362

PR #1703, opened 2025-05-09
chore: re-add query-string dependency

nsprenkle:
#1676  removed the `query-string` package. Several of our plugins still rely on this, though it should still be removed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,6 +36,7 @@
         "lodash.camelcase": "4.3.0",
         "postcss-loader": "^8.1.1",
         "prop-types": "15.8.1",
+        "query-string": "^7.1.3",
         "react": "^18.3.1",
         "react-dom": "^18.3.1",
         "react-helmet": "6.1.0",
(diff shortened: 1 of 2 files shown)


bradenmacdonald (approved):
Fine as a hotfix, but if the plugins depend on something like `query-string` they should be fixed to specify that in their own `package.json` dependencies.

nsprenkle:
> Fine as a hotfix, but if the plugins depend on something like `query-string` they should be fixed to specify that in their own `package.json` dependencies.

Question, our plugins do currently specify this as below but are still breaking. Is there a different way to specify that dependency that would cause it _not_ to break?

```json
    "peerDependencies": {
        ....
        "query-string": "*",
        "react": "*"
        ...
    },
```

nsprenkle:
> > Fine as a hotfix, but if the plugins depend on something like `query-string` they should be fixed to specify that in their own `package.json` dependencies.
> 
> Question, our plugins do currently specify this as below but are still breaking. Is there a different way to specify that dependency that would cause it _not_ to break?
> 
> ```json
>     "peerDependencies": {
>         ....
>         "query-string": "*",
>         "react": "*"
>         ...
>     },
> ```

Which is to say, in my mind this should work, but what I'm observing is that it isn't sufficient to actually ensure those dependencies are installed, based on how our framework / builds are setup.

bradenmacdonald:
I would have thought that would be enough. Perhaps the plugins aren't being installed correctly?

PR #1763, opened 2025-07-17
feat: use discount info endpoint for streak discount information

NawfalAhmed:
## Description

This PR aims to introduce an override for streak discount information based on an env variable DISCOUNT_CODE_INFO_URL

## Screenshots

StreakModal with No Discount
image

StreakModal with Discount 
image


syedsajjadkazmii (approved):

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -83,34 +93,48 @@ const StreakModal = ({
 
   // Ask ecommerce to calculate discount savings
   useEffect(() => {
-    if (streakDiscountCouponEnabled && verifiedMode && getConfig().ECOMMERCE_BASE_URL) {
-      calculateVoucherDiscount(discountCode, verifiedMode.sku, username)
-        .then(
-          (result) => {
+    (async () => {
+      let newDiscountPercentage = 0;
```
nsprenkle: Nit: naming. Calling something "new" is usually a code smell. Just call it "streakDiscountPercentage".

src/shared/streak-celebration/StreakCelebrationModal.jsx
```diff
@@ -83,34 +93,48 @@ const StreakModal = ({
...
+              // the same as the LMS. So we'll keep using the values in
+              // verifiedMode, just multiplied by the received percentage.
+              newDiscountPercentage = discountPercentage;
+              sendTrackEvent('edx.bi.course.streak_discount_enabled', {
+                course_id: courseId,
+                sku: verifiedMode.sku,
+              });
+            }
+          } else if (getConfig().ECOMMERCE_BASE_URL) {
+            const result = await calculateVoucherDiscount(discountCode, verifiedMode.sku, username);
```
nsprenkle: Nit: add better comments here explaining the two code paths. Also consider rolling the isApplicable logic, tracking, comments, etc. into a separate function.

```suggestion
// PSEUDOCODE 
        if (streakDiscountCouponEnabled && verifiedMode) {
          // If the discount service is available, use that to calculate discount.
          if (getConfig().DISCOUNT_CODE_INFO_URL) {
            const discountPercentage = await getDiscountCodeInfo(discountCode, courseId);
          } 
          // If the discount service is not available, fall back to ecommerce for discount calculation.
          else if (getConfig().ECOMMERCE_BASE_URL) {
            const result = await calculateVoucherDiscount(discountCode, verifiedMode.sku, username);
```

nsprenkle (approved):
Functionality is fine, several nits about readability naming, commenting, and making more readable functions that I would suggest before merging but are not blocking.

src/shared/streak-celebration/StreakCelebrationModal.jsx
NawfalAhmed: Sure will do that, why is it considered a code smell tho, is there a theory behind this 👀?

src/shared/streak-celebration/StreakCelebrationModal.jsx
NawfalAhmed: I refactored the code to move the discount percentage logic inside the functions themselves, so they only return the percentage, and added the comments on the if else.
Also moved the functions to utils file
everything looks good now?

nsprenkle (approved):

PR #1797, opened 2025-09-29
refactor: shift grade summary calculation to backend and display "hidden grades" label in the grade table

Anas12091101:
## Supporting tickets
Proposal: https://openedx.atlassian.net/wiki/spaces/OEPM/pages/5150769174/Proposal+Add+option+to+show+overall+assessment+results+after+due+date+but+hide+individual+assessment+results
https://github.com/openedx/platform-roadmap/issues/460

Internal ticket: https://github.com/mitodl/hq/issues/3859

## Description
This PR refactors the grade summary logic to delegate all calculation responsibilities to the backend.
Previously, the frontend was performing grade summary computations using data fetched from the API. Now, the API itself provides the fully computed grade summary, simplifying the frontend and ensuring consistent results across clients.

Additionally, a "Hidden Grades" label has been added in the grade summary table to clearly indicate sections where grades are not visible to learners.

Additionally, for visibility settings that depend on the due date, this PR adds a banner on the Progress page indicating that grades are not yet released, along with the relevant due date information.

edx-platfom and authoring MFE changes are required to test this PR. 
edx-platform PR: https://github.com/openedx/edx-platform/pull/37399
authoring MFE PR: https://github.com/openedx/frontend-app-authoring/pull/2489

## Screenshots

Screenshot 2025-09-29 at 10 05 05 PM

Screenshot 2025-09-29 at 10 09 03 PM

## Testing Instructions

1. Verify that changes don't break the logic of the progress page. 
1. Use this course tarball for testing: [course._4j9qu3f.tar.gz](https://github.com/user-attachments/files/22602706/course._4j9qu3f.tar.gz)

2. Verify that with the new visibility option:
    - Grades are included in the final course score after the due date passes.
    - Grades are not displayed in the Grade Summary table or Detailed Grades table.

3. Confirm that subsections with the new visibility option are not listed in the Detailed Grades table.

4. In the Grade Summary table, check two cases:
     - Mixed visibility: If some grades of an assignment type are hidden and some are visible, the visible grades appear with a “+ hidden grades” label (e.g., Lab exam in the screenshot).
     - Fully hidden: If all grades of an assignment type are hidden, a Lock icon is displayed (e.g., Final exam in the screenshot).

- For assignments where grades are held until after the due date, verify that a warning banner is displayed. The banner should indicate the latest due date when all grades will be released.

Screenshot 2025-09-29 at 10 20 27 PM


openedx-webhooks:
Thanks for the pull request, @Anas12091101!


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

itsjeyd:
@Anas12091101 Thank you for this contribution! Since it includes user-facing changes, you'll have to get approval on the [product proposal](https://openedx.atlassian.net/wiki/spaces/OEPM/pages/5150769174/Proposal+Add+option+to+show+overall+assessment+results+after+due+date+but+hide+individual+assessment+results) before engineering review can start.

I'll tag the team of product managers for you here.

@openedx/openedx-product-managers Could you please have a look at this PR and the corresponding [product proposal](https://openedx.atlassian.net/wiki/spaces/OEPM/pages/5150769174/Proposal+Add+option+to+show+overall+assessment+results+after+due+date+but+hide+individual+assessment+results), and guide @Anas12091101 on next steps?

This work might qualify as a [fast-track change](https://openedx.atlassian.net/wiki/spaces/COMM/pages/5089886250/Proposal+Update+the+Product+Review+Process#Fast-Track-Changes) but I'm not completely sure. So if you could check what the right product review track would be as a first step, that would be great.

--

CC @mphilbrick211 for https://github.com/openedx/edx-platform/pull/37399 and https://github.com/openedx/frontend-app-authoring/pull/2489

sarina:
@itsjeyd the propsosal is linked in the PR description (https://openedx.atlassian.net/wiki/spaces/OEPM/pages/5150769174/Proposal+Add+option+to+show+overall+assessment+results+after+due+date+but+hide+individual+assessment+results) and its been approved.

sarina:
@pdpinch remind me, did you create a Product Roadmap ticket for this? I think @itsjeyd is looking for the actual GitHub product roadmap ticket, not just the product proposal.

sarina:
Found it. @itsjeyd https://github.com/openedx/platform-roadmap/issues/460

pdpinch:
@Anas12091101 can you add some screenshots for the lower part of the progress page? We'd like to see how a hidden grade is handled there versus a non-hidden one. Along with that, it would be helpful to have more explanation of how the timing is handled differently between the old "never show assessment results" and this new "... but show overall assessment results after due date"

src/course-home/progress-tab/grades/messages.ts
```diff
@@ -21,6 +21,11 @@ const messages = defineMessages({
     defaultMessage: 'Your current grade is {currentGrade}%. A weighted grade of {passingGrade}% is required to pass in this course.',
     description: 'Alt text for the grade chart bar',
   },
+  courseGradeFooterDueDateNotice: {
+    id: 'progress.courseGrade.footer.dueDateNotice',
+    defaultMessage: 'Some assignment scores are not yet included in your total grade. These grades will be released by {dueDate}.',
+    description: 'This shown when there are pending assignments with due date in the future',
```
asadali145:
```suggestion
    description: 'This is shown when there are pending assignments with a due date in the future',
```

src/course-home/progress-tab/grades/messages.ts
```diff
@@ -148,6 +153,21 @@ const messages = defineMessages({
...
+  },
+  hiddenScoreInfoText: {
+    id: 'progress.hiddenScoreInfoText',
+    defaultMessage: 'Scores from assignments that count toward your final grade but some are not shown here.',
+    description: 'Information text about hidden score label',
+  },
+  hiddenScoreLockInfoText: {
+    id: 'progress.hiddenScoreLockInfoText',
+    defaultMessage: 'Shown when scores for an assignment type are hidden yet still counted toward the course grade.',
+    description: 'Information text about hidden score label when learner have limited access to grades feature',
```
asadali145:
```suggestion
    description: 'Information text about hidden score label when learners have limited access to grades feature',
```

src/course-home/progress-tab/grades/messages.ts
```diff
@@ -148,6 +153,21 @@ const messages = defineMessages({
...
+    description: 'Text to indicate that some scores are hidden',
+  },
+  hiddenScoreInfoText: {
+    id: 'progress.hiddenScoreInfoText',
+    defaultMessage: 'Scores from assignments that count toward your final grade but some are not shown here.',
+    description: 'Information text about hidden score label',
+  },
+  hiddenScoreLockInfoText: {
+    id: 'progress.hiddenScoreLockInfoText',
+    defaultMessage: 'Shown when scores for an assignment type are hidden yet still counted toward the course grade.',
```
asadali145:
```suggestion
    defaultMessage: 'Scores for an assignment type are hidden but still counted toward the course grade.',
```

src/course-home/progress-tab/utils.ts
```diff
@@ -5,3 +5,40 @@ export const showUngradedAssignments = () => (
...
+// Returns the subsections for an assignment type
+const getSubsectionsOfType = (assignmentType, sectionScores) => (sectionScores || []).reduce((acc, chapter) => {
+  const subs = (chapter.subsections || []).filter(
+    (s) => s.assignmentType === assignmentType,
+  );
+  return acc.concat(subs);
+}, []);
+
+// Returns True if this subsection is "hidden"
+const isSubsectionHidden = (sub) => sub.showGrades && sub.showCorrectness === 'never_but_include_grade';
```
asadali145: ShowGrades is false when due date is not passed. It results in 0% instead of a lock icon.

asadali145 (commented):
Overall looks good—just a minor enhancement and a few message updates.

Anas12091101:
> @Anas12091101 can you add some screenshots for the lower part of the progress page? We'd like to see how a hidden grade is handled there versus a non-hidden one.

@pdpinch, In the `Detailed Grades` section on the MFE progress page, we’re only displaying grades for those subsections whose scores are visible.

In my test course, there are two sections, each with five subsections, configured with different visibility settings:
- Section 1 includes subsections with “Never,” “Never but include in grade,” two “Always,” and “Always but past due” visibility.
- Section 2 includes three “Never but include in grade” subsections and two “Always” subsections.

The learner will only see the visible subsection scores in the table, as shown in the screenshot below.

Screenshot 2025-10-08 at 3 55 26 PM


> Along with that, it would be helpful to have more explanation of how the timing is handled differently between the old "never show assessment results" and this new "... but show overall assessment results after due date"

In the "Never show assessment results" option, the due date is not considered because those grades are excluded from the final results. In contrast, with the new option, the scores remain hidden from learners but are included in the final grades once the due date has passed.

src/course-home/data/api.js
```diff
@@ -3,25 +3,40 @@ import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
...
+      (points.reduce((a, b) => a + b, 0) / points.length).toFixed(4),
+    );
+    totalWeightedGrade = overallAverage * assignmentWeight;
+    if (visibleScores.length) {
+      const visibleAverage = parseFloat(
+        (visibleScores.reduce((a, b) => a + b, 0) / points.length).toFixed(4),
+      );
+      averageGrade = visibleAverage;
+      weightedGrade = averageGrade * assignmentWeight;
+    }
```
arbrandes: First: I'm always leery of manipulating critical data (such as a learner's grade) in the frontend.  This here seems like a pretty heavy lift.  And while I'm not an expert on how grading works in the backend, I expect this is duplicating backend logic.  Yes, I realize that this kind of calculation is already being done in this very file - which is unfortunate - but this is not all.

If I understand correctly from [the proposal](https://openedx.atlassian.net/wiki/spaces/OEPM/pages/5150769174/Proposal+Add+option+to+show+overall+assessment+results+after+due+date+but+hide+individual+assessment+results), the point of hiding scores is to prevent answer sharing.  However - correct me if I'm wrong - I deduce from this that the individual scores are actually being sent to the client, so that a crafty user could in fact find out the correspondence between all answers and scores.

If the whole premise is to prevent cheating, shouldn't we avoid passing all the individual grades to the frontend in the first place?

CC @pdpinch

arbrandes (commented):
The code itself seems generally fine, but as noted in the inline comment, I'm trying to understand whether the premise is being addressed properly.

src/course-home/data/api.js
pdpinch: Oh, yes, you are correct @arbrandes. We should be hiding these grades in the API.

Anas12091101:
@arbrandes, I have shifted the grades calculation logic to the backend in https://github.com/openedx/edx-platform/pull/37399 and adjusted this PR accordingly. This PR is ready for another look.

asadali145 (approved):
LGTM in testing!

Anas12091101:
@arbrandes, the grades calculation PR in edx-platform (https://github.com/openedx/edx-platform/pull/37399) has been merged. This one is ready for review as well.

arbrandes:
I'm trying to test this but am seeing some weird behavior with scores and visibility (A "show after due" problem showing the answer before the due date, a grade not being calculated correctly, etc).  Before I confirm this is a bug I'm going to reset my environment tomorrow and test from scratch.

arbrandes:
Ok, I ran into two issues, all of which you can see [in the recording of today's Frontend Working Group meeting](https://drive.google.com/file/d/1_S3hCtYwUyNE_Qr3s2Jd5-vjSbdO5g37/view) starting at 23:17.

1) If you set a section to "**Show assessment results when subsection is past due**", the results are actually shown _before_ the due date.  I'm sure this is not caused by this PR, but it may be a side effect of https://github.com/openedx/edx-platform/pull/37399.  Do you mind confirming that this also happens on your end?

2) When setting "**Never show individual assessment results, but show overall assessment results after due date**" on a subsection, if I manually alter the due dates of the assignments to one day in the past and then check progress, the assignment scores are not counted towards my total grade.  The release date on the new warning ("Some assignment scores are not yet included in your total grade. These grades will be released by...") still reflects the original date in the future.  Is this the intended behavior?  If so, how should I test this feature?  Do I have to wait for the time to actually elapse?

This is all on Tutor main with edx-platform@master updated to the latest code as of this morning, and a checkout of this PR and the corresponding one on frontend-app-authoring.  I created [course.skzk4g2n.tar.gz](https://github.com/user-attachments/files/23106375/course.skzk4g2n.tar.gz) from scratch.  And this is a screenshot of my progress after getting all the answers right and setting the corresponding due dates in the past:

image

Anas12091101:
@arbrandes, regarding the first issue: you are currently viewing the problem in staff mode, which is why the grade appears despite the due date. The relevant code responsible for displaying the correctness of a problem can be found here: [graders.py Line 491](https://github.com/openedx/edx-platform/blob/86d9b08b5d652333a446f5eeb0080d4158fca4e0/xmodule/graders.py#L491).

For the second issue, I believe that edX is not updating the due date in the LMS after changes are made in the CMS. I replicated this behavior on my local setup as well. After multiple updates to the due date, it eventually reflected correctly, and the progress page was updated accordingly. Could you please verify if this is the same issue you're encountering?

arbrandes (approved):
Alright, even though there might be an issue with resetting the due date of assignments after publishing, that would be a separate issue.  This PR is good.  Let's get it in!

Anas12091101:
Thanks @arbrandes. Could you please also merge the authoring MFE PR? https://github.com/openedx/frontend-app-authoring/pull/2489

nsprenkle:
Hey @Anas12091101 👋 If I can provide some feedback: the way this was implemented caused issues for us recently and I think there were ways it could have been avoided.

In particular, assuming that new data **will** be present in the backend while also removing the original implementation means that `frontend-app-learning` will now fail (and other operators / developers cannot author compatible changes) if `edx-platform` has not been previously deployed with the related changes ahead of time.

My suggestion, instead of assuming the new data _absolutely will be present_, this could be made more widely compatible by:
1. Checking the data returned by the API and seeing if the new data is present.
2. If the new data is present, great! It can be used and the costly frontend calculation of that data can be conditionally ignored.
3. If the new data is **not** present, gracefully fall back to the original way of calculating this data.

Still, I appreciate the work and I think this is a move in the right direction, but more care could be put into developing the change to be backward-compatible, at least for the time.

PR #1885, opened 2026-03-26
feat: decouple notifications panel using widget registry mechanism

awais-ansari:
### Summary
Removes the tightly coupled upgrade/upsell ("notifications") panel from the Learning MFE core and replaces it with a pluggable widget registry system. The right sidebar now supports dynamically-registered external widgets, making it easy to add, remove, or replace sidebar panels without forking the MFE. [Proposal](https://openedx.atlassian.net/wiki/spaces/OEPM/pages/5432344583/Proposal+Right+sidebar+refactoring)

### Motivation
The built-in upgrade panel (NotificationTray) caused several issues:

1. Misleading naming: "Notifications" implied platform-level notifications, conflicting with the separate notification tray in the header.
2. Non-Core: The upgrade panel is a non-core feature. It should not be bundled into the Open edX core.
3. Empty panel UX problem: In courses without a paid track, the panel rendered empty but remained visible.

### Architecture: Pluggable Widget Registry
The right sidebar now operates on a widget registry (SIDEBARS) built at runtime from:

Built-in widgets: only DISCUSSIONS are included by default.
External widgets: installed via npm and registered via SIDEBAR_WIDGETS in env.config.jsx
Each widget provides:
```
{
  id: string,
  priority: number, 
  Sidebar: ReactComponent,   
  Trigger: ReactComponent, 
  isAvailable: (context) => boolean,
  enabled: boolean,
  Provider?: ReactComponent, 
}
```

### Rename: notifications → upgrade
All internal references to the old "Notifications" panel are renamed for clarity:

1. WIDGETS.NOTIFICATIONS → WIDGETS.UPGRADE
2. NotificationTray → UpgradePanel
3. NotificationTrigger → UpgradeTrigger
4. NotificationIcon → UpgradeIcon
5. notificationStatus context → upgradeWidgetStatus
6. LocalStorage keys updated accordingly

### Extracted Upgrade Widget Package
The upgrade panel is extracted from the sidebar core into `upgrade` widget. It is no longer bundled by default. Instances that need it register it via env.config.jsx.

### Configuration
```
import { upgradeWidgetConfig } from './src/widgets/upgrade/src/index';

const config = {
  SIDEBAR_WIDGETS: [upgradeWidgetConfig],
};

export default config;
```
When the upgrade widget is promoted to its own npm package @edx/learning-upgrade-widget, operators will install it and update the import path. No other changes needed.

### Two-Sidebar Collaboration
The architecture docs in ARCHITECTURE.md fully describe the shared currentSidebar state between the LEFT (Course Outline) and RIGHT (widget panels) sidebars, including:

- Priority cascade: DISCUSSIONS (10) → UPGRADE (20) → external widgets → COURSE_OUTLINE (fallback)
- Auto-open logic on initial load
- Unit shift behavior for each scenario
- State persistence via localStorage / sessionStorage

### Removed
- src/courseware/course/new-sidebar -  Entire unused new-sidebar directory
- src/courseware/course/sidebar/sidebars/notifications - Built-in notification widget
- NotificationTraySlot
- NotificationWidgetSlot
- NotificationsDiscussionsSidebarSlot - Replaced by RightSidebarSlot
- NotificationsDiscussionsTriggerSlot  - Replaced by RightSidebarTriggerSlot

### Testing
- 12 new test suites covering hooks, defaultWidgets, storage utils, SidebarContextProvider, SidebarTriggers, SidebarBase, UpgradePanel, UpgradeTrigger, UpgradeIcon, UpgradeWidgetContext, and upgrade utils
- All use cases are documented in USE_CASE_VERIFICATION.md

### Migration Notes
- No breaking changes for instances that don't use the upgrade widget. Discussions and course outline work are identical.
- Instances using the upgrade panel: Add `SIDEBAR_WIDGETS: [upgradeWidgetConfig]` to env.config.jsx.
- Plugin slot consumers: 
         - Replace NotificationTraySlot / NotificationsDiscussionsSidebarSlot with RightSidebarSlot. Replace
         - Replace NotificationsDiscussionsTriggerSlot with RightSidebarTriggerSlot.
- localStorage: Old notificationStatus.{courseId} keys are not migrated to new key format upgradeWidget.{courseId} starts fresh.


awais-ansari:
@brian-smith-tcril Please review this PR.

src/external-widgets/learning-upgrade-widget/package.json
```diff
@@ -0,0 +1,35 @@
+{
```
arbrandes: Two things:

1. Let's just call this directory `widgets`, instead of `external-widgets`.  It's what [frontend-base apps are doing](https://github.com/openedx/frontend-app-learner-dashboard/tree/frontend-base/src/widgets).
2. I suggest avoiding having this as a separate package while in the repo, if for no other reason than that our Github repos not being properly tooled to be monorepos.  In other words, let's drop the package.json.

jest.config.js
```diff
@@ -13,6 +13,7 @@ const config = createConfig('jest', {
     // See https://stackoverflow.com/questions/72382316/jest-encountered-an-unexpected-token-react-markdown
     'react-markdown': '<rootDir>/node_modules/react-markdown/react-markdown.min.js',
     '@src/(.*)': '<rootDir>/src/$1',
+    '@edx/learning-mfe-widget': '<rootDir>/src/sidebar-widget-sdk.js',
```
arbrandes: The added boilerplate is another reason not to attempt a monorepo structure.

src/courseware/course/sidebar/constants.js
```diff
@@ -0,0 +1,15 @@
+export const WIDGET_PRIORITIES = {
+  UPGRADE: 20,
+  DISCUSSIONS: 10,
```
arbrandes: These priorities would ideally be defaults that the widgets themselves set.

Another way to read it is: if the sidebar is going to be truly generic, it should not know anything in advance about the Upgrade widget (or discussions, for that matter).

src/courseware/course/sidebar/defaultWidgets.js
```diff
@@ -0,0 +1,71 @@
...
+ * Check if discussions widget is available for the current unit
+ * @param {Object} context - Contains courseId, unitId, topic, courseHomeMeta
+ * @returns {boolean} Whether discussions are available
+ */
+export const discussionsIsAvailable = ({ topic }) => !!(topic?.id && topic?.enabledInContext);
+
+/**
+ * Default built-in widgets for the RIGHT sidebar.
+ * The upgrade widget is no longer built-in — configure it via SIDEBAR_WIDGETS
+ * in env.config.jsx using the @edx/learning-upgrade-widget package.
```
arbrandes: Nit: let's not even mention other widgets outside their own scope in the codebase.

src/courseware/course/sidebar/defaultWidgets.js
```diff
@@ -0,0 +1,71 @@
+import { getConfig } from '@edx/frontend-platform';
+import * as discussions from './sidebars/discussions';
+import { WIDGET_PRIORITIES, WIDGET_CONFIG } from './constants';
+
+/**
+ * Check if discussions widget is available for the current unit
+ * @param {Object} context - Contains courseId, unitId, topic, courseHomeMeta
+ * @returns {boolean} Whether discussions are available
+ */
+export const discussionsIsAvailable = ({ topic }) => !!(topic?.id && topic?.enabledInContext);
```
arbrandes: Is there a way to have this widget-specific code moved to its own scope (i.e., its own directory) in the codebase?

src/courseware/course/sidebar/README.md
```diff
@@ -0,0 +1,319 @@
+# Sidebar Widget Framework
+
+## Overview
+
+The sidebar uses a pluggable widget framework that allows instances to customise which panels appear in the course sidebar. All widgets are external — none are bundled into the core Learning MFE by default (except Discussions).
```
arbrandes: I believe the Discussions widget should be moved as part of this PR to be at the same level (conceptual and practical) as Upgrade.  I.e., under a `src/widgets/` directory.

src/courseware/course/sidebar/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+- **Built-in widget**: Discussions (always enabled)
+- **External widgets**: Installed via npm packages and configured in `env.config.jsx`
+  - Upgrade/upsell panel: `@edx/learning-upsell-widgets`
```
arbrandes: Let's move widget-specific documentation into its own scope/directory.

src/courseware/course/sidebar/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+### The `Provider` field
+
+When a widget's Panel and Trigger components need to share React state (e.g. badge visibility),
+use the optional `Provider` field instead of putting that state in `SidebarContext`.
+
+The `SidebarContextProvider` automatically wraps all children in each registered widget's
+`Provider` (in priority order), so both Panel and Trigger have access to the same widget-level
+context. The widget Provider itself can safely read `courseId` etc from `SidebarContext` since
+it mounts inside it.
```
arbrandes: [Non-blocking, just a comment.]

It's great that you arrived at this conclusion, because so did frontend-base: there's [an app-wide mechanism](https://github.com/openedx/frontend-base/blob/5387ffce835cb94238ce2f20b21c94a1cda6f0d1/types.ts#L30) to do just this.

However, this means that when this is converted to frontend-base, the API for the registry might have to change to accomodate it.

src/courseware/course/sidebar/README.md
```diff
@@ -0,0 +1,319 @@
...
+### Context Object
+
+The `isAvailable` function receives a context object with:
+
+```javascript
+{
+  courseId: string,
+  unitId: string,
+  topic: object,        // Discussion topic data from model store
+  verifiedMode: object, // Verification/upgrade data from model store
```
arbrandes: To me, these two seem brittle to have as part of the API contract of `isAvailable()`.  `topic` is not likely to be needed by any widget except Discussion, and `verifiedMode` only needed by `Upgrade` - my gut says we should attempt to be more agnostic, otherwise we'll be on the hook to maintain these two even if the corresponding widgets fall into disuse (thinking of Upgrade in particular).

On the other hand, this is server state, so the first alternative that comes to mind is to let widgets reach into it... which is not ideal, either, because then widget authors would be surprised if we ever changed the representation of server state.

I don't have a solution to offer from the top of my head, so consider this a conversation starter.

(That said, how about offering hydrated `course` and `unit` objects to the widgets, from which they can pick whatever they want - including `verifiedMode` and `topic`?)

PS: I later saw that this PR already implements reaching into server state.  So there's really no reason to _not_ be agnostic, here.

src/courseware/course/sidebar/README.md
```diff
@@ -0,0 +1,319 @@
...
+## Accessing Course Data
+
+Widgets can access course data via the model store:
+
+```javascript
+import { useModel } from '@src/generic/model-store';
+
+// In your component
+const courseData = useModel('coursewareMeta', courseId);
+const { verifiedMode, enrollmentMode } = courseData;
```
arbrandes, line 165: Ah, so there is _already_ a way for widgets to reach into the model.  Note my other comment on why this might be necessary for widget-agnosticism.

Note from the future: the current implementation of the model will soon be converted from Redux to React Query.  Not sure yet if the API will change, but instinct tells me we might be able to keep `useModel()`.

src/courseware/course/sidebar/ARCHITECTURE.md
```diff
@@ -0,0 +1,219 @@
+# Sidebar Architecture - Two Sidebar Collaboration System
+
+## Overview
+
+The Learning MFE uses a **two-sidebar system** where a left sidebar (Course Outline) and right sidebar (Discussions / External Widgets) work collaboratively to provide contextual panels to learners.
```
arbrandes, line 5: [Question] Is the collaboration a core requirement?  Or is it just needed for a particular widget (looking at Upgrade, here).

Why do I ask?  If this is only needed because of a widget intended to be extracted from the codebase, we should only provide the framework, upstream.  The implementation details of the collaboration should live in the widget itself.

src/courseware/course/sidebar/ARCHITECTURE.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Architecture Components
+
+### LEFT SIDEBAR (Course Outline)
+- **Location**: Left side of screen, adjacent to course content
+- **Component**: `CourseOutlineTray` (rendered via `CourseOutlineSidebarSlot`)
+- **Purpose**: Navigation - displays course structure, sequences, and units
+- **State Management**: Uses `useCourseOutlineSidebar()` hook
+- **ID**: `WIDGETS.COURSE_OUTLINE`
+- **Rendering**: Only renders when `currentSidebar === 'COURSE_OUTLINE'`
```
arbrandes, line 15: Is this a product requirement?  Left and right sidebars can _never_ be rendered simultaneously?  This would partly answer my question about collaboration, above.

src/courseware/course/sidebar/ARCHITECTURE.md
```diff
@@ -0,0 +1,219 @@
...
+currentSidebar: 'DISCUSSIONS' | 'COURSE_OUTLINE' | null | string
+```
+
+### State Machine
+
+| `currentSidebar` Value | LEFT Sidebar (Course Outline) | RIGHT Sidebar (Discussions/Upgrade) |
+|------------------------|-------------------------------|-------------------------------------------|
+| `null` | Hidden (returns null) | Hidden (returns null) |
+| `'DISCUSSIONS'` | Hidden | Shows Discussions panel |
+| `'UPGRADE'` | Hidden | Shows Upgrade panel |
```
arbrandes: As mentioned in this review's top comment, I haven't delved into the code much, yet, but the state machine logic should _definitely_ be agnostic as to what widgets are in there.  So much so, that I think you should make it clear in this document that these are just examples that are not hard-coded anywhere.

src/courseware/course/sidebar/ARCHITECTURE.md
```diff
@@ -0,0 +1,219 @@
...
+- `initialSidebar !== null` → RIGHT sidebar has available panels
+- `initialSidebar === null` → NO right sidebar panels available → Course Outline should open as fallback
+
+### Auto-Open Logic
+
+**RIGHT Sidebar (Desktop):**
+```javascript
+// In SidebarContextProvider
+if (shouldDisplaySidebarOpen && !shouldDisplayFullScreen) {
+  // Auto-open first available: DISCUSSIONS → UPGRADE → External Widgets
```
arbrandes: As mentioned elsewhere, we should drop the distinction between built-in and external widgets.  They should all just be widgets.

I might even drop the mention of upgrade, here, since it'll be eventually extracted from the codebase, and instead mention that this decision will be made based on the widget's self-reported priority.

src/courseware/course/sidebar/ARCHITECTURE.md
```diff
@@ -0,0 +1,219 @@
...
+  }
+}, [initialSidebar, unitId]);
+```
+
+### Priority Cascade (Desktop)
+
+1. **DISCUSSIONS** (Priority 10) - Opens if unit has discussions
+2. **UPGRADE** (Priority 20) - Opens if no discussions but verified mode exists
+3. **External Widgets** (Priority 30+) - Opens if configured and available
+4. **COURSE_OUTLINE** (Fallback) - Opens if NO right sidebar panels available
```
arbrandes: Again, make clear which ones are examples, and which ones are hard-coded defaults.  And let's drop the "external" widget distinction.

src/courseware/course/sidebar/ARCHITECTURE.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Use Case Implementation
+
+### Initial Load (Desktop > 1200px)
+
+| Available Panels | `initialSidebar` | `currentSidebar` | Result |
```
arbrandes, line 93:
```suggestion
| Available Widgets | `initialSidebar` | `currentSidebar` | Result |
```

src/courseware/course/sidebar/ARCHITECTURE.md
```diff
@@ -0,0 +1,219 @@
...
+|-----------------|------------------|------------------|---------|
+| Discussions available | `'DISCUSSIONS'` | `'DISCUSSIONS'` | Discussions opens |
+| Only Upgrade | `'UPGRADE'` | `'UPGRADE'` | Upgrade opens |
+| Neither available | `null` | `'COURSE_OUTLINE'` | Course Outline opens (fallback) |
```
arbrandes: Let's make this more generic.  For example, instead of Discussions or Upgrade, refer to generic widgets and priorities, and how they interact.  Alternatively, keep Discussions and Upgrade, but specify their default priorities, instead of assuming the reader knows.

src/courseware/course/sidebar/ARCHITECTURE.md
```diff
@@ -0,0 +1,219 @@
...
+- **Behavior**: NO auto-open, respects `localStorage`
+- **localStorage Key**: `sidebar.${courseId}`
+- **User Action**: Must manually tap trigger buttons to open panels
+
+### Unit Shift Behavior (Desktop)
+
+**Scenario 1: Previous unit had DISCUSSIONS open**
+- New unit has discussions → **Stays on Discussions**
+- New unit NO discussions, YES upgrade → **Switches to Upgrade**
+- New unit NO discussions, NO upgrade → **Closes right sidebar, Course Outline auto-opens**
```
arbrandes: Is this logic hard-coded upstream, or is it emergent behavior based on the widget's self-reported priorities and other properties?

My point, also made in other comments, is that we should never have a hard-coded `if (UPGRADE)` outside the scope of the Upgrade widget itself.

src/courseware/course/sidebar/ARCHITECTURE.md
```diff
@@ -0,0 +1,219 @@
...
+- **State Management**: Uses `useCourseOutlineSidebar()` hook
+- **ID**: `WIDGETS.COURSE_OUTLINE`
+- **Rendering**: Only renders when `currentSidebar === 'COURSE_OUTLINE'`
+- **Trigger**: `CourseOutlineTrigger` (separate location in mobile/desktop toolbar)
+
+### RIGHT SIDEBAR (Contextual Panels)
+- **Location**: Right side of screen
+- **Component**: `Sidebar` (rendered via `RightSidebarSlot`)
+- **Purpose**: Contextual tools and information panels
+- **State Management**: `SidebarContextProvider` + widget registry system
```
arbrandes, line 22: I very much like the registry idea, but my first question when seeing it was: why not offer the same registry for the left sidebar?

Then I remembered that this is exactly one of the things we tried to address with frontend-base.  I suspect that during the conversion the registry will be made obsolete - or just much thinner - due to the frontend-base provider hoisting and slot layout systems.

In other words: never mind, for now.  Just be aware that this is one of the configuration APIs that will change after the conversion.

docs/decisions/0010-upgrade-widget-extraction.md
```diff
@@ -0,0 +1,36 @@
...
+- `notificationStatus` context property → `upgradeWidgetStatus`
+- `onNotificationSeen` → `onUpgradeWidgetSeen`
+- `NotificationTray` component → `UpgradePanel`
+- `NotificationTrigger` component → `UpgradeTrigger`
+- `NotificationIcon` component → `UpgradeIcon`
+- localStorage keys: `notificationStatus.${courseId}` → `upgradeWidget.${courseId}`
+- i18n message keys (`notification.*`)
+
+
+### 3. External package `@edx/learning-upgrade-widget` is created
```
arbrandes: As mentioned elsewhere, let's just make this a subdirectory of `widgets/`: no need to have a separate package (with all the monorepo headaches it brings) until it is in a different repository.

docs/decisions/0010-upgrade-widget-extraction.md
```diff
@@ -0,0 +1,36 @@
...
+
+A standalone npm package is created with the same functionality. Instances that need the upgrade panel:
+
+1. Install `@edx/learning-upgrade-widget`
+2. Set `ENABLE_LEGACY_UPGRADE_WIDGET: false`
+3. Add the package widget to `SIDEBAR_WIDGETS` in `env.config.jsx`
+
+The package skeleton lives in `src/external-widgets/learning-upsell-widget/` until promoted to its own repository.
```
arbrandes: If my suggestion from another comment is accepted, we should add:

`### 3. Move the Discussions widget code into its own directory`

It would be a sibling to the Upgrade widget.

arbrandes (changes requested):
First off: this is amazingly well documented.  Congrats, and thank you!  I'm probably going to start using this PR as a benchmark for how complicated frontend architectural decisions should be implemented.

Second: this is mostly an architecture review based on the ADR, `ARCHITECTURE.md`,and a glance at `USE_CASE_VERIFICATION.md`.  I have not looked closely at the code, yet.

Lastly, I'm coming at this "from the future".  Prior to the Willow cutoff in October of this year, frontend-app-learning will be converted to frontend-base, and this is one of the situations where the new slot architecture will help (or, at least, affect).  Some of the inline comments are just FYIs related to that, and a couple are meant to help with the transition later.  Either way, we should keep this in mind.

src/external-widgets/learning-upgrade-widget/package.json
brian-smith-tcril: Re: avoiding separate package

This decoupling reminds me of what we did with https://github.com/openedx/frontend-app-learning/tree/master/src/plugin-slots/SequenceNavigationSlot and https://github.com/openedx/frontend-app-learning/tree/master/src/plugin-slots/CourseBreadcrumbsSlot. We kept the old components around, but instead of directly importing/using them we provided example FPF configs showing how to import and use them.

src/courseware/course/sidebar/defaultWidgets.js
awais-ansari: Yes, there is a way. I'll move it.

src/courseware/course/sidebar/ARCHITECTURE.md
awais-ansari: The left and right sidebar fallback behavior is a core framework concern. It applies uniformly to all right-panel widgets and is not specific to the Upgrade widget. Removing Upgrade would not change the collaboration logic in any way.

The framework only exposes a generic hook point. The full implementation (UpgradeWidgetProvider, badge state, and localStorage) resides in src/widgets/learning-upgrade-widget/. If Upgrade is ever extracted to its own repository, that provider will move with it. No framework changes are required.

src/courseware/course/sidebar/ARCHITECTURE.md
awais-ansari: No product requirement. This is existing behaviour. The MFE has always had a single active panel at a time. I made that behaviour explicit and consistent rather than implicit and accidental.

The single [currentSidebar](https://github.com/openedx/frontend-app-learning/blob/master/src/courseware/course/sidebar/SidebarContextProvider.jsx#L34) state variable is an implementation choice that makes the behavior consistent and predictable.

src/courseware/course/sidebar/ARCHITECTURE.md
awais-ansari: This is logic hard-coded upstream.
```
// SidebarContextProvider.jsx on master
initialSidebar = isUnitHasDiscussionTopics
  ? SIDEBARS[discussionsSidebar.ID].ID
  : verifiedMode && SIDEBARS[notificationsSidebar.ID].ID;
```

Now it is emergent behavior based on the widget's self-reported priorities. There is no if (UPGRADE) hard-coded anywhere in the registry framework logic.

arbrandes (changes requested):
Thanks for addressing my previous concerns (rename to `widgets/`, drop separate package, etc.). I tested this and it works - at least for the Dicussions widget and the outline.  Nice!

Looking at the code as it stands now (with some help from Claude), this is what I found:

---

### 1. Edge-case bug: Mismatch between what sidebar is "rendered" vs what is "open" for layout purposes

[`Sidebar.jsx:8`](https://github.com/openedx/frontend-app-learning/blob/158b2fbc/src/courseware/course/sidebar/Sidebar.jsx#L8) renders any widget in the `SIDEBARS` registry (all enabled widgets), and [`SidebarTriggers.jsx:23`](https://github.com/openedx/frontend-app-learning/blob/158b2fbc/src/courseware/course/sidebar/SidebarTriggers.jsx#L23) renders triggers for all of them too. But [`useIsSidebarOpen()`](https://github.com/openedx/frontend-app-learning/blob/158b2fbc/src/courseware/course/sequence/sequence-navigation/hooks.js#L100-L102) and [`upgradePanelVisible`](https://github.com/openedx/frontend-app-learning/blob/158b2fbc/src/courseware/course/sequence/lock-paywall/LockPaywall.jsx#L40) check `availableSidebarIds`, which only includes widgets whose `isAvailable()` returns true.

So if a widget's trigger doesn't self-hide (like `UpgradeTrigger`), a user can open a panel that's physically rendered but that layout code considers "closed". Example: Upgrade is registered but the course has no verified track - the trigger shows, clicking it opens the panel, but sequence navigation tabs don't shrink to accommodate it.

Fix: either filter `SidebarTriggers` by `availableSidebarIds`, or have `useIsSidebarOpen()` check `SIDEBARS` instead. The first is simpler.

---

### 2. Warning: The old `notificationTrayVisible` was dead code - this PR activates that layout logic

On master, [`LockPaywall`](https://github.com/openedx/frontend-app-learning/blob/a0866142/src/courseware/course/sequence/lock-paywall/LockPaywall.jsx#L28) destructures `notificationTrayVisible` from `SidebarContext`, but the old provider never sets it - it was always `undefined`. The layout variables that depended on it (`shouldDisplayGatedContentOneColumn`, etc.) were dead code.

The [new code](https://github.com/openedx/frontend-app-learning/blob/158b2fbc/src/courseware/course/sequence/lock-paywall/LockPaywall.jsx#L40) replaces it with `upgradePanelVisible = availableSidebarIds.includes(currentSidebar)`, which CAN be `true`, activating layout adjustments that previously never triggered. The variable name is also misleading: it's `true` for any open sidebar widget (including Discussions), not specifically the upgrade panel.

I have not tested this: have you?  (Is there even a way for us to test it?)

---

### 3. Nit: No-op state call in `useUnitShiftBehavior.js`

[`useUnitShiftBehavior.js:122-124`](https://github.com/openedx/frontend-app-learning/blob/158b2fbc/src/courseware/course/sidebar/hooks/useUnitShiftBehavior.js#L122-L124):

```js
} else if (currentWidget) {
  // Current RIGHT sidebar panel still available and no higher priority - keep it
  setCurrentSidebar(currentSidebar);  // no-op
```

`setCurrentSidebar(currentSidebar)` with the same value is a no-op (React bails out). The adjacent branches call `setSidebarId` too, making this one look like a missing call. Removing the line or adding a comment would clarify intent.

---

### 4. Nit: Dead guard in `useSidebarSync.js`

[`useSidebarSync.js:39`](https://github.com/openedx/frontend-app-learning/blob/158b2fbc/src/courseware/course/sidebar/hooks/useSidebarSync.js#L39):

```js
if (shouldDisplayFullScreen || previousUnitIdRef.current === null) {
  return;
}
```

React runs effects in hook declaration order. Since [`useUnitShiftBehavior` is called before `useSidebarSync`](https://github.com/openedx/frontend-app-learning/blob/158b2fbc/src/courseware/course/sidebar/SidebarContextProvider.jsx#L88-L114), the ref is always set to `unitId` before this check runs. The `=== null` branch is unreachable.

---

### 5. Nit: misleading variable name in `SequenceNavigation.jsx`

[`shouldDisplayUpgradeTriggerInSequence`](https://github.com/openedx/frontend-app-learning/blob/158b2fbc/src/courseware/course/sequence/sequence-navigation/SequenceNavigation.jsx#L44) is just a small-screen breakpoint check unrelated to the upgrade widget. Consider a generic name like `isSmallScreen`.

---

### 6. Codebase layout: `course-outline` doesn't belong in `src/widgets/`

[`src/widgets/course-outline/`](https://github.com/openedx/frontend-app-learning/blob/158b2fbc/src/widgets/course-outline) is tightly coupled to the sidebar framework via `WIDGETS.COURSE_OUTLINE` constants, dedicated plugin slots, and special-case logic in all three sidebar hooks. It's not pluggable - removing the directory would break the core layout. Placing it in `src/widgets/` alongside truly decoupled widgets like Upgrade implies a level of decoupling that doesn't exist.

(It's fine to leave Discussions in there, as it is a properly independent widget - just a default one.)

arbrandes (approved):
Alright, I think this is good enough to go.  Thanks for bearing with me!

pdpinch:
This change broke our build of the MFE. Since this involved changing the name/api for a pluginslot, it should have gone through the DEPR process and gotten some more communication. 

In particular, what is the plan for addressing this in the Verawood release?

arbrandes:
@pdpinch, you're right, we did not consider backward-compatibility at all, here.  I'm going to bring this up with the Axim engineering team, today.

arbrandes:
Revert PR: https://github.com/openedx/frontend-app-learning/pull/1898.

arbrandes:
@pdpinch, instead of reverting it, we added backward-compatibility:

https://github.com/openedx/frontend-app-learning/pull/1899

PR #1897, opened 2026-04-13
feat: move discussion topic prefetch from trigger to widget config lifecycle

awais-ansari:
## Summary

Moves data-fetching logic (`getCourseDiscussionTopics` dispatch) out of `DiscussionsTrigger` and into a new `prefetch` field on the widget config. `SidebarContextProvider` now runs `widget.prefetch()` for all enabled widgets on mount and when course metadata changes, ensuring Redux store data is populated before `isAvailable` checks and component rendering.

## Motivation

Previously, `DiscussionsTrigger` was responsible for both rendering and fetching discussion topics. This created a coupling between the trigger component and the data lifecycle. The trigger had to mount before data was available, causing timing issues with widget availability checks. Moving prefetch to the framework level ensures data is ready before any widget logic runs.

### Code (5 files)
- **`SidebarContextProvider.jsx`**: Added `useDispatch` + `useEffect` that iterates `enabledWidgets` and calls `widget.prefetch({ courseId, course, dispatch })` when course metadata changes.
- **`DiscussionsTrigger.jsx`**: Removed `useDispatch`, `useEffect`, `getCourseDiscussionTopics`, `courseHomeMeta` tabs lookup, and `getConfig().DISCUSSIONS_MFE_BASE_URL` check. Now a pure render component that only reads the `discussionTopics` model.
- **`widgetConfig.js`**: Added `discussionsPrefetch` function and `prefetch` field to the discussions widget config.
- **`SidebarContextProvider.test.jsx`**: Added `react-redux` mock, enriched `useModel` mock for `courseHomeMeta`, updated UC7 toggle tests to reflect mobile behavior (no auto-open).
- **`DiscussionsTrigger.test.jsx`**: Added `executeThunk(getCourseDiscussionTopics(...))` in `beforeEach` to pre-populate the store.

### Documentation (4 files)
- **`ARCHITECTURE.md`**: Documented the prefetch lifecycle and updated provider responsibilities.
- **`README.md`**: Added `prefetch` to the widget structure definition and usage guide.
- **`USE_CASE_VERIFICATION.md`**: Added Prefetch Effect (effect #0) to the architecture diagram.
- **`discussions/README.md`**: Added `discussionsPrefetch` export and data prefetch documentation.

## Widget Config API Addition

The `prefetch` field is a new **optional** field on the widget config:

```js
{
  id: 'MY_WIDGET',
  prefetch: ({ courseId, course, dispatch }) => {
    dispatch(fetchMyData(courseId));
  },
  // ... other fields
}

### DIFF
--- a/src/courseware/course/sidebar/ARCHITECTURE.md
+++ b/src/courseware/course/sidebar/ARCHITECTURE.md
@@ -159,27 +159,32 @@ _Example with built-in widgets:_
 - Manage `currentSidebar` state (shared by both sidebars)
 - Handle unit shift logic for RIGHT sidebar panels
 - Provide context to both left and right sidebar components
-- **Prefetch widget data** before availability checks run (via `widget.prefetch`)
+- **Prefetch widget data** after mount via `widget.prefetch` (sync logic re-evaluates availability once data arrives)
 
 ### Widget Prefetch Lifecycle
 
-Widgets can define a `prefetch` function in their config to pre-load data that their `isAvailable` or render logic depends on. The framework calls `prefetch` for every enabled widget on mount and when course metadata changes:
+Widgets can define a `prefetch` function in their config to pre-load data that their `isAvailable` or render logic depends on. The framework calls `prefetch` for every enabled widget on mount (and when `courseId` or the widget list changes):
 
 ```javascript
 // In SidebarContextProvider.jsx
+const courseMetaRef = useRef(null);
+courseMetaRef.current = { ...coursewareMeta, ...courseHomeMeta };
+
 useEffect(() => {
   enabledWidgets.forEach(widget => {
     if (widget.prefetch) {
-      widget.prefetch({ courseId, course: { ...coursewareMeta, ...courseHomeMeta }, dispatch });
+      widget.prefetch({ courseId, course: courseMetaRef.current, dispatch });
     }
   });
-}, [enabledWidgets, courseId, coursewareMeta, courseHomeMeta, dispatch]);
+}, [enabledWidgets, courseId, dispatch]);
 ```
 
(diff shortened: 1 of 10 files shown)


arbrandes (changes requested):
At first glance this makes a lot of sense: centralizing data prefetch in the provider so that `isAvailable` checks can rely on pre-populated Redux state.  But there's an issue:

The prefetch effect at [`SidebarContextProvider.jsx#L89-L96`](https://github.com/openedx/frontend-app-learning/blob/f62381fe97b5cb7c562ff36c710682192f49b689/src/courseware/course/sidebar/SidebarContextProvider.jsx#L89-L96) includes `coursewareMeta` and `courseHomeMeta` in its dependency array.  The two models settle in separate React batches during initial load, which causes the effect to fire twice after the guard in `discussionsPrefetch` starts passing.

I confirmed this produces duplicate requests to both `GET /api/discussion/v1/courses/{courseId}` and `GET /api/discussion/v2/course_topics/{courseId}` on every page load - two requests to each endpoint returning identical data ([`thunks.js#L268-L286`](https://github.com/openedx/frontend-app-learning/blob/f62381fe97b5cb7c562ff36c710682192f49b689/src/courseware/data/thunks.js#L268-L286)).

The old code in `DiscussionsTrigger` avoided this because its effect depended on `[courseId, baseUrl, edxProvider, dispatch]`, all of which were stable after the initial render.

A simple fix might be to remove `coursewareMeta` and `courseHomeMeta` from the dependency array. The effect only needs to run once per `courseId` - the `course` object is only used inside `discussionsPrefetch` to check `tabs`, and that check can happen with the latest value via a ref or by passing the models lazily.

arbrandes (approved):
Ok, the `ref` workaround solves it, but it's a bit hacky.   The real solution is going to be moving to React Query, later on.

Let's worry about that later, though!  👍🏼

PR #1909, opened 2026-05-04
feat: Created the plugin slot for UpgradeToCompleteAlert

mshet-sonata-pixel:
Creating a PluginSlot **UpgradeToCompleteAlert** at the location of the - **https://github.com/openedx/frontend-app-learning/blob/master/src/course-home/suggested-schedule-messaging/UpgradeToCompleteAlert.jsx** , The PluginSlot should render nothing by default. We will be injecting the Alert as a plugin instead. 

This will need to be reviewed by 2U owning teams as well as the open source community. It should get support from open source.

The below is the Screeshot-
image


michaelroytman (approved):

MaxFrank13:
@michaelroytman do you have permissions to run the workflows? I do not. 

Thanks for the review!

src/plugin-slots/BannerDatesUpgradeSlot/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { PluginSlot } from '@openedx/frontend-plugin-framework';
+import UpgradeToCompleteAlert from '../../course-home/suggested-schedule-messaging/UpgradeToCompleteAlert';
+
+export const BannerDatesUpgradeSlot = ({
+  courseId,
+  logUpgradeLinkClick,
+}: BannerDatesUpgradeSlotProps) => (
+  <PluginSlot
+    id="org.openedx.frontend.learning.banner_dates_upgrade.v1"
+    idAliases={['dates_upgrade_banner_slot']}
```
arbrandes: Aliases exist to support a deprecation window for slots that are being renamed.  Since this is a new slot, it shouldn't have an alias.

src/plugin-slots/BannerDatesUpgradeSlot/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+```js
+import {
+  DIRECT_PLUGIN,
+  PLUGIN_OPERATIONS,
+} from '@openedx/frontend-plugin-framework';
+
+const config = {
+  pluginSlots: {
+    dates_upgrade_banner_slot: {
```
arbrandes: Let's use the full id, here.

src/plugin-slots/BannerDatesUpgradeSlot/README.md
```diff
@@ -0,0 +1,63 @@
...
+      keepDefault: false,
+      plugins: [
+        {
+          op: PLUGIN_OPERATIONS.Insert,
+          widget: {
+            id: 'org.openedx.frontend.learning.banner_dates_upgrade.v1',
```
arbrandes: This is the widget id: it's arbitrary, and ideally it'll be different from the slot id where it's used.

arbrandes (changes requested):
Thanks for this!  Took a first pass, asked for a couple of changes.

src/plugin-slots/BannerDatesUpgradeSlot/index.tsx
mshet-sonata-pixel: Fixed @arbrandes Thanks!

src/plugin-slots/BannerDatesUpgradeSlot/README.md
mshet-sonata-pixel: Fixed Thanks!

src/plugin-slots/BannerDatesUpgradeSlot/README.md
mshet-sonata-pixel: Fixed Thanks!

arbrandes (changes requested):
Thank you for addressing my comments.

Now we have to fix the coverage regression.  There are two ways to do that: either you remove the default content in the slot (not just from the slot, but from the codebase), or you restore some of the removed tests (such as the deleted analytics test (DatesTab.test.jsx:257-281).

arbrandes (approved):

PR #1917, opened 2026-05-27
feat: Introduce slot for customizing course tab navigation

xitij2000:
Adds a new plugin slot (`CourseTabsNavigationSlot`) to enable customization, modification, or hiding of the course tab navigation.


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

src/tab-page/LoadedTabPage.tsx
```diff
@@ -1,27 +1,35 @@
...
+import useEnrollmentAlert from '@src/alerts/enrollment-alert';
+import useLogistrationAlert from '@src/alerts/logistration-alert';
+import StreakModal from '@src/shared/streak-celebration';
+import InstructorToolbar from '@src/instructor-toolbar';
 
 import ProductTours from '../product-tours/ProductTours';
 
+interface LoadedTabPageProps {
+  activeTabSlug: string;
+  children: React.ReactNode;
```
brian-smith-tcril:
```suggestion
  children?: React.ReactNode;
```

functionally equivalent to not having the `?` but this better communicates it as optional (matching what we had with `PropTypes`

src/tab-page/LoadedTabPage.tsx
```diff
@@ -1,27 +1,35 @@
...
+  courseId: string;
+  metadataModel: string;
```
brian-smith-tcril, line 22:
```suggestion
  metadataModel?: string;
```

The `PropTypes` version of this was optional

src/tab-page/LoadedTabPage.tsx
```diff
@@ -1,27 +1,35 @@
...
+  unitId: string | null;
```
brian-smith-tcril:
```suggestion
  unitId?: string | null;
```

To match the `PropTypes` logic we:
* `?` so it's optional
* `string` for the actual string type
* `| null` because `?` just supports `undefined`, not `null`

We likely don't need to match the `PropTypes` logic exactly here, and ideally we could just use `unitId?: string` and default to `undefined`, but that'd require some deeper digging (and should probably be a follow-up issue).

From a sub-components perspective, we can just check where `unitId` is used to see if defaulting to `undefined` would be a reasonable option. It seems to only be in `InstructorToolbar`, and it like `undefined` would be fine there:
https://github.com/openedx/frontend-app-learning/blob/db18c2cb456fcf887b30c9122e153c163f867886/src/instructor-toolbar/InstructorToolbar.jsx#L119
https://github.com/openedx/frontend-app-learning/blob/db18c2cb456fcf887b30c9122e153c163f867886/src/instructor-toolbar/InstructorToolbar.jsx#L126
https://github.com/openedx/frontend-app-learning/blob/db18c2cb456fcf887b30c9122e153c163f867886/src/instructor-toolbar/InstructorToolbar.jsx#L65
https://github.com/openedx/frontend-app-learning/blob/db18c2cb456fcf887b30c9122e153c163f867886/src/instructor-toolbar/InstructorToolbar.jsx#L32

The consumers of `LoadedTabPage` are where it gets messy. [Org search](https://github.com/search?q=org%3Aopenedx%20LoadedTabPage&type=code) shows it only being used in `frontend-app-learning` which limits the blast radius a bit, but there's still a lot of `null` being used. In `TabPage` for example:

https://github.com/openedx/frontend-app-learning/blob/db18c2cb456fcf887b30c9122e153c163f867886/src/tab-page/TabPage.jsx#L104
https://github.com/openedx/frontend-app-learning/blob/db18c2cb456fcf887b30c9122e153c163f867886/src/tab-page/TabPage.jsx#L96
https://github.com/openedx/frontend-app-learning/blob/db18c2cb456fcf887b30c9122e153c163f867886/src/tab-page/TabPage.jsx#L80

and then we'd need to dig into the consumers of `TabPage` and so on.

src/course-tabs/CourseTabsNavigation.tsx
```diff
@@ -20,14 +18,13 @@ interface CourseTabsNavigationProps {
...
   activeTabSlug = undefined,
-  className = null,
   tabs,
 }:CourseTabsNavigationProps) => {
   const intl = useIntl();
   const { show } = useCoursewareSearchState();
 
   return (
-    <div id="courseTabsNavigation" className={classNames('course-tabs-navigation', className)}>
+    <div id="courseTabsNavigation" className="mb-3 course-tabs-navigation">
```
brian-smith-tcril, line 27: At first I was slightly concerned about this moving in here instead of coming from the consumer in `LoadedTabPage`, but it seems the only other place that references this is

https://github.com/openedx/frontend-app-learning/blob/db18c2cb456fcf887b30c9122e153c163f867886/src/course-home/courseware-search/CoursewareSearch.jsx#L37

which is only used by the next line

https://github.com/openedx/frontend-app-learning/blob/db18c2cb456fcf887b30c9122e153c163f867886/src/course-home/courseware-search/CoursewareSearch.jsx#L38

which only cares about the `top` which isn't changed by `mb`.

src/course-tabs/CourseTabsNavigation.tsx
```diff
@@ -1,16 +1,14 @@
...
+import { CoursewareSearch, CoursewareSearchToggle } from '@src/course-home/courseware-search';
+import { useCoursewareSearchState } from '@src/course-home/courseware-search/hooks';
+import { CourseTabLinksSlot } from '@src/plugin-slots/CourseTabLinksSlot';
+import Tabs from '@src/generic/tabs/Tabs';
 import messages from './messages';
 
-interface CourseTabsNavigationProps {
+export interface CourseTabsNavigationProps {
   activeTabSlug?: string;
-  className?: string | null;
```
brian-smith-tcril, line 13: Debating if removing this makes sense. It was only used for the `mb-3` so overall it feels like cleanup. I think removing it is a reasonable choice and if someone wants to use this component in another part of the application with their own classes they can add `className` back to support that and we can discuss that choice/need in that hypothetical future PR.

brian-smith-tcril (commented):
Overall this looks great! I left a couple comments about the `PropTypes` -> `.tsx` conversion, once those are addressed this should be good to land!

src/course-tabs/CourseTabsNavigation.tsx
xitij2000: This is mainly why I removed it. It also didn't feel like a component that would be very reusable.

src/tab-page/LoadedTabPage.tsx
xitij2000: Wherever possible I tried to go for a more strict definition that would still work. Here the consumer `StreakModal` marks it as required. Since it gets a default value, I think the only situation where it would be undefined would be if someone explicitly passes undefined.

brian-smith-tcril: After digging into this a bit I agree a more strict definition works. `StreakModal` having it marked as required doesn't really play into this, as it would get the required string from `LoadedTabPage`'s default. `TabPage`, on the other hand, is the only consumer of `LoadedTabPage` and has `metadataModel` set up to be required.

With that in mind, I think it makes sense to:
* Make the switch to required from optional
* Drop the default (it's unreachable when `metadataModel` is required)
* Update any tests that rely on the default

xitij2000:
@brian-smith-tcril I've applied the suggested changes.

brian-smith-tcril:
There are a few tests that don't set `metadataModel` when rendering `LoadedTabPage` directly that should be updated:

https://github.com/openedx/frontend-app-learning/blob/db18c2cb456fcf887b30c9122e153c163f867886/src/course-home/outline-tab/OutlineTab.test.jsx#L641

https://github.com/openedx/frontend-app-learning/blob/db18c2cb456fcf887b30c9122e153c163f867886/src/course-home/outline-tab/OutlineTab.test.jsx#L657

https://github.com/openedx/frontend-app-learning/blob/db18c2cb456fcf887b30c9122e153c163f867886/src/course-home/progress-tab/ProgressTab.test.jsx#L1434

https://github.com/openedx/frontend-app-learning/blob/db18c2cb456fcf887b30c9122e153c163f867886/src/course-home/progress-tab/ProgressTab.test.jsx#L1448

https://github.com/openedx/frontend-app-learning/blob/db18c2cb456fcf887b30c9122e153c163f867886/src/course-home/progress-tab/ProgressTab.test.jsx#L1463

https://github.com/openedx/frontend-app-learning/blob/db18c2cb456fcf887b30c9122e153c163f867886/src/course-home/progress-tab/ProgressTab.test.jsx#L1476

Once those are updated this should be good to land!

xitij2000:
@brian-smith-tcril  I've applied your suggestions. This is ready for another look.

brian-smith-tcril (approved):

diana-villalvazo-wgu:
Hi @xitij2000 are you planning to backport this to Verawood? 😅

xitij2000:
@diana-villalvazo-wgu It's already backported to ulmo in this branch: https://github.com/open-craft/frontend-app-learning/tree/kshitij/plugin-slots-exemplar.ulmo

But upstreaming it to verawood shouldn't be much of an effort. I am not sure if it'll qualify as a post-release backport though. I'll look into it.

xitij2000:
@diana-villalvazo-wgu  Create a PR here: https://github.com/openedx/frontend-app-learning/pull/1931

jacobo-dominguez-wgu:
> @diana-villalvazo-wgu It's already backported to ulmo in this branch: https://github.com/open-craft/frontend-app-learning/tree/kshitij/plugin-slots-exemplar.ulmo
> 
> But upstreaming it to verawood shouldn't be much of an effort. I am not sure if it'll qualify as a post-release backport though. I'll look into it.

Is it backported to upstream ulmo?

xitij2000:
@jacobo-dominguez-wgu No, and I think it's a bit too late to do that now since the next release is cut.

PR #1720, opened 2025-05-29
feat: hide sidebar on screen resize

holaontiveros:
fixes #1400 

## Description
Adds an event handler on the window resize to check if the sidebar isOpen and the size of the viewport is smaller than the sidebar display to hide the sidebar and prevent it from blocking the course view.

This doesn't have any debounce given that the execution only happens when the sidebar is open and the size it's met so the code get's only executed once.

## Video
Before the change:

https://github.com/user-attachments/assets/8a71c0dc-e335-4785-9bd0-6280da3fdabf

After the change:

https://github.com/user-attachments/assets/27010517-1534-4633-a833-2e97ddfa8794

PR #1375, opened 2024-04-30
[FC-0056] Course outline sidebar

ihor-romaniuk:
**Settings**

```yaml
EDX_PLATFORM_REPOSITORY: https://github.com/raccoongang/edx-platform.git
EDX_PLATFORM_VERSION: NiedielnitsevIvan/FC-0056/feature/waffle-flag-for-displaying-discussions-tab
```

### Description
This pull request adds an important feature to our platform: displaying a navigation sidebar within a given course.

Interaction with feature flag `courseware.show_default_right_sidebar` has been added to control the default display of the discussion sidebar and `courseware.disable_navigation_sidebar` to control of enabling the course outline sidebar. 

Discussions or Notifications sidebar shouldn't be opened on Learning MFE by default. If waffle flag enabled - Discussions always opens on the pages with discussions, if user is in Audit and course has verified mode - show Notifications.

### Depends on BE PRs
- [feat: [FC-0056] Implement Sidebar Navigation](https://github.com/openedx/edx-platform/pull/34457)
- [feat: [FC-0056] Waffle flag for opening discussions/notifications by default and disabling the course outline sidebar](https://github.com/openedx/edx-platform/pull/34650)

### Design
https://www.figma.com/file/gew5tORDX4Q7wxOS8vjqZu/side-nav-OEX?type=design&node-id=318-3234&mode=design&t=rBe1ToNYP8RY6QOp-0

image

### Testing instructions
1. Run master devstack.
2. Start platform make dev.up.lms+cms+frontend-app-course-authoring and make checkout on this branch.
3. Go to Course Unit page from the Course Outline page.

PR #1920, opened 2026-05-27
feat: Course outline restructure with new plugin slots

xitij2000:
Splits `useCourseOutlineSidebar` into `useCourseOutlineData` + `useCourseOutlineSidebar`, adds `CourseOutlineSidebarHeadingSlot`, `CourseOutlineSidebarSectionCompletionIconSlot`, and `CourseOutlineSidebarSequenceCompletionIconSlot`, and introduces CSS variables for theming the sidebar.

### DIFF
--- a/src/courseware/course/sidebar/ARCHITECTURE.md
+++ b/src/courseware/course/sidebar/ARCHITECTURE.md
@@ -8,9 +8,9 @@ The Learning MFE uses a **two-sidebar system** where a left sidebar (Course Outl
 
 ### LEFT SIDEBAR (Course Outline)
 - **Location**: Left side of screen, adjacent to course content
-- **Component**: `CourseOutlineTray` (rendered via `CourseOutlineSidebarSlot`)
+- **Component**: `CourseOutlineTray` (rendered via `CourseOutlineSidebarSlot`). This is now a thin wrapper (`CourseOutlineTray.tsx`) that gates on sidebar state and renders the presentational `CourseOutline` (`CourseOutline.tsx`).
 - **Purpose**: Navigation - displays course structure, sequences, and units
-- **State Management**: Uses `useCourseOutlineSidebar()` hook
+- **State Management**: Split across two hooks — `useCourseOutlineData()` (course data + unit-click tracking) and `useCourseOutlineSidebar()` (sidebar open/collapse state)
 - **ID**: `WIDGETS.COURSE_OUTLINE`
 - **Rendering**: Only renders when `currentSidebar === 'COURSE_OUTLINE'`
 - **Trigger**: `CourseOutlineTrigger` (separate location in mobile/desktop toolbar)
@@ -200,22 +200,36 @@ if (!firstAvailable) {
 }
 ```
 
-### useCourseOutlineSidebar Hook
+### Course Outline Hooks (`hooks.js`)
+
+The former single `useCourseOutlineSidebar` hook is split into two, separating
+course data from sidebar context so the outline can be reused outside a sidebar:
+
+#### `useCourseOutlineData()`
 
 **Responsibilities:**
-- Detect when RIGHT sidebar has no available panels (`!initialSidebar`)
-- Auto-open Course Outline as fallback (unless manually collapsed)
-- Handle Course Outline specific interactions (unit clicks, toggle, resize)
+- Provide course outline data from Redux: `sections`, `sequences`, `units`, `courseOutlineStatus`, `activeSequenceId`, `sequenceStatus`
+- Provide `isEnabledCompletionTracking` and `isActiveEntranceExam`
+- Provide `handleUnitClick` (analytics + `checkBlockCompletion`) and trigger the outline-structure load effect
+- Reads Redux + `useParams` only — **does not** read `SidebarContext`, so it is reusable outside a sidebar
+
+#### `useCourseOutlineSidebar()`
+
+**Responsibilities:**
+- Expose sidebar context state: `currentSidebar`, `shouldDisplayFullScreen`, `handleToggleCollapse`
+- Handle the resize → collapse behaviour when the viewport drops below the desktop breakpoint
 
 **Key Logic:**
 ```javascript
-const isOpenSidebar = !initialSidebar && !isCollapsedOutlineSidebar;
-
-useEffect(() => {
-  if (isOpenSidebar && currentSidebar !== ID) {
-    toggleSidebar('COURSE_OUTLINE');
-  }
-}, [initialSidebar, unitId]);
+useLayoutEffect(() => {
+  const handleResize = () => {
+    if (currentSidebar === ID && global.innerWidth < breakpoints.large.maxWidth) {
+      collapseSidebar();
+    }
+  };
+  global.addEventListener('resize', handleResize);
+  return () => global.removeEventListener('resize', handleResize);
+}, [currentSidebar]);
 ```
 
 ### Sidebar.jsx (RIGHT Sidebar Renderer)
@@ -231,15 +245,44 @@ if (!currentSidebar || !SIDEBARS || !SIDEBARS[currentSidebar]) {
 }
 ```
 
-### CourseOutlineTray.jsx (LEFT Sidebar Renderer)
+### CourseOutlineTray.tsx / CourseOutline.tsx (LEFT Sidebar Renderer)
 
-**Responsibilities:**
-- Render course outline navigation
-- Only show when `currentSidebar === 'COURSE_OUTLINE'`
+The renderer is split into a sidebar-aware wrapper and a presentational component:
 
-**Key Logic:**
+**`CourseOutlineTray.tsx` (wrapper):**
+- Reads `useCourseOutlineSidebar()`
+- Only renders when `currentSidebar === 'COURSE_OUTLINE'`, otherwise `null`
+- Renders `<CourseOutline>`, passing `shouldDisplayFullScreen` and `onToggleCollapse`
+- Still carries `CourseOutlineTray.ID` and is the component registered in the slot
+
 ```javascript
-if (isActiveEntranceExam || currentSidebar !== ID) {
+if (currentSidebar !== ID) {
   return null;
 }
+return <CourseOutline shouldDisplayFullScreen={shouldDisplayFullScreen} onToggleCollapse={handleToggleCollapse} />;
 ```
+
+**`CourseOutline.tsx` (presentational):**
+- Reads course data via `useCourseOutlineData()` + `useParams()`
+- Returns `null` only when `isActiveEntranceExam`
+- Renders the heading through `CourseOutlineSidebarHeadingSlot` (see Extension Points below)
+
+### Course Outline Extension Points (Plugin Slots)
+
+The refactor extracts the heading and completion icon into standalone components
+exposed through plugin slots, so operators can customise them via `env.config.jsx`:
+
+- **`CourseOutlineSidebarHeadingSlot`** — ID `org.openedx.frontend.learning.course_outline_sidebar_heading.v1`. Wraps the extracted `components/CourseOutlineHeading.tsx`. Props: `isDisplaySequenceLevel`, `backButton?`, `onToggleCollapse?`. See `src/plugin-slots/CourseOutlineSidebarHeadingSlot/README.md`.
+- **`CourseOutlineSidebarCompletionIconSlot`** — ID `org.openedx.frontend.learning.course_outline_sidebar_completion_icon.v1`. Wraps `CompletionIcon.tsx` (now TypeScript, exporting `CompletionIconProps`). Consumed by **both** `SidebarSection` and `SidebarSequence`; the `variant` prop (`'section' | 'sequence'`) tells a plugin which location it is rendering, and `active` reflects whether that section/sequence is current. See `src/plugin-slots/CourseOutlineSidebarCompletionIconSlot/README.md`.
+
+The mobile "collapse the outline after selecting a unit" behaviour now lives in
+`components/UnitLinkWrapper.tsx`, which consumes both hooks (data for the click
+handler, sidebar for `shouldDisplayFullScreen`/`handleToggleCollapse`).
+
+### Styling
+
+The outline sidebar styling (`CourseOutlineTray.scss`) uses Paragon theme tokens
+rather than hard-coded colours — e.g. borders use `var(--pgn-color-light-700)`.
+The active-section highlight is driven by an `.active-section` class on
+`.course-sidebar-section` (`background-color: var(--pgn-color-info-100)`) instead
+of an inline `bg-info-100` class on the button.
--- a/src/courseware/course/sidebar/sidebars/course-outline/CourseOutline.tsx
+++ b/src/courseware/course/sidebar/sidebars/course-outline/CourseOutline.tsx
@@ -2,33 +2,43 @@ import { useIntl } from '@edx/frontend-platform/i18n';
 import { useToggle } from '@openedx/paragon';
 import { LOADING } from '@src/constants';
 
+import {
+  useCourseOutlineData,
+} from '@src/courseware/course/sidebar/sidebars/course-outline/hooks';
 import PageLoading from '@src/generic/PageLoading';
 import { CourseOutlineSidebarHeadingSlot } from '@src/plugin-slots/CourseOutlineSidebarHeadingSlot';
 import classNames from 'classnames';
 import { useState } from 'react';
-import SidebarSection from './components/SidebarSection';
+import { useParams } from 'react-router-dom';
 import SidebarSequence from './components/SidebarSequence';
-import { ID } from './constants';
-import { useCourseOutlineSidebar } from './hooks';
+import SidebarSection from './components/SidebarSection';
 import messages from './messages';
 
-export const CourseOutline = () => {
+interface CourseOutlineProps {
+  shouldDisplayFullScreen?: boolean;
+  onToggleCollapse?: () => void;
+}
+
+interface CoursePageParams extends Record<string, string> {
+  courseId: string;
+  unitId: string;
+}
+
+export const CourseOutline = ({
+  shouldDisplayFullScreen = false,
+  onToggleCollapse,
+}: CourseOutlineProps) => {
   const intl = useIntl();
   const [selectedSection, setSelectedSection] = useState<string | null>(null);
   const [isDisplaySequenceLevel, setDisplaySequenceLevel, setDisplaySectionLevel] = useToggle(true);
-
+  const { unitId, courseId } = useParams<CoursePageParams>();
   const {
-    courseId,
-    unitId,
-    currentSidebar,
-    isActiveEntranceExam,
     courseOutlineStatus,
     activeSequenceId,
     sections,
     sequences,
-    shouldDisplayFullScreen,
-    handleToggleCollapse,
-  } = useCourseOutlineSidebar();
+    isActiveEntranceExam,
+  } = useCourseOutlineData();
 
   const resolvedSectionId = selectedSection
         || Object.keys(sections).find(
@@ -47,19 +57,16 @@ export const CourseOutline = () => {
     setDisplaySequenceLevel();
     setSelectedSection(id);
   };
-
   const sidebarHeading = (
     <CourseOutlineSidebarHeadingSlot
-      onToggleCollapse={handleToggleCollapse}
+      onToggleCollapse={onToggleCollapse}
       isDisplaySequenceLevel={isDisplaySequenceLevel}
       backButton={backButtonTitle ? { title: backButtonTitle, onClick: handleBackToSectionLevel } : undefined}
     />
   );
-
-  if (isActiveEntranceExam || currentSidebar !== ID) {
+  if (isActiveEntranceExam) {
     return null;
   }
-
   if (courseOutlineStatus === LOADING) {
     return (
       <div className={classNames('outline-sidebar-wrapper', {
@@ -108,5 +115,3 @@ export const CourseOutline = () => {
     </div>
   );
 };
-
-export default CourseOutline;
--- a/src/courseware/course/sidebar/sidebars/course-outline/CourseOutline.tsx
+++ b/src/courseware/course/sidebar/sidebars/course-outline/CourseOutline.tsx
@@ -97,10 +97,10 @@ export const CourseOutline = ({
             ? sequenceIds.map((sequenceId: string) => (
               <SidebarSequence
                 key={sequenceId}
-                courseId={courseId!}
+                courseId={courseId}
                 sequence={sequences[sequenceId]}
                 defaultOpen={sequenceId === activeSequenceId}
-                activeUnitId={unitId!}
+                activeUnitId={unitId}
               />
             ))
             : sectionsIds.map((sectionId) => (
--- a/src/courseware/course/sidebar/sidebars/course-outline/CourseOutlineTray.tsx
+++ b/src/courseware/course/sidebar/sidebars/course-outline/CourseOutlineTray.tsx
@@ -5,12 +5,14 @@ import { useCourseOutlineSidebar } from './hooks';
 const CourseOutlineTray = () => {
   const {
     currentSidebar,
+    shouldDisplayFullScreen,
+    handleToggleCollapse,
   } = useCourseOutlineSidebar();
 
   if (currentSidebar !== ID) {
     return null;
   }
-  return <CourseOutline />;
+  return <CourseOutline shouldDisplayFullScreen={shouldDisplayFullScreen} onToggleCollapse={handleToggleCollapse} />;
 };
 
 CourseOutlineTray.ID = ID;
--- a/src/courseware/course/sidebar/sidebars/course-outline/components/SidebarSection.jsx
+++ b/src/courseware/course/sidebar/sidebars/course-outline/components/SidebarSection.jsx
@@ -8,7 +8,7 @@ import { Button, Icon } from '@openedx/paragon';
 import { ChevronRight as ChevronRightIcon } from '@openedx/paragon/icons';
 
 import courseOutlineMessages from '@src/course-home/outline-tab/messages';
-import { useCourseOutlineSidebar } from '../hooks';
+import { useCourseOutlineData } from '../hooks';
 
 const SidebarSection = ({ section, handleSelectSection }) => {
   const intl = useIntl();
@@ -20,7 +20,7 @@ const SidebarSection = ({ section, handleSelectSection }) => {
     completionStat,
   } = section;
 
-  const { activeSequenceId, isEnabledCompletionTracking } = useCourseOutlineSidebar();
+  const { activeSequenceId, isEnabledCompletionTracking } = useCourseOutlineData();
   const isActiveSection = sequenceIds.includes(activeSequenceId);
 
   const sectionTitle = (
--- a/src/courseware/course/sidebar/sidebars/course-outline/components/SidebarSequence.jsx
+++ b/src/courseware/course/sidebar/sidebars/course-outline/components/SidebarSequence.jsx
@@ -8,7 +8,7 @@ import { useIntl } from '@edx/frontend-platform/i18n';
 import { Collapsible } from '@openedx/paragon';
 
 import courseOutlineMessages from '@src/course-home/outline-tab/messages';
-import { useCourseOutlineSidebar } from '../hooks';
+import { useCourseOutlineData } from '../hooks';
 import SidebarUnit from './SidebarUnit';
 import { UNIT_ICON_TYPES } from './UnitIcon';
 
@@ -30,7 +30,7 @@ const SidebarSequence = ({
   } = sequence;
 
   const [open, setOpen] = useState(defaultOpen);
-  const { activeSequenceId, units, isEnabledCompletionTracking } = useCourseOutlineSidebar();
+  const { activeSequenceId, units, isEnabledCompletionTracking } = useCourseOutlineData();
   const isActiveSequence = id === activeSequenceId;
 
   const sectionTitle = (
--- a/src/courseware/course/sidebar/sidebars/course-outline/components/UnitLinkWrapper.tsx
+++ b/src/courseware/course/sidebar/sidebars/course-outline/components/UnitLinkWrapper.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { Link, useLocation } from 'react-router-dom';
 
-import { useCourseOutlineSidebar } from '../hooks';
+import { useCourseOutlineData, useCourseOutlineSidebar } from '../hooks';
 
 interface Props {
   courseId: string;
@@ -26,17 +26,25 @@ const UnitLinkWrapper: React.FC<Props> = ({
   courseId,
   children,
 }) => {
-  const { handleUnitClick } = useCourseOutlineSidebar();
+  const { handleUnitClick } = useCourseOutlineData();
+  const { shouldDisplayFullScreen, handleToggleCollapse } = useCourseOutlineSidebar();
   const { pathname } = useLocation();
   const isPreview = pathname.startsWith('/preview');
   const baseUrl = `/course/${courseId}/${sequenceId}/${id}`;
   const link = isPreview ? `/preview${baseUrl}` : baseUrl;
+  const handleClick = React.useCallback(() => {
+    // Hide the sidebar after selecting a unit on a mobile device.
+    if (shouldDisplayFullScreen) {
+      handleToggleCollapse();
+    }
+    handleUnitClick({ sequenceId, activeUnitId, id });
+  }, [handleUnitClick, sequenceId, activeUnitId, id, shouldDisplayFullScreen, handleToggleCollapse]);
 
   return (
     <Link
       to={link}
       className="row w-100 m-0 d-flex align-items-center text-gray-700"
-      onClick={() => handleUnitClick({ sequenceId, activeUnitId, id })}
+      onClick={handleClick}
     >
       {children}
     </Link>
--- a/src/courseware/course/sidebar/sidebars/course-outline/hooks.js
+++ b/src/courseware/course/sidebar/sidebars/course-outline/hooks.js
@@ -1,27 +1,26 @@
+import { sendTrackEvent, sendTrackingLogEvent } from '@edx/frontend-platform/analytics';
+import { breakpoints } from '@openedx/paragon';
+import { LOADED } from '@src/constants';
 import {
-  useContext, useEffect, useLayoutEffect, useState,
+  getCourseOutline,
+  getCourseOutlineShouldUpdate,
+  getCourseOutlineStatus,
+  getCoursewareOutlineSidebarSettings,
+  getSequenceId,
+  getSequenceStatus,
+} from '@src/courseware/data/selectors';
+import { checkBlockCompletion, getCourseOutlineStructure } from '@src/courseware/data/thunks';
+
+import { useModel } from '@src/generic/model-store';
+import {
+  useContext, useEffect, useLayoutEffect,
 } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { useParams } from 'react-router-dom';
-import { sendTrackEvent, sendTrackingLogEvent } from '@edx/frontend-platform/analytics';
-import { breakpoints } from '@openedx/paragon';
-
-import { useModel } from '@src/generic/model-store';
-import { LOADED } from '@src/constants';
-import { checkBlockCompletion, getCourseOutlineStructure } from '@src/courseware/data/thunks';
-import {
-  getCoursewareOutlineSidebarSettings,
-  getCourseOutlineShouldUpdate,
-  getCourseOutlineStatus,
-  getSequenceId,
-  getCourseOutline,
-  getSequenceStatus,
-} from '@src/courseware/data/selectors';
 import SidebarContext from '../../SidebarContext';
 import { ID } from './constants';
 
-// eslint-disable-next-line import/prefer-default-export
-export const useCourseOutlineSidebar = () => {
+export const useCourseOutlineData = () => {
   const dispatch = useDispatch();
   const {
     enableCompletionTracking: isEnabledCompletionTracking,
@@ -30,21 +29,14 @@ export const useCourseOutlineSidebar = () => {
   const courseOutlineStatus = useSelector(getCourseOutlineStatus);
   const sequenceStatus = useSelector(getSequenceStatus);
   const activeSequenceId = useSelector(getSequenceId);
-  const { sections = {}, sequences = {}, units = {} } = useSelector(getCourseOutline);
+  const {
+    sections = {},
+    sequences = {},
+    units = {},
+  } = useSelector(getCourseOutline);
 
   const { courseId } = useParams();
   const course = useModel('coursewareMeta', courseId);
-
-  const {
-    unitId,
-    currentSidebar,
-    toggleSidebar,
-    shouldDisplayFullScreen,
-  } = useContext(SidebarContext);
-
-  // Course outline state is now fully controlled by SidebarContextProvider
-  // This component only renders when currentSidebar === 'COURSE_OUTLINE'
-  const [isOpen, setIsOpen] = useState(true);
 
   const {
     entranceExamEnabled,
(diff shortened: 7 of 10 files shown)


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

xitij2000:
@brian-smith-tcril Did you get a chance to look at this PR?

brian-smith-tcril:
> @brian-smith-tcril Did you get a chance to look at this PR?

Not quite yet. I've been focused on a release testing bug fix. I definitely haven't forgotten about it and it's still high on my todo list.

src/courseware/course/sidebar/sidebars/course-outline/components/CompletionIcon.tsx
```diff
@@ -1,12 +1,19 @@
...
   LmsCompletionSolid as LmsCompletionSolidIcon,
 } from '@openedx/paragon/icons';
 
 import { DashedCircleIcon } from '../icons';
 
-const CompletionIcon = ({ completionStat: { completed = 0, total = 0 }, enabled }) => {
+export interface CompletionIconProps {
+  completionStat: {
+    completed: number;
+    total: number;
```
brian-smith-tcril: The propTypes didn't require these, they just required a `completionStat` object be passed in.


```suggestion
    completed?: number;
    total?: number;
```

src/courseware/course/sidebar/sidebars/course-outline/components/CourseOutlineHeading.tsx
```diff
@@ -0,0 +1,51 @@
...
+export const CourseOutlineHeading = ({
+  isSequenceLevel,
+  title,
+  onClickBack,
+  onToggleCollapse,
+}:CourseOutlineHeadingProps) => {
+  const intl = useIntl();
+  return (
+    <div
+      className="outline-sidebar-heading-wrapper sticky d-flex justify-content-between align-self-start align-items-center p-2.5 pl-4"
```
brian-smith-tcril: This doesn't match the classes that existed in the `sidebarHeading` const in `CourseOutlineTray.jsx`. Could you explain why the removed classes are being removed? Adding a suggestion here for putting the old classes back just to make it easy to see what changed:

```suggestion
      className="outline-sidebar-heading-wrapper sticky d-flex justify-content-between align-self-start align-items-center bg-light-200 p-2.5 pl-4"
```

src/courseware/course/sidebar/sidebars/course-outline/components/CourseOutlineHeading.tsx
```diff
@@ -0,0 +1,51 @@
...
+    >
+      {isSequenceLevel && title ? (
+        <Button
+          variant="link"
+          iconBefore={ChevronLeftIcon}
+          className="outline-sidebar-heading p-0 mb-0 text-left"
```
brian-smith-tcril: This doesn't match the classes that existed in the `sidebarHeading` const in `CourseOutlineTray.jsx`. Could you explain why the removed classes are being removed? Adding a suggestion here for putting the old classes back just to make it easy to see what changed:


```suggestion
          className="outline-sidebar-heading p-0 mb-0 text-left text-dark-500"
```

src/courseware/course/sidebar/sidebars/course-outline/components/CourseOutlineHeading.tsx
```diff
@@ -0,0 +1,51 @@
...
+          onClick={onClickBack}
+        >
+          {title}
+        </Button>
+      ) : (
+        <span className="outline-sidebar-heading mb-0 h4">
```
brian-smith-tcril: This doesn't match the classes that existed in the `sidebarHeading` const in `CourseOutlineTray.jsx`. Could you explain why the removed classes are being removed? Adding a suggestion here for putting the old classes back just to make it easy to see what changed:

```suggestion
        <span className="outline-sidebar-heading mb-0 h4 text-dark-500">
```

src/courseware/course/sidebar/sidebars/course-outline/components/CourseOutlineHeading.tsx
```diff
@@ -0,0 +1,51 @@
...
+          {intl.formatMessage(messages.courseOutlineTitle)}
+        </span>
+      )}
+      {onToggleCollapse && (
+        <IconButton
+          alt={intl.formatMessage(messages.toggleCourseOutlineTrigger)}
+          className="outline-sidebar-toggle-btn flex-shrink-0 text-dark"
```
brian-smith-tcril: This doesn't match the classes that existed in the `sidebarHeading` const in `CourseOutlineTray.jsx`. Could you explain why the removed classes are being removed? Adding a suggestion here for putting the old classes back just to make it easy to see what changed:

```suggestion
          className="outline-sidebar-toggle-btn flex-shrink-0 text-dark bg-light-200"
```

src/courseware/course/sidebar/sidebars/course-outline/components/CourseOutlineHeading.tsx
```diff
@@ -0,0 +1,51 @@
...
+  ChevronLeft as ChevronLeftIcon,
+} from '@openedx/paragon/icons';
+import messages from '../messages';
+
+export interface CourseOutlineHeadingProps {
+  isSequenceLevel: boolean;
+  title?: string;
+  onClickBack: () => void;
+  onToggleCollapse?: () => void;
+}
```
brian-smith-tcril, line 16:
```suggestion
export interface CourseOutlineHeadingProps {
  backButton?: {
    title: string;
    onClick: () => void;
  };
  onToggleCollapse?: () => void;
}
```

src/courseware/course/sidebar/sidebars/course-outline/components/CourseOutlineHeading.tsx
```diff
@@ -0,0 +1,51 @@
...
+  title,
+  onClickBack,
+  onToggleCollapse,
+}:CourseOutlineHeadingProps) => {
+  const intl = useIntl();
+  return (
+    <div
+      className="outline-sidebar-heading-wrapper sticky d-flex justify-content-between align-self-start align-items-center p-2.5 pl-4"
+    >
+      {isSequenceLevel && title ? (
```
brian-smith-tcril: The `sidebarHeading` const in `CourseOutlineTray.jsx` had more descriptive names here (suggestion used to make diff easy to see)

```suggestion
      {isDisplaySequenceLevel && backButtonTitle ? (
```

`isDisplaySequenceLevel` does feel like a confusing name to me, but `title` is more confusing here as it's only used for the back button text.

I think a clean way to handle this would be to reshape the props (see https://github.com/openedx/frontend-app-learning/pull/1920/changes#r3416972228) and instead just evaluate the ternary on `backButton`

src/courseware/course/sidebar/sidebars/course-outline/components/CourseOutlineHeading.tsx
```diff
@@ -0,0 +1,51 @@
...
+  onClickBack: () => void;
+  onToggleCollapse?: () => void;
+}
+
+export const CourseOutlineHeading = ({
+  isSequenceLevel,
+  title,
+  onClickBack,
+  onToggleCollapse,
+}:CourseOutlineHeadingProps) => {
```
brian-smith-tcril: nit:


```suggestion
}: CourseOutlineHeadingProps) => {
```

src/courseware/course/sidebar/sidebars/course-outline/hooks.js
brian-smith-tcril, line 1: This split feels like a good idea in general, but I'd want to do a thorough review of it. I also don't want to block the slots landing on this split happening, so I'd prefer this split happen in a follow-up PR.

src/courseware/course/sidebar/sidebars/course-outline/components/SidebarSection.jsx
```diff
@@ -18,13 +20,17 @@ const SidebarSection = ({ section, handleSelectSection }) => {
...
   const isActiveSection = sequenceIds.includes(activeSequenceId);
 
   const sectionTitle = (
     <>
       <div className="col-auto p-0">
-        <CompletionIcon completionStat={completionStat} enabled={isEnabledCompletionTracking} />
+        <CourseOverviewSectionCompletionIconSlot
+          completionStat={completionStat}
+          enabled={isEnabledCompletionTracking}
+          active={isActiveSection}
```
brian-smith-tcril, line 33: This isn't used by the default content (the `CompletionIcon` component) or any examples in the slot `README`. I think this should either be removed, or a slot `README` example showing why this is needed should be added.

src/courseware/course/sidebar/sidebars/course-outline/components/SidebarSection.jsx
```diff
@@ -41,13 +47,10 @@ const SidebarSection = ({ section, handleSelectSection }) => {
...
   return (
-    <li className="mb-2 course-sidebar-section">
+    <li className={classNames('course-sidebar-section', { 'active-section': isActiveSection })}>
       <Button
         variant="tertiary"
-        className={classNames(
-          'd-flex align-items-center w-100 px-4 py-3.5 rounded-0 justify-content-start',
-          { 'bg-info-100': isActiveSection },
-        )}
+        className="d-flex align-items-center w-100 rounded-0 justify-content-start"
```
brian-smith-tcril: Curious about the `bg-info-100` removal here

src/courseware/course/sidebar/sidebars/course-outline/components/SidebarSection.jsx
```diff
@@ -41,13 +47,10 @@ const SidebarSection = ({ section, handleSelectSection }) => {
   );
 
   return (
-    <li className="mb-2 course-sidebar-section">
+    <li className={classNames('course-sidebar-section', { 'active-section': isActiveSection })}>
```
brian-smith-tcril: `active-section` doesn't appear to be defined anywhere. I couldn't find it in existing stylesheets (https://github.com/search?q=repo%3Aopenedx%2Ffrontend-app-learning%20active-section&type=code) and it doesn't appear to be added in any stylesheets in this PR.

src/courseware/course/sidebar/sidebars/course-outline/components/SidebarSection.jsx
```diff
@@ -41,13 +47,10 @@ const SidebarSection = ({ section, handleSelectSection }) => {
   );
 
   return (
-    <li className="mb-2 course-sidebar-section">
```
brian-smith-tcril, line 51: Curious about the `mb-2` removal here

src/courseware/course/sidebar/sidebars/course-outline/components/SidebarSequence.jsx
```diff
@@ -28,13 +30,17 @@ const SidebarSequence = ({
...
   const isActiveSequence = id === activeSequenceId;
 
   const sectionTitle = (
     <>
       <div className="col-auto p-0" style={{ fontSize: '1.1rem' }}>
-        <CompletionIcon completionStat={completionStat} enabled={isEnabledCompletionTracking} />
+        <CourseOverviewSequenceCompletionIconSlot
+          completionStat={completionStat}
+          enabled={isEnabledCompletionTracking}
+          active={isActiveSequence}
```
brian-smith-tcril, line 43: Same comment as https://github.com/openedx/frontend-app-learning/pull/1920/changes#r3419773864

This isn't used by the default content (the `CompletionIcon` component) or any examples in the slot `README`. I think this should either be removed, or a slot `README` example showing why this is needed should be added.

src/courseware/course/sidebar/sidebars/course-outline/components/SidebarUnit.jsx
```diff
@@ -28,7 +28,7 @@ const SidebarUnit = ({
   const completeAndEnabled = complete && isCompletionTrackingEnabled;
 
   return (
-    <li className={classNames({ 'bg-info-100': isActive, 'border-top border-light': !isFirst })}>
+    <li className={classNames({ 'active-unit bg-info-100': isActive, 'border-top border-light': !isFirst })}>
```
brian-smith-tcril: `active-unit` doesn't appear to exist.

src/courseware/course/sidebar/sidebars/course-outline/components/UnitLinkWrapper.tsx
```diff
@@ -26,17 +26,25 @@ const UnitLinkWrapper: React.FC<Props> = ({
...
   const isPreview = pathname.startsWith('/preview');
   const baseUrl = `/course/${courseId}/${sequenceId}/${id}`;
   const link = isPreview ? `/preview${baseUrl}` : baseUrl;
+  const handleClick = React.useCallback(() => {
+    // Hide the sidebar after selecting a unit on a mobile device.
+    if (shouldDisplayFullScreen) {
+      handleToggleCollapse();
+    }
+    handleUnitClick({ sequenceId, activeUnitId, id });
+  }, [handleUnitClick, sequenceId, activeUnitId, id]);
```
brian-smith-tcril: This might not be needed after https://github.com/openedx/frontend-app-learning/pull/1923 lands, worth double checking once that's in.

src/plugin-slots/CourseOutlineSidebarSectionCompletionIconSlot/index.tsx
brian-smith-tcril, line 1: This seems very similar to the sequence icon slot. I'm wondering if having 1 slot to handle both makes more sense. Doing that would mean plugin authors that want to have different components based on sequence vs section would likely need a prop, but that feels lighter than having 2 different slots.

brian-smith-tcril (commented):
Overall these extension points seem like good ones to add. Looking forward to continuing to discuss these changes and finding a good path for them to land!

src/plugin-slots/CourseOutlineSidebarSectionCompletionIconSlot/index.tsx
xitij2000: I agree, I think I can consolidate this and use a prop to select where the icon rendered since I think in most cases people can ignore that prop and render the same component.

src/courseware/course/sidebar/sidebars/course-outline/hooks.js
xitij2000: Sure! The main cause for making this split was to allow the outline component to be relocatable outside of the sidebar. We needed it rendered in the header. 

I wonder if there is any value in have a generalisable approach for components that can be relocatable.

src/courseware/course/sidebar/sidebars/course-outline/components/CourseOutlineHeading.tsx
xitij2000: I had moved these to the SCSS file applied to the outline-sidebar-heading-wrapper class but it seems during the subsequent changes to SCSS and rebasing those changes got accidentally removed. It was to make this easier to theme since `bg-light-200` applies a `!important`. Likewise for the other removed classes. 

Is that an acceptable change to make or should I revert these changes?

src/courseware/course/sidebar/sidebars/course-outline/components/SidebarSection.jsx
xitij2000: It's to make it themeable based on whether the section is active. While the default theme might not use it yet, this change will allow other themes to use it.

xitij2000:
@brian-smith-tcril 

Just giving context on some of the issues:
- I removed some of the padding, and colors classes so that themes can more easily override those. Unfortunately I had move them to the CSS file but those changes are lost. I can reimplement it, or revert to the original classes if what I'm planning is unacceptable. 
- I've added `active-unit` and `active-section` to allow themes to override these styles even if the default theme doesn't do it or does it by simply applying a class. 


I tried splitting this PR into two parts, however there were quite a few dependencies between the commits so it wasn't as easy a split as I hoped. I'll apply the other fixes first.

brian-smith-tcril:
> I removed some of the padding, and colors classes so that themes can more easily override those. Unfortunately I had move them to the CSS file but those changes are lost. I can reimplement it, or revert to the original classes if what I'm planning is unacceptable.

I'm open to discussing those changes, but I'd like to keep this PR focused on adding the plugin slots. By keeping the PR focused we won't have changes blocking each other from landing.

> I've added active-unit and active-section to allow themes to override these styles even if the default theme doesn't do it or does it by simply applying a class.

That's probably fine, but I'd still prefer to have this as a "just adding slots" PR.

> I tried splitting this PR into two parts, however there were quite a few dependencies between the commits so it wasn't as easy a split as I hoped. I'll apply the other fixes first.

I'm sorry about that. I know it can be frustrating to split things up, but I think the benefits of focused PRs are very real.

xitij2000:
@brian-smith-tcril I've split out the slots into this PR: https://github.com/openedx/frontend-app-learning/pull/1933

A test is failing locally so I still have work to do, but I'll rebase this PR once that one is merged.

src/courseware/course/sidebar/sidebars/course-outline/components/SidebarSection.jsx
```diff
@@ -48,13 +48,10 @@ const SidebarSection = ({ section, handleSelectSection }) => {
...
+    <li className={classNames('course-sidebar-section', { 'active-section': isActiveSection })}>
       <Button
         variant="tertiary"
-        className={classNames(
-          'd-flex align-items-center w-100 px-4 py-3.5 rounded-0 justify-content-start',
-          { 'bg-info-100': isActiveSection },
```
xitij2000, line 56: This is now applied via SCSS

xitij2000:
@brian-smith-tcril  I've updated this PR. 

FYI: the update to the `ARCHITECTURE.md` file was made using claude.

itsjeyd:
Hi @brian-smith-tcril, would you have time to take another look at this PR?

src/courseware/course/sidebar/sidebars/course-outline/components/SidebarSection.jsx
brian-smith-tcril: this was addressed in https://github.com/openedx/frontend-app-learning/pull/1933

src/courseware/course/sidebar/sidebars/course-outline/components/SidebarSection.jsx
brian-smith-tcril: this appears to no longer be removed.

src/courseware/course/sidebar/sidebars/course-outline/components/SidebarSequence.jsx
brian-smith-tcril: this was addressed in #1933

src/courseware/course/sidebar/sidebars/course-outline/components/CourseOutlineHeading.tsx
brian-smith-tcril: Resolving as this file is no longer modified by this PR.

src/courseware/course/sidebar/sidebars/course-outline/components/CourseOutlineHeading.tsx
brian-smith-tcril: Resolving as this file is no longer modified by this PR.

src/courseware/course/sidebar/sidebars/course-outline/components/CourseOutlineHeading.tsx
brian-smith-tcril: Resolving as this file is no longer modified by this PR.

src/courseware/course/sidebar/sidebars/course-outline/components/CourseOutlineHeading.tsx
brian-smith-tcril: Resolving as this file is no longer modified by this PR.

src/courseware/course/sidebar/sidebars/course-outline/components/CourseOutlineHeading.tsx
brian-smith-tcril: Resolving as this file is no longer modified by this PR.

src/courseware/course/sidebar/sidebars/course-outline/components/CourseOutlineHeading.tsx
brian-smith-tcril: Resolving as this file is no longer modified by this PR.

src/courseware/course/sidebar/sidebars/course-outline/components/CourseOutlineHeading.tsx
brian-smith-tcril: Resolving as this file is no longer modified by this PR.

src/courseware/course/sidebar/sidebars/course-outline/components/UnitLinkWrapper.tsx
```diff
@@ -26,17 +26,25 @@ const UnitLinkWrapper: React.FC<Props> = ({
...
   const isPreview = pathname.startsWith('/preview');
   const baseUrl = `/course/${courseId}/${sequenceId}/${id}`;
   const link = isPreview ? `/preview${baseUrl}` : baseUrl;
+  const handleClick = React.useCallback(() => {
+    // Hide the sidebar after selecting a unit on a mobile device.
+    if (shouldDisplayFullScreen) {
+      handleToggleCollapse();
+    }
+    handleUnitClick({ sequenceId, activeUnitId, id });
+  }, [handleUnitClick, sequenceId, activeUnitId, id, shouldDisplayFullScreen, handleToggleCollapse]);
```
brian-smith-tcril, line 41: In https://github.com/openedx/frontend-app-learning/pull/1920#discussion_r3419908256 I mentioned
> This might not be needed after https://github.com/openedx/frontend-app-learning/pull/1923 lands, worth double checking once that's in.

https://github.com/openedx/frontend-app-learning/pull/1923 has landed now, I haven't looked to deeply into if that would change what needs to happen here or not. Anything stand out to you as no longer needed or as something that should change now that it has landed?

src/courseware/course/sidebar/sidebars/course-outline/components/UnitLinkWrapper.tsx
brian-smith-tcril: continued in https://github.com/openedx/frontend-app-learning/pull/1920/changes#r3775369034

src/plugin-slots/CourseOutlineSidebarSectionCompletionIconSlot/index.tsx
brian-smith-tcril: Resolving as this file is no longer modified by this PR.

brian-smith-tcril:
@itsjeyd I took another look and resolved a lot of my old comments.

@xitij2000 this PR has a conflict in `SidebarUnit.jsx`, knowing how that gets resolved would make the next review pass much smoother, so if you could rebase this that would be perfect!

src/courseware/course/sidebar/sidebars/course-outline/components/SidebarSection.jsx
brian-smith-tcril: I think for the purposes of this PR I would prefer to keep the restructure and the "add some extra classes to help people with themes" concerns separate.

I'm happy to discuss adding `active-section` and `active-unit` classes, but I'd prefer that discussion not be buried in a review of a restructure.

A tiny PR changing 2 lines feels like a much better place to have that conversation.

xitij2000: Sure, i've kept them as separate commits to split them out to separate PRs. will do that tomorrow. Github is acting up for me right now.

src/courseware/course/sidebar/sidebars/course-outline/components/UnitLinkWrapper.tsx
xitij2000: I looked at that PR, but I don't think that should have any impact here. 

This code already existing even after that PR, it's just moved from here: https://github.com/openedx/frontend-app-learning/pull/1920/changes#diff-b879d0478da5bf04c130f250097f28b8c38d39e6607241497af78671d3940610L91-L96

xitij2000:
@brian-smith-tcril  That conflict had to do with the styling changes which I've removed. This is ready now. I'll prepare another PR for the styling changes.

src/courseware/course/sidebar/sidebars/course-outline/CourseOutline.tsx
```diff
@@ -2,33 +2,43 @@ import { useIntl } from '@edx/frontend-platform/i18n';
...
+}: CourseOutlineProps) => {
   const intl = useIntl();
   const [selectedSection, setSelectedSection] = useState<string | null>(null);
   const [isDisplaySequenceLevel, setDisplaySequenceLevel, setDisplaySectionLevel] = useToggle(true);
-
+  const { unitId, courseId } = useParams<CoursePageParams>();
   const {
-    courseId,
-    unitId,
-    currentSidebar,
```
brian-smith-tcril, line 23: Taking notes here:
* `currentSidebar` was only ever used in this file on L59 as part of `if (isActiveEntranceExam || currentSidebar !== ID)`
* https://github.com/openedx/frontend-app-learning/pull/1933 added a gate to `CourseOutlineTray.tsx` that does that exact check
* `CourseOutlineTray` is the only thing that renders this `CourseOutline` component

with that in mind removing `currentSidebar` is reasonable.

src/courseware/course/sidebar/sidebars/course-outline/CourseOutline.tsx
```diff
@@ -2,33 +2,43 @@ import { useIntl } from '@edx/frontend-platform/i18n';
...
-export const CourseOutline = () => {
+interface CourseOutlineProps {
+  shouldDisplayFullScreen?: boolean;
+  onToggleCollapse?: () => void;
+}
+
+interface CoursePageParams extends Record<string, string> {
+  courseId: string;
+  unitId: string;
+}
```
brian-smith-tcril, line 25: Claude called this out

> The non-optional declarations do nothing. `courseId: string; unitId: string` come back through `Partial<>` as `string | undefined` regardless. That's precisely why the author still had to write `courseId!` and `unitId!` at lines 100 and 103 — the interface claims the params are always present, the type system disagrees, and the `!` papers over the gap. The interface is stating a guarantee it can't deliver, and the route `/course/:courseId/:sequenceId` (`constants.ts:16`) means `unitId` genuinely is absent sometimes.

src/courseware/course/sidebar/sidebars/course-outline/hooks.js
```diff
@@ -1,27 +1,26 @@
-import {
-  useContext, useEffect, useLayoutEffect, useState,
-} from 'react';
```
brian-smith-tcril, line 3: Taking notes:
* This was moved down to `L15`
* `useState` is no longer being imported
* `useState` was only used by `const [isOpen, setIsOpen] = useState(true);` on `L47` which has been removed.

src/courseware/course/sidebar/sidebars/course-outline/hooks.js
```diff
@@ -1,27 +1,26 @@
...
-import { useDispatch, useSelector } from 'react-redux';
```
brian-smith-tcril, line 4: Taking notes:
* This was moved down to `L18`, no changes

src/courseware/course/sidebar/sidebars/course-outline/hooks.js
```diff
@@ -1,27 +1,26 @@
...
-import { useParams } from 'react-router-dom';
```
brian-smith-tcril, line 5: Taking notes:
* This was moved down to `L19`, no changes

src/courseware/course/sidebar/sidebars/course-outline/hooks.js
```diff
@@ -1,27 +1,26 @@
...
 import { sendTrackEvent, sendTrackingLogEvent } from '@edx/frontend-platform/analytics';
 import { breakpoints } from '@openedx/paragon';
-
-import { useModel } from '@src/generic/model-store';
```
brian-smith-tcril, line 9: Taking notes:
* This was moved down to `L14`, no changes

src/courseware/course/sidebar/sidebars/course-outline/hooks.js
```diff
@@ -1,27 +1,26 @@
...
 import { LOADED } from '@src/constants';
-import { checkBlockCompletion, getCourseOutlineStructure } from '@src/courseware/data/thunks';
```
brian-smith-tcril, line 11: Taking notes:
* This was moved down to `L12`, no changes

src/courseware/course/sidebar/sidebars/course-outline/hooks.js
```diff
@@ -1,27 +1,26 @@
...
 import {
-  getCoursewareOutlineSidebarSettings,
+  getCourseOutline,
   getCourseOutlineShouldUpdate,
   getCourseOutlineStatus,
+  getCoursewareOutlineSidebarSettings,
   getSequenceId,
-  getCourseOutline,
   getSequenceStatus,
 } from '@src/courseware/data/selectors';
```
brian-smith-tcril, line 11: Taking notes:
* This is just reordering, no changes
  * `getCourseOutline` `L17` -> `L5`
  * `getCoursewareOutlineSidebarSettings` `L13` -> `L8`

src/courseware/course/sidebar/sidebars/course-outline/hooks.js
```diff
@@ -30,41 +29,26 @@ export const useCourseOutlineSidebar = () => {
...
 
   const { courseId } = useParams();
   const course = useModel('coursewareMeta', courseId);
 
-  const {
-    unitId,
-    currentSidebar,
-    toggleSidebar,
-    shouldDisplayFullScreen,
-  } = useContext(SidebarContext);
```
brian-smith-tcril, line 43: Taking notes:
* Moved to `useCourseOutlineSidebar` `L98..102`
* `unitId` no longer part of destructure

src/courseware/course/sidebar/sidebars/course-outline/hooks.js
```diff
@@ -30,41 +29,26 @@ export const useCourseOutlineSidebar = () => {
...
-
-  // Course outline state is now fully controlled by SidebarContextProvider
-  // This component only renders when currentSidebar === 'COURSE_OUTLINE'
-  const [isOpen, setIsOpen] = useState(true);
```
brian-smith-tcril, line 47: Taking notes:
* `isOpen` and `setIsOpen` appear to have no consumers, so exporting them as part of `useCourseOutlineSidebar` isn't serving any purpose anymore.

Claude digging:
>   #1375 (1c3610e9, "create course outline sidebar") — introduced const [isOpen, setIsOpen] = useState(true); and returned both from the hook. Even at birth, nothing read them: the only three references in the whole sidebar directory were the declaration and the two return-object lines. And setIsOpen was never called anywhere, ever — the only commit in courseware history matching setIsOpen( is #1184, an unrelated
  modal-height change. So isOpen was hardcoded true from day one.
>
>  #1720 (d1dede56, "hide sidebar on screen resize") — added the resize effect and keyed it }, [isOpen]);. That's the one and only time isOpen did anything: it served as a dependency. Since the value never changed, the effect simply registered once — so the dep was decorative, but the reference existed.
>
>  #1885 (0664dc38, "decouple notifications panel using widget registry") — this is your answer. It moved sidebar-open ownership into SidebarContextProvider, deleting initialSidebar, isOpenSidebar, and the auto-open effect from the hook, and swapping the resize effect's dep from [isOpen] → [currentSidebar]. That removed the last reference. From that commit onward, isOpen/setIsOpen were declared and returned to nobody, and
  the author left the two-line comment as the explanation for why they were still there.
>
>  So: #1885 killed it, and left it in place with a tombstone. Pre-#1920, the pair was still being returned purely because nobody deleted them — not because anything depended on the return shape. #1920 is the cleanup, and since setIsOpen was never called in any version, there's no behavior anywhere that could have depended on the state changing.

src/courseware/course/sidebar/sidebars/course-outline/hooks.js
```diff
@@ -30,41 +29,26 @@ export const useCourseOutlineSidebar = () => {
...
-
   const {
     entranceExamEnabled,
     entranceExamPassed,
   } = course.entranceExamData || {};
   const isActiveEntranceExam = entranceExamEnabled && !entranceExamPassed;
 
-  const collapseSidebar = () => {
-    toggleSidebar(null);
-  };
```
brian-smith-tcril, line 57: Taking notes:
* Moved to new `useCourseOutlineSidebar` `L104..106`

src/courseware/course/sidebar/sidebars/course-outline/hooks.js
```diff
@@ -30,41 +29,26 @@ export const useCourseOutlineSidebar = () => {
...
-
-  const handleToggleCollapse = () => {
-    if (currentSidebar === ID) {
-      collapseSidebar();
-    } else {
-      toggleSidebar(ID);
-    }
-  };
```
brian-smith-tcril, line 65: Taking notes:
* Moved to new `useCourseOutlineSidebar` `L108..114`

src/courseware/course/sidebar/sidebars/course-outline/hooks.js
```diff
@@ -88,11 +72,6 @@ export const useCourseOutlineSidebar = () => {
 
     logEvent('edx.ui.lms.sequence.tab_selected', 'left');
     dispatch(checkBlockCompletion(courseId, sequenceId, activeUnitId));
-
-    // Hide the sidebar after selecting a unit on a mobile device.
-    if (shouldDisplayFullScreen) {
-      handleToggleCollapse();
-    }
```
brian-smith-tcril, line 95: Taking notes:
* `handleToggleCollapse` no longer exists here, it is part of the new `useCourseOutlineSidebar` not ` useCourseOutlineData`
* `handleUnitClick` was/is only ever called by `UnitLinkWrapper`

src/courseware/course/sidebar/sidebars/course-outline/components/UnitLinkWrapper.tsx
brian-smith-tcril, line 41: Looking at this block a few things stand out to me:
* It's no longer inline
  * This makes sense, we used to have the `handleToggleCollapse` in `handleUnitClick`, now that needs to be called here instead, making it a named function seems reasonable.
* The `if (shouldDisplayFullScreen) { handleToggleCollapse` block is _before_ `handleUnitClick`
  * That block was at the _end_ of `handleUnitClick` before. I don't think there's a behavioral difference with it moving, but it stood out to me as a difference to investigate.
* It is now using `useCallback` instead of a plain arrow function.
  * It's not clear to me what this is buying us. I'm open to the change if there's clear justification, but I'd think sticking to a plain arrow function here would be fine.

Claude dive into why `useCallback` isn't buying us anything
>  1. Something re-renders `UnitLinkWrapper`.
>  2. `UnitLinkWrapper()` runs again, so `useCourseOutlineData()` runs again.
>  3. That hook's line `const handleUnitClick = (...) => {...}` executes again → allocates a different function object than last render.
>  4. `useCallback` compares this render's deps to last render's with `Object.is`. `handleUnitClick` is a different object, so that check fails.
>  5. Cache miss → `useCallback` returns a newly created `handleClick`.

src/courseware/course/sidebar/sidebars/course-outline/hooks.js
```diff
@@ -102,6 +81,38 @@ export const useCourseOutlineSidebar = () => {
...
+    isEnabledCompletionTracking,
+    isActiveEntranceExam,
+    courseOutlineStatus,
+    activeSequenceId,
+    sections,
+    sequences,
+    units,
+    handleUnitClick,
+    sequenceStatus,
+  };
```
brian-smith-tcril, line 94: Taking notes:
* Comparing to the full `return` block from the old `useCourseOutlineSidebar`
  * `courseId`: removed, things get it with `useParams` now
  * `unitId`: removed, things get it with `useParams` now
  * `isOpen`/`setIsOpen`: removed, see https://github.com/openedx/frontend-app-learning/pull/1920#discussion_r3805425675
  * `currentSidebar`/`shouldDisplayFullScreen`/`handleToggleCollapse`: moved to new `useCourseOutlineSidebar`
  * Everything else: still here

src/courseware/course/sidebar/sidebars/course-outline/CourseOutline.tsx
```diff
@@ -2,33 +2,43 @@ import { useIntl } from '@edx/frontend-platform/i18n';
...
+
+export const CourseOutline = ({
+  shouldDisplayFullScreen = false,
+  onToggleCollapse,
+}: CourseOutlineProps) => {
   const intl = useIntl();
   const [selectedSection, setSelectedSection] = useState<string | null>(null);
   const [isDisplaySequenceLevel, setDisplaySequenceLevel, setDisplaySectionLevel] = useToggle(true);
-
+  const { unitId, courseId } = useParams<CoursePageParams>();
```
brian-smith-tcril, line 34: I don't see any test changes in this PR. Were there tests assuming we were getting a `unitId` from `useCourseOutlineSidebar` before that need to be updated to have one provided by params?

src/courseware/course/sidebar/sidebars/course-outline/CourseOutlineTray.tsx
```diff
@@ -5,12 +5,14 @@ import { useCourseOutlineSidebar } from './hooks';
...
     currentSidebar,
+    shouldDisplayFullScreen,
+    handleToggleCollapse,
   } = useCourseOutlineSidebar();
 
   if (currentSidebar !== ID) {
     return null;
   }
-  return <CourseOutline />;
+  return <CourseOutline shouldDisplayFullScreen={shouldDisplayFullScreen} onToggleCollapse={handleToggleCollapse} />;
```
brian-smith-tcril, line 15: Claude question about this:
> Why props here instead of having CourseOutline call useCourseOutlineSidebar() directly? Every other consumer in this subtree — the trigger, UnitLinkWrapper — reads the sidebar hook itself, so CourseOutline becomes the only component being fed these two values from outside. If the intent is to make CourseOutline renderable outside a sidebar, could the props fall back to the context values when they aren't passed, rather than to false/undefined? That way the sidebar case keeps working through the hook as it does today (and CourseOutlineTray doesn't need to pass anything at all), and the props become an explicit override for the out-of-sidebar case.

src/courseware/course/sidebar/ARCHITECTURE.md
brian-smith-tcril, line 1: I haven't read through this file fully but it looks like it might need to be updated, so this is just a note to make sure this file accurately reflects the current state of the repo with the changes in this PR.

brian-smith-tcril (commented):
I finally feel like I was able to wrap my head around these changes. Overall it's looking pretty good. I left a few comments in there with suggestions/questions, looking forward to hearing your thoughts!

src/courseware/course/sidebar/sidebars/course-outline/CourseOutlineTray.tsx
xitij2000: Having CourseOuline call `useCourseOutlineSidebar` is the exact reason for this refactoring in the first place, we want to make it independent of any sidebar logic. Having it still get it from the SidebarContext would partially defeat the purpose of the split.

src/courseware/course/sidebar/sidebars/course-outline/CourseOutline.tsx
xitij2000: Yes, that is a good point. I think the fact that not tests fail after moving the source from Sidebar context to useParams  means that the tests were probably not testing how unitId is used.

src/courseware/course/sidebar/sidebars/course-outline/CourseOutlineTray.tsx
brian-smith-tcril: > Having CourseOuline call `useCourseOutlineSidebar` is the exact reason for this refactoring in the first place, we want to make it independent of any sidebar logic. Having it still get it from the SidebarContext would partially defeat the purpose of the split.

Could you elaborate on this a bit? My feeling is that regardless of _where_ `CourseOutline` gets `shouldDisplayFullScreen` and `handleToggleCollapse` from, by caring about those _at all_ it's including "sidebar logic."

I'm open to these being passed as props, I just want to understand the motivation a bit better.

xitij2000: I see your point, that is how those values are currently used. However for a client we extracted this component and injected it in the header as a dropdown. It can still use the toggle collapse, but its context is different from that of a sidebar. This work was triggered by the need for hosting this component outside the sidebar. Having the values passed explicitly seemed to be a cleaner separation.

That said it's still in the sidebars folder so without further refactoring it's still somewhat tied to the sidebar and keeping the values by default won't break anything. 

We could do further refactoring to move this out and make the fullscreen code part of the sidebar wrapper  rather than the outline component, but I think for now I can make these params default to the sidebar context values.

PR #1713, opened 2025-05-23
feat: remove waffle flags for managing course outline sidebar

ihor-romaniuk:
**DEPR:** https://github.com/openedx/public-engineering/issues/316#issuecomment-2847839209

## Description

This PR removes support for the deprecated waffle flags:

- `courseware.enable_navigation_sidebar`
- `courseware.always_open_auxiliary_sidebar`

These flags have been replaced by [PluginSlots introduced in PR #1543](https://github.com/openedx/frontend-app-learning/pull/1543), available starting from the **Teak** release.

The default behavior is now:
- Navigation sidebar is shown by default
- Auxiliary sidebar is closed by default

All customization should now be handled via the **Frontend Plugin Framework** using `PluginSlot`.

This change simplifies the sidebar logic and reduces technical debt.

## Testing Instructions

1. Open a course in courseware view.
2. Confirm the following:
   - Navigation sidebar is visible.
   - Auxiliary sidebar is hidden unless triggered explicitly.
3. Ensure the behavior remains the same regardless of any waffle flag presence or value.
4. Validate that plugin-based sidebar overrides via `PluginSlot` still function correctly.

## Changelog

- Removed legacy waffle flags: `courseware.enable_navigation_sidebar`, `courseware.always_open_auxiliary_sidebar`.
- Unified sidebar behavior under plugin-based architecture.

PR #1923, opened 2026-06-09
fix: don't auto-open right sidebar by default

brian-smith-tcril:
## Summary

Restores the navigation (course outline) sidebar as the default when no panel is stored, adds a session-scoped flag to track user-initiated close state so it survives unit navigation and full-page remounts, and drops the now-subsumed narrower `hideCourseOutlineSidebar` flag. Aligns the test suite with the new behavior and adds Course-level integration coverage.

## Context

PR #1713 removed the `alwaysOpenAuxiliarySidebar` waffle flag guard around the right (auxiliary) sidebar's auto-open branch, but didn't invert the default — effectively hardcoding the flag to `true`. The PR description ("Auxiliary sidebar is closed by default") and the resulting code disagree.

Refs: https://github.com/openedx/wg-build-test-release/issues/574

### Default state & user-closed tracking
- **`useInitialSidebar`** — defaults to `COURSE_OUTLINE` on desktop when nothing is stored. Honors a stored RIGHT panel preference even before async widget data has loaded (avoids a transient flicker to `COURSE_OUTLINE` that downstream effects would otherwise write back to storage).
- **`storage.js` / `constants.js`** — new session-scoped `sidebarClosedByUser` flag. Set when `toggleSidebar(null)` runs, cleared on open.
- **`SidebarContextProvider`** — `toggleSidebar` sets/clears the flag based on `newSidebar === null`.
- **`useResponsiveBehavior`, `useUnitShiftBehavior`, `useSidebarSync`** — respect the flag so the user-closed state survives unit navigation and full-page remounts within the same tab session.

### Drop `hideCourseOutlineSidebar`
The previous `hideCourseOutlineSidebar` flag tracked only "user collapsed the outline" via the outline trigger. Because every writer of that flag also calls `toggleSidebar(null)` (which now writes the new closed-by-user flag), `outline-hidden=true ⇒ closed-by-user=true` always — making every consumer of the narrower flag reach via paths gated by closed-by-user. The narrower flag and its helpers (`isOutlineSidebarCollapsed` / `setOutlineSidebarCollapsed`) are removed; `ARCHITECTURE.md` and the course-outline README are updated.

### Unit-navigation behavior
- **`useUnitShiftBehavior` CASE 1** (COURSE_OUTLINE open) — keep it open instead of switching to a RIGHT panel. The browser behavior was "correct" only because `useSidebarSync` was reverting the switch via stale-memo cancellation; dropping the switching branch removes two redundant state updates / storage writes per unit nav.
- **`useUnitShiftBehavior` CASE 3** (`currentSidebar=null` on wide viewport) — recover to `COURSE_OUTLINE` (the default) instead of auto-opening a RIGHT panel.

### Tests
- New `describe('sidebar behavior')` in `Course.test.jsx` covering: default outline, closed-by-user persistence, stored discussions/outline on render, and the four click transitions (close outline; open outline closing discussions; open discussions closing outline; close discussions).
- Hook tests restructured by input state with explicit precondition wrappers; outline-collapsed coverage replaced with closed-by-user coverage.
- `storage.test.js` updated for the new helpers.
- `SidebarContextProvider.test.jsx` `UC7*` tests assert the new `setSidebarClosedByUser` calls.

## Test plan

- [x] On first visit (clean localStorage), navigation sidebar opens by default.
- [x] If discussions sidebar is open, advancing to next/prev unit keeps discussions open.
- [x] If navigation sidebar is open, advancing to next/prev unit keeps navigation open.
- [x] If user closes the sidebar (X button), advancing to next/prev unit keeps it closed.
- [x] If user closes the sidebar then triggers a full-page reload within the same tab, the sidebar stays closed.
- [x] Stored discussions preference survives a full-page reload (no transient flicker to navigation).
- [x] Opening any panel via a trigger clears the closed state.
- [x] Unit + integration tests pass: ``npm test -- src/courseware/course/Course.test.jsx src/courseware/course/sidebar/``.

### DIFF
--- a/src/courseware/course/Course.test.jsx
+++ b/src/courseware/course/Course.test.jsx
@@ -3,9 +3,10 @@ import React from 'react';
 import { Factory } from 'rosie';
 
 import { breakpoints } from '@openedx/paragon';
+import userEvent from '@testing-library/user-event';
 
 import {
-  fireEvent, getByRole, initializeTestStore, loadUnit, render, screen, waitFor,
+  cleanup, fireEvent, getByRole, initializeTestStore, loadUnit, render, screen, waitFor,
 } from '../../setupTest';
 import * as celebrationUtils from './celebration/utils';
 import { handleNextSectionCelebration } from './celebration';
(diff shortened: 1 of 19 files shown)


awais-ansari (approved):

arbrandes (approved):
Nice work!  I very much like the refactor.  Plus, it works as per the expected behavior in the https://github.com/openedx/wg-build-test-release/issues/574 description.  👍🏼

PR #1986, opened 2026-08-07
refactor: convert TabPage to TypeScript

brian-smith-tcril:
## Summary

Converts the shared `src/tab-page/TabPage.jsx` → `TabPage.tsx` as a **standalone, behavior-preserving** refactor. Part of the Redux → React Query migration (#1946), Phase 3 (course-home), **stacked on the CTA-toast conversion** (#1982, the branch below). Tracked by #1985.

Isolating the mechanical `.jsx → .tsx` conversion in its own layer keeps it a clean, reviewable diff — so later course-home changes to `TabPage` build on an already-TypeScript component instead of mixing a conversion together with feature changes.

Principle: a **pure conversion** — only the changes TypeScript actually forces. Structure, control flow, the string-status constants, and the destructured props signature are kept as-is, so the diff reads ~1:1.

## What changed

- **`TabPage.jsx` → `TabPage.tsx`:** typed `TabPageProps` (`courseId?`/`unitId?` optional, `courseStatus: StatusValue`, `children?: ReactNode`); structure and control flow unchanged.
- **Typed selectors:** the two `useSelector` reads use the store's exported `RootState` (`ReturnType<typeof store.getState>`), imported type-only — not a bespoke interface or a `Partial`.
- **The `&& courseId` guard** (the one behavior-adjacent change): `LoadedTabPage` requires `courseId: string`, but `TabPage`'s `courseId` is optional (`null` until the course loads). Rendering it under `&& courseId` narrows the type to `string` **without a cast and without loosening `LoadedTabPage`'s contract**. Behavior-identical in the app — `loaded`/`denied` always imply a `courseId`.
- **Status checks:** kept the `LOADING`/`LOADED`/`DENIED` constants, compared with `===` — not `[].includes`, which won't type-check against literal-typed constants without an `as StatusValue[]` cast.
- **Dependency in #1982 (below):** `ToastContent.message` is narrowed `ReactNode → string` there so Paragon's string-only `<Toast>` slot type-checks once `TabPage` is TypeScript. It lives in the toast PR because that's where `ToastContext` is introduced, and the mismatch only surfaces once `TabPage` is TS.

## Testing

[...]

PR #1982, opened 2026-08-06
refactor: convert course-home CTA toast from Redux to a React ToastProvider

brian-smith-tcril:
## Summary

Extracts the course-home **call-to-action toast** from Redux to a React context, and converts the two POSTs that feed it to React Query mutations. Part of the Redux → React Query migration (#1946), Phase 3 (course-home), **stacked on the courseware-search conversion** (#1970, the branch below). This is the prerequisite that lets the course-home tabs convert without a Redux toast bridge.

Behavior is preserved — same toast content, action link, auto-hide, and manual close, plus the same post-action data refresh — verified with the full test suite **and** a live manual smoke pass on both `master` and this branch (with a small fade-out improvement noted below).

## What changed

- **Toast state → React context:** new `ToastProvider`/`useToast` (`src/generic/ToastContext.tsx`), mounted once at the route root; `TabPage` renders the single Paragon `<Toast>` from it. Message and visibility are **decoupled** (`setToastContent` vs `openToast`/`closeToast`), removing the slice's `show={!!toastHeader}` content-vs-visibility fusion.
- **Writers → React Query mutations:** `useResetDeadlines` / `usePostEvent` in `src/course-home/data/apiHooks.ts`, following the `product-tours` mutation recipe. The hooks own only the POST + toast and are **Redux-free**; the transitional model-store refresh (`dispatch(getTabData)`) stays in each caller's `mutate` `onSuccess` until that data becomes an RQ query.
- **Redux removed:** the `resetDeadlines` + `processEvent` thunks (and the `resetDeadlines` re-export), and the `setCallToActionToast` reducer + `toastHeader`/`toastBodyText`/`toastBodyLink` fields from the course-home slice.
- **Courseware touch:** the `processEvent` postMessage parse/guard moves into `useIFrameBehavior` (`handlePostMessageEvent`), so `window.onmessage` is a bare reference; `eventTypes` is exported from `thunks.js`.
- **Bonus:** decoupling content from visibility removes a small glitch where the toast text blanked for a frame on auto-hide before fading.

## Testing

**Automated:** `npm run types`, `npm run lint`, `npm run build`, and the full `npm test` suite (106 suites, 896 passing) pass. New unit tests: `ToastContext.test.tsx` (incl. the decoupling) and `course-home/data/apiHooks.test.tsx` (axios-level POST → toast mapping + both `onError` paths). Own-wrapper suites that render `TabPage` (`DatesTab`, `DiscussionTab`, `CoursewareContainer`) get the provider.

[...]

Issue #1977, opened 2026-08-05
Dissolve the model-store normalized cache


brian-smith-tcril:
Part of #1946 — Redux → React Query migration (Stage 1).

**Goal:** delete the normalized `models` cache now that React Query is the cache.

**Tasks**
- After the feature conversions, audit remaining `state.models.*` selectors (`generic/model-store`, `src/data/hooks.ts` `useContextId`, …) and move any residual reads to query data.
- Provide a small replacement for `useContextId` (read the active `courseId` from route/params or a context) — it currently reads `courseware` / `courseHome` Redux state.
- Delete the `generic/model-store` slice and the `models` reducer.

**Verify:** no `state.models` references remain; full suite green.

---

**Note (size):** most `useModel` readers (204 sites / 82 files today) fall away as courseware converts (#1976 and its siblings, esp. the metadata/sequence layers, keep them alive via the model-store bridge). Reassess this issue's residual scope once the courseware decomposition lands — it may split (e.g. `useContextId` replacement vs. bridge removal). The `useContextId` replacement also overlaps with the courseware teardown (#1976); coordinate so it isn't done twice.

Issue #1975, opened 2026-08-05
Convert course-home tab data to React Query


brian-smith-tcril:
Part of #1946 — Redux → React Query migration (Stage 1).

**Goal:** all course-tab data fetching on React Query; remove the `courseHome` reducer.

Two prerequisites were tracked as their own sub-issues: courseware search (#1979) and the CTA toast → `ToastProvider` (#1980) — the shared `TabPage` had to be off Redux before the tabs converted.

**Done** (each its own sub-issue of #1946, stacked in #2062)
- The five tabs: dates #1984, outline #1991 (incl. `dismissWelcomeMessage`), progress #1998, live #2002, discussion #2003 — the `fetchTab` machinery and `TabContainer.jsx` orchestration are gone.
- `requestCert` → mutation (#1994).
- `fetchExamAttemptsData` → `useExamAttemptsData` + the plugin-facing `useExamsData()` (#2075 → PR #2077); split out because `courseHome.examsData` was a plugin-facing read (PR #1829), so its removal is a breaking change on its own.
- `course-home/data/thunks.js`, `redux.test.js`, `slice.test.js` deleted along the way; `api.test.js` and the Pact test (`course-home/data/pact-tests/lmsPact.test.jsx`) stay — they test `api.js`, which is unchanged.

**Remaining — one layer, closes this issue**
- The slice is down to one live field, `proctoringPanelStatus`: written by `ProctoringInfoPanel` when its `getProctoringInfoData` fetch settles, read by `OutlineTab` (defers the weekly-goal card) and `ProductTours` (holds the outline tour until the card exists). Four other initial-state fields (`courseStatus`, `courseId`, `errorMessage`, `errorCode`) are dead.
- Convert `getProctoringInfoData` to a `useProctoringInfoData` query; both readers read "settled" off the query's status (same key → one fetch); the panel derives its display state from `data` instead of six `useState`s.
- Delete `slice.js` and `course-home/data/index.js`; drop `courseHome` from `store.ts` and `setupTest.js`'s `initializeTestStore`.

**Verify:** `courseHome` reducer gone (`store.ts` = `models` + `specialExams` + `plugins`); outline tab still defers the weekly-goal card until the proctoring panel settles; outline-tab tour still waits for it; `git grep courseHome src/store.ts src/setupTest.js` empty; Pact test green.

> [!NOTE]
> The plan below was written by Claude (Claude Code) and reviewed before posting.

<details>
[...]

PR #1987, opened 2026-08-08
refactor: convert the dates tab to React Query

### LINKED ISSUE #1984: Convert the dates tab to React Query
Part of #1975 (course-home tab data → React Query), Phase 3 of the epic #1946.

## Summary

Convert the **dates** tab from Redux thunks + `model-store` to React Query, and **establish the per-tab conversion pattern** for the rest of course-home: each tab becomes self-wrapping — it renders `<TabPage>` itself and owns its data-loading via query hooks (the shape `CoursewareContainer` already uses) — and its wrapper line is removed from `index.jsx`. `TabContainer` is **not** modified; it keeps serving the unconverted tabs + course-exit and is deleted later (Phase 4, with course-exit).

Dates is the pattern-setter because its existing `DatesTab.test.jsx` (365 lines) mocks at the **axios layer** and is already wired with `QueryClientProvider` + a `/course/:courseId/dates` router, so it survives the thunk→RQ conversion with its assertions intact and acts as the behavior-parity guard.

## Tasks

- [ ] New `useCourseHomeMeta`, `useDatesTabData`, `useCourseHomeTab` query hooks (+ `src/course-home/data/queryKeys.ts`).
- [ ] `DatesTab` self-wraps `<TabPage>`; `courseId` via `useParams`; `courseStatus` derived from the queries.
- [ ] Dates subtree (`Timeline`, `Day`, `ShiftDatesAlert`, `UpgradeToShiftDatesAlert`) switches `courseId` from `useSelector(state.courseHome)` to `useParams`.
- [ ] `ShiftDatesAlert`'s `fetch` (thunk) prop → an `onReset` callback: dates invalidates the RQ query; outline keeps dispatching `fetchOutlineTab` (one-line call-site change, unchanged behavior).
- [ ] Remove the dates wrapper from `index.jsx`; delete the now-dead `fetchDatesTab` thunk + re-export.
- [ ] Transitional `model-store` bridges for `courseHomeMeta` and `dates` so the subtree keeps its `useModel(...)` reads unchanged (removed in Phase 5 / when `TabPage` reads meta from RQ).
- [ ] Preserve `DatesTab.test.jsx` teeth (only drop the `<TabContainer>` wrapper from its harness); add `apiHooks.test` cases (status matrix + both bridges); update the one `index.test` route assertion; keep `OutlineTab.test` green.

## Notes / decisions

- **Model-store stays a transitional read cache.** This PR moves *fetching* to React Query; the `courseHomeMeta`/`dates` bridges keep the existing readers working. Removing model-store for course-home is Phase 5.
- **`courseStatus` derivation** matches the thunk today: error → FAILED; fetching → LOADING; metadata `!courseAccess.hasAccess` → DENIED; else LOADED (`getDatesTabData` swallows 401/403, so access is authoritative from the metadata call).
- **403 failure detail:** on a rare `FAILED`, `TabPage` shows the generic failure message rather than the 403 `errorMessage`/`errorCode` (access errors go through DENIED). Accepted for now.

> [!NOTE]

<details>
<summary>Claude Plan — dates tab conversion</summary>

### Approach
One tab per PR. The tab component becomes self-wrapping (renders `<TabPage>` and owns its data via query hooks); `TabContainer` is untouched and only the converting tab's wrapper line leaves `index.jsx`. Stacked on the CTA-toast PR (#1982), which already took `TabPage` off Redux for toast.

### Query hooks (`src/course-home/data/apiHooks.ts`; keys in a new `queryKeys.ts`)
```ts
export const useCourseHomeMeta = (courseId) => useQuery({
  queryKey: courseHomeQueryKeys.metadata(courseId),
  queryFn: () => getCourseHomeCourseMetadata(courseId, 'outline'),
});

export const useDatesTabData = (courseId) => {              // query + transitional 'dates' bridge
  const query = useQuery({
    queryKey: courseHomeQueryKeys.datesTab(courseId),
    queryFn: () => getDatesTabData(courseId),
  });
  useEffect(() => {                                          // BRIDGE: subtree still reads useModel('dates')
    if (query.data) { dispatch(addModel({ modelType: 'dates', model: { id: courseId, ...query.data } })); }
  }, [query.data, courseId]);
  return query;
};

export const useCourseHomeTab = (courseId, tabQuery) => {   // generic: meta bridge + status
  const meta = useCourseHomeMeta(courseId);
  useEffect(() => {                                          // BRIDGE: TabPage/LoadedTabPage/subtree read useModel('courseHomeMeta')
    if (meta.data) { dispatch(addModel({ modelType: 'courseHomeMeta', model: { id: courseId, ...meta.data } })); }
  }, [meta.data, courseId]);

  let courseStatus = LOADED;
  if (meta.isError || tabQuery.isError) courseStatus = FAILED;
  else if (meta.isLoading || tabQuery.isLoading) courseStatus = LOADING;
  else if (!meta.data.courseAccess.hasAccess) courseStatus = DENIED;
  return { courseStatus };
};
```

### Self-wrapping DatesTab
```jsx
const DatesTab = () => {
  const dates = useDatesTabData(courseId);
  const { courseStatus } = useCourseHomeTab(courseId, dates);
  const queryClient = useQueryClient();
  const { isSelfPaced, org } = useModel('courseHomeMeta', courseId);  // bridged
  const { courseDateBlocks } = useModel('dates', courseId);           // bridged
  return (
    <TabPage activeTabSlug="dates" courseId={courseId} courseStatus={courseStatus} metadataModel="courseHomeMeta">
      {courseId && <OuterExamTimer courseId={courseId} />}
      {isSelfPaced && hasDeadlines && (
        <>
          <ShiftDatesAlert model="dates"
            onReset={() => queryClient.invalidateQueries({ queryKey: courseHomeQueryKeys.datesTab(courseId) })} />
          <SuggestedScheduleHeader />
          <BannerDatesUpgradeSlot courseId={courseId} logUpgradeLinkClick={logUpgradeLinkClick} />
          <UpgradeToShiftDatesAlert logUpgradeLinkClick={logUpgradeLinkClick} model="dates" />
        </>
      )}
      <Timeline />
    </TabPage>
  );
};
```

### Files
- **new** `src/course-home/data/queryKeys.ts` — `courseHomeQueryKeys` (`metadata`, `datesTab`).
- **edit** `src/course-home/data/apiHooks.ts` — `useCourseHomeMeta`, `useDatesTabData`, `useCourseHomeTab`.
- **edit** `src/course-home/dates-tab/DatesTab.jsx` — self-wrap; `useParams`; query hooks; `onReset` invalidate.
- **edit** `src/course-home/dates-tab/timeline/Timeline.jsx`, `timeline/Day.jsx` — `courseId` via `useParams`.
- **edit** `src/course-home/suggested-schedule-messaging/ShiftDatesAlert.jsx` — `useParams`; `fetch` prop → `onReset`; drop `useDispatch`.
- **edit** `src/course-home/suggested-schedule-messaging/UpgradeToShiftDatesAlert.jsx` — `courseId` via `useParams`.
- **edit** `src/course-home/outline-tab/OutlineTab.jsx` — `ShiftDatesAlert` `fetch` → `onReset` (still dispatches `fetchOutlineTab`).
- **edit** `src/index.jsx` — dates route → `<DatesTab />`; drop `fetchDatesTab` import.
- **edit** `src/course-home/data/thunks.js` + `index.js` — remove `fetchDatesTab` + re-export.

### Tests
- `DatesTab.test.jsx` — drop the `<TabContainer>` wrapper only; keep all assertions incl. "handles shift due dates click" (invalidate-driven refetch clears the banner; toast via `ToastProvider`).
- `apiHooks.test.tsx` — `useDatesTabData`/`useCourseHomeMeta` fetch, `useCourseHomeTab` status matrix, both bridges (`addModel`).
- `index.test.jsx` — dates route now renders the `Dates Tab` mock (was `Tab Container`).
- `OutlineTab.test.jsx` — green-check after the `ShiftDatesAlert` prop swap.

### Verification
`nvm use && npm run types && npm run lint && npm test` (targeted first: `DatesTab`, `apiHooks`, `OutlineTab`, `index`, `TabContainer`), then full suite + `npm run build`. `git grep fetchDatesTab` → gone. Manual: timeline + suggested-schedule alerts render; "Shift due dates" refreshes the banner + toast; loading/denied/failed via `TabPage`; other tabs unaffected.

</details>

brian-smith-tcril:
## Summary

Converts the **dates tab** from Redux thunks to React Query, per [OEP-0067 ADR-0010](https://docs.openedx.org/projects/openedx-proposals/en/latest/best-practices/oep-0067/decisions/frontend/0010-react-query.html). Part of the Redux → React Query migration (#1946), Phase 3 (course-home), **stacked on the TabPage TypeScript conversion** (#1986, itself on the CTA-toast conversion #1982). The dates tab becomes **self-wrapping** — it renders `TabPage` itself and owns its data loading via query hooks — the shape `CoursewareContainer` already uses.

Behavior is preserved — timeline, suggested-schedule alerts, access-denied redirects, and the "Shift due dates" refresh — verified with the full test suite **and** a live manual pass (the banner clears and dates shift via the RQ refetch, plus the toast).

## What changed

- **Self-wrapping `DatesTab`:** renders `<TabWithTimer>` (→ `TabPage`) and owns its data via `useCourseHomeMeta` + `useDatesTabData` (`course-home/data/apiHooks.ts`); `courseId` from `useParams`. Its wrapper line leaves `index.jsx`; `TabContainer` is untouched and keeps serving the unconverted tabs.
- **`TabPage` status derivation:** `courseStatus` becomes a union — `StatusValue | { metadataQuery, tabDataQuery }`. A converted tab passes its queries and `TabPage` derives the view (`deriveView` → loading/error/denied/loaded), reading access straight off the metadata query. Not-yet-converted (Redux) callers still pass a status string; that branch drops when courseware converts. Builds on the TypeScript conversion in #1986.
- **`TabWithTimer`:** a small wrapper rendering `TabPage` + `OuterExamTimer`, so the shared `TabPage` doesn't import `@edx/frontend-lib-special-exams` (which the Stage-2 frontend-base port is gated on). `TabContainer` renders `TabWithTimer`; `CoursewareContainer` keeps rendering plain `TabPage` (no timer, exactly as today).
- **`ShiftDatesAlert`** invalidates the dates query on reset — it owns the `resetDeadlines` mutation, so it owns invalidating the data that mutation affects. Its `fetch` prop is now **optional** (the still-Redux outline tab passes `fetchOutlineTab` as its transitional refresh; the dates tab passes nothing). The dates subtree (`Timeline`, `Day`, `ShiftDatesAlert`, `UpgradeToShiftDatesAlert`, `UpgradeToCompleteAlert`) reads `courseId` from `useParams`.
- **Transitional model-store bridge:** `data/queryKeys.ts` + `data/modelStoreBridge.ts` mirror query results into the existing `useModel` readers via a `QueryCache` `onSuccess` keyed off a `meta` tag, so the subtree keeps its `useModel` reads unchanged mid-migration (removed with model-store in Phase 5). `createTestQueryClient` wires the same bridge when given a store.
- **Redux removed:** the `fetchDatesTab` thunk and its `data/index.js` re-export.

## Testing

**Automated:** `npm run types`, `npm run lint`, and the full `npm test` suite (106 suites, 888 passing, 3 pre-existing skips) pass. `DatesTab.test.jsx` (the axios-level parity guard) keeps its assertions — only the `TabContainer` harness is dropped and it renders `<DatesTab />` directly, incl. "handles shift due dates click" (the `invalidate`-driven refetch clears the banner). `OutlineTab.test.jsx` gains a `/course/:courseId/home` route so the now-`useParams` shared alerts resolve, plus a "shift due dates click" test covering outline's transitional `dispatch(fetchOutlineTab)` refresh; `createTestQueryClient(store)` wires the model-store bridge for the subtree's `useModel` reads; the `redux.test.js` `fetchDatesTab` block is removed (its `fetchTab` coverage lives on via `fetchOutlineTab`).

**Manual (self-paced course with a seeded missed deadline):** the timeline and suggested-schedule alerts render, and clicking **"Shift due dates"** shows the success toast **and** clears the banner + shifts the dates on its own — the `invalidateQueries(datesTab)` → refetch → bridge → `missedDeadlines: false` path, matching the old thunk-dispatch behavior, with no console errors.

## Decisions

<details>

# Decisions — Redux → React Query: the dates tab (#1984)

#1946). **Not checked in** — referenced when opening the PR. First slice of
#1975 (course-home tab data), stacked on the CTA-toast conversion (#1982). This
is the **pattern-setter** for the remaining course-home tabs.

## Scope: one tab per PR, each tab self-wrapping

**Decision.** Convert one tab at a time. The tab component becomes
*self-wrapping* — it renders `<TabPage>` itself and owns its data-loading via
React Query hooks. `TabContainer` is **not** modified; the only `index.jsx`
change is removing the converting tab's wrapper line.

**Why.** Today `index.jsx` wraps each tab in a generic `TabContainer` that
(1) dispatches the tab's fetch thunk on mount and (2) reads `courseStatus`/
`courseId` from the slice. Under React Query the fetch-on-mount role
*evaporates* — a component that calls a query hook triggers the fetch itself —
so `TabContainer`'s remaining job is just deriving status and rendering
`<TabPage>`. `CoursewareContainer` already renders `<TabPage>` directly (no
`TabContainer`), so "the page owns its data and renders TabPage" is the
established courseware shape; the course-home tabs converge onto it.

**Alternatives rejected.**
- *An upfront "move every wrapper into its tab" setup PR.* It'd be a
  behavior-neutral refactor that also plants 5–6 copies of transitional Redux
  fetch-on-mount boilerplate, each rewritten later. The one thing it buys —
  future PRs not touching `index.jsx` — is a one-line route swap that belongs
  with each tab's own conversion anyway.
- *An injected `useTabData` seam on `TabContainer`.* Keeping `TabContainer` and
  parameterizing it by a hook adds coupling and a transitional Redux adapter for
  the courseware route; self-wrapping needs neither. `TabContainer` simply
  lingers for the unconverted routes and is deleted with course-exit in Phase 4.

## Lead tab: dates (after a detour through live)

**Decision.** Dates is the pattern-setter.

**How we got here.** We first picked dates ("one tab end-to-end"), then went
looking for a *cheaper* pattern-setter and explored **live**: it's the smallest
tab that still exercises a tab-data query (discussion has *no* tab-data fetch, so
it wouldn't prototype the `useXTabData` hook the heavier tabs need). But live
turns out to have **zero test coverage** — no `LiveTab.test.jsx`, and `index.test`
only mocks it as a string. Converting it would mean writing its safety net from
scratch, which is a weak pattern-setter and cuts against leaning on existing
tests. Dates, by contrast, has a 365-line `DatesTab.test.jsx` that mocks at the
**axios layer** and is already wired with `QueryClientProvider` + a
`/course/:courseId/dates` router — so it survives the thunk→RQ conversion with
its assertions intact and becomes the behavior-parity guard. So we came back to
dates; **live is deferred to a later "backfill the tests, then convert" PR.**

## Data bridge: a synchronous `QueryCache` sync into the model store

**Decision.** The dates render subtree (`DatesTab`, `Timeline`, `Day`,
`ShiftDatesAlert`, `UpgradeToShiftDatesAlert`, and `UpgradeToCompleteAlert` —
reached via `BannerDatesUpgradeSlot`) keeps its existing `useModel(...)`
reads unchanged. React Query is the fetch source and mirrors both `courseHomeMeta`
and the `dates` payload into `model-store` via a **global `QueryCache` `onSuccess`
callback** (in `data/modelStoreBridge.ts`, wired into the app `QueryClient` in
`index.jsx`): any query that tags itself `meta: { modelType, courseId }` gets a
`dispatch(addModel(...))` on success. The query hooks stay pure `useQuery` (one
`meta` line, no Redux), the components never touch the bridge, and `courseStatus`
is derived purely from the query state — final form, no model-store dependency.

**Why a `QueryCache` callback, not a per-hook `useEffect`.** The store must be
populated *before* the render that flips `courseStatus` to `LOADED`, or the shared
readers crash — `Timeline` does `courseDateBlocks.forEach(...)` and `LoadedTabPage`
does `tabs.filter(...)`, both unguarded. A per-hook `useEffect` bridge writes
*after* that render → a one-frame gap → crash. The cache callback fires as the
query resolves, before observers re-render, so `useModel` is populated in time.
(The old thunk had this invariant too — it called `addModel` *before* dispatching
`LOADED`; the cache callback restores it.)

**Why not gate `courseStatus` on the store instead.** Deriving `LOADED` from "is
the data in the store yet" also avoids the crash, but it invents a *new*
model-store dependency for loaded-state — the exact coupling we're removing.
Keeping `courseStatus` query-derived + a synchronous bridge fixes the timing
without that regression.

**Why `meta` + a global cache callback (the two reasons this is worth it).**
1. **It's fully transitional — once Redux / `model-store` is gone, the `meta`
   tags and the entire `QueryCache` config go with it** (Phase 5, #1977). It adds
   no permanent API surface; it exists only to keep `useModel` readers alive
   mid-migration.
2. **It keeps the migration clean and localized.** The subtree keeps its
   `useModel` reads (only `courseId` → `useParams`), and the shared
   `TabPage`/`LoadedTabPage`, the shared alerts, and the outline tab are all left
   untouched. The whole bridge is one file + a one-line `QueryClient` change, so
   each subsequent tab is a small PR too.

**Ecosystem note.** `meta` and `QueryCache`/`MutationCache` callbacks are core,
documented React Query (v4+; we're on v5.101), but `git grep` across the openedx
RQ adopters (`learner-dashboard`, `authn`, `authoring`, `course-authoring`) finds
**no** use of `meta` or global cache callbacks — they use RQ per-hook, and even
its canonical use (global error handling) is absent. So this is the first place to
introduce that class of RQ pattern here; it's chosen for the two reasons above,
not for local precedent. The precedent-matching alternative — prop-threading
`courseHomeMeta` into `TabPage`/`LoadedTabPage` and the banner data into the shared
alerts — was rejected because it balloons this PR into shared-component + outline
changes.

## `courseId` from `useParams`

**Decision.** The subtree reads `courseId` from `useParams()` rather than
`useSelector(state.courseHome)`.

**Why.** Once the dates route stops dispatching its thunk, the slice no longer
carries the loaded `courseId` for this route. `useParams` is the source the
already-loaded page has (decoded via `DecodePageRoute`, as `CoursewareSearch`
relies on today). The existing test already renders through a
`/course/:courseId/dates` route, so this is covered without new test files.

## Reset-deadlines refresh: `ShiftDatesAlert` invalidates its own queries

**Decision.** `ShiftDatesAlert` owns the `resetDeadlines` mutation, so it also owns
the refresh: on success it **invalidates the dates query itself**
(`queryClient.invalidateQueries(datesTab)`). Its `fetch` prop is kept but made
**optional and transitional** — the still-Redux outline tab passes
`fetch={fetchOutlineTab}` so its model-store data is refetched too; the dates tab
passes nothing.

**Why.** The component doing the write is the right place to invalidate the queries
that write affects — the canonical React Query pattern, not coupling (the alert
already owns the deadline data it's mutating). Shifting deadlines invalidates the
dates data *regardless of which tab triggered it*, so invalidating the dates query
even from the outline tab is correct (it refetches when dates is next viewed). This
also lets `DatesTab` pass **no** refresh prop and leaves `OutlineTab`'s call site
**unchanged**. (The toast still fires from inside `useResetDeadlines`, unchanged from
#1982.)

**Rejected: a refresh callback the tab supplies** (`onReset`/`fetch`). It pushes the
"invalidate the query for this mutation" responsibility onto callers that don't own
the mutation, keeps the shared alert ignorant of data it's literally changing, and
read oddly at the call site (`onReset` for a "Shift due dates" button).

**End state.** When outline converts to RQ, `ShiftDatesAlert` invalidates the outline
query too (invalidate all affected queries), and the `fetch` prop + `useDispatch` are
deleted — the alert goes fully Redux-free.

## `OuterExamTimer`: a `TabWithTimer` wrapper, not a `TabPage` prop

**Decision.** The proctored-exam `OuterExamTimer` — which `TabContainer` renders on
every tab it wraps — moves into a small `TabWithTimer` component (`TabPage` + the
timer as its first child). `TabContainer` and the self-wrapping tabs (`DatesTab`)
render `TabWithTimer`; `CoursewareContainer` keeps rendering plain `TabPage`.

**Why not a `withTimer` prop on `TabPage`.** `OuterExamTimer` comes from
`@edx/frontend-lib-special-exams`, and the Stage-2 frontend-base port (#1905) is gated
on that library. A prop makes the shared `TabPage` import special-exams directly; a
wrapper keeps that dependency out of the component that has to port cleanly, and reads
as composition ("TabPage plus the timer") rather than a boolean that conditionally
injects one specific external child. Rendering it per-tab was also rejected — it's
cross-cutting, and duplicating the line into every self-wrapping tab scatters it.

**Behavior is preserved exactly.** Today `OuterExamTimer` renders only in
`TabContainer`; `CoursewareContainer` (the in-unit courseware view) has none. Moving
it into `TabWithTimer` keeps that split — every `TabContainer` tab and converted tab
gets it, `CoursewareContainer` (plain `TabPage`) still doesn't. Hoisting it into
`TabPage` unconditionally would have newly rendered it on courseware.

## `apiHooks.ts` stays pure; `TabPage` owns `courseStatus` derivation

`apiHooks.ts` holds only thin `useQuery`/`useMutation` wrappers around `api.js`
(each query hook adds a single `meta: { modelType, courseId }` tag so the
`QueryCache` bridge mirrors it — no Redux, no status constants in the data layer).
This follows the Phase-0 recommendations precedent (#1967): the data layer stays a
bare `useQuery`, and status handling lives with the *consumer* (there,
`CourseRecommendations.jsx` branches on `isPending`/`isError`; `track.js` maps
`isError ? FAILED : LOADED`).

**`courseStatus` is derived in `TabPage`, via a union-typed prop.** `TabPage` is
the component that renders from the status, so it owns the derivation:

```ts
type CourseStatus = StatusValue | {
  metadataQuery: UseQueryResult<{ courseAccess?: { hasAccess: boolean } }>;
  tabDataQuery: UseQueryResult;
};
```

A converted tab hands `TabPage` its two queries
(`courseStatus={{ metadataQuery, tabDataQuery }}`); a not-yet-converted (Redux) caller
(`TabContainer`, courseware) still passes a plain status **string**.
`deriveView(courseStatus)` normalizes either input into three booleans —
`{ isLoading, isError, isDenied }` — and the render tree branches on those. Access is
read straight off the metadata query (`metadataQuery.data?.courseAccess?.hasAccess`), so
`deriveView` doesn't depend on the model-store bridge — the bridge's timing only still
matters for the *subtree's* `useModel` reads. The query's data is typed **inline** to
just the field this file reads: the metadata is untyped JS, and a named one-field
`CourseMetadata` interface would misrepresent the real shape (and rot as a stub). Named
`metadataQuery`/`tabDataQuery`, not `meta` — which collides with React Query's own `meta`.

**Booleans, not a status constant (don't re-manufacture the Redux vocabulary).**
`TabPage` never passes the status to a child — it only branches on it to decide what to
render. So the query path derives the render booleans directly and never mentions the
`LOADING`/`LOADED`/`DENIED` constants; those appear in exactly one place — the
transitional `typeof courseStatus === 'string'` branch of `deriveView`, which maps a
legacy Redux status string onto the same booleans. That branch, the `StatusValue` union
member, and the `constants` import are deleted together when courseware (the last string
caller) converts, leaving `deriveView` purely query-native.

**Ordering: access is resolved before tab-data (matches the thunk's short-circuit).**
The metadata call is authoritative for access, so `deriveView` resolves the metadata
phase — metadata error → failed; metadata in-flight → loading; `!hasAccess` → denied —
*before* it looks at `tabData` at all. This mirrors the `fetchTab` thunk, which denied
*"regardless of the tabDataResult"* (thunks.js). It matters in two cases a combined
`metadataQuery.isError || tabDataQuery.isError` ordering gets wrong: (1) **no access + a non-auth
`tabData` error** (e.g. a 500 while metadata says `hasAccess: false`) — deny, don't
show failed; and (2) **no access while `tabData` is still loading** — deny immediately
rather than flashing loading first. (`getDatesTabData` swallows 401/403 → `{}`, so the
*common* no-access path never errors on tab-data anyway; this ordering covers the
residual cases.) Keep the two phases separate — don't recombine the guards.

**Render tree: a redirect gate, then a render function per part.** `isDenied` drives
the access-denied redirect (`getAccessDeniedRedirectUrl` returns no URL for enroll/auth
errors on the **outline** tab, so that case falls through and renders the page — where
the outline shows its enroll/upgrade CTAs). The page content keys off
`shouldRenderContent = !isLoading && !isError` (loaded, or denied-without-redirect).
Each part is a render function — `renderToast`, `renderTourButton`, `renderLoading`,
`renderLoadedTabPage`, `renderError` — and the JSX gates each call with the relevant
boolean (`shouldRenderContent` / `isLoading` / `isError`), so the return reads as the
render structure top-to-bottom and the primary conditions aren't buried inside the
functions. The functions are called in their original positions, so DOM order is
unchanged; the two with a *secondary* condition keep it internally (`renderTourButton`
checks `metadataModel`, `renderLoadedTabPage` narrows `courseId`). Loading / content /
error are mutually exclusive by construction, so at most one renders.

**Why `TabPage` reads `courseAccess` now (it didn't in the string path).** In the
Redux flow this derivation never lived in `TabPage`: the `fetchTab` thunk
(`data/thunks.js`) computed the status — metadata rejected → `FAILED`;
`!courseAccess.hasAccess` → `fetchTabDenied` → `DENIED`; tab-data rejected →
`FAILED`; else `LOADED` — and stored the *resolved string* in the slice, which
`TabContainer` passed down. So `TabPage` only ever rendered a finished status and
never inspected `courseAccess` for it (it read `courseAccess` **solely** for the
access-denied redirect URL). The RQ path hands `TabPage` the raw queries instead of
a pre-resolved string, so `deriveView` is exactly that thunk branch relocated
into `TabPage` — which is why the `courseAccess.hasAccess` → `DENIED` check now
appears here. The value is the same one `TabPage` already reads from
`useModel('courseHomeMeta')`, so no new data source is introduced.

**Why in `TabPage`, not elsewhere.** Tried and rejected, in order:
- **an inline block in each tab** — a four-branch computation wedged into the
  component body reads badly and would be copy-pasted across five tabs;
- **a standalone `courseTabStatus.ts` helper** — a whole file for one derivation
  felt like overkill;
- **a `useCourseHomeTab` composing hook** — drags the status constants (and, in an
  earlier design, the bridge) back into the data layer — the clutter we'd just
  separated out.

Putting it in `TabPage` writes the logic once for every tab, keeps each tab body to
`courseStatus={{ metadataQuery, tabDataQuery }}`, and adds no new file.

**`TabPage` is already TypeScript (#1986).** The `.jsx → .tsx` conversion landed in the
lower PR (#1986); this PR only adds the `CourseStatus` union member and `deriveView` on
top of it. `TabPage` is shared and now goes **dual-mode** during the migration
(string | queries); the string branch is removed once courseware — the last string
caller — converts. `getAccessDeniedRedirectUrl` and `useModel` are untyped JS, so adding
the union doesn't ripple into them.

## Accepted trade-off: 403 detail on a hard failure

Today `fetchTabFailure` stashes a 403 body's `errorMessage`/`errorCode` into the
slice and `TabPage` renders it. On the RQ path `TabPage` falls back to the
generic failure message for the dates route. Access errors normally surface via
`DENIED` (from metadata), so this only affects rare unexpected failures —
accepted for this PR rather than threading the detail through.

## Tests (preserve the teeth)

- **`DatesTab.test.jsx`** is the parity guard. Because it mocks at the **axios
  layer** and already renders through `QueryClientProvider` + a
  `/course/:courseId/dates` route, the only harness change is dropping the
  `<TabContainer>` wrapper (render `<DatesTab/>` directly) and removing the
  now-unused `fetchDatesTab`/`TabContainer` imports. **All assertions stay
  verbatim** — including "handles shift due dates click," where the
  `invalidate`-driven refetch picks up the swapped `missedDeadlines: false` mock
  and the banner clears. Adjust `await` timing only if needed; do **not** weaken
  assertions.
- **`course-home/data/apiHooks.test.tsx`** — cover `useCourseHomeMeta`/
  `useDatesTabData` fetch (axios MockAdapter): the hooks return the mapped data.
  Mirroring into the model store is the `QueryCache` bridge's job (in
  `modelStoreBridge.ts`), **not** the hooks', so it's proven end-to-end by
  `DatesTab.test.jsx` (the subtree only renders real content if the bridge
  populated `useModel`); a focused `modelStoreBridge` test for the
  `meta` → `addModel` mapping is optional.
- **`TabPage.test`** — the existing string-`courseStatus` cases stay valid (the
  union still accepts a string). Add the query/combo path: `deriveView` producing
  `isLoading`/`isError`/`isDenied` from `{ meta, tabData }` + `courseAccess` —
  this is where the status matrix has its teeth (also exercised end-to-end via
  `DatesTab.test`).
- **`index.test.jsx`** — the dates route now renders the `Dates Tab` mock instead
  of the `Tab Container` mock; update that one assertion (still has teeth: it
  asserts the dates path renders `DatesTab` directly).
- **`OutlineTab.test.jsx`** — its `fetchAndRender` now wraps `OutlineTab` in a
  `/course/:courseId/home` route so the shared alerts' `useParams` resolves, and a
  new "shift due dates click" test covers outline's transitional
  `dispatch(fetchOutlineTab)` refresh — the one line the dates path no longer
  exercises (dates invalidates its query instead).

## Verification

`nvm use && npm run types && npm run lint && npm test && npm run build` — `types`
matters more now that `TabPage` is TypeScript. Targeted first (`DatesTab`,
`TabPage`, `apiHooks`, `OutlineTab`, `index`, and `TabContainer` still green), then
the full suite + build. `git grep "fetchDatesTab"` comes back clean (gone from
source), `TabPage.jsx` is gone (only `TabPage.tsx`), and the dates subtree no
longer reads `useSelector(state.courseHome)` for `courseId`.

## Manual testing

The dates tab renders against mocked queries in the suite; for a live sanity pass,
the "Shift due dates" banner needs a self-paced course with a missed
suggested-schedule deadline (seeded via `edx-when` per the toast PR's notes, since
the authoring MFE can't set relative due dates). Compare master vs. branch:
timeline + suggested-schedule alerts render identically, and "Shift due dates"
refreshes the banner (now via `invalidateQueries` → refetch → bridge) and shows
the toast. Other tabs are untouched.

</details>

Closes #1984


src/tab-page/TabPage.tsx
```diff
@@ -21,15 +22,51 @@ import LoadedTabPage from './LoadedTabPage';
...
+    return { ...view, isError: true };
+  }
+
+  // Access is read from the metadata query, resolved before tabData is considered.
+  const { metadataQuery, tabDataQuery } = courseStatus;
+  if (metadataQuery.isError) { return { ...view, isError: true }; }
+  if (metadataQuery.isLoading) { return { ...view, isLoading: true }; }
+  if (!metadataQuery.data?.courseAccess?.hasAccess) { return { ...view, isDenied: true }; }
+  if (tabDataQuery.isError) { return { ...view, isError: true }; }
+  if (tabDataQuery.isLoading) { return { ...view, isLoading: true }; }
```
arbrandes: Suggestion: maybe use `isPending` instead.

React Query's `isLoading` is `false` for a query that has no data and is not currently fetching. `isPending` is the check that means "no data yet".

arbrandes (approved):
Approved, with one suggestion.

brian-smith-tcril:

🤖 _Comment via Claude Code._

src/tab-page/TabPage.tsx
brian-smith-tcril: Good catch! Addressed in https://github.com/openedx/frontend-app-learning/compare/d70ead96d5c27e4b187c1ff2ece70d84972ba21f..0232601f58397469f15786bbd0b9b78d49478938

brian-smith-tcril:
> 

addressed in https://github.com/openedx/frontend-app-learning/compare/0232601f58397469f15786bbd0b9b78d49478938..db60f162b1c6e5a25b36968f0f669f53feebdbd1

brian-smith-tcril:
🤖 _Decision notes via Claude Code._

<details>
<summary>Decision log — <code>isPending</code> review fix + #2022 query error logging</summary>

# Decisions — dates tab → React Query (#1987), + review fix + query error logging (#2022)

Working notes for the #1987 branch (part of the Redux → React Query migration, #1946).
Originally just the dates-tab conversion; now also carries a review fix and the query
error-logging infra (#2022), because #1987 is the lowest open layer.

---

## `isPending` over `isLoading` in `TabPage.deriveView` (review fix)

Review on #1987 (arbrandes) suggested `isPending` instead of `isLoading` for the query-status
gates in `deriveView`. This is a correctness fix, not just style:

- v5 status is one of `pending | error | success`. `isPending` = "no data yet". `isLoading`
  is a *conjunction*: `isPending && isFetching` — so it's **false** for a pending query that
  isn't actively fetching (a **disabled** `enabled: false` query, or a **paused**/offline one).
- With `isLoading`, such a pending-but-not-fetching query falls through the loading gate to the
  `!data?.courseAccess?.hasAccess` check → `data` is `undefined` → treated as **denied**
  (a redirect). That's wrong for a query that simply hasn't loaded.
- This matters directly for the `enabled: !!courseId` guards used elsewhere in the migration:
  `isLoading` + `enabled` are contradictory (the guard means "don't fetch yet", which
  `isLoading` doesn't catch); `isPending` is the correct partner.
- Framed positively: `isPending` makes `deriveView` a clean state machine — per query,
  `error → error`, `pending → loading`, else (`success`, data present) → proceed.

Applied to both the metadata and tab-data gates. Test mocks updated `isLoading: true` →
`isPending: true`.

**Rebase conflict (discussion tab).** The discussion-tab layer made `tabDataQuery?` optional
(it passes no tab-data query) and added optional chaining. Resolved by keeping **both**: the
optional chaining *and* `isPending` → `tabDataQuery?.isPending`.

---

## Query error logging via a global `QueryCache.onError` (#2022)

### Why here, and why not per-hook

Converting a thunk to a `useQuery` drops the `logError` the thunk did on failure. v5 removed
`onError`/`onSuccess` from `useQuery` (they live only on `useMutation` and the `QueryCache`),
so there's no per-hook place for it. The v5 home for cross-cutting query error logging is the
global `QueryCache.onError`. This lands in **#1987** because it's the first open layer that
removed any logging (nothing below it did), so a layer below would be infra with nothing to log.

### Where it lives: a permanent `src/queryClient.ts` (not the bridge)

- New `src/queryClient.ts` builds the app QueryCache with `onError: logError` and
  `onSuccess: bridgeToModelStore(store, …)`, plus `createQueryClient(store)`.
- `modelStoreBridge.ts` is reduced to just `bridgeToModelStore(store, data, query)` (the
  onSuccess body, a plain function). Nothing we intend to keep lives in the bridge — it's named
  for model-store bridging and is deleted with the model store (#1977).
- The `store` param is **transitional**: it only feeds the bridge. Once the bridge is gone, the
  cache — and `createQueryClient` — no longer need a store.
- `index.jsx` and `setupTest`'s `createTestQueryClient` use the new module (`onError` logging is
  therefore active in store-backed tests).

### Scope: plain `logError` is enough here; the 403-as-info nuance is deferred

All course-home tab getters (`getDatesTabData`, `getProgressTabData`, `getOutlineTabData`)
**swallow 401/403/404 in the API layer** (return `{}`) — the expected access-denied cases are
handled via course metadata, not logging. So their *query* errors are only genuine failures
(500/network), and a plain `logError` restores exactly what the thunks logged.

The `403 → logInfo` (New Relic page-action, not a noticed error) nuance that `fetchCourse` had
is only needed by a query that surfaces a 403 *as an error* — the courseware outline
(`getLearningSequencesOutline`, which throws 403). That's a **#2010** concern, so the
meta-driven "quiet statuses" mechanism is deferred to that layer, not built speculatively here.

### Typing note

`bridgeToModelStore`'s `query` param is `Query<unknown, unknown>` — matching the `TError = unknown`
that the QueryCache `onSuccess` callback provides. (Bare `Query` defaults `TError` to `Error`,
which mismatched once the handler was extracted from the inline callback and lost contextual
typing.)

</details>

brian-smith-tcril:
@arbrandes this has changed a little bit since you approved it, and I'd like to make sure you're still happy with it before landing it. It should be decently easy to see what has changed using the diff links I put in earlier comments.

src/queryClient.ts
```diff
@@ -0,0 +1,16 @@
+import { logError } from '@edx/frontend-platform/logging';
+import { QueryCache, QueryClient } from '@tanstack/react-query';
+import { Store } from 'redux';
+
+import { bridgeToModelStore } from './course-home/data/modelStoreBridge';
```
arbrandes: It seems `modelStoreBridge` isn't really course-home-specific anymore.  Having a root-level `queryClient.ts` reach down into a feature directory for it feels a little backwards.

Since both files are new here, maybe colocate them next to `addModel` in `src/generic/model-store/`, or just beside `queryClient.ts`.

Not blocking.

arbrandes (approved):

src/queryClient.ts
brian-smith-tcril: moved to `/src/data` in https://github.com/openedx/frontend-app-learning/compare/db60f162b1c6e5a25b36968f0f669f53feebdbd1..0e034bb1573d604fa4953954b08f798bbc5987d9

arbrandes: Still approved!

PR #1995, opened 2026-08-12
refactor: convert the requestCert writer to a React Query mutation

brian-smith-tcril:
## Summary

Convert the shared `requestCert` POST thunk to a `useRequestCert` React Query mutation and delete the thunk. Part of the Redux → React Query migration (#1946). Prerequisite **below the outline-tab conversion (#1991)** — outline's `CertificateStatusAlert` is one of this thunk's three callers, so converting `requestCert` first lets outline build on an already-converted base. Closes #1994.

`requestCert` is a fire-and-forget "request certificate" POST (`postRequestCert` → `/courses/{courseId}/generate_user_cert`) shared by three tabs, so all three callers convert together — the only way to *delete* the thunk rather than leave a redundant POST path.

## What changed

- **`apiHooks.ts`** — add `useRequestCert` (`mutationFn: ({ courseId }) => postRequestCert(courseId)`, `onError: logError`), mirroring the other course-home mutations.
- **Convert the three callers** from `dispatch(requestCert(courseId))` to `useRequestCert().mutate({ courseId })`, dropping each file's `useDispatch` + `requestCert` thunk import:
  - `course-home/outline-tab/alerts/certificate-status-alert/CertificateStatusAlert.jsx`
  - `courseware/course/course-exit/CourseCelebration.jsx`
  - `course-home/progress-tab/certificate-status/CertificateStatus.jsx`
- **`data/thunks.js`** — delete the `requestCert` thunk and its now-unused `postRequestCert` import. (`requestCert` isn't re-exported from `data/index.js`, so no re-export change.)

## Behavior

No user-facing change. The thunk and the mutation both call the same `postRequestCert(courseId)` → identical fire-and-forget POST, no response handling. The only delta is `onError: logError` (the thunk left failures unhandled) — a strict improvement.

## Testing

`npm run types`, `npm run lint`, and the full `npm test` suite (106 suites, 3 pre-existing skips) pass. New `useRequestCert` coverage in `apiHooks.test.tsx` (asserts the POST + error logging). All three callers' click paths are now exercised: `ProgressTab.test.jsx` and `CourseExit.test.jsx` click the request-certificate button and assert the `generate_user_cert` POST fires; `OutlineTab.test.jsx` already clicks it. (The `CourseExit` click test was added here — its button previously rendered but was never clicked, leaving the `mutate` line uncovered.)

## Decisions

<details>
<summary>Full decision log</summary>

# Decisions — convert the requestCert writer to a React Query mutation (#1994)

[...]

Issue #1978, opened 2026-08-05
Reduce the Redux store to the external specialExams reducer


brian-smith-tcril:
Part of #1946 — Redux → React Query migration (Stage 1). Partial teardown.

**Goal:** shrink the Redux store to the last external dependency — `specialExams`.

**Tasks**
- Reduce `store.ts` to only `specialExams` (external, from `@edx/frontend-lib-special-exams`). All *local* reducers leave first, including the `plugins` slice (**#2017** — its de-Redux is Stage-1, not a Stage-2 concern).
- This `specialExams`-only endpoint needs **no external dependency** — it depends only on learning's own remaining conversions (tabs, courseware, model-store #1977, plugins #2017).
- **Final dep removal is separate and library-gated:** dropping `react-redux` / `redux` / `@reduxjs/toolkit`, deleting `store.ts`, and removing `AppProvider store=` requires the special-exams components to no longer need a host-supplied react-redux `Provider` + store. The approach for that (self-contained library store vs. React Query conversion) is being investigated in **#2018**; a concrete `frontend-lib-special-exams` change is filed only if/when that investigation says so.

**Verify:** `store.ts` registers only `specialExams`; `git grep react-redux` in learning's own source returns only special-exams-adjacent files; exams still function (timer, gating, access token).

Issue #1973, opened 2026-08-05
Convert remaining small / api-only features to React Query


brian-smith-tcril:
Part of #1946 — Redux → React Query migration (Stage 1).

**Goal:** convert the remaining small, self-contained features to React Query — a catch-all for conversions not owned by another sub-issue.

**Scope note:** the features currently here use **no Redux**, so converting them removes no reducer — it is an optional React Query caching/dedupe improvement, not de-Redux work. Keep this issue scoped to whatever isn't covered elsewhere; prune as items move.

**Current contents**
- `courseware/course/celebration` (single POST)
- `alerts/enrollment-alert` (api-only)
- `preferences-unsubscribe/data` (query + mutation)
- `instructor-toolbar/masquerade-widget/data` (query + mutation; its `data/` is already just `api.ts`)

**Explicitly not here**
- `recommendations` — done in the infrastructure / pattern-setter issue.
- `bookmark` — moves to the courseware/data issue (its `bookmarked` state lives in the shared courseware `units` model read by `UnitTitleSlot` / `UnitButton`, so it can't be cut cleanly until then).

**Verify:** each converted feature works against mocked queries; no new Redux introduced.

brian-smith-tcril:
> [!NOTE]
> This comment was authored by Claude (Claude Code) and reviewed before posting.

## Note for the `courseware/course/celebration` item, from #2086

#2086 (the last `courseHomeMeta` layer of the #1977 model-store dissolution) moves the two celebration writers off the model store and onto the query cache as a literal port: `dispatch(updateModel({ modelType: 'courseHomeMeta', … }))` becomes `queryClient.setQueryData(courseHomeQueryKeys.metadata(courseId), (meta) => ({ ...meta, celebrations: { ...celebrations, <flag> } }))`, with every signature kept (`dispatch` → `queryClient` in the same slot) and `postCelebrationComplete` untouched as a fire-and-forget POST outside React Query. Doing the cache write by hand beside the POST was accepted as the minimal change for #2086 on the understanding that the real shape — a `useMutation` whose callback does the write — is this issue's.

Two things for that conversion to know:

[...]

PR #1997, opened 2026-08-14
refactor: convert outline tab from Redux to React Query

### LINKED ISSUE #1991: Convert the outline tab to React Query

brian-smith-tcril:
## Summary

Convert the course-home **outline tab** from Redux to React Query. Part of the Redux → React Query migration (#1946), and the next course-home tab through the door the dates-tab conversion opened. Closes #1991.

`OutlineTab` becomes self-wrapping — it owns its data (`useCourseHomeMeta` + a new `useOutlineTabData`) and renders `TabWithTimer` itself, so the `index.jsx` route drops its `<TabContainer tab="outline" fetch={fetchOutlineTab}>` wrapper for a bare `<OutlineTab />` and `fetchOutlineTab` is deleted.

## What changed

- **`apiHooks.ts` / `queryKeys.ts`** — add `useOutlineTabData` (tagged `meta: { modelType: 'outline', courseId }` so the model-store bridge still populates the subtree's `useModel('outline')` reads) and the `outlineTab(courseId)` key.
- **`OutlineTab.jsx`** — split into a thin data-owning `OutlineTab` (runs the two queries, renders `<TabWithTimer courseStatus={{ metadataQuery, tabDataQuery }}><OutlineTabContent/></TabWithTimer>`) and an `OutlineTabContent` in the same file. The split keeps the `useModel('outline')` destructure out of the loading phase, where the model is still `{}`. `courseId` from `useParams`; `proctoringPanelStatus` stays on `useSelector`.
- **Subtree `courseId` → `useParams`** — the seven slice-reading widgets (`DateSummary`, `CourseDates`, `CourseHandouts`, `CourseTools`, `StartOrResumeCourseCard`, `WeeklyLearningGoalCard`, `ProctoringInfoPanel`) and the three `section-outline` components that reached the slice via `useContextId` (`Section`, `SequenceDueDate`, `SequenceTitle`). Required: the route no longer runs a thunk to set `state.courseHome.courseId`.
- **`ShiftDatesAlert.jsx`** — `refreshTabData` now invalidates **both** `datesTab` and `outlineTab`; drops its `fetch`/`useDispatch` (outline was the last `fetch=` caller).
- **Outline-only POST writers → mutations** — `dismissWelcomeMessage` → `useDismissWelcomeMessage` (thunk deleted) and `saveWeeklyLearningGoal` → `useSaveWeeklyLearningGoal` (moved out of `thunks.js`, where it was a loose non-thunk helper).

## Behavior

No user-facing change. `proctoringPanelStatus` deliberately stays in Redux (client UI state, out of scope for a data-layer conversion). The mutations add `onError: logError`, which the thunks lacked — a strict improvement.

## Testing

`npm run types`, `npm run lint`, and the full `npm test` suite (106 suites, 3 pre-existing skips) pass. `OutlineTab.test.jsx` and `ProductTours.test.jsx` now render the real self-wrapping tab through the bridged query client; new mutation coverage (`useOutlineTabData`, `useDismissWelcomeMessage`, `useSaveWeeklyLearningGoal`) in `apiHooks.test.tsx`. Manual browser verification and the test-coverage split are documented in the decision log below.

## Decisions

<details>

# Decisions — Redux → React Query: the outline tab (#1991)

(course-home tab data), stacked on the dates-tab conversion (#1984), which set
the self-wrapping pattern.

## Scope: same self-wrapping conversion the dates tab established

**Decision.** `OutlineTab` becomes self-wrapping — it renders `<TabWithTimer>`
itself and owns its data via `useCourseHomeMeta` + a new `useOutlineTabData`
hook. `courseId` comes from `useParams`. The `index.jsx` route drops its
`<TabContainer tab="outline" fetch={fetchOutlineTab}>` wrapper for a bare
`<OutlineTab />`, and `fetchOutlineTab` is deleted.

**Why.** This is the pattern the dates tab set (#1984), which itself converges
on the shape `CoursewareContainer` already uses. Nothing tab-specific here —
the outline tab is just the next tab through the same door.

## `OutlineTab` splits into a thin wrapper + `OutlineTabContent`

**Decision.** `OutlineTab` is now just the data-owning wrapper: it runs the two
queries and renders `<TabWithTimer courseStatus={{ metadataQuery, tabDataQuery }}>
<OutlineTabContent /></TabWithTimer>`. The heavy body — everything that reads
`useModel('outline', courseId)` — moves into an `OutlineTabContent` component in
the same file.

**Why.** Self-wrapping means `OutlineTab` renders *before* its data resolves (the
spinner phase lives inside `TabWithTimer`). The body destructures the `outline`
model (`const { courseBlocks, … } = useModel('outline', courseId)`), and
`useModel` returns `{}` for a missing model, so that destructure ran against `{}`
during loading and crashed. `TabPage` only renders its children once the data has
loaded, so moving the body into a child (`OutlineTabContent`) means the
destructure never runs empty — no defensive defaults, no guard, and the wrapper
stays a thin data-owner. Same file, so it reads as one unit.

## The `outline` model stays on the transitional bridge — it isn't outline-only

**Decision.** Keep every `useModel('outline'|'courseHomeMeta', …)` read in the
subtree exactly as-is, populated by the model-store bridge (the `meta:
{ modelType, courseId }` tag on the query drives `addModel` on success). Do
**not** convert any of these reads to consume `useOutlineTabData`'s data
directly in this PR. Dissolving the model store is its own coordinated epic
step (#1977).

**Why — the mechanism.** We looked into who actually reads these models, to see
if any were outline-tab-only and could be pulled forward:
- `courseHomeMeta` is read all over the app (courseware, every tab, dozens of
  shared alerts, `TabPage`) — plainly shared.
- `outline` looked outline-scoped — every reader is under `outline-tab/`
  **except two**: `useEnrollmentAlert` and `useLogistrationAlert`. Those run
  inside `LoadedTabPage`, which renders on **every** tab and courseware, and
  they read `useModel('outline', courseId)` **passively and defensively**
  (`outline && outline.courseBlocks && …`) — on non-outline pages the model is
  empty and the alert stays hidden.

The distinction that matters: **`useModel` is a passive read** (returns whatever
is cached, fetches nothing); **`useOutlineTabData(courseId)` is an active query**
(calling it *triggers* the outline fetch). Swapping those shared alert hooks to
the query would fire the outline API on the dates tab, the progress tab, and
courseware — a real regression. The passive-read-plus-bridge is exactly what
makes the shared read safe today, which is why untangling it is a coordinated
store dissolution (#1977), not per-tab work.

**Alternative rejected.** *Convert just the outline-tab-only readers to
`useOutlineTabData` (child calls dedupe against `OutlineTab`'s query, so no
extra fetch), leaving the bridge for the two shared alerts.* It can't remove the
bridge or the `meta` tag (the shared alerts still need the model store
populated), and it leaves the `outline` model read two different ways for #1977
to reconcile anyway — churn without retiring the transitional mechanism.

## `proctoringPanelStatus` stays in Redux

**Decision.** `OutlineTab` keeps reading `proctoringPanelStatus` from
`useSelector(state.courseHome)`; only `courseId` moves to `useParams`.
`ProctoringInfoPanel` keeps its `useDispatch(fetchProctoringInfoResolved())`.

**Why.** It's client UI state (a reducer flips it to `loaded` when the panel
resolves, gating when the goal widget appears), not server data — out of scope
for a data-layer conversion. It's retired when the store is reduced to the
external specialExams reducer (#1978).

## Subtree `courseId`: slice / `useContextId` reads → `useParams`

**Decision.** Every subtree component that derived `courseId` from the slice
moves to `useParams`:
- direct `useSelector(state.courseHome)` readers — `DateSummary`, `CourseDates`,
  `CourseHandouts`, `CourseTools`, `StartOrResumeCourseCard`,
  `WeeklyLearningGoalCard`, `ProctoringInfoPanel` (keeps its `useDispatch`).
- the `section-outline` readers — `Section`, `SequenceDueDate`, `SequenceTitle`
  — which reached the slice *indirectly* through `useContextId()`
  (`state.courseware.courseId ?? state.courseHome.courseId`). Only these three
  outline-only callers move; the shared `useContextId` itself is left alone
  (~30 progress/course-exit components rely on it).

**Why — required, not cosmetic.** `state.courseHome.courseId` is only ever set
by a tab's fetch thunk (`fetchTabRequest`). Once the outline route stops running
one, any descendant still reading it — directly or via `useContextId()` — gets
`undefined`, so its `useModel(…, courseId)` misses or its `sequences[id]` lookup
crashes: the same breakage class as the `UpgradeToCompleteAlert` slot child in
the dates PR.

## `ShiftDatesAlert`: invalidate all date-bearing caches, and finish its bridge

**Decision.** `refreshTabData` invalidates **both** `datesTab(courseId)` and
`outlineTab(courseId)`, independent of which tab the alert is on. The `model`
prop stays only for the `useModel(model)` visibility gate and the
`resetDeadlines.mutate({ courseId, model })` call. Outline is the last
`fetch={fetchOutlineTab}` caller, so the `fetch` prop and `useDispatch` are
removed entirely.

**Why.** The alert is about *dates*; shifting them changes date info wherever
it's cached — the `dates` model (`datesTab` query) **and** the outline model's
`datesWidget`/`SequenceDueDate` (`outlineTab` query). So invalidation is
model-independent: both keys, always.

**Not a fix to the dates PR.** The dates PR invalidating only `datesTab` was
correct for its moment — outline was still Redux/thunk-fetched then, so no
`outlineTab` query existed to invalidate; its on-mount `fetchOutlineTab` handled
staleness. This PR is where the outline cache first exists, so it's where it
joins the alert's invalidation set.

## Bucket 3: outline-only POST writers become mutations

**Decision.** Convert the outline tab's fire-and-forget POST writers to React
Query mutations, removing `useDispatch` from their components:
- `dismissWelcomeMessage` (`WelcomeMessage`) → `useDismissWelcomeMessage`; the
  thunk is deleted.
- `saveWeeklyLearningGoal` (`WeeklyLearningGoalCard`) →
  `useSaveWeeklyLearningGoal`. This one wasn't even a thunk — it was a plain
  `postWeeklyLearningGoal` wrapper sitting in `thunks.js`. Since we're converting
  the outline writers anyway, it moves out of `thunks.js` and becomes a mutation
  like the others rather than a loose helper.

**Why.** They only POST, dispatch no action, and touch no state — the same *kind*
of writer the CTA-toast PR (#1982) converted (`resetDeadlines`/`processEvent` →
mutations). Converting them here gets the outline tab's genuine remaining Redux
off the store rather than leaving it for a later sweep (#1973).

**`requestCert` was peeled into a prerequisite PR (#1995), not converted here.**
It looked like a third outline writer (`CertificateStatusAlert`), but it's shared
— `CourseCelebration` (course-exit) and `CertificateStatus` (progress) dispatch
it too. Converting a shared thunk mid-outline-PR would drag two unrelated
subtrees into this diff, so it became its own lower stack layer (`useRequestCert`,
all three callers converted); this PR builds on top of it.

**Minor behavior note.** The mutations use `onError: logError` to match the
sibling mutations in `apiHooks.ts`; the original thunks logged nothing (a failed
POST was an unhandled rejection), so this is a small improvement, not a
regression.

## Verification: what was checked in a browser vs. left to the tests

The risky part of this PR is mechanical — the query conversion plus the
`courseId` rewrites, where a wrong `courseId` makes a `useModel` read miss
silently or crashes a `sequences[id]` lookup. So the manual pass targeted the
paths where that failure mode is *only* visible at runtime, and leaned on the
automated suite for the rest.

**Manually verified in a browser:**
- Core load — spinner → content with no mid-load crash (the `OutlineTabContent`
  split), including a hard-reload directly on the outline URL.
- The course outline structure — the `section-outline` `useContextId` →
  `useParams` move: sections expand/collapse, sequence titles link into
  courseware, due dates and completion state render.
- The `useParams`-fed sidebar widgets that a normal course surfaces — dates
  widget, course handouts, course tools, Start/Resume card.
- Smoke: launch tour from outline, outline ↔ dates full-page navigation,
  masquerade-as-learner, and a no-access course.

**Left to the automated tests** (each renders the real tab through the bridged
query path — the same code path production uses — so they exercise the actual
`useParams`/query wiring, not a mock of it):
- Shift due dates from outline — `OutlineTab.test` "handles shift due dates
  click" seeds `missed_deadlines: true`, swaps the refetch to `false`, clicks,
  and asserts the button disappears; it only can if `refreshTabData` invalidated
  `outlineTab` and refetched. This is the one genuinely new behavior in the PR,
  and it's directly asserted.
- Welcome message render + dismiss — `OutlineTab.test` "Welcome Message" block
  plus `apiHooks.test` `useDismissWelcomeMessage` (POST + error logging).
- Weekly learning goal — `OutlineTab.test` "Weekly Learning Goal" block (with the
  proctoring endpoint mocked, so the proctoring-gated render is exercised) plus
  `apiHooks.test` `useSaveWeeklyLearningGoal`.
- Proctoring info panel — its `courseId` is the same `useParams` value proven
  good by the widgets verified above; the proctoring GET is mocked in the tab
  tests.

**Why that split is safe.** What a browser pass would add beyond the tests is
narrow: whether the *live* backend persists a dismiss / goal / shift across a
real reload. But those three endpoints (`dismiss_welcome_message`,
`save_course_goal`, `reset_course_deadlines`) are unchanged by this PR — the same
`api.js` functions, now called from a mutation instead of a thunk — so that path
is the platform's behavior, not this diff's. Standing up the fixtures they need
(a set welcome message, a self-paced course with a past-due deadline, proctored
exams) buys little signal for this change.

</details>

### DIFF
--- a/src/course-home/outline-tab/OutlineTab.jsx
+++ b/src/course-home/outline-tab/OutlineTab.jsx
@@ -1,5 +1,5 @@
 import { useEffect, useRef, useState } from 'react';
-import { useLocation, useNavigate } from 'react-router-dom';
+import { useLocation, useNavigate, useParams } from 'react-router-dom';
 import { useSelector } from 'react-redux';
 import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 import { getAuthenticatedUser } from '@edx/frontend-platform/auth';
@@ -13,7 +13,7 @@ import CourseHandouts from './widgets/CourseHandouts';
 import StartOrResumeCourseCard from './widgets/StartOrResumeCourseCard';
 import WeeklyLearningGoalCard from './widgets/WeeklyLearningGoalCard';
 import CourseTools from './widgets/CourseTools';
-import { fetchOutlineTab } from '../data';
+import { useCourseHomeMeta, useOutlineTabData } from '../data/apiHooks';
 import messages from './messages';
 import ShiftDatesAlert from '../suggested-schedule-messaging/ShiftDatesAlert';
 import UpgradeToShiftDatesAlert from '../suggested-schedule-messaging/UpgradeToShiftDatesAlert';
@@ -27,13 +27,12 @@ import WelcomeMessage from './widgets/WelcomeMessage';
 import ProctoringInfoPanel from './widgets/ProctoringInfoPanel';
 import AccountActivationAlert from '../../alerts/logistration-alert/AccountActivationAlert';
 import CourseHomeSectionOutlineSlot from '../../plugin-slots/CourseHomeSectionOutlineSlot';
+import { TabWithTimer } from '../../tab-page';
 
-const OutlineTab = () => {
+const OutlineTabContent = () => {
   const intl = useIntl();
-  const {
-    courseId,
-    proctoringPanelStatus,
-  } = useSelector(state => state.courseHome);
+  const { courseId } = useParams();
+  const { proctoringPanelStatus } = useSelector(state => state.courseHome);
 
   const {
     isSelfPaced,
@@ -146,7 +145,7 @@ const OutlineTab = () => {
           />
           {isSelfPaced && hasDeadlines && (
             <>
-              <ShiftDatesAlert model="outline" fetch={fetchOutlineTab} />
+              <ShiftDatesAlert model="outline" />
               <UpgradeToShiftDatesAlert model="outline" logUpgradeLinkClick={logUpgradeToShiftDatesLinkClick} />
             </>
           )}
@@ -188,6 +187,22 @@ const OutlineTab = () => {
         )}
       </div>
     </>
+  );
+};
+
+const OutlineTab = () => {
+  const { courseId } = useParams();
+  const metadataQuery = useCourseHomeMeta(courseId);
+  const tabDataQuery = useOutlineTabData(courseId);
+
+  return (
+    <TabWithTimer
+      activeTabSlug="outline"
+      courseId={courseId}
+      courseStatus={{ metadataQuery, tabDataQuery }}
+    >
+      <OutlineTabContent />
+    </TabWithTimer>
   );
 };
 
--- a/src/course-home/data/apiHooks.test.tsx
+++ b/src/course-home/data/apiHooks.test.tsx
@@ -6,7 +6,10 @@ import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 
 import { initializeMockApp } from '../../setupTest';
 import { ToastProvider, useToast } from '../../generic/ToastContext';
-import { useResetDeadlines, usePostEvent, useRequestCert } from './apiHooks';
+import {
+  useOutlineTabData, useResetDeadlines, usePostEvent, useRequestCert, useDismissWelcomeMessage,
+  useSaveWeeklyLearningGoal,
+} from './apiHooks';
 
 const { loggingService } = initializeMockApp();
 
@@ -137,6 +140,83 @@ describe('course-home apiHooks', () => {
 
       await act(async () => {
         await result.current.mutateAsync({ courseId: 'course-1' }).catch(() => {});
+      });
+
+      await waitFor(() => expect(loggingService.logError).toHaveBeenCalled());
+    });
+  });
+
+  describe('useDismissWelcomeMessage', () => {
+    const dismissUrl = `${getConfig().LMS_BASE_URL}/api/course_home/dismiss_welcome_message`;
+
+    it('POSTs to the dismiss url', async () => {
+      axiosMock.onPost(dismissUrl).reply(201);
+      const { wrapper } = buildWrapper();
+      const { result } = renderHook(() => useDismissWelcomeMessage(), { wrapper });
+
+      await act(async () => { await result.current.mutateAsync({ courseId: 'course-1' }); });
+
+      expect(axiosMock.history.post[0].url).toEqual(dismissUrl);
+    });
+
+    it('logs the error when the POST fails', async () => {
+      axiosMock.onPost(dismissUrl).reply(500);
+      const { wrapper } = buildWrapper();
+      const { result } = renderHook(() => useDismissWelcomeMessage(), { wrapper });
+
+      await act(async () => {
+        await result.current.mutateAsync({ courseId: 'course-1' }).catch(() => {});
+      });
+
+      await waitFor(() => expect(loggingService.logError).toHaveBeenCalled());
+    });
+  });
+
+  describe('useOutlineTabData', () => {
+    const outlineUrl = `${getConfig().LMS_BASE_URL}/api/course_home/outline/course-1`;
+
+    it('resolves to an empty object on a 403 (access is handled via the metadata request)', async () => {
+      axiosMock.onGet(outlineUrl).reply(403, {});
+      const { wrapper } = buildWrapper();
+      const { result } = renderHook(() => useOutlineTabData('course-1'), { wrapper });
+
+      await waitFor(() => expect(result.current.isSuccess).toBe(true));
+      expect(result.current.data).toEqual({});
+    });
+
+    it('surfaces the error on a non-403 failure', async () => {
+      axiosMock.onGet(outlineUrl).reply(500);
+      const { wrapper } = buildWrapper();
+      const { result } = renderHook(() => useOutlineTabData('course-1'), { wrapper });
+
+      await waitFor(() => expect(result.current.isError).toBe(true));
+    });
+  });
+
+  describe('useSaveWeeklyLearningGoal', () => {
+    const goalUrl = `${getConfig().LMS_BASE_URL}/api/course_home/save_course_goal`;
+
+    it('POSTs the weekly learning goal', async () => {
+      axiosMock.onPost(goalUrl).reply(200, {});
+      const { wrapper } = buildWrapper();
+      const { result } = renderHook(() => useSaveWeeklyLearningGoal(), { wrapper });
+
+      await act(async () => {
+        await result.current.mutateAsync({ courseId: 'course-1', daysPerWeek: 3, subscribedToReminders: true });
+      });
+
+      expect(axiosMock.history.post[0].url).toEqual(goalUrl);
+    });
+
+    it('logs the error when the POST fails', async () => {
+      axiosMock.onPost(goalUrl).reply(500);
+      const { wrapper } = buildWrapper();
+      const { result } = renderHook(() => useSaveWeeklyLearningGoal(), { wrapper });
+
+      await act(async () => {
+        await result.current.mutateAsync(
+          { courseId: 'course-1', daysPerWeek: 3, subscribedToReminders: true },
+        ).catch(() => {});
       });
 
       await waitFor(() => expect(loggingService.logError).toHaveBeenCalled());
--- a/src/course-home/data/apiHooks.ts
+++ b/src/course-home/data/apiHooks.ts
@@ -3,7 +3,14 @@ import { useMutation, useQuery } from '@tanstack/react-query';
 
 import { useToast, ToastContent } from '@src/generic/ToastContext';
 import {
-  executePostFromPostEvent, getCourseHomeCourseMetadata, getDatesTabData, postCourseDeadlines, postRequestCert,
+  executePostFromPostEvent,
+  getCourseHomeCourseMetadata,
+  getDatesTabData,
+  getOutlineTabData,
+  postCourseDeadlines,
+  postDismissWelcomeMessage,
+  postRequestCert,
+  postWeeklyLearningGoal,
 } from './api';
 import { courseHomeQueryKeys } from './queryKeys';
 
@@ -61,7 +68,25 @@ export const useDatesTabData = (courseId: string) => useQuery({
   meta: { modelType: 'dates', courseId },
 });
 
+export const useOutlineTabData = (courseId: string) => useQuery({
+  queryKey: courseHomeQueryKeys.outlineTab(courseId),
+  queryFn: () => getOutlineTabData(courseId),
+  meta: { modelType: 'outline', courseId },
+});
+
 export const useRequestCert = () => useMutation({
   mutationFn: ({ courseId }: { courseId: string }) => postRequestCert(courseId),
+  onError: (error) => logError(error),
+});
+
+export const useDismissWelcomeMessage = () => useMutation({
+  mutationFn: ({ courseId }: { courseId: string }) => postDismissWelcomeMessage(courseId),
+  onError: (error) => logError(error),
+});
+
+export const useSaveWeeklyLearningGoal = () => useMutation({
+  mutationFn: ({ courseId, daysPerWeek, subscribedToReminders }: {
+    courseId: string; daysPerWeek: number; subscribedToReminders: boolean;
+  }) => postWeeklyLearningGoal(courseId, daysPerWeek, subscribedToReminders),
   onError: (error) => logError(error),
 });
--- a/src/course-home/data/index.js
+++ b/src/course-home/data/index.js
@@ -1,8 +1,6 @@
 export {
-  fetchOutlineTab,
   fetchProgressTab,
   deprecatedSaveCourseGoal,
-  saveWeeklyLearningGoal,
 } from './thunks';
 
 export { reducer } from './slice';
--- a/src/course-home/data/queryKeys.ts
+++ b/src/course-home/data/queryKeys.ts
@@ -4,4 +4,5 @@ export const courseHomeQueryKeys = {
   all: [appId, 'courseHome'] as const,
   metadata: (courseId: string) => [...courseHomeQueryKeys.all, 'metadata', courseId] as const,
   datesTab: (courseId: string) => [...courseHomeQueryKeys.all, 'datesTab', courseId] as const,
+  outlineTab: (courseId: string) => [...courseHomeQueryKeys.all, 'outlineTab', courseId] as const,
 };
--- a/src/course-home/data/redux.test.js
+++ b/src/course-home/data/redux.test.js
@@ -42,45 +42,6 @@ describe('Data layer integration tests', () => {
     store = initializeStore();
   });
 
-  describe('Test fetchOutlineTab', () => {
-    const outlineBaseUrl = `${getConfig().LMS_BASE_URL}/api/course_home/outline`;
-    const outlineUrl = `${outlineBaseUrl}/${courseId}`;
-
-    it('Should result in fetch failure if error occurs', async () => {
-      axiosMock.onGet(courseMetadataUrl).networkError();
-      axiosMock.onGet(outlineUrl).networkError();
-
-      await executeThunk(thunks.fetchOutlineTab(courseId), store.dispatch);
-
-      expect(loggingService.logError).toHaveBeenCalled();
-      expect(store.getState().courseHome.courseStatus).toEqual('failed');
-    });
-
-    it('Should fetch, normalize, and save metadata', async () => {
-      const outlineTabData = Factory.build('outlineTabData', { courseId });
-
-      axiosMock.onGet(courseMetadataUrl).reply(200, courseHomeMetadata);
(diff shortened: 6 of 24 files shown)


brian-smith-tcril:

🤖 _Comment via Claude Code._

brian-smith-tcril:
> 

addressed in https://github.com/openedx/frontend-app-learning/compare/163da20974d71e07b00f4f16f4d00fea48b725d8..12b91126ad54fb3322aea8dd0c4653338dfaa827 (part of a stack rebase, see https://github.com/openedx/frontend-app-learning/pull/1987)

src/course-home/outline-tab/OutlineTab.jsx
```diff
@@ -27,13 +27,12 @@ import WelcomeMessage from './widgets/WelcomeMessage';
 import ProctoringInfoPanel from './widgets/ProctoringInfoPanel';
 import AccountActivationAlert from '../../alerts/logistration-alert/AccountActivationAlert';
 import CourseHomeSectionOutlineSlot from '../../plugin-slots/CourseHomeSectionOutlineSlot';
+import { TabWithTimer } from '../../tab-page';
 
-const OutlineTab = () => {
+const OutlineTabContent = () => {
```
arbrandes, line 32: This appears to cause a regression: for a learner without access `OutlineTabContent` mounts while the `outline` model is still `{}`, and throws.

Suggestion: add `= {}` defaults to the `courseBlocks` and `datesWidget` destructures just below (OutlineTab.jsx:45-58).

It's a rare path, but possibly worth a unit test.

arbrandes (approved):
Pre-approved, but Claude seems to have found a regression in one of the not-so-happy paths.  Fix and merge at your discretion.

src/course-home/outline-tab/OutlineTab.jsx
brian-smith-tcril: ended up addressing this by changing the `deriveView` order in `TabPage` to better mach the prior redux logic instead of adding `{}` defaults all over the place https://github.com/openedx/frontend-app-learning/compare/5d59c67e820f943785a7b7f443c7a677137437ed..f8788fb8b77bdb881cd05113f8d46a6fdacd0fbd

brian-smith-tcril: Decision log for this fix, written by Claude 🤖:

<details>
<summary>Why the <code>deriveView</code> reorder instead of <code>= {}</code> defaults</summary>

**Regression**

A learner without access and **no redirect** (this only happens on the outline tab — `getAccessDeniedRedirectUrl` returns `null` for `activeTabSlug === 'outline'`) renders `OutlineTabContent` while the `outline` model is still `{}`, and the undefaulted `courseBlocks` / `datesWidget` destructures throw.

**Why it regressed (vs. the Redux behavior)**

The old `fetchTab` did `Promise.allSettled([metadata, outline])` and dispatched *any* status — including the `!courseAccess.hasAccess` "denied regardless of tabData" short-cut — only after **both** settled, so the `outline` model was always populated before the page rendered. The React Query split runs metadata and `useOutlineTabData` as independent queries, and `deriveView` returned `denied` from the metadata query alone, so content could mount before the outline query resolved.

**Options**

- **`= {}` defaults on the destructures** (the original suggestion) — the same empty-model read is duplicated across ~8 spots (`useCourseEndAlert`, the certificate-status and scheduled-content alert hooks, plus several widgets), so defaults would be whack-a-mole and easy to reintroduce.
- **Reorder `deriveView`** ✅ — move the `tabDataQuery.isPending` check *above* the access check. One line; restores the Redux "both loaded before any status" invariant; makes *every* reader safe at once. Access is still derived from the metadata query alone — this only gates rendering on the tab data being **loaded**, it doesn't *consider* tab data for the access decision.

**Scope**

- Metadata-only tabs (discussion) are unaffected — `tabDataQuery?` is undefined, so the check is skipped.
- Denied-**redirect** tabs (dates/progress) now wait for their query before redirecting, which also matches the old thunk (it awaited both before denying).

**Tests**

- `OutlineTab.test.jsx`: an integration test that reproduces the actual throw — a fresh store (genuinely empty `outline` model) plus a deferred outline response that stays pending → it crashes without the fix, and shows loading then the private-course alert with it.
- `TabPage.test.jsx`: a `deriveView`-contract test (denied + pending tab data → loading).

</details>

arbrandes: Works for me!

Issue #2000, opened 2026-08-15
Progress tab heading displays "Your progress" instead of "Course progress for <username>" when masquerading


brian-smith-tcril:
> [!NOTE]
> This issue was drafted by Claude.

### Steps to reproduce
1. As course staff, open a course's Progress tab.
2. In the masquerade bar, choose "View this course as" → **Specific student**, and enter an enrolled learner.
3. Observe the Progress tab heading.

### Expected
Heading reads `Course progress for <username>` (the masqueraded learner).

### Actual
Heading reads "Your progress", even though the grades/section scores shown below are the masqueraded learner's.

### Additional context
Navigating directly to `/course/<courseId>/progress/<targetUserId>/` **does** correctly show the other learner's heading (`Course progress for <username>`) — only the masquerade-bar path is affected.

Issue #1999, opened 2026-08-15
Source the access-expiration masquerade banner from the tab query, not useModel(tab)


brian-smith-tcril:
Part of the Redux → React Query migration, epic #1946. **Related to the model-store dissolution (#1977)** — it lands as a piece of that, *not* as a standalone prerequisite for any single tab. See **Sequencing** below.

## Summary

`useAccessExpirationMasqueradeBanner(courseId, tab)` — rendered by the shared `InstructorToolbar` — reads `accessExpiration.masqueradingExpiredCourse` via `useModel(tab, courseId)` to show the "this learner's access expired" masquerade banner. It's the **only** dynamic `useModel(<var>)` reader that resolves to a course-home tab's data model (a literal `useModel('progress'` grep misses it). Because it reads the tab **model store**, any converted tab has to keep its data on the transitional bridge (`meta` tag) just to feed this one banner.

## Sequencing: with #1977 (after the tabs convert), not as a per-tab prerequisite

We tried building this as a prerequisite *below* the progress-tab conversion (#1998), so progress could be fully bridge-free. Attempting it surfaced that the migration is **awkward at any position in the current transition** — it's a data-flow problem, not a stack-ordering one. `accessExpiration` lives in **per-tab data**, its consumer is the **shared, tab-generic** `InstructorToolbar`, and only some tabs are query-backed yet, so every mid-transition shape is bad:

- **Prop-thread it** (`TabPage` → `LoadedTabPage` → `InstructorToolbar` → hook): `useAccessExpirationMasqueradeBanner` ends up *taking `accessExpiration` in*, inverting the responsibility of a hook named for the data it ought to source.
- **Hook reads the React Query cache itself**: a generic `src/alerts/` hook has to import `course-home/data/queryKeys` and special-case progress's `targetUserId` — worse coupling.

Both also still need a transitional `useModel(activeTabSlug)` fallback for the not-yet-converted tabs (courseware/live/discussion).

It's only clean at the **north star** — every tab query-backed and the model store gone — when the banner sources `accessExpiration` from a uniform query-backed source with no fallback. So this lands **with the model-store dissolution (#1977)**, once the tabs are converted; it's a piece of that effort, not a blocker for progress.

**In the meantime:** the progress tab (#1998) keeps a **scoped one-reader `meta` bridge**. The model store persists until #1977 regardless, so that one line is just one more consumer of a bridge that's staying anyway — and it lets progress ship without waiting on this.

## Target (at the north star)

[...]

PR #2001, opened 2026-08-15
refactor: convert progress tab from Redux to React Query

### LINKED ISSUE #1998: Convert the progress tab to React Query
Part of #1975 (convert course-home tab data to React Query), Phase 3 of the epic #1946. Stacks directly on the outline-tab conversion (#1997), reusing the self-wrapping pattern the dates (#1987) and outline (#1997) tabs set.

## Summary

Convert the progress tab off Redux to React Query, following the **self-wrapping** pattern: `ProgressTab` renders `<TabWithTimer>` itself and owns its data via `useCourseHomeMeta` + a new `useProgressTabData` hook (meta-tagged `progress` so the transitional model-store bridge keeps `useModel('progress', …)` populated for the whole grade/certificate subtree). The body moves into a `ProgressTabContent` child so the `useModel('progress')` destructure never runs against the empty model mid-load (the `OutlineTabContent` split). The `index.jsx` route drops its `<TabContainer tab="progress" fetch={fetchProgressTab} isProgressTab>` wrapper for a bare `<ProgressTab />`.

Progress has a staff "view a learner" route (`/course/:courseId/progress/:targetUserId/`), so the hook takes `useProgressTabData(courseId, targetUserId)` — `targetUserId` from `useParams`, in both the query key and the query function (so switching the viewed learner refetches).

Because the route no longer runs `fetchProgressTab`, `state.courseHome.courseId` is never set for it, so **every subtree component that read `courseId` via `useContextId()` moves to `useParams`** — 20 progress-only components plus the `ProgressTabGradeBreakdownSlot` plugin slot and `ProgressTab` itself (21 files). Same breakage class as outline's section-outline components, just wider.

`fetchProgressTab` is deleted (its only caller was the route wrapper), and the now-dead `isProgressTab` prop + branch are removed from `TabContainer` (progress was its only consumer). The shared `fetchTab` helper stays — live and discussion still use it.

## Out of scope (deliberately preserved)

- **`examsData` stays.** `ProgressTab` → `useGetExamsData` → `fetchExamAttemptsData` → `state.courseHome.examsData` (fed by `getExamsData`) has no in-repo reader **by design**: PR #1829 added it so "children of the root `ProgressPage` or downstream plugin slots [can] access this data from the Redux store." Those consumers are out-of-tree plugins in the `ProgressTab*` slots, so deleting it would break a public contract. Kept as-is, like the model store; only the `courseId` it's called with moves to `useParams`.

## Tasks

- [ ] `queryKeys.ts` — add `progressTab(courseId, targetUserId)`.
- [ ] `apiHooks.ts` — add `useProgressTabData(courseId, targetUserId)` (`getProgressTabData`, `meta: { modelType: 'progress', courseId }`).
- [ ] `ProgressTab.jsx` — self-wrap `TabWithTimer` (`activeTabSlug="progress"`, `courseStatus={{ metadataQuery, tabDataQuery }}`); split the body into `ProgressTabContent`; `courseId`/`targetUserId` from `useParams`; keep the `useGetExamsData` dispatch.
- [ ] Move the 20 subtree `useContextId()` reads + the `ProgressTabGradeBreakdownSlot` to `useParams` (all verified progress-only).
- [ ] `ProgressHeader.jsx` — `courseId`/`targetUserId` from `useParams` instead of `state.courseHome`.
- [ ] `index.jsx` — progress route → `<ProgressTab />`; drop the `fetchProgressTab` import/wiring.
- [ ] `TabContainer` — remove the `isProgressTab` prop + its `targetUserId` fetch branch.
- [ ] Delete `fetchProgressTab` (thunk + `data/index.js` re-export).
- [ ] Tests: `ProgressTab.test.jsx` renders `<ProgressTab />` directly through the bridged query client (mirroring `OutlineTab.test.jsx`); the `examsData` assertions stay (path unchanged). Drop the `Test fetchProgressTab` block from `redux.test.js`, and confirm the shared `fetchTab` helper stays covered by a live/discussion test (add one if needed). Add a `useProgressTabData` block to `apiHooks.test.tsx` (success + error path).

## Notes / decisions

- **The `progress` model stays on the transitional bridge.** All readers are within the progress render, but one is the `ProgressTabGradeBreakdownSlot` plugin slot, so the model is plausibly a plugin-facing read surface too. Converting the in-repo readers to the query would leave the model read two ways while the bridge still has to exist for the slot — churn without retiring the mechanism (that's #1977).
- **Subtree `courseId` → `useParams` is required, not cosmetic.** The slice's `courseId` is only ever set by a tab's fetch thunk; once the route stops running one, `useContextId()` returns `undefined` and every `useModel('progress', courseId)` read misses (or `sectionScores[...]` crashes).
- **Per-file `useParams`, not a `courseId` context.** These leaf components already each source their own `courseId` via `useContextId`, so `useParams` is a 1:1 swap. A context would be a larger refactor than the conversion and departs from the direct-hook pattern outline/dates set.

> [!NOTE]

<details>
<summary>Claude Plan — progress tab → React Query</summary>

### Approach

A self-wrapping conversion mirroring outline: the page owns its data via query hooks and renders `<TabWithTimer>` directly; `TabContainer` stops wrapping this route (and loses its now-unused `isProgressTab` path). The model-store bridge (meta-tagged queries → `addModel`) keeps the large `useModel('progress'|'courseHomeMeta', …)` subtree working unchanged, so the conversion is behavior-preserving. `examsData` is left entirely in place (plugin contract).

### Data layer

```ts
// queryKeys.ts
progressTab: (courseId: string, targetUserId?: string) =>
  [...courseHomeQueryKeys.all, 'progressTab', courseId, targetUserId ?? null] as const,

// apiHooks.ts
export const useProgressTabData = (courseId: string, targetUserId?: string) => useQuery({
  queryKey: courseHomeQueryKeys.progressTab(courseId, targetUserId),
  queryFn: () => getProgressTabData(courseId, targetUserId),
  meta: { modelType: 'progress', courseId },
});
```

### `ProgressTab.jsx`

```jsx
const ProgressTab = () => {
  const { courseId, targetUserId } = useParams();
  const tabDataQuery = useProgressTabData(courseId, targetUserId);
  return (
    <TabWithTimer
      activeTabSlug="progress"
      courseId={courseId}
    >
      <ProgressTabContent />
    </TabWithTimer>
  );
};

// ProgressTabContent: today's ProgressTab body — useParams courseId, useModel('progress'),
// the sequenceIds useMemo, and useGetExamsData(courseId, sequenceIds) (examsData path unchanged).
```

### The `useContextId` → `useParams` sweep (21 files)

`ProgressTab`, `ProgressTabGradeBreakdownSlot`, and 19 progress-only leaves under `progress-tab/` (course-completion, credit-information, related-links, certificate-status, and the grades/* tree). Each swaps `const courseId = useContextId()` for `const { courseId } = useParams()` and updates imports. Verified none render outside `progress-tab/`, so `useParams` yields the right `courseId` everywhere they mount.

</details>

brian-smith-tcril:
## Summary

Convert the course-home **progress tab** from Redux to React Query. Part of the Redux → React Query migration (#1946), stacked on the outline-tab conversion (#1997). Closes #1998.

`ProgressTab` becomes self-wrapping — it owns its data (`useCourseHomeMeta` + a new `useProgressTabData`) and renders `TabWithTimer` itself, so the `index.jsx` route drops its `<TabContainer tab="progress" fetch={fetchProgressTab} isProgressTab>` wrapper for a bare `<ProgressTab />` and `fetchProgressTab` is deleted.

## What changed

- **`apiHooks.ts` / `queryKeys.ts`** — add `useProgressTabData(courseId, targetUserId?)` and the `progressTab(courseId, targetUserId?)` key. It keeps a scoped `meta: { modelType: 'progress', courseId }` so the model-store bridge still feeds the one shared holdout reader (below).
- **`ProgressTab.jsx`** — split into a thin data-owning `ProgressTab` (runs the two queries, renders `<TabWithTimer courseStatus={{ metadataQuery, tabDataQuery }}><ProgressTabContent/></TabWithTimer>`) and a `ProgressTabContent` in the same file. The split keeps the query-data reads out of the loading phase, where the data is still empty.
- **`useProgressData()` hook** (`progress-tab/hooks.jsx`) — reads the tab data from `useProgressTabData(courseId, targetUserId)`; subtree readers call it instead of `useModel('progress')`, so they share one fetch (keyed by the same `targetUserId` — critical on the staff `/progress/:targetUserId/` route).
- **Subtree `courseId` → `useParams`** — the ~20 grade/certificate/related-links components + `ProgressTabGradeBreakdownSlot` move off `useContextId()`; most drop `courseId` entirely and read `useProgressData()`. Required: the route no longer runs a thunk to set `state.courseHome.courseId`.
- **`ProgressHeader.jsx`** — reads `targetUserId`/`username` from `useParams` + `useProgressData()` instead of the slice + `useModel('progress')`; the `targetUserId !== userId` int comparison is preserved.
- **Route + `TabContainer` cleanup** — `fetchProgressTab` deleted; the now-dead `isProgressTab` prop + its `targetUserId` fetch branch removed from `TabContainer` (progress was its only consumer). The shared `fetchTab` helper stays (live/discussion still use it).

## The one retained bridge

Every *progress-specific* reader is off the model store. `useProgressTabData` keeps its `meta` bridge for exactly one shared, tab-generic holdout: `useAccessExpirationMasqueradeBanner` (rendered by `InstructorToolbar`) does a dynamic `useModel(tab)` that resolves to `'progress'`. Migrating it cleanly needs every tab to be query-backed, so it's deferred to the model-store dissolution (#1977, tracked in #1999) — the `meta` line retires with the rest of the bridge then. Full rationale in the decision log.

## Behavior

No user-facing change. `examsData` (the exam-attempts Redux write that plugin slots read) stays as-is; only its `courseId` source moves to `useParams`.

Manual testing surfaced a **pre-existing** bug (present on `master`): masquerading as a specific learner via the masquerade bar shows "Your progress" instead of "Course progress for &lt;username&gt;", because that flow leaves the URL (and `:targetUserId`) unchanged. Filed as #2000 — out of scope here; this conversion preserves the behavior exactly.

## Testing

`npm run types`, `npm run lint`, and the full `npm test` suite pass (the one `Course.test.jsx` failure is a pre-existing parallel-run flake — green in isolation). `ProgressTab.test.jsx` now renders the real self-wrapping tab through the bridged query client; new `useProgressTabData` coverage (success/transform, `targetUserId` URL, 401/403/404, error) in `apiHooks.test.tsx`; `redux.test.js` drops the `fetchProgressTab` block (`fetchTab` stays covered by `DiscussionTab.test`). Manual browser verification and the test-coverage split are documented in the decision log below.

## Decisions

<details>

# Decisions — Redux → React Query: the progress tab (#1946)

(course-home tab data), stacked on the outline-tab conversion (#1997).

Unlike the dates/outline conversions, this one goes **bridge-free for every
progress-specific reader**: the whole grade/certificate subtree reads the tab
data straight from the React Query cache, not from the Redux model store. One
shared, tab-generic reader forces a single, narrowly-scoped exception (below).

## Scope: self-wrapping conversion

**Decision.** `ProgressTab` becomes self-wrapping — it owns its data via
`useCourseHomeMeta` + a new `useProgressTabData` hook and renders `<TabWithTimer>`
itself. The `index.jsx` route drops `<TabContainer tab="progress"
fetch={fetchProgressTab} isProgressTab>` for a bare `<ProgressTab />`, and
`fetchProgressTab` is deleted.

## `ProgressTab` splits into a thin wrapper + `ProgressTabContent`

**Decision.** `ProgressTab` runs the two queries and renders `<TabWithTimer
courseStatus={{ metadataQuery, tabDataQuery }}><ProgressTabContent /></TabWithTimer>`;
the body moves to `ProgressTabContent`. Same rationale as `OutlineTabContent`:
`TabPage` only renders children once the data has loaded, so the body reading the
query data (`useProgressData()`, and `sectionScores.flatMap(...)`) never runs
before the fetch resolves.

## Progress-specific readers read the query directly — no bridge for them

**Decision.** Every progress-subtree reader of the `progress` model — the ~20
grade/certificate/related-links components, the `ProgressTabGradeBreakdownSlot`
plugin slot, `ProgressHeader`, and `ProgressTabContent` — stops calling
`useModel('progress', courseId)` and reads the tab data from `useProgressTabData`
(via the `useProgressData` hook below). No model-store round-trip.

**Why.** The goal of the epic is to remove Redux, not to keep relocating reads
onto a transitional bridge. We're already editing every one of these files to
move `courseId` off `useContextId()` (required — see below), so converting the
read in the same pass costs almost nothing and actually retires the Redux
dependency instead of deferring it. React Query dedupes on the query key, so the
wrapper's `useProgressTabData` and every reader share one fetch.

## `useProgressData()` reader hook

**Decision.** Add `useProgressData()` (in `progress-tab/hooks.jsx`):
`const { courseId, targetUserId } = useParams(); return useProgressTabData(courseId,
targetUserId).data;`. Subtree readers call `useProgressData()` and destructure.

**Why.** React Query keys by every argument, so a reader must pass the *same*
`targetUserId` as the wrapper to hit the same cache entry — otherwise, on the
staff `/progress/:targetUserId/` route, it would key on `undefined`, fire a second
fetch, and show the requesting user's grades instead of the learner's. Threading
`targetUserId` through ~20 leaves is noise; the model store used to hide it (one
`progress` model keyed by `courseId`, the target user's data baked in). This hook
restores that ergonomic: callers name neither `targetUserId` nor (usually)
`courseId`. The wrapper still uses raw `useProgressTabData` because it needs the
full query object (`isLoading`/`isError`) for gating.

## The one shared reader keeps the `meta` bridge (its migration is deferred)

**Decision.** `useProgressTabData` keeps `meta: { modelType: 'progress', courseId }`,
so the model-store bridge still populates `useModel('progress', courseId)` — **for
exactly one reader**: `useAccessExpirationMasqueradeBanner(courseId, tab)` (rendered
by the shared `InstructorToolbar`) does `useModel(tab, courseId)` and reads
`accessExpiration.masqueradingExpiredCourse` for the "this learner's access expired"
masquerade banner. It's the only dynamic `useModel(<var>)` reader that resolves to
`'progress'` (the others are dates/outline/courseware); a literal `useModel('progress'`
grep misses it. Every *progress-specific* reader is already off the bridge (above) —
this is the lone shared, tab-generic holdout.

**Why keep it rather than migrate it here.** We tried migrating this banner first,
as a lower stack layer, so progress could be fully bridge-free — it doesn't work
cleanly. `accessExpiration` lives in per-tab data, the banner's consumer is the
shared, tab-generic `InstructorToolbar`, and only some tabs are query-backed, so
every mid-transition shape is bad: thread it down → the hook *takes its own data
in* (inverting the responsibility of a hook named for that data); have the hook
read the cache → a generic `alerts/` hook has to import `course-home` query keys
and special-case progress's `targetUserId`. Both still need a `useModel` fallback
for the unconverted tabs. It's only clean once every tab is query-backed. So the
banner migration is **deferred to the model-store dissolution (#1977)** and tracked
in **#1999**; the `meta` here retires with the rest of the bridge at that point.
The model store persists until #1977 regardless, so this one `meta` line is just
one more consumer of a bridge that's staying anyway — not worth holding up progress
or contorting the banner for.

## Subtree `courseId`: `useContextId()` → `useParams` (required)

**Decision.** The 20 progress-only subtree components + the
`ProgressTabGradeBreakdownSlot` move `courseId` from `useContextId()` to
`useParams` (most drop `courseId` entirely, reading `useProgressData()` instead;
the handful that use `courseId` elsewhere keep `const { courseId } = useParams()`).

**Why — required, not cosmetic.** `useContextId()` reads
`state.courseware.courseId ?? state.courseHome.courseId`; the `courseHome` one is
only set by the fetch thunk. Once the route stops running it, `useContextId()`
returns `undefined` and every reader misses. All 20 were verified progress-only.

## `targetUserId` from `useParams`, with the int comparison preserved

**Decision.** `ProgressHeader` reads `targetUserId` from `useParams` (a string)
and parses it for its one numeric use — the `targetUserId !== userId` title check
(`userId` is a number). The old thunk `parseInt`'d `targetUserId` into the slice,
so a faithful conversion keeps that an int comparison rather than switching to a
string compare.

## Masquerade heading is a pre-existing bug, left out of scope (#2000)

**Decision.** `ProgressHeader` keeps deriving "viewing another student" from the URL
`:targetUserId` segment (the faithful conversion above — `parseInt(useParams()
.targetUserId, 10)`). We do **not** extend it to also cover the masquerade bar.

**Why.** Manual testing surfaced that masquerading as a *specific student* via the
masquerade bar shows "Your progress" instead of "Course progress for &lt;username&gt;":
the bar sets a server-side session and leaves the URL (and `:targetUserId`) unchanged,
so the check is false. Confirmed **pre-existing on `master`** — the conversion preserves
the behavior exactly (old: slice `targetUserId` undefined → falsy; new:
`parseInt(undefined, 10)` → `NaN` → falsy). It's an independent bug whose real fix is
non-trivial (distinguishing specific-student masquerade from generic-role masquerade
needs a signal beyond the URL — the masquerade *target identity*), so bundling it would
break the faithful-conversion principle. Filed standalone as **#2000**. A `username`-based
proxy was considered and rejected as too indirect. (The deep-link `/progress/:targetUserId/`
path is unaffected and works.)

## `examsData` stays for now — a separate Redux write to retire later

**Decision.** Leave the exam-attempts path (`useGetExamsData` →
`fetchExamAttemptsData` → `state.courseHome.examsData`, via `getExamsData`) in
place; only its `courseId` source moves to `useParams`.

**Why (and not "it's a plugin contract").** It's a fire-and-forget Redux *write*
with no in-repo reader — it exists so plugin slots can read `examsData` from the
store (PR #1829). The Redux store isn't a supported plugin API, so that's not a
reason to preserve it; it's just another Redux surface to remove. But removing it
means giving plugins a query-based replacement, which is its own task — so it's
out of scope here, not something we protect. (An earlier note calling it a
"contract to keep" was wrong.)

## Route + `TabContainer` cleanups

**Decision.** Delete `fetchProgressTab` (only the route wrapper called it), and
remove the now-dead `isProgressTab` prop + its `targetUserId` fetch branch from
`TabContainer` (progress was its only consumer). The shared `fetchTab` helper
stays (live and discussion still use it).

## Coverage: preserve the shared `fetchTab` helper's coverage

**Decision.** When removing the `fetchProgressTab` redux test, keep the shared
`fetchTab` helper covered by a remaining tab test (live/discussion still use it),
adding one if needed. Give `useProgressTabData` its own `apiHooks.test` (success +
error). Same indirect-coverage trap the outline PR hit in `getOutlineTabData`.

## Test conversion

**`ProgressTab.test.jsx` renders the self-wrapping tab through the bridged query
client.** The `fetchAndRender` helper drops `executeThunk(fetchProgressTab)` +
bare `render(<ProgressTab />)` for the `OutlineTab.test` pattern: an `AppProvider`
+ `MemoryRouter` (route `/course/:courseId/progress/:targetUserId?`) +
`createTestQueryClient(store)` (the model-store bridge, so the one remaining
`useModel('progress')` reader still resolves) + `UserMessagesProvider`/`ToastProvider`,
then waits for the loading `status` role to clear. The masquerade / course-start
banner tests previously hand-rendered `<LoadedTabPage …>...</LoadedTabPage>`; they
now render the full `<ProgressTab />` and assert on the `instructor-toolbar` the
tab draws itself — same as the outline access-expiration tests.

**Two existing selectors had to disambiguate now that the full page (with tab
navigation) renders.** Rendering the whole tab, not just its content, adds the
header + `CourseTabsNavigationSlot` links to the DOM:
- The "Dates" related-link is scoped to the "Related links" `section` via
  `within(...)`, because the tab nav also has a "Dates" link.
- The two "locked feature" tests asserted on `getAllByRole('link', 'Unlock now')`
  — `'Unlock now'` is a positional string (an ignored options arg), so the query
  actually matched *every* link and only passed because the content-only render
  had exactly three. They now target the real button, `{ name: 'Upgrade now' }`
  (there is exactly one), keeping the existing `getAllByRole`/index/`fireEvent`
  style.

**The exam re-fetch test drives the data change through the query.** "Re-fetch
exam data when section scores change" used to re-run `fetchProgressTab`; it now
calls `queryClient.invalidateQueries()` (the helper returns the client) so the tab
query refetches the updated mock and `useGetExamsData` re-fires.

**`redux.test.js`** drops the whole `Test fetchProgressTab` block (thunk deleted).
`fetchTab` stays covered by `DiscussionTab.test` (its `TabContainer` mount dispatches
`fetchDiscussionTab` → `fetchTab`); `getProgressTabData`'s success/transform,
`targetUserId` URL, and 401/403/404/other-error branches move to the
`useProgressTabData` block in `apiHooks.test`. The 404 branch calls
`global.location.replace`, so that one case swaps `window.location` (non-configurable
`replace` can't be spied) inside a `try/finally` that always restores it.

**`TabContainer.test.jsx`** drops the `isProgressTab` `targetUserId` test — the prop
and its fetch branch are gone, and that behavior now lives in the progress query's
`useParams`/`useProgressTabData` path (covered by the `/progress/10/` route test).

## Manual testing (in-browser) vs. what we lean on the automated suite for

Tested in a real browser against a live backend (tutor local), scoped to what this
PR changes.

**Verified by hand:**
- Cold load + hard-reload directly on the progress URL (the `useProgressTabData`
  fetch path; nothing pre-fetched).
- The whole grade subtree renders off the `useParams` `courseId` — the highest-risk
  mechanical rewrite: course-grade header, grade summary, detailed grades + the
  per-problem drawer + subsection navigation, related-links + their analytics event,
  and the completion graph.
- Staff deep-link `/progress/<numericId>/`: heading shows "Course progress for
  &lt;username&gt;", the body shows *that* learner's grades/certificate, and the Studio
  link shows for admins.
- Access-expiration masquerade banner correctly **absent** for a masqueraded learner
  with active access.
- 404 on the progress data → redirect to the legacy `/courses/<id>/progress`.

**Left to the automated suite (not re-done by hand):**
- Both *positive* masquerade banners (access-expired shows; course-start shows) —
  `ProgressTab.test` renders them through the bridged query client, which exercises
  the retained one-reader `meta` bridge end-to-end in-test.
- Certificate states + "Request certificate" — `ProgressTab.test` Certificate Status suite.
- Locked/limited "Upgrade now" preview + its analytics — `ProgressTab.test` locked-feature tests.
- Exam-data integration (`examsData`) — `ProgressTab.test` exam-integration suite + `redux.test`'s `fetchExamAttemptsData`.
- Per-learner query keying (switching `targetUserId`) — the query key includes
  `targetUserId` (`queryKeys` + the `apiHooks` targetUserId-URL test); not re-checked in-browser.
- Denied / unenrolled path — shared `TabPage` denied handling.

The masquerade-bar heading is the known pre-existing bug (#2000 above), not covered here.

</details>

### DIFF
--- a/src/course-home/data/apiHooks.test.tsx
+++ b/src/course-home/data/apiHooks.test.tsx
@@ -1,5 +1,6 @@
 import { act, renderHook, waitFor } from '@testing-library/react';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { Factory } from 'rosie';
 import MockAdapter from 'axios-mock-adapter';
 import { getConfig } from '@edx/frontend-platform';
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
@@ -7,8 +8,8 @@ import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 import { initializeMockApp } from '../../setupTest';
 import { ToastProvider, useToast } from '../../generic/ToastContext';
 import {
-  useOutlineTabData, useResetDeadlines, usePostEvent, useRequestCert, useDismissWelcomeMessage,
-  useSaveWeeklyLearningGoal,
+  useOutlineTabData, useProgressTabData, useResetDeadlines, usePostEvent, useRequestCert,
+  useDismissWelcomeMessage, useSaveWeeklyLearningGoal,
 } from './apiHooks';
 
 const { loggingService } = initializeMockApp();
@@ -188,6 +189,68 @@ describe('course-home apiHooks', () => {
       axiosMock.onGet(outlineUrl).reply(500);
       const { wrapper } = buildWrapper();
       const { result } = renderHook(() => useOutlineTabData('course-1'), { wrapper });
+
+      await waitFor(() => expect(result.current.isError).toBe(true));
+    });
+  });
+
+  describe('useProgressTabData', () => {
+    const progressUrl = `${getConfig().LMS_BASE_URL}/api/course_home/progress/course-1`;
+
+    it('transforms the server response', async () => {
+      axiosMock.onGet(progressUrl).reply(200, Factory.build('progressTabData'));
+      const { wrapper } = buildWrapper();
+      const { result } = renderHook(() => useProgressTabData('course-1'), { wrapper });
+
+      await waitFor(() => expect(result.current.isSuccess).toBe(true));
+      expect(result.current.data.studioUrl).toEqual('http://studio.edx.org/settings/grading/course-v1:edX+Test+run');
+      expect(result.current.data.gradesFeatureIsFullyLocked).toBe(false);
+    });
+
+    it('appends the targetUserId to the request URL', async () => {
+      axiosMock.onGet(`${progressUrl}/7/`).reply(200, Factory.build('progressTabData'));
+      const { wrapper } = buildWrapper();
+      const { result } = renderHook(() => useProgressTabData('course-1', '7'), { wrapper });
+
+      await waitFor(() => expect(result.current.isSuccess).toBe(true));
+      expect(axiosMock.history.get[0].url).toEqual(`${progressUrl}/7/`);
+    });
+
+    it.each([401, 403])(
+      'resolves to an empty object on a %s (access is handled via the metadata request)',
+      async (status) => {
+        axiosMock.onGet(progressUrl).reply(status, {});
+        const { wrapper } = buildWrapper();
+        const { result } = renderHook(() => useProgressTabData('course-1'), { wrapper });
+
+        await waitFor(() => expect(result.current.isSuccess).toBe(true));
+        expect(result.current.data).toEqual({});
+      },
+    );
+
+    it('redirects to the legacy progress page and resolves to an empty object on a 404', async () => {
+      // jsdom's location.replace is non-configurable, so we swap the whole location for the
+      // duration of this test (restored in finally so it can never bleed into another test).
+      const originalLocation = window.location;
+      const replace = jest.fn();
+      Object.defineProperty(window, 'location', { configurable: true, value: { replace } });
+      try {
+        axiosMock.onGet(progressUrl).reply(404, {});
+        const { wrapper } = buildWrapper();
+        const { result } = renderHook(() => useProgressTabData('course-1'), { wrapper });
+
+        await waitFor(() => expect(result.current.isSuccess).toBe(true));
+        expect(result.current.data).toEqual({});
+        expect(replace).toHaveBeenCalledWith(`${getConfig().LMS_BASE_URL}/courses/course-1/progress`);
+      } finally {
+        Object.defineProperty(window, 'location', { configurable: true, value: originalLocation });
+      }
+    });
+
+    it('surfaces the error on a non-handled failure', async () => {
+      axiosMock.onGet(progressUrl).reply(500);
+      const { wrapper } = buildWrapper();
+      const { result } = renderHook(() => useProgressTabData('course-1'), { wrapper });
 
       await waitFor(() => expect(result.current.isError).toBe(true));
     });
--- a/src/course-home/data/apiHooks.ts
+++ b/src/course-home/data/apiHooks.ts
@@ -7,6 +7,7 @@ import {
   getCourseHomeCourseMetadata,
   getDatesTabData,
   getOutlineTabData,
+  getProgressTabData,
   postCourseDeadlines,
   postDismissWelcomeMessage,
   postRequestCert,
@@ -72,6 +73,12 @@ export const useOutlineTabData = (courseId: string) => useQuery({
   queryKey: courseHomeQueryKeys.outlineTab(courseId),
   queryFn: () => getOutlineTabData(courseId),
   meta: { modelType: 'outline', courseId },
+});
+
+export const useProgressTabData = (courseId: string, targetUserId?: string) => useQuery({
+  queryKey: courseHomeQueryKeys.progressTab(courseId, targetUserId),
+  queryFn: () => getProgressTabData(courseId, targetUserId),
+  meta: { modelType: 'progress', courseId },
 });
 
 export const useRequestCert = () => useMutation({
--- a/src/course-home/data/index.js
+++ b/src/course-home/data/index.js
@@ -1,6 +1,3 @@
-export {
-  fetchProgressTab,
-  deprecatedSaveCourseGoal,
-} from './thunks';
+export { deprecatedSaveCourseGoal } from './thunks';
 
 export { reducer } from './slice';
--- a/src/course-home/data/queryKeys.ts
+++ b/src/course-home/data/queryKeys.ts
@@ -5,4 +5,5 @@ export const courseHomeQueryKeys = {
   metadata: (courseId: string) => [...courseHomeQueryKeys.all, 'metadata', courseId] as const,
   datesTab: (courseId: string) => [...courseHomeQueryKeys.all, 'datesTab', courseId] as const,
   outlineTab: (courseId: string) => [...courseHomeQueryKeys.all, 'outlineTab', courseId] as const,
+  progressTab: (courseId: string, targetUserId?: string) => [...courseHomeQueryKeys.all, 'progressTab', courseId, targetUserId] as const,
 };
--- a/src/course-home/data/redux.test.js
+++ b/src/course-home/data/redux.test.js
@@ -42,61 +42,6 @@ describe('Data layer integration tests', () => {
     store = initializeStore();
   });
 
-  describe('Test fetchProgressTab', () => {
-    const progressBaseUrl = `${getConfig().LMS_BASE_URL}/api/course_home/progress`;
-
-    it('Should result in fetch failure if error occurs', async () => {
-      axiosMock.onGet(courseMetadataUrl).networkError();
-      axiosMock.onGet(`${progressBaseUrl}/${courseId}`).networkError();
-
-      await executeThunk(thunks.fetchProgressTab(courseId), store.dispatch);
-
-      expect(loggingService.logError).toHaveBeenCalled();
-      expect(store.getState().courseHome.courseStatus).toEqual('failed');
-    });
-
-    it('Should fetch, normalize, and save metadata', async () => {
-      const progressTabData = Factory.build('progressTabData', { courseId });
-
-      const progressUrl = `${progressBaseUrl}/${courseId}`;
-
-      axiosMock.onGet(courseMetadataUrl).reply(200, courseHomeMetadata);
-      axiosMock.onGet(progressUrl).reply(200, progressTabData);
-
-      await executeThunk(thunks.fetchProgressTab(courseId), store.dispatch);
-
-      const state = store.getState();
-      expect(state.courseHome.courseStatus).toEqual('loaded');
-    });
-
-    it('Should handle the url including a targetUserId', async () => {
-      const progressTabData = Factory.build('progressTabData', { courseId });
-      const targetUserId = 2;
-      const progressUrl = `${progressBaseUrl}/${courseId}/${targetUserId}/`;
-
-      axiosMock.onGet(courseMetadataUrl).reply(200, courseHomeMetadata);
-      axiosMock.onGet(progressUrl).reply(200, progressTabData);
-
-      await executeThunk(thunks.fetchProgressTab(courseId, 2), store.dispatch);
-
-      const state = store.getState();
-      expect(state.courseHome.targetUserId).toEqual(2);
-    });
-
-    it.each([401, 403, 404])(
-      'should result in fetch denied for expected errors and failed for all others',
-      async (errorStatus) => {
-        const progressUrl = `${progressBaseUrl}/${courseId}`;
-        axiosMock.onGet(courseMetadataUrl).reply(200, courseHomeAccessDeniedMetadata);
-        axiosMock.onGet(progressUrl).reply(errorStatus, {});
-
-        await executeThunk(thunks.fetchProgressTab(courseId), store.dispatch);
-
-        expect(store.getState().courseHome.courseStatus).toEqual('denied');
-      },
-    );
-  });
-
   describe('Test saveCourseGoal', () => {
     it('Should save course goal', async () => {
       const goalUrl = `${getConfig().LMS_BASE_URL}/api/course_home/save_course_goal`;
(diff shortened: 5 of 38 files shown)


brian-smith-tcril:

🤖 _Comment via Claude Code._

brian-smith-tcril:
> 

addressed in https://github.com/openedx/frontend-app-learning/compare/18309da89020f7ad867a58aef166eee7f758d862..6209a71958254f3c20fb68bad4187bfd632e774b (part of a stack rebase, see https://github.com/openedx/frontend-app-learning/pull/1987)

arbrandes (approved):
Pre-approved, but found an issue you can fix at your leisure:

The progress-tab plugin-slot READMEs document a pattern that's now broken.  Fixable by updating the `useContextId()` examples in these five READMEs to read `courseId` from `useParams()`, or pass `courseId` into the slots as a prop the way `CourseOutlineTabNotificationsSlot` and `ProgressCertificateStatusSlot` already do:

- `src/plugin-slots/ProgressTabGradeBreakdownSlot/README.md`
- `src/plugin-slots/ProgressTabRelatedLinksSlot/README.md`
- `src/plugin-slots/ProgressTabCourseGradeSlot/README.md`
- `src/plugin-slots/ProgressTabCertificateStatusMainBodySlot/README.md`
- `src/plugin-slots/ProgressTabCertificateStatusSidePanelSlot/README.md`

brian-smith-tcril:
> The progress-tab plugin-slot READMEs document a pattern that's now broken. Fixable by updating the `useContextId()` examples in these five READMEs to read `courseId` from `useParams()`, or pass `courseId` into the slots as a prop the way `CourseOutlineTabNotificationsSlot` and `ProgressCertificateStatusSlot` already do:
> 
> * `src/plugin-slots/ProgressTabGradeBreakdownSlot/README.md`
> * `src/plugin-slots/ProgressTabRelatedLinksSlot/README.md`
> * `src/plugin-slots/ProgressTabCourseGradeSlot/README.md`
> * `src/plugin-slots/ProgressTabCertificateStatusMainBodySlot/README.md`
> * `src/plugin-slots/ProgressTabCertificateStatusSidePanelSlot/README.md`

addressed in https://github.com/openedx/frontend-app-learning/compare/82c3c3e09f2d7a872379b743997de0f440a1971c..c87bdc322b726ee3062ac9c5ed71fd21016029e6

PR #2005, opened 2026-08-19
refactor: convert the discussion tab from Redux to React Query

brian-smith-tcril:
## Summary

Convert the course-home **discussion tab** from Redux to React Query. Part of the Redux → React Query migration (#1946), stacked on the progress-tab conversion (#2001), below the live-tab conversion (#2002). Closes #2003.

The discussion tab is **metadata-only** (no tab-data endpoint), so it self-wraps `<TabWithTimer>` with just `useCourseHomeMeta` and needs no model-store bridge. The other `fetchDiscussionTab` caller (`CourseAccessErrorPage`) is converted too, so the thunk is deleted outright.

## What changed

- **`tab-page/TabPage.tsx`** — `tabDataQuery` is now optional in `CourseStatus`; `deriveView` guards its two `tabDataQuery` checks. A metadata-only tab passes `courseStatus={{ metadataQuery }}`; the data-bearing tabs still pass both.
- **`course-home/discussion-tab/DiscussionTab.jsx`** — self-wrapping: a thin data-owning `DiscussionTab` (runs `useCourseHomeMeta`, renders `<TabWithTimer>`) + a `DiscussionTabContent` child holding the iframe and the `useIFrameHeight`/`useIFramePluginEvents('discussions.navigate')` wiring, so it mounts post-load. `courseId`/`path` from `useParams`.
- **`generic/CourseAccessErrorPage.jsx`** — the other `fetchDiscussionTab` caller (a metadata-only access probe) moves to `useCourseHomeMeta`; its four-state `courseStatus` branch maps 1:1 to the query states.
- **`index.jsx`** — discussion route → bare `<DiscussionTab />`; `fetchDiscussionTab` import/wiring dropped.
- **`course-home/data/thunks.js`** — `fetchDiscussionTab` deleted (both callers converted). The shared `fetchTab` helper and `fetchLiveTab` stay — the live tab consumes them until #2002, where `fetchTab` retires.

## Behavior

No user-facing change. No `discussion` model bridge is added: `fetchDiscussionTab` never wrote a `discussion` model, so `useModel('discussion')` was already empty and the access-expiration masquerade banner never rendered on this tab — preserved exactly. `TabWithTimer` is retained, so the outer exam timer still mounts as it did via `TabContainer`.

## Testing

[...]

PR #2006, opened 2026-08-19
refactor: convert the live tab from Redux to React Query

### LINKED ISSUE #2002: Convert the live tab to React Query
Part of #1975 (convert course-home tab data to React Query), Phase 3 of the epic #1946. The **top** course-home tab layer — stacks on the discussion-tab conversion (#2003), which sits above progress (#2001) — reusing the self-wrapping pattern the dates (#1987), outline (#1997), and progress (#2001) tabs set.

## Summary

Convert the live tab off Redux to React Query, following the **self-wrapping** pattern: `LiveTab` renders `<TabWithTimer>` itself and owns its data via `useCourseHomeMeta` + a new `useLiveTabData` hook. The iframe HTML moves into a `LiveTabContent` child so its styling `useEffect` runs when the content mounts (post-load), not when the outer tab mounts (mid-load). The `index.jsx` route drops its `<TabContainer tab="lti_live" fetch={fetchLiveTab} slice="courseHome">` wrapper for a bare `<LiveTab />`, and `fetchLiveTab` is deleted.

## No transitional bridge needed

Unlike progress (which keeps a scoped `meta` bridge for the masquerade banner), `useLiveTabData` needs **no `meta` tag**:

- The **only** reader of `state.models.live` is `LiveTab` itself (`state.models.live[courseId]?.iframe`), and it moves to the query.
- The shared masquerade banner (`useAccessExpirationMasqueradeBanner(courseId, tab)` → `useModel(tab, courseId)`) is passed `tab="lti_live"` on this route, but `fetchLiveTab` writes its data under model **`live`**, not `lti_live` — so that `useModel('lti_live', …)` already resolves empty and the banner never renders on the live tab today. The conversion preserves that exactly (the `TabWithTimer activeTabSlug` stays `"lti_live"`), so nothing needs the bridge.

So this tab retires its Redux with no `#1999`-style holdout — cleaner than progress.

## fetchTab retires here

This is the last course-home tab on the shared `fetchTab` helper. Discussion (#2003, below) already removed the other consumer, so deleting `fetchLiveTab` here orphans `fetchTab` entirely — both go together, and the `redux.test.js` `Test fetchTab` block is removed wholesale. Ordering discussion below live is what makes this a clean deletion instead of retargeting that test block from `fetchLiveTab` to `fetchDiscussionTab` and back.

## Tasks

- [ ] `queryKeys.ts` — add `liveTab(courseId)`.
- [ ] `apiHooks.ts` — add `useLiveTabData(courseId)` (`getLiveTabIframe`, **no `meta`**).
- [ ] `LiveTab.jsx` — self-wrap `TabWithTimer` (`activeTabSlug="lti_live"`, `courseStatus={{ metadataQuery, tabDataQuery }}`); split the iframe body + its `getElementById('lti-tab-embed')` `useEffect` into a `LiveTabContent` child; `courseId` from `useParams`.
- [ ] `index.jsx` — live route → `<LiveTab />`; drop the `fetchLiveTab` import/wiring.
- [ ] Delete `fetchLiveTab` **and** the now-orphaned shared `fetchTab` helper (discussion, its last other consumer, converted in #2003).
- [ ] Tests: `LiveTab.test.jsx` renders `<LiveTab />` directly through the bridged query client (mirroring `ProgressTab.test.jsx`); add a `useLiveTabData` block to `apiHooks.test.tsx` (success/iframe + 404→`{}`); **delete** the `Test fetchTab` block from `redux.test.js` wholesale (its subject `fetchTab` is deleted here).

## Notes / decisions

- **The `LiveTab`/`LiveTabContent` split is required, not cosmetic.** Today `LiveTab` is `TabContainer`'s child, so it mounts only once the tab data is loaded — its mount `useEffect` finds the injected `#lti-tab-embed` iframe and applies the sizing classes. In the self-wrapping shape the data-owning component mounts *before* load, so the effect must live in the gated `LiveTabContent` child to keep the same timing.
- **`fetchTab` is deleted here, not retargeted.** Ordering discussion below live (#2003) means `fetchLiveTab` is `fetchTab`'s last consumer; removing it orphans the helper, so both go together and the `Test fetchTab` block is removed wholesale — no back-and-forth churn in `redux.test.js`.

> [!NOTE]

<details>
<summary>Claude Plan — live tab → React Query</summary>

### Approach

A self-wrapping conversion mirroring dates/outline/progress: the page owns its data via query hooks and renders `<TabWithTimer>` directly; `TabContainer` stops wrapping this route. No model-store bridge — the sole `state.models.live` reader is the tab itself, and the masquerade banner never resolves to this tab's model (route slug `lti_live` ≠ data model `live`), so the conversion is behavior-preserving with the query as the only cache.

### Data layer

```ts
// queryKeys.ts
liveTab: (courseId: string) => [...courseHomeQueryKeys.all, 'liveTab', courseId] as const,

// apiHooks.ts
export const useLiveTabData = (courseId: string) => useQuery({
  queryKey: courseHomeQueryKeys.liveTab(courseId),
  queryFn: () => getLiveTabIframe(courseId),
});
```

### `LiveTab.jsx`

```jsx
const LiveTab = () => {
  const tabDataQuery = useLiveTabData(courseId);
  return (
    <TabWithTimer
      activeTabSlug="lti_live"
      courseId={courseId}
    >
      <LiveTabContent html={tabDataQuery.data?.iframe} />
    </TabWithTimer>
  );
};

// LiveTabContent: the dangerouslySetInnerHTML div (prop named `html` to avoid shadowing
// the effect's `iframe` DOM node) + the existing getElementById('lti-tab-embed') sizing
// useEffect (renders only once loaded).
```

</details>

brian-smith-tcril:
## Summary

Convert the course-home **live tab** from Redux to React Query. Part of the Redux → React Query migration (#1946), stacked on the discussion-tab conversion (#2005) as the **top** course-home tab layer. Closes #2002.

The live tab has real tab data (the LTI iframe HTML) but needs no model-store bridge — its only reader was the tab itself. It's the **last** course-home tab on the shared `fetchTab` helper, so this layer also retires `fetchTab`.

## What changed

- **`course-home/data/queryKeys.ts` / `apiHooks.ts`** — add `useLiveTabData(courseId)` (`getLiveTabIframe`, **no `meta`** tag) and the `liveTab(courseId)` key.
- **`course-home/live-tab/LiveTab.jsx`** — self-wrapping: a thin data-owning `LiveTab` (runs `useCourseHomeMeta` + `useLiveTabData`, renders `<TabWithTimer>`) + a `LiveTabContent` child holding the `dangerouslySetInnerHTML` div and the `getElementById('lti-tab-embed')` sizing `useEffect`, so it mounts post-load. `courseId` from `useParams`.
- **`index.jsx`** — live route → bare `<LiveTab />`; the `fetchLiveTab` import/wiring dropped.
- **`course-home/data/thunks.js`** — delete `fetchLiveTab` **and** the now-orphaned shared `fetchTab` helper (discussion, its other consumer, was converted in #2005), plus their now-unused imports. `eventTypes`, `deprecatedSaveCourseGoal`, and `fetchExamAttemptsData` remain.
- **`course-home/data/slice.js`** — remove the three now-dead reducers orphaned by the `fetchTab` deletion (`fetchTabRequest` / `fetchTabDenied` / `fetchTabSuccess`); `fetchTabFailure` stays (still dispatched by `TabPage.test.jsx` for the `errorMessage` display).

## Behavior

No user-facing change. No `live` model bridge: the masquerade banner reads `useModel('lti_live')`, but `fetchLiveTab` wrote model `live` (slug ≠ model), so that read was already empty and the banner never rendered here — preserved. `TabWithTimer` is retained, so the outer exam timer still mounts as it did via `TabContainer`. The three `fetchTab` request/denied/success reducers in `slice.js`, orphaned by the deletion, are removed; `fetchTabFailure` stays (still exercised by `TabPage.test.jsx`) and retires with the `courseHome` reducer in #1975.

## Testing

`npm run types`, `npm run lint`, and the affected suites pass. New `LiveTab.test.jsx` renders `<LiveTab />` through the bridged query client; `apiHooks.test.tsx` gains a `useLiveTabData` block (success / 404→`{}` / error); `redux.test.js` drops the `Test fetchTab` block wholesale (its subject `fetchTab` is deleted here). Manual browser verification (HAR-confirmed clean load) is documented in the decision log below.

## Decisions

<details>

# Decisions — Redux → React Query: the live tab (#1946)

(course-home tab data), stacked on the discussion-tab conversion (#2005) as the
**top** course-home tab layer. Closes #2002.

The live tab has real tab data (the LTI iframe HTML) but still needs **no
model-store bridge** — its only reader was the tab itself. It's the **last**
course-home tab on the shared `fetchTab` helper, so this layer also retires
`fetchTab`.

## Scope: self-wrapping conversion

**Decision.** `LiveTab` becomes self-wrapping — it owns its data via
`useCourseHomeMeta` + a new `useLiveTabData` hook and renders `<TabWithTimer>`
itself. The `index.jsx` route drops `<TabContainer tab="lti_live"
fetch={fetchLiveTab} slice="courseHome">` for a bare `<LiveTab />`.

## No transitional bridge (cleaner than progress)

**Decision.** `useLiveTabData` carries **no `meta` tag**:

- The only reader of `state.models.live` is `LiveTab` itself
  (`state.models.live[courseId]?.iframe`), which moves to the query.
- The shared masquerade banner reads `useModel(tab, courseId)` with
  `tab="lti_live"`, but `fetchLiveTab` wrote its data under model **`live`**, not
  `lti_live` — so `useModel('lti_live', …)` already resolved empty and the banner
  never rendered on this tab. `TabWithTimer activeTabSlug` stays `"lti_live"`, so
  that's preserved exactly. No `#1999`-style holdout.

## `LiveTab` splits into a thin wrapper + `LiveTabContent`

**Decision.** `LiveTab` runs the two queries and renders `<TabWithTimer
courseStatus={{ metadataQuery, tabDataQuery }}><LiveTabContent /></TabWithTimer>`;
the iframe `<div>` **and** its `getElementById('lti-tab-embed')` sizing `useEffect`
move to `LiveTabContent`. `TabPage` renders children only once loaded, so putting
the effect in the gated child preserves today's timing (where `LiveTab` was
`TabContainer`'s already-loaded child); in the self-wrapping shape the data-owning
component mounts *before* load.

## The iframe HTML is a prop named `html`, not `iframe`

**Decision.** `LiveTabContent` takes `html` (the iframe markup) and renders it via
`dangerouslySetInnerHTML`. It's named `html` rather than `iframe` so it doesn't
shadow the effect's `const iframe = document.getElementById('lti-tab-embed')`
DOM-node variable — which keeps that effect byte-identical to the original.

## `getLiveTabIframe` reused unchanged

**Decision.** `useLiveTabData`'s `queryFn` is the existing `getLiveTabIframe`
(`GET /api/course_live/iframe/<courseId>/`; 404 → `{}`, else throw), reused
as-is. Its name doesn't match the `get<Tab>TabData` siblings; a rename was
considered and **rejected** as out-of-scope churn for a conversion.

## `fetchTab` retires here (the last consumer)

**Decision.** Deleting `fetchLiveTab` leaves the shared `fetchTab` with no callers
(discussion, its other consumer, was converted in #2005). So `fetchTab` and its
now-unused imports (`getCourseHomeCourseMetadata`, `addModel`,
`fetchTab{Request,Denied,Failure,Success}`) are removed from `thunks.js`. What's
left there is `eventTypes`, `deprecatedSaveCourseGoal`, and `fetchExamAttemptsData`.

**Slice cleanup:** deleting `fetchTab` orphaned three of its reducers —
`fetchTabRequest`, `fetchTabDenied`, and `fetchTabSuccess` had no remaining
dispatcher, so they're removed from `slice.js` (along with the now-unused
`LOADING` / `DENIED` imports). `fetchTabFailure` **stays**: `TabPage.test.jsx`
still dispatches it to exercise `TabPage`'s `errorMessage` display. The residual
`courseStatus` / `errorMessage` fields retire with the `courseHome`-reducer
teardown (#1975), not this layer.

## `redux.test.js`: `Test fetchTab` block removed wholesale

**Decision.** With `fetchTab` gone, the `Test fetchTab` block (which drove it via
`fetchLiveTab`) is removed entirely, along with its now-orphaned `courseMetadataUrl`
/ `courseHomeAccessDeniedMetadata` declarations and the `appendBrowserTimezoneToUrl`
import. This is the clean, single deletion the below-live ordering was chosen to
enable — no retarget churn.

## Tests

- `LiveTab.test.jsx` (new) renders `<LiveTab />` through the bridged query client
  and asserts the iframe HTML from `useLiveTabData` renders — exercising `LiveTab`,
  `LiveTabContent`, and the sizing effect.
- `apiHooks.test.tsx` adds a `useLiveTabData` block: success (iframe payload),
  404 → `{}`, and non-404 → error. The 404/500 mocks attach `customAttributes`
  (via `Object.assign` on a real `Error`, so it's typed and lint-clean) because
  `getLiveTabIframe` reads `error.customAttributes.httpErrorStatus` — matching the
  approach in `api.test.js`.
- `redux.test.js` — `Test fetchTab` block removed (see above).

## Manual testing (in-browser)

**Verified** (no LTI live provider configured — the realistic demo case): a direct,
cold load of `/course/<id>/live` renders cleanly. The HAR shows
`GET /api/course_home/course_metadata/…` → 200 (`useCourseHomeMeta`) and
`GET /api/course_live/iframe/<id>/` → 200 (`useLiveTabData`), the latter with an
empty payload, so `#live_tab` renders empty — no crash, full page render. The
`OuterExamTimer` path fired (`…/proctored_exam/attempt/…` → 200), confirming
`TabWithTimer` is preserved. Zero 4xx/5xx, and nothing hit a `fetchTab` /
`courseStatus` / `courseware/sequence` path — i.e. deleting the shared `fetchTab`
helper broke nothing on this render.

**Not exercised:** a fully-configured LTI live provider (visible iframe content +
the `#lti-tab-embed` sizing effect) — none available locally. The automated
`LiveTab.test.jsx` covers the iframe-render + effect path with a mocked payload.

</details>

### DIFF
--- a/src/course-home/data/apiHooks.ts
+++ b/src/course-home/data/apiHooks.ts
@@ -6,6 +6,7 @@ import {
   executePostFromPostEvent,
   getCourseHomeCourseMetadata,
   getDatesTabData,
+  getLiveTabIframe,
   getOutlineTabData,
   getProgressTabData,
   postCourseDeadlines,
@@ -73,6 +74,12 @@ export const useOutlineTabData = (courseId: string) => useQuery({
   queryKey: courseHomeQueryKeys.outlineTab(courseId),
   queryFn: () => getOutlineTabData(courseId),
   meta: { modelType: 'outline', courseId },
+});
+
+export const useLiveTabData = (courseId: string) => useQuery({
+  queryKey: courseHomeQueryKeys.liveTab(courseId),
+  queryFn: () => getLiveTabIframe(courseId),
+  refetchOnWindowFocus: false,
 });
 
 export const useProgressTabData = (courseId: string, targetUserId?: string) => useQuery({
--- a/src/course-home/live-tab/LiveTab.jsx
+++ b/src/course-home/live-tab/LiveTab.jsx
@@ -1,21 +1,48 @@
 import React, { useEffect } from 'react';
-import { useSelector } from 'react-redux';
+import PropTypes from 'prop-types';
+import { useParams } from 'react-router-dom';
 
-const LiveTab = () => {
-  const { courseId } = useSelector(state => state.courseHome);
-  const liveModel = useSelector(state => state.models.live);
+import { useCourseHomeMeta, useLiveTabData } from '../data/apiHooks';
+import { TabWithTimer } from '../../tab-page';
+
+const LiveTabContent = ({ html }) => {
   useEffect(() => {
     const iframe = document.getElementById('lti-tab-embed');
     if (iframe) {
       iframe.className += ' vh-100 w-100 border-0';
     }
-  }, []);
+  }, [html]);
   return (
     <div
       id="live_tab"
       // eslint-disable-next-line react/no-danger
-      dangerouslySetInnerHTML={{ __html: liveModel[courseId]?.iframe }}
+      dangerouslySetInnerHTML={{ __html: html }}
     />
+  );
+};
+
+LiveTabContent.propTypes = {
+  html: PropTypes.string,
+};
+
+LiveTabContent.defaultProps = {
+  html: undefined,
+};
+
+const LiveTab = () => {
+  const { courseId } = useParams();
+
+  const metadataQuery = useCourseHomeMeta(courseId);
+  const tabDataQuery = useLiveTabData(courseId);
+
+  return (
+    <TabWithTimer
+      activeTabSlug="lti_live"
+      courseId={courseId}
+      courseStatus={{ metadataQuery, tabDataQuery }}
+    >
+      <LiveTabContent html={tabDataQuery.data?.iframe} />
+    </TabWithTimer>
   );
 };
 
--- a/src/course-home/data/apiHooks.test.tsx
+++ b/src/course-home/data/apiHooks.test.tsx
@@ -8,7 +8,7 @@ import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 import { initializeMockApp } from '../../setupTest';
 import { ToastProvider, useToast } from '../../generic/ToastContext';
 import {
-  useOutlineTabData, useProgressTabData, useResetDeadlines, usePostEvent, useRequestCert,
+  useOutlineTabData, useLiveTabData, useProgressTabData, useResetDeadlines, usePostEvent, useRequestCert,
   useDismissWelcomeMessage, useSaveWeeklyLearningGoal,
 } from './apiHooks';
 
@@ -189,6 +189,46 @@ describe('course-home apiHooks', () => {
       axiosMock.onGet(outlineUrl).reply(500);
       const { wrapper } = buildWrapper();
       const { result } = renderHook(() => useOutlineTabData('course-1'), { wrapper });
+
+      await waitFor(() => expect(result.current.isError).toBe(true));
+    });
+  });
+
+  describe('useLiveTabData', () => {
+    const liveUrl = `${getConfig().LMS_BASE_URL}/api/course_live/iframe/course-1/`;
+
+    it('resolves to the iframe payload', async () => {
+      axiosMock.onGet(liveUrl).reply(200, { iframe: 'https://example.com/live' });
+      const { wrapper } = buildWrapper();
+      const { result } = renderHook(() => useLiveTabData('course-1'), { wrapper });
+
+      await waitFor(() => expect(result.current.isSuccess).toBe(true));
+      expect(result.current.data).toEqual({ iframe: 'https://example.com/live' });
+    });
+
+    it('resolves to an empty object on a 404', async () => {
+      axiosMock.onGet(liveUrl).reply(() => Promise.reject(
+        Object.assign(new Error('Request failed with status code 404'), {
+          response: { status: 404, data: {} },
+          customAttributes: { httpErrorStatus: 404 },
+        }),
+      ));
+      const { wrapper } = buildWrapper();
+      const { result } = renderHook(() => useLiveTabData('course-1'), { wrapper });
+
+      await waitFor(() => expect(result.current.isSuccess).toBe(true));
+      expect(result.current.data).toEqual({});
+    });
+
+    it('surfaces the error on a non-404 failure', async () => {
+      axiosMock.onGet(liveUrl).reply(() => Promise.reject(
+        Object.assign(new Error('Request failed with status code 500'), {
+          response: { status: 500, data: {} },
+          customAttributes: { httpErrorStatus: 500 },
+        }),
+      ));
+      const { wrapper } = buildWrapper();
+      const { result } = renderHook(() => useLiveTabData('course-1'), { wrapper });
 
       await waitFor(() => expect(result.current.isError).toBe(true));
     });
--- a/src/course-home/data/queryKeys.ts
+++ b/src/course-home/data/queryKeys.ts
@@ -5,5 +5,6 @@ export const courseHomeQueryKeys = {
   metadata: (courseId: string) => [...courseHomeQueryKeys.all, 'metadata', courseId] as const,
   datesTab: (courseId: string) => [...courseHomeQueryKeys.all, 'datesTab', courseId] as const,
   outlineTab: (courseId: string) => [...courseHomeQueryKeys.all, 'outlineTab', courseId] as const,
+  liveTab: (courseId: string) => [...courseHomeQueryKeys.all, 'liveTab', courseId] as const,
   progressTab: (courseId: string, targetUserId?: string) => [...courseHomeQueryKeys.all, 'progressTab', courseId, targetUserId] as const,
 };
--- a/src/course-home/data/redux.test.js
+++ b/src/course-home/data/redux.test.js
@@ -6,7 +6,7 @@ import { getConfig } from '@edx/frontend-platform';
 
 import * as thunks from './thunks';
 
-import { appendBrowserTimezoneToUrl, executeThunk } from '../../utils';
+import { executeThunk } from '../../utils';
 
 import { initializeMockApp } from '../../setupTest';
 import initializeStore from '../../store';
@@ -18,20 +18,6 @@ const axiosMock = new MockAdapter(getAuthenticatedHttpClient());
 describe('Data layer integration tests', () => {
   const courseHomeMetadata = Factory.build('courseHomeMetadata');
   const { id: courseId } = courseHomeMetadata;
-  let courseMetadataUrl = `${getConfig().LMS_BASE_URL}/api/course_home/course_metadata/${courseId}`;
-  courseMetadataUrl = appendBrowserTimezoneToUrl(courseMetadataUrl);
-
-  const courseHomeAccessDeniedMetadata = Factory.build(
-    'courseHomeMetadata',
-    {
-      id: courseId,
-      course_access: {
-        has_access: false,
-        error_code: 'bad codes',
-        additional_context_user_message: 'your Codes Are BAD',
-      },
-    },
-  );
 
   let store;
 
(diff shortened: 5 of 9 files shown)


brian-smith-tcril:

🤖 _Comment via Claude Code._

brian-smith-tcril:
> 

addressed in https://github.com/openedx/frontend-app-learning/compare/5005683e0cee3acad4dbd29809bd6b8d70418d67..4568fedb91eb6f7e926ca5c7a40057968d1226fe (part of a stack rebase, see https://github.com/openedx/frontend-app-learning/pull/1987)

src/course-home/data/apiHooks.ts
```diff
@@ -75,6 +76,11 @@ export const useOutlineTabData = (courseId: string) => useQuery({
   meta: { modelType: 'outline', courseId },
 });
 
+export const useLiveTabData = (courseId: string) => useQuery({
+  queryKey: courseHomeQueryKeys.liveTab(courseId),
+  queryFn: () => getLiveTabIframe(courseId),
+});
```
arbrandes, line 83: I think we might need to add `refetchOnWindowFocus: false` to `useLiveTabData`.

The payload isn't stable across requests, it looks like. (The LMS builds the iframe `srcdoc` from `lti_embed`, so every response carries a fresh `oauth_nonce`/`oauth_timestamp`/`oauth_signature`). React compares `dangerouslySetInnerHTML.__html` by value, so a differing string means the div's contents are replaced and the `#lti-tab-embed` iframe is destroyed and relaunched. With current React Query's defaults, that fires whenever the learner switches windows and comes back (which could drop, say, an in-progress meeting embedded in the tab). The old thunk fetched once per navigation, so this is new.

src/course-home/live-tab/LiveTab.jsx
```diff
@@ -1,9 +1,11 @@
...
-const LiveTab = () => {
-  const { courseId } = useSelector(state => state.courseHome);
-  const liveModel = useSelector(state => state.models.live);
+import { useCourseHomeMeta, useLiveTabData } from '../data/apiHooks';
+import { TabWithTimer } from '../../tab-page';
+
+const LiveTabContent = ({ html }) => {
   useEffect(() => {
     const iframe = document.getElementById('lti-tab-embed');
     if (iframe) {
```
arbrandes, line 11: Change the dependency array to `[html]`.  Same root cause.

arbrandes (changes requested):
Looks like we might have an actual blocker, on this one.  Let me know.

src/course-home/data/apiHooks.ts
brian-smith-tcril: addressed (and added tests to cover it) in https://github.com/openedx/frontend-app-learning/compare/420557299201b3a155260f6f9e0be33c530af344..a61c3f6222ae8091c1ef70810e831a2b8d6f8913

src/course-home/live-tab/LiveTab.jsx
brian-smith-tcril: addressed (and added tests to cover it) in https://github.com/openedx/frontend-app-learning/compare/420557299201b3a155260f6f9e0be33c530af344..a61c3f6222ae8091c1ef70810e831a2b8d6f8913

brian-smith-tcril:
Decision log for the review fixes, written by Claude 🤖:

<details>
<summary>Decisions — #2006 live-tab review: LTI iframe stability (TDD)</summary>

Working notes for addressing arbrandes's CHANGES_REQUESTED review on #2006 (convert
the live tab from Redux to React Query). Not checked in.

## The blocker (arbrandes)

Two related review comments, one root cause — the LTI payload is **not stable across
requests**. The LMS builds the iframe `srcdoc` from `lti_embed`, so every response
carries a fresh `oauth_nonce`/`oauth_timestamp`/`oauth_signature`. `LiveTabContent`
renders it via `dangerouslySetInnerHTML`, and React compares `__html` by value — a
differing string replaces the div's contents, so the `#lti-tab-embed` iframe is
**destroyed and relaunched**. Any spurious refetch therefore drops an in-progress
meeting embedded in the tab.

The old Redux thunk fetched **once per navigation**, so this never happened. React
Query's defaults reintroduce it in two ways:

1. **`refetchOnWindowFocus` (default `true`)** — switching away from the browser and
   back refetches the (stale, `staleTime: 0`) live query, producing a new payload →
   iframe relaunched. `src/course-home/data/apiHooks.ts:82` — fix: add
   `refetchOnWindowFocus: false` to `useLiveTabData`.
2. **The class-styling effect's `[]` dependency** — `LiveTabContent` adds
   `vh-100 w-100 border-0` to the iframe once on mount. When the html legitimately
   changes and a *new* iframe node is created (navigation, invalidation, reconnect),
   the effect doesn't re-run, so the fresh iframe loses its sizing classes.
   `src/course-home/live-tab/LiveTab.jsx` — fix: change the dep array to `[html]`.

The two fixes are complementary: (1) stops the *unwanted* refetch; (2) keeps styling
correct for the *wanted* html changes.

## Approach: TDD

Per the review, write tests that reproduce each failure first (RED), then apply the
suggested changes and confirm they pass (GREEN) — rather than making the changes blind.

### Test 1 — no refetch on window focus
`does not refetch the live embed when the window regains focus`. Renders LiveTab,
waits for the embed, records the live-endpoint call count, drives a focus transition
via `focusManager.setFocused(false) → setFocused(true)` inside `act`, and asserts the
call count is unchanged. `createTestQueryClient` only sets `retry: false`, so RQ's
default `refetchOnWindowFocus: true` + `staleTime: 0` apply and the query is stale at
focus time. `afterEach` resets `focusManager.setFocused(undefined)` so the global
focus state doesn't leak between tests.

### Test 2 — sizing classes survive an html change
`re-applies the sizing classes when the iframe html changes`. Serves iframe A first
(`replyOnce`) then iframe B, waits for A + asserts it has the classes, forces a refetch
with `queryClient.invalidateQueries()`, waits for B, and asserts B *also* has the
classes. Distinguishes A/B by `src` so the assertion targets the recreated node.

## Log

- **RED (confirmed)** — both new tests fail against the unfixed code, each for the
  intended reason:
  - Test 1 (`does not refetch the live embed when the window regains focus`):
    `expect(liveCallCount()).toBe(1)` → `Expected: 1, Received: 2`. The focus
    transition triggered a second GET to the live endpoint — the refetch that would
    relaunch the iframe.
  - Test 2 (`re-applies the sizing classes when the iframe html changes`): after the
    payload changes to iframe B, `toHaveClass('vh-100 w-100 border-0')` fails with an
    empty class list. The `src=live-b` `waitFor` passed first, proving the iframe node
    was recreated; the `[]`-dep effect didn't re-run, so the new node lost its classes.
- **GREEN (confirmed)** — applied both fixes and reran; all 3 tests pass (the two new
  ones + the original render test):
  - `src/course-home/data/apiHooks.ts` — added `refetchOnWindowFocus: false` to
    `useLiveTabData`. Test 1 now records a single live-endpoint call after the focus
    transition.
  - `src/course-home/live-tab/LiveTab.jsx` — changed the styling effect's dep array
    from `[]` to `[html]`. Test 2's recreated iframe (B) now carries the sizing classes.
  - (The `404` AxiosErrors in the run output are pre-existing console noise from other
    unmocked endpoints rendered by `TabWithTimer`/`InstructorToolbar` via `logError`,
    not test failures — present before these changes too.)
- **Fix-isolation matrix** — reverted each fix independently to confirm each test maps
  1:1 to its own fix, with no cross-coupling:

  | apiHooks `refetchOnWindowFocus:false` | LiveTab dep | Test 1 (focus) | Test 2 (classes) |
  | --- | --- | --- | --- |
  | ✗ | `[]`     | fail | fail |
  | ✗ | `[html]` | fail | **pass** |
  | ✓ | `[]`     | **pass** | fail |
  | ✓ | `[html]` | pass | pass |

  Test 1 tracks only the `refetchOnWindowFocus` fix (it drives focus, unaffected by the
  dep array); Test 2 tracks only the `[html]` dep (it drives an `invalidateQueries`
  refetch, unaffected by the focus setting). Both fixes are therefore necessary and
  each is independently verified.

</details>

arbrandes (approved):
Thanks for addressing!  👍🏼

Issue #2019, opened 2026-08-20
Peel: convert CoursewareContainer to TypeScript (fast-follow to de-class #2008)


brian-smith-tcril:
Part of #1946 — Redux → React Query migration (Stage 1). Part of the #1976 courseware decomposition ([plan](https://github.com/openedx/frontend-app-learning/issues/1946#issuecomment-5348891139)) — **fast-follow to #2008**, stacked directly above the de-class and below the RQ peels (#2010+).

**Goal:** convert `src/courseware/CoursewareContainer.jsx` to `.tsx` immediately after it's de-classed (#2008), so the courseware React Query peels all happen in TypeScript-land.

**Why a separate PR from #2008:** the de-class is already a churn-y structural diff; layering a rename + type annotations on top makes it harder to review. Keeping them as two stacked PRs lets each review as a single idea.

**Approach — don't type the doomed Redux stuff.** The Redux-sourced selectors (`currentCourseSelector`, etc.) and the redirect-helper positional args are about to be deleted/replaced by the RQ peels, so type them loosely (`any`/minimal) now and fill in real types per-peel as each selector becomes a typed `useQuery` return.

**Tasks**
- Rename `CoursewareContainer.jsx` → `.tsx`; add the minimum annotations to compile.
- `any`/loose-type the Redux-sourced values as needed (in practice reselect infers the redirect helpers + `createSelector` selectors, so only the two internal refs need `useRef<any>`).

> ~~Add a reusable `AppDispatch` type + `useAppDispatch` hook to `src/store.ts`~~ — **dropped.** Typed thunk dispatch only serves the Redux we're deleting (RQ code doesn't dispatch), and plain `useDispatch()` type-checks fine since the thunks are still JS. `store.ts` is left untouched. See the PR decision log.

**Verify:** `npm run types` + `npm run lint` clean; `CoursewareContainer.test.jsx` still green; no behavior change.

*Scoped-but-light; a full plan gets added when this is picked up.*

Issue #2016, opened 2026-08-20
Convert getCourseDiscussionTopics to React Query


brian-smith-tcril:
Part of #1946 — Redux → React Query migration (Stage 1). Part of the #1976 courseware decomposition ([plan](https://github.com/openedx/frontend-app-learning/issues/1946#issuecomment-5348891139)) — **Target 5 (leaf)**. Independent — can slot anywhere in the courseware stack.

**Goal:** convert the discussion-topics fetch to React Query.

**Tasks**
- Convert `getCourseDiscussionTopics` → query (writes the `discussionTopics` model; `getCourseDiscussionConfig` gate for the `openedx` provider preserved).
- Readers (`SidebarContextProvider`, `DiscussionsSidebar`, `DiscussionsTrigger`) read from the query / bridge.

**Verify:** discussion topics load in the sidebar for openedx-provider courses; legacy-provider courses unaffected.

## Plan

> [!NOTE]
> The findings and plan below were generated by Claude (Claude Code) and reviewed before posting.

Investigation findings that adjust the task list above:

- **A prefetch function, not a hook.** No reader converts in this issue (`useModel` → query-result read conversions are #1977's home), and the thunk's only dispatcher is imperative — the widget-registry `prefetch` effect in `SidebarContextProvider` — so `courseware/data/apiHooks.ts` gains `prefetchDiscussionTopics(queryClient, courseId)` wrapping `queryClient.prefetchQuery(...)` rather than a `useQuery` hook. The three `useModel('discussionTopics', unitId)` readers are untouched and keep working through the bridge.
- **The provider gate moves inside `queryFn`.** Same two-call sequence as the thunk: fetch the discussion config, and only for `config.provider === 'openedx'` fetch topics (comment and `usageKey` filter preserved). Non-openedx resolves to `[]`, which bridges as a no-op — the same end state as the thunk's dispatch-nothing path. `prefetchQuery` never rejects; failures log through the app QueryCache `onError` → `logError`, the thunk's catch behavior.
- **The model-store bridge gains an `idField` pass-through.** The `discussionTopics` model is keyed by `usageKey` (the unit id — each topic keeps its own `id`, the discussion topic id, which readers also check), and `ModelMirror` can't express a non-`id` key today. The model-store slice already accepts `idField` on every add/update action, so the bridge is the only gap.
[...]

Issue #1976, opened 2026-08-05
Tear down the courseware Redux slice + replace useContextId


brian-smith-tcril:
Part of #1946 — Redux → React Query migration (Stage 1). Final teardown of the courseware conversion (**Target 6** in the [decomposition plan](https://github.com/openedx/frontend-app-learning/issues/1946#issuecomment-5348891139)).

> This issue was originally "Convert courseware/data to React Query + de-class CoursewareContainer" — the too-big item. It has been **decomposed into direct sub-issues of #1946** and re-scoped to just the teardown at the end. The rest of the courseware conversion is carried by these siblings, in dependency order:
>
> 1. #2008 — peel: de-class `CoursewareContainer` (structural)
> 2. #2019 — peel: convert `CoursewareContainer` to TypeScript (fast-follow to #2008)
> 3. #2009 — peel: extend the model-store bridge to collections
> 4. #2010 — courseware metadata → RQ
> 5. #2011 — sequence data → RQ
> 6. #2012 — peel: `checkBlockCompletion` → mutation
> 7. #2013 — outline sidebar → RQ + context
> 8. #2014 — bookmarking → RQ + de-class `UnitButton`
> 9. #2015 — `saveIntegritySignature` + `saveSequencePosition` → mutations
> 10. #2016 — `getCourseDiscussionTopics` → RQ

**Goal (this issue):** after the siblings above land, remove the `courseware` Redux slice.

**Tasks**
- Migrate the residual `state.courseware.*` reads (`courseId` / `courseStatus` / `sequenceId` / `sequenceStatus`) to query state / `useParams`. `courseware.courseId` is kept *written* transitionally by the earlier layers (it's a read-hub, not a shared writer) and is removed here.
- Replace `useContextId` (`src/data/hooks.ts`, reads `state.courseware.courseId ?? state.courseHome.courseId`) with a route/params- or context-based courseId.
- Remove the `courseware` reducer from `store.ts` once no reader remains.

**Coordinate with #1977** (model-store dissolution) — this is where the two meet.

**Verify:** no `state.courseware.*` references remain; `useContextId` callers work off the replacement; suite green.

## Plan

> [!NOTE]
> The findings and plan below were generated by Claude (Claude Code) and reviewed before posting. Updated as layers land: this revision reflects the B review reshapes — the redirect machinery extracted to `courseware/redirects.ts` (typed options objects, seven rules including the outline-failure redirect) and the B-prep `getErrorDetail` shape.

[...]

PR #2020, opened 2026-08-20
refactor: de-class CoursewareContainer

### LINKED ISSUE #2008: Peel: de-class CoursewareContainer (structural, no data-layer change)
Part of #1946 — Redux → React Query migration (Stage 1). Part of the #1976 courseware decomposition ([plan](https://github.com/openedx/frontend-app-learning/issues/1946#issuecomment-5348891139)) — **Target 1 (courseware player), peel #1**.

**Goal:** convert `CoursewareContainer` from a class + `connect` component to a functional component using `useDispatch`/`useSelector`, still dispatching today's thunks in effects. **No data-layer change.**

**Why first:** the container is the single orchestrator that *calls* `fetchCourse`/`fetchSequence`; a class can't call the query hooks the later layers introduce. De-classing it (structurally, keeping Redux) unblocks incremental hook adoption without touching the data layer — the courseware analog of the TabPage→TS prep layer.

**Fast-follow:** the TypeScript conversion is split into a separate stacked PR — **#2019**, landing directly above this de-class and below the RQ peels — so this one reviews as a pure structural change.

> [!NOTE]

<details>
<summary> Claude Plan </summary>

# De-class CoursewareContainer (structural, no data-layer change)

This peel does *only* the structural change — class→function, `connect`→hooks, HOC inlined — with **no data-layer or behavior change**.

Decisions locked during planning:
- **Effects: faithful, not idiomatic.** Keep the reselect `memoize` guards as the de-dupe mechanism (not React dependency arrays). We prove equivalence bit-by-bit as each fetch becomes a `useQuery` later, rather than swapping the de-dupe engine now.
- **Inline the router hooks**, delete `utils.jsx` (its only consumer).
- **Stay `.jsx`.** TypeScript is the fast-follow #2019, keeping each PR reviewable as one idea.
- **Tests unchanged** — behavior is identical; fix only if the de-class forces it.

### 1. `src/courseware/CoursewareContainer.jsx` — class → function component (stays `.jsx`)

Rewrite the `class CoursewareContainer` + its `connect`/`mapStateToProps` wiring as a functional component. **Unchanged:** the six module-level `memoize`d redirect helpers and all the `createSelector` selectors stay exactly as they are.

- **Route inputs (inline the HOC):** `useParams()` → `routeCourseId`/`routeSequenceId`/`routeUnitId`, `useNavigate()` → `navigate`, `useLocation()` → `isPreview = pathname.startsWith('/preview')`. Replicates `withParamsAndNavigation` inline.
- **Store reads (`connect` → `useSelector`):** `courseId`, `sequenceId`, `courseStatus`, `sequenceStatus`, `sequenceMightBeUnit` from `state.courseware`; and `course` / `sequence` / `nextSequence` / `firstSequenceId` / `sectionViaSequenceId` via the existing selectors.
- **Dispatch (`connect` map → `useDispatch`):** `const dispatch = useDispatch()`; `dispatch(fetchCourse(...))`, `dispatch(fetchSequence(...))`, `dispatch(saveSequencePosition(...))`, `dispatch(checkBlockCompletion(...))`.
- **Preserve the `memoize` guards faithfully.** `checkFetchCourse`, `checkFetchSequence`, `checkSaveSequencePosition` were instance fields with last-args de-dupe. Recreating them per render would reset that memoization, so hold them in a `useRef` (stable identity, created once). They read *current* values via a `propsRef` refreshed every render — the exact analog of `this.props` always being current (so `checkSaveSequencePosition` reads the live `courseId`/`sequenceId`/`sequence`, never a stale closure).
- **Lifecycle → one no-dependency `useEffect`.** `componentDidMount` + `componentDidUpdate` collapse into a single `useEffect(() => { ... })` with **no dependency array** (runs after every render). Body, in order: the two fetch guards, `checkSaveSequencePosition(routeUnitId)`, the ids-mismatch early-return, then the six `checkXxxRedirect(...)` calls — identical to today's `componentDidUpdate`.
  - *Verified equivalence:* the class ran only the two fetches on mount and everything on update; a no-dep effect runs the whole body on mount too, but that's harmless — `saveSequencePosition` and all redirects gate on `sequenceStatus`/`courseStatus === 'loaded'` (false at mount) and the ids-mismatch bail returns early at mount anyway. Net behavior unchanged.
  - Use `useEffect` (standard). `useLayoutEffect` would match `componentDidUpdate`'s pre-paint timing more strictly, but the redirects only fire after async fetches resolve, so the timing difference isn't user-perceptible.
- **Handlers:** `handleUnitNavigationClick` → `dispatch(checkBlockCompletion(...))`; `handleNextSequenceClick` unchanged; `handlePreviousSequenceClick` stays an empty `() => {}` (still a **required** prop of `Course` → `Sequence`).
- **Drop dead code found during planning:**
  - Remove the `previousSequenceSelector` and the `previousSequence` prop/read — the selector result is never consumed (the handler is empty). Keep the empty `previousSequenceHandler` prop passed to `Course`.
  - Remove `PropTypes` / `defaultProps` — the component now takes no props.
- **Default export:** `export default CoursewareContainer;` (plain function component — no `connect`, no `withParamsAndNavigation`).

### 2. `src/courseware/data/slice.js` — drop dead exports

Remove the `fetchCourseRecommendations{Request,Success,Failure}` re-exports. Confirmed no readers anywhere outside `slice.js`.

### 3. `src/courseware/utils.jsx` — delete

`withParamsAndNavigation` is now inlined; `CoursewareContainer` was its only consumer.

## Tests

Leave `src/courseware/CoursewareContainer.test.jsx` **as-is**. It renders `<CoursewareContainer />` with no props inside `<AppProvider store>` + `<Routes>`, driven by URL + store + axios mocks, and separately unit-tests the exported redirect helpers (unchanged). Behavior is identical, so it should pass untouched. Adjust only if something actually breaks.

## Verification

- `npm run test -- src/courseware/CoursewareContainer.test.jsx` — green.
- `npm run lint` (let eslint autofix spacing).
- Manual smoke on the courseware player: player + sequence nav + outline load identically; resume redirect (`/course/:id` → last active unit), section→sequence, unit-marker `first`/`last`, and unit-position save all behave as before. No data-layer/behavior change.

</details>

brian-smith-tcril:
## Summary

De-class `CoursewareContainer`: convert it from a `class` + `connect` component to a functional component using hooks. **Purely structural — no data-layer or behavior change** (same thunks, same store reads). First peel of the courseware-player decomposition (#1976 → #2008–#2016), part of the Redux → React Query migration (#1946, Stage 1). Stacked on the live-tab conversion (#2006) as the bottom of the courseware chain. Closes #2008.

Why first: `CoursewareContainer` is the single orchestrator that *calls* `fetchCourse` / `fetchSequence`; a class can't call the `useQuery` hooks the later peels introduce, so it has to become a function component before any of them. TypeScript is deliberately **not** part of this PR — it's the fast-follow #2019 (stacked directly above), so this one reads as a pure structural change.

## What changed

- **`src/courseware/CoursewareContainer.jsx`** — `class` → function component.
  - `connect` → `useSelector` / `useDispatch`; route inputs via inlined `useParams` / `useNavigate` / `useLocation`.
  - The reselect `memoize` fetch/save guards are preserved (held in a `useRef`, reading a per-render `latest` ref — the analog of `this.props` always being current), inside a single no-dependency `useEffect` that reproduces `componentDidMount` + `componentDidUpdate` exactly. The de-dupe mechanism is unchanged (memoize, not dependency arrays).
  - Drop the dead `previousSequenceSelector` + `previousSequence` prop (unused; `nextSequence` stays). The empty `previousSequenceHandler` is kept — `Course` → `Sequence` still requires it.
  - `PropTypes` / `defaultProps` removed — the component now takes no props.
- **`src/courseware/utils.jsx`** — deleted. `withParamsAndNavigation` existed only to feed route params into the class; `CoursewareContainer` was its only consumer.
- **`src/courseware/course/sequence/Unit/hooks/useIFrameBehavior.ts`** — guard `activeSequence.unitIds?.length > 0` (see Behavior).
- **`src/courseware/data/slice.js`** — drop the dead `fetchCourseRecommendations{Request,Success,Failure}` exports (orphaned from the recommendations RQ conversion in #1967; no readers).

## Behavior

No user-facing change — the de-class preserves the exact fetch orchestration, URL-normalization redirects, sequence-position save, and store reads.

One downstream guard is required. Removing `connect` means `useIFrameBehavior`'s read of the **global** store id (`getSequenceId` → `state.courseware.sequenceId`) can briefly lead the rendered `sequenceId` prop during a navigation, so the sequence model it looks up can transiently be the course-outline stub (no `unitIds`). The unconditional `activeSequence.unitIds.length` then threw `Cannot read properties of undefined (reading 'length')`. Guarding with `?.length > 0` makes that transient render a no-op (`activeUnitId = null`) that resolves once the sequence loads.

This is isolated, not the tip of an iceberg: `useIFrameBehavior` is the **only** component in the courseware render path that reads the global id and `useModel('sequences', …)` on it. Every other unguarded `.unitIds` reader takes `sequenceId` as a prop/arg — the value that passed `Sequence`'s `sequenceStatus === 'loaded'` gate — so it's always consistent with the render that mounted it. `connect` kept the global id and the rendered prop in lockstep, which masked this latent fragility; the full analysis is in the decision log.

## Testing

`npm run types`, `npm run lint`, and the full suite pass. `CoursewareContainer.test.jsx`'s existing cases are unchanged (they render `<CoursewareContainer />` with no props, driven by URL + store + mocks); **two tests were added** to cover the relocated lines codecov's patch flagged: (1) `handleUnitNavigationClick` → `checkBlockCompletion`, via a `mock`-prefixed opt-in flag that renders the real unit nav only in that test so a `Next` click exercises the handler (asserted through the `get_completion` POST); (2) `firstSequenceIdSelector`'s empty-`sectionIds` branch, via a sectionless course that asserts the resume redirect doesn't fire (no first sequence to pick). The other cases are untouched. The one remaining flagged line is the `handleNextSequenceClick` celebration branch, which codecov's diff shows unmoved (byte-identical, not counted as patch), pre-existing and equally uncovered on `master`. In-browser parity sweep verified core load, all URL-normalization redirects, resume, next/prev unit + sequence, and unit completion; plugin/preview/access cases were not exercised in-browser (details in the manual-testing log). Full logs below.

## Decisions

<details>

# Decisions — de-class CoursewareContainer (#2008)

#1946). **Not checked in** — referenced when opening the PR. First peel of the
courseware-player decomposition (#1976 → #2008–#2016); stacked on the live-tab
conversion (#2006) as the bottom of the courseware chain. Closes #2008.

This layer is **purely structural**: `CoursewareContainer` goes from a
`class` + `connect` to a function component using hooks. **No data-layer or
behavior change** — it dispatches the same thunks and reads the same store. It
exists so the later peels can call `useQuery` hooks a class can't.

## Stays `.jsx`; TypeScript is the fast-follow #2019

**Decision.** The de-class is already a churn-y structural diff; layering a
`.tsx` rename + type annotations on top would make it harder to review. TypeScript
is split into #2019 (stacked directly above this), so the RQ peels happen in
TS-land while this PR reads as a single idea.

## Effects: keep the reselect `memoize` guards (faithful, not dep-arrays)

**Decision.** The class de-duped its work with reselect `memoize` (last-args
guards `checkFetchCourse` / `checkFetchSequence` / `checkSaveSequencePosition`),
not React dependency arrays. We preserve that exact mechanism:

- The three guards live in a `useRef` (stable identity, created once) — recreating
  them per render would reset their last-args memoization.
- They read current values through a per-render `latest` ref, the analog of
  `this.props` always being current (so `checkSaveSequencePosition` reads the live
  `sequence`/ids, never a stale closure).
- `componentDidMount` + `componentDidUpdate` collapse into a **single
  `useEffect` with no dependency array** (runs after every render), body identical
  to the old `componentDidUpdate`.

**Why not idiomatic dep-array effects.** Swapping the de-dupe engine to React deps
now would mean proving every dependency array exactly right across a redirect block
that reads many inputs — a large correctness argument for no functional gain. We
prove equivalence bit-by-bit as each fetch becomes a `useQuery` later, when the
guard for that fetch falls away on its own.

**Mount-runs-everything is safe.** The class ran only the two fetches on mount and
everything on update; the no-dep effect runs the whole body on mount too. Harmless:
`saveSequencePosition` and all redirects gate on `sequenceStatus`/`courseStatus ===
'loaded'` (false at mount), and the ids-mismatch bail returns early at mount anyway.

## Inline the router hooks; delete `utils.jsx`

**Decision.** `withParamsAndNavigation` (`src/courseware/utils.jsx`) existed only to
feed route params/`navigate` into the class. The function component calls
`useParams` / `useNavigate` / `useLocation` directly, so `utils.jsx` is deleted —
`CoursewareContainer` was its only consumer (verified repo-wide).

## Dead code removed: `previousSequence` selector + prop

**Decision.** Remove `previousSequenceSelector` and the `previousSequence` prop it
fed. Keep the empty `previousSequenceHandler` (`() => {}`) — that's a *different*
thing, a required prop of `Course` → `Sequence`.

**We specifically questioned this during review** — is `previousSequence` actually
dead, or a plugin/top-nav extension point we'd be breaking? Verified it is dead:

- The selector is **not exported**, and repo-wide `previousSequence\b` has **zero**
  readers outside this file. Its only consumer was `mapStateToProps`
  (`previousSequence: previousSequenceSelector(state)`), feeding a prop that
  `render()`/handlers never read. Chain: `selector → prop → nowhere`.
- Contrast `nextSequence`, which *is* live — `handleNextSequenceClick` reads it for
  the next-section celebration. So next stays, previous goes.
- **The top-nav plugin does not consume it.** The sequence-navigation slots
  (`SequenceNavigationSlot`, `NextUnitTopNavTriggerSlot`,
  `SequenceBottomNavigationSlot`) compute their own prev/next in
  `src/courseware/course/sequence/sequence-navigation/hooks.js`
  (`previousSequenceId = sequenceIndex > 0 ? sequenceIds[sequenceIndex - 1] : null`
  → `previousLink`). Previous-sequence navigation runs through
  `previousSequenceHandler` (kept) + that hook's own computation, never through the
  container's `previousSequence` data.

## Dead exports removed: `fetchCourseRecommendations*`

**Decision.** Drop the `fetchCourseRecommendations{Request,Success,Failure}`
re-exports from `src/courseware/data/slice.js` — no readers anywhere outside
`slice.js` (verified).

Strictly unrelated to the de-class — these are orphaned leftovers from the
course-recommendations React Query conversion (#1967, already merged), which
dropped the reducers but left the dead destructured exports (so they've resolved
to `undefined` since). Because the PR that should have removed them has already
landed on `master`, there's no cleaner owner elsewhere in the stack, so this
trivial deletion rides along here rather than manufacturing a separate layer.

## Guard `useIFrameBehavior`'s sequence read (crash surfaced by the de-class)

**Decision.** In `useIFrameBehavior.ts`, change the unconditional
`activeSequence.unitIds.length > 0` to `activeSequence.unitIds?.length > 0` — add
the `?.` null-guard while preserving the exact `> 0` predicate. This is the one
downstream change this PR makes outside `CoursewareContainer`.

**Why — the crash.** Manual testing (jumping between units via the course-outline
sidebar) hit `Uncaught TypeError: Cannot read properties of undefined (reading
'length')` at `useIFrameBehavior.ts:42`, flashing the ErrorBoundary then
recovering. Instrumentation caught the exact state: `useIFrameBehavior` was
rendering while `sequenceStatus === 'loaded'`/mid-transition against a sequence
whose model was still the **outline stub** `fetchCourse` seeds (`keys=[id, title,
sectionId]`, no `unitIds`). A 22k-line master log of the same navigation never
crashes; my branch hit it on essentially every jump.

**Why it's isolated, not the tip of an iceberg.** `useIFrameBehavior` is the
**only** component in the courseware render path that reads the *global* store id
(`getSequenceId` → `state.courseware.sequenceId`) and then `useModel('sequences',
…)` on it. Every other unguarded `.unitIds` reader (`Sequence`,
`SequenceNavigation`, `UnitNavigationEffortEstimate`,
`useSequenceNavigationMetadata`) takes `sequenceId` as a **prop/arg** — the same
value that passed `Sequence`'s `sequenceStatus === 'loaded'` gate — so their model
is always the loaded one, consistent with the render that mounted them. Only
`useIFrameBehavior` bypasses the prop, so during a transition its global id can
lead the rendered prop and point at the not-yet-loaded stub.

`connect` (the pre-de-class wiring) delivered store updates to `CoursewareContainer`
and its subtree in lockstep, so the global id and the rendered prop never
disagreed and this latent fragility stayed hidden. Removing `connect` in favor of
independent `useSelector` subscriptions lets them momentarily diverge, exposing it.
So this is a "should have been guarded all along" fix — reading the global
`sequenceId` and assuming a fully-loaded model was never safe — not a symptom of
broad breakage. The grep confirming this (single global-id render-path reader) is
the basis for that confidence.

## Tests: existing suite unchanged + one added coverage test

**Decision.** The existing `CoursewareContainer.test.jsx` cases are left as-is —
they render `<CoursewareContainer />` with no props inside `<AppProvider store>` +
`<Routes>`, driven by URL + store + axios mocks, and separately unit-test the
exported redirect helpers. Behavior is identical, so they pass without edits.

**Two tests added** to cover the relocated lines codecov's patch flagged (both
rewritten/moved by the de-class, so codecov sees them as new; both were equally
uncovered on `master`):

- `handleUnitNavigationClick` → `checkBlockCompletion` (was
  `this.props.checkBlockCompletion(...)` → `dispatch(...)`). The existing suite
  mocks `Unit` as a bare div with no navigation, and 7 cases assert
  `assertNoSequenceNavigation`, so the mock can't render nav globally. A
  `mock`-prefixed opt-in flag (`mockRenderUnitNav`, reset in `afterEach`) makes the
  `Unit` mock render the real `renderUnitNavigation()` **only** in the new test; a
  `Next` click drives `handleNext → unitNavigationHandler → handleUnitNavigationClick`,
  asserted via the `get_completion` POST. Existing cases untouched (flag defaults off).
- `firstSequenceIdSelector`'s empty-`sectionIds` branch — a course with no sections
  makes `firstSequenceId` resolve to `null`, so the resume redirect can't pick a
  first sequence. The test loads a sectionless course at `/course/:courseId` and
  asserts no redirect happens (URL stays at the course root, no unit rendered) —
  exercising the branch via its observable behavior, not just for the line hit.

The one remaining codecov-flagged line is the `handleNextSequenceClick` celebration
branch, which codecov's diff shows unmoved (byte-identical, no +/− markers) so it
isn't counted as patch — pre-existing, equally uncovered on `master`, left as-is.

</details>

<details>
<summary>Manual testing sweep</summary>

# Manual testing — de-class CoursewareContainer (#2008)

Scoped to what **this PR** changes: `CoursewareContainer` goes from a class +
`connect` to a function component on hooks. It is a **purely structural** change —
same thunks, same store reads, same behavior. So this list is a **parity sweep**:
everything the container orchestrated should work exactly as it did on `master`.

The risky part is that `componentDidMount` + `componentDidUpdate` (data fetches,
the URL-normalization redirects, and sequence-position saving) now run in a single
render `useEffect`, and the `memoize` de-dupe guards moved into refs. Focus testing
on those flows.

Mark results as you go — `[x]` pass, `[!]` problem (add a note), `[-]` not tested (with reason).

Automated tests already cover (68 passing in `CoursewareContainer.test.jsx`,
unchanged): initial spinner, successful course/sequence render, the
`course_access` error path, and all six redirect helpers as units. So this list
targets the real-browser navigation/timing behaviors those can't fully exercise.

## 1. Core load

- [x] Open a course at a full `/course/:courseId/:sequenceId/:unitId` URL: spinner →
      course header + sequence + unit render, no flash/crash/double-fetch.
- [x] Hard-reload directly on that URL (cold cache, nothing pre-fetched) — loads clean.

## 2. URL normalization redirects (the effect's main job)

Each of these should land on the normalized `/course/:courseId/:sequenceId/:unitId`
URL, same as `master`:

- [x] Bare `/course/:courseId` → **resume redirect** to the last active
      sequence/unit (or the first sequence if none active).
- [x] `/course/:courseId/:sequenceId` (no unit) → fills in the active-or-first unit.
- [x] `/course/:courseId/:sequenceId/first` and `.../last` → resolve to the first /
      last unit of the sequence.
- [x] `/course/:courseId/:sectionId` (a section id where a sequence is expected) →
      redirects to the section's first sequence.
- [x] `/course/:courseId/:unitId` (bare unit) → fills in the parent sequence id.
- [x] A URL whose "sequence" is actually a unit (the 422 `sequenceMightBeUnit`
      path) → redirects to the correct parent sequence, then the unit.
- [x] Confirm none of the above ends up in a redirect loop or a flicker between URLs.

## 3. Sequence-position save + resume

`saveSequencePosition` now fires from the effect via the ref'd guard.

- [x] Navigate across a few units in a sequence, then hard-reload the course root
      (`/course/:courseId`) — resume lands on the **last unit you were on**.
- [-] Rapidly click through units — position saves once per unit (no duplicate/
      stale saves; the memoize guard should still de-dupe on the unit id).
      *(Not specifically exercised; resume itself verified above.)*

## 4. In-sequence and cross-sequence navigation

Exercises the nav slots (top + bottom) and the `next/previous` handlers passed to
`Course`:

- [x] Next / Previous **unit** buttons move within the sequence.
- [x] Next / Previous **sequence** (stepping off the first/last unit) moves across
      sequences and lands correctly.
- [-] The top-nav "next unit" trigger works (the previously-questioned nav path).
      *(Not tested — it's a plugin slot requiring `env.config.jsx` setup. The nav
      computes prev/next itself in `sequence-navigation/hooks.js`, independent of
      the container.)*

## 5. Next-section celebration

`handleNextSequenceClick` still fires `handleNextSectionCelebration`.

- [-] On a course with first-section celebration enabled, advancing from the last
      unit of the first section into the next section shows the celebration modal
      (same as `master`). *(Not tested — needs celebration setup. Note the handler
      is unchanged from the class version.)*

## 6. Unit completion

`handleUnitNavigationClick` → `checkBlockCompletion`.

- [x] Completing/advancing a unit marks it complete (check icon / progress updates),
      same as before.

## 7. Preview mode

`isPreview` is now derived from `useLocation().pathname` instead of the HOC.

- [-] Load a `/preview/course/...` URL — the same redirects fire but **keep the
      `/preview` prefix** on every normalized URL. *(Not tested.)*

## 8. Access / error states

- [-] A course you can't access (`courseStatus: 'denied'`) shows the denied/redirect
      behavior, unchanged. *(Not tested.)*
- [-] A `course_access` error renders the error state, not a crash. *(Not tested
      in-browser; covered by the automated `course_access` test in `CoursewareContainer.test.jsx`.)*

</details>


src/courseware/CoursewareContainer.jsx
```diff
@@ -148,63 +147,135 @@ export const checkSequenceUnitMarkerToSequenceUnitRedirect = memoize(
...
+  const sectionViaSequenceId = useSelector(sectionViaSequenceIdSelector);
+
+  const latest = useRef();
+  latest.current = {
+    courseId,
+    sequenceId,
+    sequenceStatus,
+    sequence,
+    isPreview,
+  };
```
arbrandes: The ref is written on every render. React recommends explicitly _not_ doing this.

This should be easy to fix: assigning at the top of the effect body should be exactly equivalent, in this case.

src/courseware/CoursewareContainer.jsx
```diff
@@ -1,6 +1,6 @@
-import React, { Component } from 'react';
-import PropTypes from 'prop-types';
-import { connect } from 'react-redux';
+import React, { useEffect, useRef } from 'react';
```
arbrandes: Nit: drop the unused `React` import.

arbrandes (approved):
Pre-approved.  Adopt suggestions at your discretion.

src/courseware/CoursewareContainer.jsx
brian-smith-tcril: addressed in https://github.com/openedx/frontend-app-learning/compare/0200dd815ddbb8929d2ccc6de4d81595204fa5ee..c7bef2f721b879e05a4239c72b7744c8d440a8c3

src/courseware/CoursewareContainer.jsx
brian-smith-tcril: addressed in https://github.com/openedx/frontend-app-learning/compare/0200dd815ddbb8929d2ccc6de4d81595204fa5ee..c7bef2f721b879e05a4239c72b7744c8d440a8c3

Issue #2010, opened 2026-08-20
Convert courseware metadata to React Query


brian-smith-tcril:
Part of #1946 — Redux → React Query migration (Stage 1). Part of the #1976 courseware decomposition ([plan](https://github.com/openedx/frontend-app-learning/issues/1946#issuecomment-5348891139)) — **Target 1 (courseware player)**. Stacked on the de-class and bridge-extension peels.

**Goal:** convert the courseware metadata fetch to React Query.

**Tasks**
- Convert `getCourseMetadata` + `getLearningSequencesOutline` → query hooks feeding `coursewareMeta` / `sections` / `sequences` via the extended bridge.
- **Reuse the existing `useCourseHomeMeta`** for the `courseHomeMeta` model that `fetchCourse` currently double-produces — don't re-fetch it.
- **Peel `getCoursewareOutlineSidebarToggles` aside** — it feeds only the outline sidebar; convert it there (the sidebar layer), not here.
- Move the access-gating decision (`courseAccess.hasAccess` + outline result → success / denied / failure) into the functional container.
- Keep the `courseware` slice status (`courseId`/`courseStatus`) written transitionally.

**Verify:** player renders from the query; `useModel` readers still served via the bridge; access-denied/redirect behavior unchanged.

*Scoped-but-light; a full plan gets added when this is picked up.*

Issue #2015, opened 2026-08-20
Convert saveIntegritySignature + saveSequencePosition to React Query mutations


brian-smith-tcril:
Part of #1946 — Redux → React Query migration (Stage 1). Part of the #1976 courseware decomposition ([plan](https://github.com/openedx/frontend-app-learning/issues/1946#issuecomment-5348891139)) — **Target 4 (remaining writers)**. Stacked on the metadata + sequence conversions (models already exist; no new peel).

**Goal:** convert the two remaining courseware write thunks to React Query mutations.

**Tasks**
- `saveIntegritySignature` (`HonorCode.jsx`) → `useMutation`; on success patch `coursewareMeta.userNeedsIntegritySignature`. Preserve the masquerade skip (no backend POST when masquerading).
- `saveSequencePosition` (`CoursewareContainer`) → `useMutation`; preserve the optimistic `sequences.activeUnitIndex` update + rollback on error.

**Verify:** honor-code acceptance dismisses the modal and clears the integrity prompt; sequence position persists across reloads; rollback on failure.

## Plan

> [!NOTE]
> The findings and plan below were generated by Claude (Claude Code) and reviewed before posting.

Investigation findings that adjust the task list above:

- **One stack layer, not a #2014-style split.** Both conversions are the same kind of change in the same files — `courseware/data/apiHooks.ts` gains two mutations, and `courseware/data/thunks.js` / `redux.test.js` shrink and die together — so there's nothing to peel and no benefit to splitting; a single PR closes this issue.
- **The models stay the write target until #1977.** Every reader of the written state is a model read (`sequences.activeUnitIndex` via `CoursewareContainer`'s selectors and `useIFrameBehavior`; `coursewareMeta.userNeedsIntegritySignature` via `useShouldDisplayHonorCode`'s `useModel`), so both mutations keep dispatching `updateModel` rather than patching the sequence/metadata query caches — the same correction verified for #2014: the bridge rewrites models only on a real fetch `onSuccess`, and a real refetch carries server truth *including* the saved position / signature. (One pre-existing caveat, unchanged by the port: after a masqueraded dismissal the server still reports the signature as needed, so a metadata refetch can resurrect the prompt — the thunk behaved identically.)
[...]

Issue #2011, opened 2026-08-20
Convert sequence data to React Query


brian-smith-tcril:
Part of #1946 — Redux → React Query migration (Stage 1). Part of the #1976 courseware decomposition ([plan](https://github.com/openedx/frontend-app-learning/issues/1946#issuecomment-5348891139)) — **Target 1 (courseware player)**. Stacked on the metadata conversion (#2023).

**Goal:** convert the sequence fetch to React Query.

**Tasks**
- Convert `fetchSequence` → query hook feeding the `sequences` + `units` models via the extended bridge.
- Preserve the 422 `sequenceMightBeUnit` path (CoursewareContainer relies on the failed request to detect a unit).
- Keep the `courseware` slice `sequenceId`/`sequenceStatus` written transitionally.

**Verify:** sequence + units render from the query; unit-vs-sequence (422) detection still works; `useModel('units')`/`useModel('sequences')` readers unaffected.

<details>
<summary>Full implementation plan (drafted by Claude 🤖)</summary>

## Context

`fetchSequence` is the last big courseware *fetch* thunk still on Redux: it loads one sequence's metadata + its units, mirrors them into the `sequences`/`units` model store, and drives the `courseware` slice's `sequenceId`/`sequenceStatus`/`sequenceMightBeUnit` fields.

The conversion is a faithful continuation of the pattern #2023 established for metadata/outline:
- an RQ query hook whose `meta.models` mirrors the result into the model store via the `bridgeToModelStore` QueryCache `onSuccess` (`src/data/modelStoreBridge.ts`), and
- a small transitional **status-bridge** hook (`src/courseware/data/statusBridge.ts`) that writes the still-Redux slice status fields, so the many not-yet-converted readers (the container's redirect helpers, `Sequence.jsx`, breadcrumbs, sequence-navigation, sequence-alerts, the outline sidebar) keep working unchanged.

The one subtlety unique to sequences: `CoursewareContainer` relies on the sequence fetch **failing** (a backend **422**) to detect that a URL segment is actually a unit, not a sequence (`sequenceMightBeUnit`) — and on a *fast* failure, because the redirect that resolves the unit's parent sequence only fires once `sequenceStatus === 'failed'`.

## Key files

- `src/courseware/data/thunks.js` — `fetchSequence`: the thunk being converted.
- `src/courseware/data/apiHooks.ts` — where `useCoursewareMetadata`/`useCoursewareOutline` live; add `useSequenceMetadata` here.
[...]

PR #2021, opened 2026-08-20
refactor: convert CoursewareContainer to TypeScript

### LINKED ISSUE #2019: Peel: convert CoursewareContainer to TypeScript (fast-follow to de-class #2008)

brian-smith-tcril:
## Summary

Convert `CoursewareContainer` from `.jsx` to `.tsx`. The diff is tiny — a file rename plus a handful of annotations, no behavior change — and that's the point: **the value isn't in this PR, it's in what it unlocks.** With the container in TypeScript, the courseware React Query peels (#2010+) convert the data layer *in TS-land*, so every `useQuery`/`useMutation` and the code consuming it is type-checked as it lands — instead of writing the conversions in untyped `.jsx` and retrofitting types later.

Fast-follow to the de-class ([#2020](https://github.com/openedx/frontend-app-learning/pull/2020)); part of the Redux → React Query migration (#1946, Stage 1), stacked above #2020 and below the RQ peels. Split from the de-class so each PR reviews as one idea (structural change vs. type annotations). Closes #2019.

## What changed

Only `src/courseware/CoursewareContainer.jsx` → `.tsx` (git tracks it as a rename; the diff is just the annotations):

- `state: RootState` on the inline `useSelector`s (matching the `TabPage.tsx` precedent); `useRef<any>()` on the two loose refs. The redirect helpers and `createSelector` selectors needed **no** annotation — reselect infers them.
- Two small **interim** bridges to `TabPage`'s already-typed props: `courseStatus as CourseStatus` at the read, and `courseId ?? undefined` at the prop. Both fall away when the peels move this component to RQ (it'll pass `courseStatus={{ metadataQuery, tabDataQuery }}` and take `courseId` from `useParams`, like the five course-home tabs already do).

The decision log covers the full rationale — the two bridges, the alternatives weighed (incl. changing the slice), and why `store.ts` is left untouched.

## Testing

`npm run types` (0 errors), `npm run lint` (clean), and `CoursewareContainer.test.jsx` (70 passing) all green — the test file stays `.jsx` and unchanged, resolving to the `.tsx`. No behavior change to exercise; a manual click-through of the courseware player confirmed it renders as before.

## Decisions

<details>

# Decisions — convert CoursewareContainer to TypeScript (#2019)

#1946). **Not checked in** — referenced when opening the PR. Fast-follow to the
de-class #2008/#2020, stacked directly above it and below the courseware RQ peels
(#2010+). Closes #2019.

This layer is a **pure TypeScript conversion**: rename
`src/courseware/CoursewareContainer.jsx` → `.tsx` and add the minimum annotations
to compile. **No behavior change.** It exists so the courseware RQ peels all happen
in TS-land; it's split from the de-class (#2020) so each PR reviews as one idea
(structural change vs. type annotations).

## Don't type the doomed Redux stuff (`any`/loose)

**Decision.** The Redux-sourced values are about to be deleted/replaced as each
fetch becomes a typed `useQuery` in the RQ peels, so type them loosely now and
tighten per-peel. In practice this needed very little:

- The six exported redirect helpers (`checkResumeRedirect`, …) and the
  `createSelector` selectors (`currentCourseSelector`, …) needed **no**
  annotation — reselect's `defaultMemoize` / `createSelector` types infer their
  parameters, so `tsc` was already satisfied. No `any` added there.
- The two internal refs (`latest`, `guards`) hold doomed shapes (the mirrored
  props and the memoized guard fns), so they're `useRef<any>()`.

## Dispatch stays plain `useDispatch()` — no typed-dispatch primitive

**Decision.** Do **not** add an `AppDispatch` type / `useAppDispatch` hook (`store.ts`
is untouched). Typed thunk dispatch only ever serves the Redux layer we're deleting:
the RQ-converted code uses `useQuery`/`useMutation` and doesn't dispatch at all, and
as each peel converts a fetch its `dispatch(...)` calls disappear — so there's no
future where typed dispatch helps the RQ code. Adding it would be investing type
effort in the doomed Redux stuff, against the migration's whole point.

It's also unnecessary to compile: the thunks live in `thunks.js`, so `tsc` sees them
loosely and plain `useDispatch()` type-checks the `dispatch(fetchCourse(id))` calls
without complaint. (The #2019 issue's task list included adding the primitive — that
premise doesn't hold, so it's dropped; `store.ts` keeps only its existing `RootState`.)

## `useSelector` state typed as `RootState`

**Decision.** The five inline `useSelector((state: RootState) => state.courseware.…)`
calls are typed with `RootState` (matching the `TabPage.tsx` precedent), which is
enough for `state.courseware.*` to resolve. The courseware slice is still JS, so
those values come back loosely inferred — fine for now.

## Two interim bridges to `TabPage`'s (RQ-era) props

`TabPage`'s prop types are designed for the converted world (`CourseStatus =
StatusValue | { metadataQuery, tabDataQuery }`, `courseId?: string`), while this
component is still a Redux caller. Typing the state surfaced two gaps — handled
differently, because they are different kinds of gap:

- **`courseStatus` — assert once at the read.** The slice value genuinely *is* a
  valid `StatusValue`, but because `slice.js` is JS, `RootState` infers the field as
  the wider `string`, which won't land in `CourseStatus` on its own. So it's typed
  `as CourseStatus` **at the `useSelector`** (`import type` from `../tab-page/TabPage`),
  making it a valid `CourseStatus` from its source — not at the prop. The value is
  what the type says; the assertion just tells TS what the JS slice can't.
- **`courseId` — coerce at the `TabPage` prop only.** `courseId={courseId ?? undefined}`.
  The slice uses `string | null` (null = no course yet); `TabPage` uses `string | undefined`.
  This one *can't* move to the read: the effect's ids-mismatch bail
  (`courseId !== (routeCourseId || null)`) depends on `courseId` staying `string | null`
  — coercing it to `undefined` globally would make `undefined !== null` fire the bail
  on every unloaded render (a behavior change). So `courseId` stays `string | null`
  everywhere and is coerced (a real null→undefined value transform, not a type
  assertion) only for the one consumer that wants `undefined`. Behavior-safe:
  `TabPage` treats a falsy `courseId` the same either way.

Both are interim. When the courseware peels convert this component's data, it'll
pass `courseStatus={{ metadataQuery, tabDataQuery }}` (the object arm of
`CourseStatus`, no cast — exactly as the five course-home tabs already do) and take
`courseId` from `useParams`/`useContextId` (`string | undefined`, matching directly),
and both bridges drop.

### Alternatives considered for the `courseId` null→undefined bridge

We deliberately kept the inline `courseId ?? undefined` at the `TabPage` prop after
weighing the alternatives:

- **Make the slice return `undefined` instead of `null`.** Tempting as "prepping the
  RQ shape," but rejected: the RQ-era `courseId` does **not** come from this slice —
  the #1976 teardown *deletes* `state.courseware.courseId` and sources it from
  `useContextId`/`useParams` (already `string | undefined`). So changing the slice
  makes a soon-deleted field briefly mimic the RQ shape without building toward it —
  i.e. investing in Redux we're removing. It would also touch the behavior-critical
  bail (below), require editing `sequenceId` for consistency, and need a repo-wide
  audit of ~14 readers of `state.courseware.courseId`/`.sequenceId`. Not worth it to
  save one coercion; out of scope for a TS-only PR.
- **Type the whole slice (`slice.js` → `.ts`).** Would remove the `courseStatus`
  assertion too, but same objection — typing effort spent on the slice that's deleted
  in #1976, and out of scope here.
- **Coerce `courseId` at the read (globally), like `courseStatus`.** Rejected: the
  effect's bail normalizes both sides to null-when-absent —
  `courseId !== (routeCourseId || null)` — so it relies on `courseId` being
  `string | null`. Coercing it to `string | undefined` makes the unloaded case
  `undefined !== null` → `true`, firing the bail when it shouldn't (a behavior change).
  Unlike `courseStatus` (a type assertion of an already-valid value), this is a *value
  transform* that's only correct for the `TabPage` consumer, so it must stay local to
  that consumer.
- **Named intermediate (`const tabPageCourseId = courseId ?? undefined`).** Equivalent
  to the inline form; chose inline since there's a single consumer and it's one line.
  The "why" lives here in the doc rather than as a code comment.

## Tests unchanged

**Decision.** `CoursewareContainer.test.jsx` stays `.jsx` and unchanged — it
imports `./CoursewareContainer`, which now resolves to the `.tsx`, and the 70 cases
(incl. the two coverage tests added in #2020) pass as-is. No behavior change, so no
test edits. `npm run types` and `npm run lint` are clean.

</details>

<details>

# Manual testing — convert CoursewareContainer to TypeScript (#2019)

A type-only conversion (`.jsx` → `.tsx`) with **no runtime/behavior change**, so
there's nothing new to exercise beyond confirming the courseware player still
renders. The real safety net is tooling: `npm run types` (0 errors),
`npm run lint` (clean), and `CoursewareContainer.test.jsx` (70 passing, unchanged).

Mark results as you go — `[x]` pass, `[!]` problem (add a note).

- [x] Clicked around the courseware player (load, unit navigation) — renders and
      behaves as before. No functional change expected, none observed.

</details>

### DIFF
--- a/src/courseware/CoursewareContainer.tsx
+++ b/src/courseware/CoursewareContainer.tsx
@@ -13,6 +13,8 @@ import {
   saveSequencePosition,
 } from './data';
 import { TabPage } from '../tab-page';
+import type { CourseStatus } from '../tab-page/TabPage';
+import type { RootState } from '../store';
 
 import Course from './course';
 import { handleNextSectionCelebration } from './course/celebration';
@@ -221,20 +223,20 @@ const CoursewareContainer = () => {
   } = useParams();
   const isPreview = pathname.startsWith('/preview');
 
-  const courseId = useSelector((state) => state.courseware.courseId);
-  const sequenceId = useSelector((state) => state.courseware.sequenceId);
-  const courseStatus = useSelector((state) => state.courseware.courseStatus);
-  const sequenceStatus = useSelector((state) => state.courseware.sequenceStatus);
-  const sequenceMightBeUnit = useSelector((state) => state.courseware.sequenceMightBeUnit);
+  const courseId = useSelector((state: RootState) => state.courseware.courseId);
+  const sequenceId = useSelector((state: RootState) => state.courseware.sequenceId);
+  const courseStatus = useSelector((state: RootState) => state.courseware.courseStatus) as CourseStatus;
+  const sequenceStatus = useSelector((state: RootState) => state.courseware.sequenceStatus);
+  const sequenceMightBeUnit = useSelector((state: RootState) => state.courseware.sequenceMightBeUnit);
   const course = useSelector(currentCourseSelector);
   const sequence = useSelector(currentSequenceSelector);
   const nextSequence = useSelector(nextSequenceSelector);
   const firstSequenceId = useSelector(firstSequenceIdSelector);
   const sectionViaSequenceId = useSelector(sectionViaSequenceIdSelector);
 
-  const latest = useRef();
+  const latest = useRef<any>();
 
-  const guards = useRef();
+  const guards = useRef<any>();
   if (!guards.current) {
     guards.current = {
       checkFetchCourse: memoize((id) => {
@@ -392,7 +394,7 @@ const CoursewareContainer = () => {
   return (
     <TabPage
       activeTabSlug="courseware"
-      courseId={courseId}
+      courseId={courseId ?? undefined}
       unitId={routeUnitId}
       courseStatus={courseStatus}
     >


src/courseware/CoursewareContainer.tsx
```diff
@@ -221,18 +223,18 @@ const CoursewareContainer = () => {
...
+  const sequenceStatus = useSelector((state: RootState) => state.courseware.sequenceStatus);
+  const sequenceMightBeUnit = useSelector((state: RootState) => state.courseware.sequenceMightBeUnit);
   const course = useSelector(currentCourseSelector);
   const sequence = useSelector(currentSequenceSelector);
   const nextSequence = useSelector(nextSequenceSelector);
   const firstSequenceId = useSelector(firstSequenceIdSelector);
   const sectionViaSequenceId = useSelector(sectionViaSequenceIdSelector);
 
-  const latest = useRef();
+  const latest = useRef<any>();
```
arbrandes, line 237: **Nit:** `latest` could carry its real shape instead of `any`.  But I figure this might be going away, so totally optional.

arbrandes (approved):
Pre-approved, with an optional nit.

src/courseware/CoursewareContainer.tsx
brian-smith-tcril: It is definitely going away, so using `any` was a "fewer lines to remove later" thing.

It is currently used in:
* `checkFetchSequence` (will be removed as part of https://github.com/openedx/frontend-app-learning/issues/2011)
* `checkSaveSequencePosition` (will be removed as part of https://github.com/openedx/frontend-app-learning/issues/2015)

Issue #2022, opened 2026-08-20
Restore dropped query error logging via a global QueryCache.onError


brian-smith-tcril:
🤖 _Drafted by Claude Code._

Part of #1946.

## Background

React Query v5 removed `onError`/`onSuccess`/`onSettled` from `useQuery` (they remain only on `useMutation` and the `QueryCache`). As the migration converts thunks → query hooks, each converted **query** silently loses the `logError`/`logInfo` its thunk emitted — there's no per-hook place for it in v5. Converted **mutations** are unaffected (they keep `onError`).

## Query conversions that dropped logging without replacement

| PR | area | old thunk did | now |
|---|---|---|---|
| #1987 | dates tab (`fetchDatesTab`) | `logError` on 401/403/404 | dropped; hook swallows 403→`{}`, others unlogged |
| #1997 | outline tab (`fetchOutlineTab`) | `logError` | dropped |
| #2001 | progress tab (`fetchProgressTab`) | `logError` | dropped |
| #2005 | discussion tab | (no query logging) | nothing removed |
| #2006 | live tab (shared `fetchTab`, deleted here) | `catch` → 403-detail extract + `logError` | dropped |
| courseware metadata (`fetchCourse`, in-flight) | `logError` metadata/courseHomeMeta/outline; `logInfo` outline-403 | dropped (toggles logging kept via thunk) |

*Adjacent (verify separately, not this issue):* those `catch` blocks also extracted 403 `errorMessage`/`error_code` for the error page. The courseware work preserves that in the container; the course-home tab conversions should be confirmed.

<details><summary>Proposed approach</summary>

- Add a global `QueryCache.onError(error, query)` — the v5-recommended home for cross-cutting query error logging; permanent infra that survives the bridge/model-store removal (#1977).
- Meta-driven nuance: queries tag `meta` (e.g. `{ quietStatuses: [403] }`) so the handler emits `logInfo`/skips for expected statuses and `logError` for the rest — restoring the 403-as-info behavior.
- Lift `QueryClient`/`QueryCache` construction into a permanent `queryClient` module; the transitional bridge contributes only the `onSuccess` mirror there.
- One handler restores logging for every converted query uniformly.

</details>

## Stack placement

The logging layer lands **in #1987** — the lowest open layer and the first conversion that
removes any logging (nothing below it has dropped a log, so a separate layer *before* #1987
would be infra with nothing to log yet).

[...]

PR #2063, opened 2026-09-15
refactor: convert checkBlockCompletion to a React Query mutation

### LINKED ISSUE #2012: Peel: convert checkBlockCompletion to a React Query mutation
Part of #1946 — Redux → React Query migration (Stage 1). Part of the #1976 courseware decomposition ([plan](https://github.com/openedx/frontend-app-learning/issues/1946#issuecomment-5348891139)) — **Target 2 (outline sidebar), peel**. Stacked on the sequence conversion.

**Goal:** convert `checkBlockCompletion` to a React Query mutation.

**Why its own layer:** it's the courseware `requestCert` — a shared writer that patches **both** the `units` model (`complete`) and the sidebar's `courseware.courseOutline` completion rollups (`updateCourseOutlineCompletion`), and it has two consumers in different layers (`CoursewareContainer` on unit render, and the outline sidebar on unit click). Peeling it below both consumers keeps them from fighting over completion state once the sidebar is also React-Query-backed.

**Tasks**
- Convert `checkBlockCompletion` → `useMutation`; on success update the `units` cache (via the bridge) and the outline cache (`setQueryData`, replacing `updateCourseOutlineCompletion`).
- Preserve the "already complete → no-op" short-circuit.

**Verify:** completing a unit updates the unit indicator and the sidebar completion rollups (incl. the `courseOutlineShouldUpdate` refetch trigger for locked sequences).

## Plan

> [!NOTE]

- **No outline `setQueryData` at this layer.** The sidebar outline (`getCourseOutlineStructure` → `state.courseware.courseOutline`) is still Redux until #2013 converts it — there is no outline query to write into yet. This layer keeps dispatching `updateCourseOutlineCompletion` (transitional); #2013 swaps that dispatch for a cache update. Likewise the `units` write stays a direct `updateModel` dispatch rather than going "via the bridge": the model-store bridge is wired to the QueryCache only, and the payload is shaped from mutation variables + result (`{ id: unitId, complete }`), which the declarative `meta.models` mirror can't express. #1977 later moves that write to the sequence query cache.
- **The transitional dispatches must live in the hook-level `onSuccess`, not `mutate()`-site callbacks.** TanStack v5 skips `mutate(vars, { onSuccess })` callbacks if the observing component unmounted, and the sidebar unmounts routinely before the POST resolves (mobile collapses it immediately on unit click; any unit click navigates away from the old sequence's `UnitLinkWrapper`s). The thunk always ran to completion, so hook-level `onSuccess` — which runs on the mutation itself regardless of unmount — is the faithful home.
- **A hidden behavior to preserve:** when the sidebar outline was never fetched (`courseOutline: {}` — sidebar disabled or unopened) and a unit completes, the `updateCourseOutlineCompletion` reducer throws (`state.courseOutline.units` is undefined). The thunk's catch-all logged that error *after* the units-model write had already landed. The conversion keeps an explicit try/catch reproducing exactly that, plus a test pinning it.
- **The "already complete → no-op" guard needs a call-time store read**: `useStore().getState().models.units[unitId]?.complete`, not `useSelector` (which would re-render both consumers on every unit write).
- **Consumer detail:** the container's call fires on unit *navigation click* (`handleUnitNavigationClick`, passed down as `unitNavigationHandler`), not on unit render; it checks the unit being navigated away from. The sidebar's `handleUnitClick` checks the previously-active unit.

<details>
<summary>Full plan</summary>

# Plan: #2012 — Convert `checkBlockCompletion` to a React Query mutation

## Context

decomposition — **Target 2 (outline sidebar), peel**. New stack layer on top of #2061
(sequence data → RQ). `checkBlockCompletion` is a shared *writer*: after the learner
views/leaves a unit it asks the LMS `get_completion` handler whether the unit is now
complete, and on `true` fans the answer out to **two** stores:

1. `updateModel({ modelType: 'units', ... })` — the model-store unit's `complete` flag
   (read by sequence-navigation unit buttons / `UnitButton`), and
2. `updateCourseOutlineCompletion` (courseware slice) — the sidebar outline's
   completion rollups (unit → sequence `completionStat`/`complete` → section
   `completionStat`/`complete`, plus the `courseOutlineShouldUpdate` refetch flag for
   sections containing a locked/prerequisite sequence).

It has two consumers in different layers:
- `CoursewareContainer.tsx` — `handleUnitNavigationClick` (passed down as
  `unitNavigationHandler` through `Course` → `Sequence` → sequence-navigation), checking
  the unit being navigated away from (`routeUnitId`).
- the outline sidebar — `useCourseOutlineSidebar`'s `handleUnitClick`
  (`course-outline/hooks.js`), fired from `UnitLinkWrapper` on unit click, checking the
  previously-active unit (`activeUnitId`).

Peeling it below both consumers now means #2013 (sidebar outline → RQ) and the
container teardown don't fight over completion state later.

**Correction to the issue body.** #2012's body says on success to update "the outline
cache (`setQueryData`, replacing `updateCourseOutlineCompletion`)" — that's not possible
at this layer: the sidebar outline (`getCourseOutlineStructure` →
`state.courseware.courseOutline`) is still Redux until #2013 converts it. There is no
outline query to `setQueryData` into. So this layer keeps dispatching
`updateCourseOutlineCompletion` (transitional), and #2013 swaps that dispatch for a
cache update when the outline itself moves. Same for the units write: the model store
is still the merged source of truth for `units` (written by the sequence-query bridge +
this mutation), so the mutation dispatches `updateModel` directly; #1977 later moves
that to the sequence query cache.

## Key files (all read during investigation)

- `src/courseware/data/thunks.js` — `checkBlockCompletion` (lines 40–63): the thunk being converted.
- `src/courseware/data/api.js` — `getBlockCompletion(courseId, sequenceId, usageKey)` (POST `.../handler/get_completion`, returns `data.complete === true`). Unchanged.
- `src/courseware/data/slice.js` — `updateCourseOutlineCompletion` (lines 78–124). Kept; now dispatched by the hook.
- `src/courseware/data/apiHooks.ts` — where the queries live; add `useCheckBlockCompletion` here (mutation precedent: course-home's `usePostCourseDeadlines`/`useRequestCert`, which do hook-level `onSuccess` side effects + `onError: logError`).
- `src/courseware/CoursewareContainer.tsx` — consumer 1 (`handleUnitNavigationClick`, line 375–377; `checkBlockCompletion` import from `./data`).
- `src/courseware/course/sidebar/sidebars/course-outline/hooks.js` — consumer 2 (`handleUnitClick`, line 90; import from `@src/courseware/data/thunks`).
- `src/courseware/data/index.js` — `checkBlockCompletion` re-export (remove).
- `src/store.ts` — `RootState` (for the guard's `useStore<RootState>()` read).
- `src/courseware/data/redux.test.js` — `Test checkBlockCompletion` describe (lines 79–149) to port.
- `src/courseware/data/apiHooks.test.tsx` — where the ported cases land.
- The five sidebar test files (`CourseOutlineTray`, `CourseOutlineTrigger`, `SidebarSection`, `SidebarSequence`, `SidebarUnit`) — raw `@testing-library/react` `render` with hand-built providers, **no QueryClientProvider**; all six components that call `useCourseOutlineSidebar` will now transitively mount `useMutation`.

### 1. `apiHooks.ts` — `useCheckBlockCompletion()`

```ts
export const useCheckBlockCompletion = () => {
  const store = useStore<RootState>();
    mutationFn: ({ courseId, sequenceId, unitId }: CheckBlockCompletionVars) => (
      getBlockCompletion(courseId, sequenceId, unitId)
    ),
    onSuccess: (isComplete, { sequenceId, unitId }) => {
      dispatch(updateModel({ modelType: 'units', model: { id: unitId, complete: isComplete } }));
      try {
        dispatch(updateCourseOutlineCompletion({ sequenceId, unitId, isComplete }));
      } catch (error) {
        logError(error); // reducer throws when the sidebar outline isn't loaded
      }
    },
    onError: (error) => logError(error),
  });
  return useCallback((courseId, sequenceId, unitId) => {
    if (store.getState().models.units[unitId]?.complete) {
      return; // things don't get uncompleted after they are completed
    }
    mutate({ courseId, sequenceId, unitId });
  }, [mutate, store]);
};
```

Points that make this faithful:

- **The returned callback keeps the thunk's call signature** `(courseId, sequenceId,
  unitId)`, so both call sites are a rename-on-read (`const checkBlockCompletion =
  useCheckBlockCompletion();` and the call line loses only its `dispatch(...)`
  wrapper). Param types follow what the container actually passes (nullable
  courseId/sequenceId from the slice selectors) — no `as any`.
- **The already-complete short-circuit** read `getState().models.units[unitId]?.complete`
  at call time; `useStore()` (first use in the repo, standard react-redux) reproduces a
  call-time read without subscribing the caller to every model change (a `useSelector`
  on `models.units` would re-render both consumers on every unit write).
- **Transitional dispatches live in the hook-level `onSuccess`, not `mutate`-site
  callbacks.** This is load-bearing: TanStack v5 skips `mutate(vars, { onSuccess })`
  callbacks if the observing component unmounted, and the sidebar unmounts routinely
  before the POST resolves (mobile: `handleUnitClick` collapses the sidebar
  immediately; any unit click navigates and unmounts the old sequence's
  `UnitLinkWrapper`s). Hook-level `onSuccess` runs on the mutation itself regardless,
  matching the thunk, which always ran to completion.
- **The `try/catch` around the outline dispatch preserves the thunk's catch-all.** When
  the sidebar outline was never fetched (`courseOutline: {}` — sidebar disabled or not
  yet opened) and `isComplete` is true, the `updateCourseOutlineCompletion` reducer
  throws (`state.courseOutline.units` is undefined). The thunk's single `try/catch`
  caught that and `logError`'d it, *after* the units-model write had already landed.
  The explicit catch reproduces exactly that (one log, units write intact, no reliance
  on TanStack's onSuccess-throw → onError routing).
- **No retry config needed**: mutations default to `retry: 0` in TanStack (the client's
  `shouldRetryQuery` and #2024 apply to queries only), matching the thunk's
  single-attempt behavior.
- **No `meta.models`**: the model-store bridge is wired to the QueryCache only, and the
  units payload is shaped from *variables + result* (`{ id: unitId, complete }`), which
  the declarative mirror can't express. Explicit dispatch, like the thunk did.

Placement considered and rejected: `statusBridge.ts` — its charter is mirroring query
*status* into slice status fields; this is a writer, and course-home already set the
precedent of mutations with side-effectful hook-level `onSuccess` living in
`apiHooks.ts`. The hook survives #2013/#1977; only its `onSuccess` internals shift from
dispatches to `setQueryData`.

### 2. `CoursewareContainer.tsx`
- Remove `checkBlockCompletion` from the `./data` import; add
  `const checkBlockCompletion = useCheckBlockCompletion();` (import from
  `./data/apiHooks`).
- `handleUnitNavigationClick` body becomes
  `checkBlockCompletion(courseId, sequenceId, routeUnitId);` — otherwise untouched.

### 3. `course-outline/hooks.js`
- Drop `checkBlockCompletion` from the `@src/courseware/data/thunks` import (keep
  `getCourseOutlineStructure` — that's #2013).
- `const checkBlockCompletion = useCheckBlockCompletion();` at the top of
  `useCourseOutlineSidebar`; `handleUnitClick` line 90 becomes
  `checkBlockCompletion(courseId, sequenceId, activeUnitId);`.

### 4. Delete the thunk
- Remove `checkBlockCompletion` from `thunks.js`, plus its now-unused imports there
  (`getBlockCompletion` from `./api`, `updateCourseOutlineCompletion` from `./slice`;
  `updateModel` stays — `saveSequencePosition`/`saveIntegritySignature` still use it).
- Remove the `checkBlockCompletion` re-export from `courseware/data/index.js`.
- `slice.js` keeps `updateCourseOutlineCompletion` (now hook-dispatched; #2013 deletes it).
- The pact test (`lmsPact.test.jsx`) drives `getBlockCompletion` (the api fn) directly — unaffected.

### 5. Port the `redux.test.js` describe → `apiHooks.test.tsx`
Delete `Test checkBlockCompletion` (lines 79–149) and rebuild the cases as hook tests
(`renderHook` with a wrapper providing **both** the Redux `Provider`/`AppProvider` and
`QueryClientProvider` with `createTestQueryClient(store)` — the mutation dispatches, so
it needs the store context, unlike the existing query tests in that file):

- **complete=true, outline loaded** (seed via `executeThunk(getCourseOutlineStructure)`
  against the axios mocks, as the old test did): callback fires the POST;
  `models.units[unitId].complete` is true; outline unit/sequence/section `complete` and
  `completionStat.completed` rollups update — same assertions as redux.test.js lines
  120–125.
- **complete=false**: no model write, no outline change (old lines 128–148).
- **network error**: POST attempted, `logError` called once, no writes (old lines 83–100).
- **already complete** (new — the guard is now hook logic): seed
  `models.units[unitId].complete = true`, call the callback, assert **no POST** in
  `axiosMock.history.post`.
- **complete=true, outline never loaded** (new — pins the preserved catch-all):
  `models.units[unitId].complete` becomes true, `logError` called once,
  `state.courseware.courseOutline` stays `{}`.

### 6. Sidebar test wrappers
All five sidebar test files build providers by hand around raw `render`; add
`<QueryClientProvider client={createTestQueryClient(store)}>` (from `@src/setupTest`)
inside their `AppProvider` so the hook's `useMutation` mounts. Mechanical, one wrapper
edit per file; no assertion changes expected.

### 7. Existing coverage that must stay green untouched
- `CoursewareContainer.test.jsx` — already renders with `createTestQueryClient`; the
  "marks the current unit complete when navigating to the next unit" test (line 406)
  clicks next and asserts the `get_completion` POST — exercises the converted path
  end-to-end.
- `redux.test.js` — the shared "Thunks that require fetched sequences" seeding stays
  (still used by `saveSequencePosition` / `saveIntegritySignature` tests, #2015).

Conventions: `userEvent`, no `eslint-disable`, rationale in the decision doc not in
comments (the one inline comment above is the terse per-case kind: the reducer-throw
branch is the surprising one).

## Decision doc

Capture in the decision doc (folded into the PR at submit time): the issue-body
correction (no outline `setQueryData` until #2013, and why); hook-level vs mutate-site
callbacks (unmount semantics); the `useStore` call-time guard; the preserved catch-all
around the outline reducer (and that the outline-not-loaded `logError` is pre-existing
behavior, kept deliberately); placement in `apiHooks.ts`; and what #2013/#1977 will
each replace.

## Stack

New layer stacked on #2061 (sequence data → RQ); submitted once green.

## Verification

- `npm run test -- src/courseware/data src/courseware/CoursewareContainer.test.jsx src/courseware/course/sidebar` — ported hook tests + container matrix + sidebar suites.
- Manual smoke (tutor local, DemoX): navigate past a unit → its nav button gets the
  completion check *and* the open sidebar's rollups tick up; click a unit in the
  sidebar (previous unit gets checked); sidebar closed on mobile → completion still
  lands after collapse. The locked-sequence `courseOutlineShouldUpdate` refetch needs a
  prereq-gated course — verify via the existing unit tests if no such course is handy.

</details>

brian-smith-tcril:
## Summary

Convert the `checkBlockCompletion` thunk to a `useCheckBlockCompletion` React Query mutation, following the mutation pattern the course-home conversions established (`useRequestCert`, `usePostCourseDeadlines`). **No user-facing change** — on success the hook writes the same two places the thunk wrote: the `units` model's `complete` flag (sequence-nav indicators) and the sidebar outline's completion rollups (`updateCourseOutlineCompletion`), so the two consumers (`CoursewareContainer`'s unit-navigation handler and the outline sidebar's unit click) behave exactly as before.

Part of the Redux → React Query migration (#1946, Stage 1) and the #1976 courseware decomposition, stacked above #2061 (the sequence-data conversion). Resolves #2012.

One correction to the issue's task list: the outline write can't move to `setQueryData` yet — the sidebar outline is still Redux until #2013 converts it — so this layer keeps dispatching `updateCourseOutlineCompletion` transitionally, and #2013 swaps that dispatch for a cache update (details in the decision log).

## What changed

- **`useCheckBlockCompletion`** (`apiHooks.ts`) — the mutation. The transitional Redux writes live in the **hook-level** `onSuccess` (not `mutate`-site callbacks): TanStack v5 skips call-site callbacks when the calling component unmounts, and the sidebar routinely unmounts with the POST in flight (mobile collapse on unit click, unit navigation). The thunk ran to completion regardless of any component, so the hook must too.
- **The "already complete → no-op" guard** is a call-time `useStore()` read, preserving the thunk's exact `getState` check without subscribing either consumer to every model write. The returned callback keeps the thunk's `(courseId, sequenceId, unitId)` signature, so both call sites just lose their `dispatch(...)` wrapper.
- **The outline dispatch keeps the thunk's catch-all** — `updateCourseOutlineCompletion` throws when the sidebar outline was never loaded (sidebar disabled/unopened); the thunk logged that and kept the units write. An explicit try/catch reproduces exactly that.
- **Consumers** — `CoursewareContainer.tsx` and the outline sidebar's `useCourseOutlineSidebar` swap the thunk dispatch for the hook callback; nothing else in either changes.
- **Cleanup** — the thunk and its `courseware/data` re-export are deleted; the `updateCourseOutlineCompletion` slice action stays (now hook-dispatched, deleted in #2013).
- **Tests** — the three `redux.test.js` thunk cases move to `apiHooks.test.tsx` as hook tests, plus three new ones: the already-complete guard (no POST), the outline-never-loaded path (units write lands, one log), and unmount survival (a deferred POST resolves after the hook's component unmounts; the writes must still land — verified to fail if the callbacks move to the `mutate` call site). The five course-outline sidebar test files gain a `QueryClientProvider`, since every component there now mounts the hook transitively.

## Testing

`npm run types` (0 errors), `npm run lint` (clean), and the full suite (112 suites, 1106 passed, 3 pre-existing skips, 0 failures) all green; `thunks.js` line coverage is unchanged by the test removal. Manual pass on tutor local (DemoX) verified unit-nav completion + sidebar rollups, sidebar unit click, the no-repeat-POST short-circuit, and the sidebar-never-opened path (with its expected pre-existing `logError`); the unmount-under-latency mechanism is pinned by the new jest test, and the locked-sequence `courseOutlineShouldUpdate` refetch needs a prereq-gated course (reducer untouched by this layer).

## Decisions

<details>

# Decisions: #2012 — Convert `checkBlockCompletion` to a React Query mutation

## 1. No outline `setQueryData` at this layer (issue-body correction)

The issue said to update "the outline cache (`setQueryData`, replacing
`updateCourseOutlineCompletion`)". That's not possible yet: the sidebar outline
(`getCourseOutlineStructure` → `state.courseware.courseOutline`) is still Redux until
#2013 converts it — there is no outline query to write into. This layer keeps
dispatching `updateCourseOutlineCompletion` (transitional); #2013 swaps that dispatch
for a cache update. Likewise the `units` write stays a direct `updateModel` dispatch
rather than "via the bridge": the model-store bridge is wired to the QueryCache only,
and the payload is shaped from mutation variables + result (`{ id: unitId, complete }`),
which the declarative `meta.models` mirror can't express. #1977 later moves that write
to the sequence query cache.

## 2. Transitional dispatches live in the hook-level `onSuccess`

TanStack v5 skips `mutate(vars, { onSuccess })` callbacks if the observing component
unmounted, and the sidebar unmounts routinely before the POST resolves (mobile collapses
it immediately on unit click; any unit click navigates away from the old sequence's
`UnitLinkWrapper`s). The thunk always ran to completion, so the dispatches sit on the
mutation options (`useMutation({ onSuccess })`), which run on the mutation itself
regardless of unmount.

## 3. The "already complete → no-op" guard is a call-time `useStore()` read

The thunk read `getState().models.units[unitId]?.complete` at dispatch time. The hook
reproduces that with `useStore()` (first use in the repo; standard react-redux) instead
of `useSelector`, which would subscribe both consumers to every model write and
re-render them constantly. The guard chains on `units` itself
(`units?.[unitId]?.complete`), which the thunk did not: `models.units` is undefined
until the first unit write, and the outline sidebar can fire this before any sequence
has loaded. In the thunk that threw inside an async function, so the click handler
returned normally and `Link` still navigated client-side, with an unhandled rejection in
the console. In the hook the callback is synchronous, so the same throw would escape
`onClick` before `Link`'s own handler ran, and the browser would follow the `href` as a
full page load (review, arbrandes). With the chaining, a bare store means "not
complete", the mutation fires, and its success handler creates the map. A hook test
covers the bare-store case; the others seed `models.units` via `seedSequenceModels`,
mirroring the old `redux.test.js` `beforeEach`.

## 4. The outline dispatch keeps the thunk's catch-all

When the sidebar outline was never fetched (`courseOutline: {}` — sidebar disabled or
unopened) and `isComplete` is true, the `updateCourseOutlineCompletion` reducer throws
(`state.courseOutline.units` is undefined). The thunk's single `try/catch` caught that
and `logError`'d it, *after* the units-model write had already landed. The hook wraps
that one dispatch in an explicit `try/catch` + `logError` — reproducing exactly that
(one log, units write intact) without relying on TanStack's onSuccess-throw → onError
routing, which would also flip the mutation to error state despite the successful POST. This pre-existing outline-not-loaded `logError` is kept deliberately; a new
test pins the ordering (units write survives, outline stays `{}`, one log).

## 5. Placement: `apiHooks.ts`, one hook

Considered `statusBridge.ts` (the transitional-Redux-writes file) and rejected it: its
charter is mirroring query *status* into slice status fields, and course-home already
set the precedent of mutations with side-effectful hook-level `onSuccess` living in
`apiHooks.ts` (`usePostCourseDeadlines`, `useRequestCert`). The hook survives
#2013/#1977 — only its `onSuccess` internals shift from dispatches to `setQueryData`.

## 6. The returned callback keeps the thunk's call signature

`useCheckBlockCompletion()` returns `(courseId, sequenceId, unitId) => void`, so both
call sites are a rename-on-read: `const checkBlockCompletion = useCheckBlockCompletion();`
and the call line just loses its `dispatch(...)` wrapper.

## 7. Typing: nullable `courseId` / `sequenceId`, optional `unitId`

The container passes `courseId`/`sequenceId` read from the still-untyped Redux slice —
TS infers those fields as `null` from the slice's JS `initialState` — and `routeUnitId`
from `useParams()` is `string | undefined`. The callback and mutation vars are typed to
what the call sites actually pass (`string | null`, `unitId?: string`) instead of
asserting; the nullability disappears when those readers convert (#1976). The catch
variable is `unknown` in TS, so the reducer throw is logged as `logError(error as
Error)`. No `as any` anywhere; the one cast through `unknown` (the test's seeded-outline
accessor) is because the slice's inferred `courseOutline: {}` initial type doesn't
overlap the fully-loaded shape the reducer replaces it with.

## 8. No retry configuration

Mutations default to `retry: 0` in TanStack (the client's `shouldRetryQuery` and the
#2024 smart-retry policy apply to queries only), matching the thunk's single attempt.

## 9. Tests

Where the thunk's behavioral contract is now pinned:

| Behavior the thunk guaranteed | Where it's tested now |
|---|---|
| Fires the `get_completion` POST for the given course/sequence/unit | All three ported hook tests in `apiHooks.test.tsx` (same URL assertion) |
| `complete: true` → units model marked complete and outline rolls up (unit → sequence → section) | "marks the unit complete in the units model and rolls the outline up" |
| `complete: false` → units model written `false`, outline untouched | "writes complete: false to the units model and leaves the outline untouched" |
| Request failure → one log, no writes | "logs the error and writes nothing when the completion request fails" |
| Already complete → no request at all | "skips the request entirely when the unit is already complete" (thunk behavior that had no test) |
| Outline never loaded → units write still lands, error logged | "still marks the unit complete, and logs, when the outline was never loaded" (thunk behavior that had no test) |
| Writes land even if the caller unmounts mid-POST | "still writes completion when the caller unmounts before the response arrives" (guaranteed structurally in the thunk world; a deliberate choice here — decision 2) |
| Container wiring: unit-nav click → completion check | `CoursewareContainer.test.jsx` "marks the current unit complete when navigating to the next unit" (unchanged; now exercises the hook end-to-end) |

Untested before and after (no coverage lost): the sidebar's `handleUnitClick` wiring
(sidebar suites only verify rendering) and the `courseOutlineShouldUpdate`
locked-sequence flag (reducer behavior, unchanged by this layer; needs a prereq-gated
course to verify by hand). The old network-error test also incidentally *executed*
`getCourseOutlineStructure`'s failure branch without asserting on it; that branch is
still executed by other suites (`thunks.js` line coverage confirmed unchanged).

- The three `redux.test.js` thunk cases moved to `apiHooks.test.tsx` as hook tests
  (`renderHook` under `AppProvider` + `QueryClientProvider`), keeping the old
  assertions: complete=true rolls up unit → sequence → section; complete=false writes
  `complete: false` to the units model and leaves the outline untouched; a network
  error logs once and writes nothing.
- Two new cases pin behavior that used to live inside the thunk: the already-complete
  guard fires **no POST**, and the outline-never-loaded path still marks the unit
  complete while logging once (decision 4).
- A third new case pins the unmount-survival guarantee (decision 2): a deferred POST
  resolves only after the hook's component unmounts, and the writes must still land.
  The thunk had this for free (store-owned promise chain, no lifecycle); the hook has
  it only because the dispatches live in the hook-level `onSuccess`. Verified to fail
  if the callbacks move to the `mutate` call site — the equivalent-looking refactor
  it exists to catch.
- All five course-outline sidebar test files got a `QueryClientProvider` (with
  `createTestQueryClient(store)`) inside their hand-built `AppProvider` wrappers —
  every component there mounts `useCourseOutlineSidebar`, which now calls
  `useMutation`. No assertion changes.
- `executeThunk`'s optional third arg is dropped at the new TSX call site
  (`getCourseOutlineStructure` never reads `getState`, and the JS default parameter
  types the arg as `undefined` under TS).

## 10. Logging assertions capture `loggingService` from `initializeMockApp`

Unlike the query describes in `apiHooks.test.tsx` (queries don't log — the QueryCache
`onError` does, and that's tested in `queryClient.test.ts`), this mutation owns its
logging (`onError: logError` + the outline-dispatch catch), so its tests assert it.
The assertion handle is `const { loggingService } = initializeMockApp()`, which is how
this migration's already-converted mutations are tested
(`course-home/data/apiHooks.test.tsx`: `useRequestCert`, `usePostCourseDeadlines`,
`usePostEvent` — same capture, same `beforeEach` reset, same
`loggingService.logError` assertions). Considered `jest.mock('@edx/frontend-platform/logging')`
+ `jest.mocked(logError)` instead — a more direct assertion on the function the hook
calls — but jest hoists module mocks file-wide (the query describes' error tests also
route through the logging module via the QueryCache), and it would make the two
mutation apiHooks test files disagree on pattern.

## 11. Verification performed

Automated: full jest suite green (112 suites, 1106 passed, 3 pre-existing skips),
`npm run types` and `npm run lint` clean. `thunks.js` line coverage confirmed unchanged
by the test removal (only pre-existing gap: `getCourseDiscussionTopics`' catch, #2016's
territory).

Manual (tutor local, DemoX, staff user), verified by hand:
- Unit navigation marks the left-behind unit complete — POST fires, sequence-nav
  indicator updates, open sidebar's sequence/section rollups tick up.
- Sidebar unit click checks the previously-active unit — same POST + indicator updates.
- Already-complete short-circuit — navigating back and forth across a completed unit
  fires **no** repeat `get_completion` POST.
- Sidebar never opened — the unit still gets marked complete in the sequence nav, with
  the expected (pre-existing, decision 4) `logError` page action for the outline
  reducer.

Not re-verified by hand: unmount-under-real-latency (the mechanism is jest-pinned —
decision 9's unmount test) and the locked-sequence `courseOutlineShouldUpdate` refetch
(needs a prereq-gated course; the reducer is unchanged by this layer).

## 12. Error logging is hook-level for mutations, centralized for queries

The repo has two error-logging conventions, split by kind, and this hook follows the
mutation one. Queries never log in the hook: their errors surface to the QueryCache
`onError` in `queryClient.ts` (#2022), where `meta.logStatusAs` picks the level — that's
why the query hooks in the same file carry only `meta` and no catch blocks. Every
converted mutation logs in the hook via `onError: (error) => logError(error)`:
`usePostCourseDeadlines`, `usePostEvent`, `useRequestCert`
(`course-home/data/apiHooks.ts`) and both product-tours mutations
(`product-tours/data/apiHooks.ts`). No MutationCache equivalent of the #2022 handler
exists — a handful of three-line `onError`s hasn't justified the machinery — so
`useCheckBlockCompletion` uses the same hook-level `onError`. The inner `try/catch`
around the outline dispatch (decision 4) is separate from this convention and has no
precedent: no other mutation's `onSuccess` dispatches into a reducer that can throw.

</details>


src/courseware/data/apiHooks.ts
```diff
@@ -45,3 +52,38 @@ export const useSequenceMetadata = (sequenceId: string | undefined, isPreview: b
...
+      } catch (error) {
+        logError(error as Error); // the reducer throws when the sidebar outline isn't loaded
+      }
+    },
+    onError: (error) => logError(error),
+  });
+
+  return useCallback((courseId: string | null, sequenceId: string | null, unitId?: string) => {
+    const { units } = (store.getState() as { models: { units: Record<string, { complete?: boolean }> } }).models;
+    if (unitId && units[unitId]?.complete) {
```
arbrandes: Worth `units?.[unitId]?.complete` here: `models.units` is undefined until a unit model is written, and unlike the async thunk (which just left a rejected promise) this callback is synchronous, so the TypeError escapes into `onClick` and react-router's `Link` skips its own handler, turning the click into a full page load.

arbrandes (changes requested):
One minor change, but worth doing:

arbrandes (approved):
(Changing to approved.  Merge at your discretion!)

src/courseware/data/apiHooks.ts
brian-smith-tcril: addressed in https://github.com/openedx/frontend-app-learning/compare/a58013e2935a39fe14d9b71e1a02f4407b84aab6..482a9eb96344577a3439bd27c96dd92f8ea5d327

PR #2082, opened 2026-09-21
refactor!: move the plugin override registry off Redux onto React context

### LINKED ISSUE #2017: Convert the plugin-store override registry off Redux to React context

**Goal:** convert the plugin-store override registry off Redux to a React context / module-level registry, so the `plugins` reducer can leave `src/store.ts`.

**Context — this is *not* `@openedx/frontend-plugin-framework` (FPF).** FPF is the declarative `PluginSlot` system (`PluginSlot` / `DIRECT_PLUGIN` / `PLUGIN_OPERATIONS`), used in `src/plugin-slots/*` and config-driven — it never touches `state.plugins`. `generic/plugin-store` is a separate, bespoke Redux "override-method" registry: `registerOverrideMethod` stores function references by plugin+method (hence the serializable-check exemption in `store.ts`), and `usePluginsCallback(name, default)` folds registered overrides over a default. Its only in-repo consumer is the iframe-URL override in `src/courseware/course/sequence/Unit/index.jsx`.

**Why Stage 1:** this is learning's *own* Redux. Getting `redux` / `react-redux` / `@reduxjs/toolkit` out of learning's `package.json` must happen before the frontend-base move, so this slice has to go regardless of the special-exams blocker. Only the *later* reshaping of the override API onto the frontend-base `Slot` system is Stage 2.

**Tasks**
- Replace the `plugin` slice + `registerOverrideMethod` + `usePluginsCallback` with a React context / module-level registry.
- Remove the `plugins` reducer from `store.ts` (and its `ignoredActions` serializable exemption).
- Keep the single in-repo consumer (`Unit/index.jsx` iframe-URL override) working.

**Open question (settle when scoped):** whether the external `registerOverrideMethod` dispatch contract can be preserved in Stage 1, or whether that reshaping must wait for the Stage-2 Slot migration — the redux removal itself does not.

**Verify:** iframe-URL override still applies; `plugins` reducer gone from the store; no `react-redux` in the plugin-store.

Prerequisite for #1978's "specialExams-only" store endpoint.

## Plan

> [!NOTE]

Investigation findings that settle the open question above and shape the task list:

- **This is client state, so it goes to context, not React Query.** Nothing in the registry is fetched, cached or invalidated: plugins (descendants) contribute overrides and the host component (an ancestor) reads them. That is the React-context half of OEP-0067 ADR-0010. A `PluginOverridesProvider` holds the registry in `useState` and is mounted once at the route root beside `ToastProvider`, the migration's established client-state pattern (state in the provider, use in the consumer).
- **The generic contract is preserved; only the transport changes.** The `registerOverrideMethod({ pluginName, methodName, method })` payload, the fold (`defaultMethod()` first, then each registered override receives the previous result, in registration order), overwrite on re-register, and persistence after the registrar unmounts all stay exactly as today. The registry is a public extension surface and operator `env.config.jsx` files are unsearchable, so replacing it with per-behavior typed hooks would break plugins we can't see. Functions in state were only awkward under Redux's serializability rule (hence the `store.ts` middleware exemption); context has no such rule, so the generic shape costs nothing.
- **Plugins register from their own effect, imperatively.** `useDispatch()` + `dispatch(registerOverrideMethod(...))` becomes `usePluginOverrides()` + `registerOverrideMethod(...)`; the plugin keeps ownership of its effect deps. An effect-owning convenience hook can come later without breaking this.
- **No implicit cleanup on unmount, but an opt-in `unregisterOverrideMethod`.** `SequenceContent` renders `<Unit key={unitId}>`, so the slot widget remounts per unit; for a registrar gated behind an async fetch (the known external plugin), implicit cleanup would make the next unit's iframe load un-overridden first and then reload once the plugin re-registers. Today the previous unit's override covers that first render. So the provider does not clean up on its own, and plugins that want cleanup return `unregisterOverrideMethod` from their effect.
- **Names drop the Redux vocabulary.** `generic/plugin-store` → `generic/plugin-overrides`; `PluginOverridesProvider` / `usePluginOverrides()`. `registerOverrideMethod` keeps its name (it describes what the plugin does, not the transport), and the host-side `usePluginsCallback(methodName, defaultMethod)` keeps its signature, so `Unit/index.jsx` changes only an import path.
- **`store.ts` loses its only middleware option along with the reducer** (the serializable-check exemption existed for this slice alone), leaving `models` + `specialExams`.
- **BREAKING (`refactor!:` + `BREAKING CHANGE:` footer), same class as #2077.** `registerOverrideMethod` can no longer be `dispatch`ed and the import path moves. The one known external registrar (`@edx/unit-translation-selector-plugin`; see the baseline-example comment below) also imports the model store's `useModel`, which #1977 breaks — the two should ship as one plugin release. A `generic/plugin-overrides/README.md` carries the plugin-author contract and the before/after migration note.
- **Tests run the real code for the first time.** The module has no tests today and `setupTest.js` mocks `usePluginsCallback` because the test store never mounted the `plugins` reducer. The test `render` wrapper gains the provider, the mock is deleted, a new test file pins fold order / overwrite / skip / per-call default / persistence / unregister / throw-outside-provider, and `Unit/index.test.jsx` gains a consumer-level case where a registered override reaches the iframe `src`. The manual test is the baseline `env.config.jsx` ported to the new hook, plus a next-unit navigation check for the no-cleanup decision.

<details>
<summary>Full plan</summary>

# Plan: #2017 — Move the plugin override registry off Redux onto React context

## Context

Part of epic #1946 (Redux → React Query, Stage 1). `generic/plugin-store` is
the last learning-owned reducer besides `models` (which #1977 removes), so this
is a prerequisite for #1978's "store = `specialExams` only" endpoint. It is
**not** `@openedx/frontend-plugin-framework` (FPF): FPF decides *where a plugin
renders*; this registry is a side channel a rendered plugin uses to *change host
behavior elsewhere in the tree*. Both arrived in the same commit (PR #1330,
2024-03-27, which also added FPF to `package.json`).

**Which half of OEP-0067 ADR-0010 this is.** The registry holds no server
state — nothing is fetched, cached or invalidated — so React Query does not
apply. It is client state contributed by descendants (plugins) and read by an
ancestor (the host component), which is exactly what React context is for.
The provider owns the state; plugins write to it from effects; the host reads
it during render. Same split as `ToastProvider` (state in the provider, use in
the consumer).

**Generality is preserved on purpose.** We know of one registrar and one
override point, but the registry is a public extension surface: any operator's
`env.config.jsx` plugin may register other method names or override
`getIFrameUrl` differently, and those files are untracked and unsearchable.
Replacing the registry with per-behavior typed hooks would force a new host
hook for every such use and break plugins we can't see. So the *contract* stays
generic — `(pluginName, methodName) → method`, folded over a default — and only
the *transport* changes (Redux dispatch → context). What made functions-in-state
awkward was Redux's serializability rule (hence the middleware exemption in
`store.ts`); React context has no such rule, so the generic shape costs nothing.

The baseline example (`env.config.jsx`, posted on the issue) is the reference
for "functional equivalent": ported to the new API it must behave identically.

### How it works today

Three files, no tests of their own:

- `slice.js` — `createSlice({ name: 'plugin', initialState: {} })` with one
  reducer, `registerOverrideMethod({ pluginName, methodName, method })`, which
  sets `state[pluginName][methodName] = method`. Registering the same pair
  again **overwrites**. Nothing removes an entry.
- `hooks.js` — `usePluginsCallback(methodName, defaultMethod)`:
  `useSelector(state => () => fold(state.plugins), shallowEqual)`. The fold runs
  `defaultMethod()` first, then every registered `plugin[methodName]` receives
  the previous result, in `Object.values` insertion order. The selector returns
  a fresh function each run, so `shallowEqual` never matches and the consumer
  re-renders on every store change.
- `store.ts` — mounts the reducer as `plugins` and carries the file's **only**
  middleware option: the serializable-check exemption for
  `plugin/registerOverrideMethod` / `plugins` (the payload is a function).

**Consumer (one in-repo):** `courseware/course/sequence/Unit/index.jsx` —
`const getUrl = usePluginsCallback('getIFrameUrl', () => getIFrameUrl({ id, view, format, examAccess, jumpToId, preview }))`.
The default closure is re-created per render, so overrides compose over the
unit's current props.

**Registrar (none in-repo; one known external):** the published
`@edx/unit-translation-selector-plugin` (npm 0.5.0; source repo is private).
An FPF `DIRECT_PLUGIN` in `org.openedx.frontend.learning.unit_title.v1`, so it
renders inside the app's react-redux `Provider`. It calls `useDispatch()` and,
in a `useEffect` keyed on `[courseLanguage, selectedLanguage]`, dispatches
`registerOverrideMethod({ pluginName: id, methodName: 'getIFrameUrl', method })`,
importing the action creator from the **`@src/generic/plugin-store` alias** (it
compiles against learning's source via `env.config.jsx`). `id` is the widget id
FPF passes from the slot config. Registration is gated behind an async config
fetch (the widget renders nothing until `fetchTranslationConfig` resolves).

**Semantics observed (the contract to preserve):**

1. Registration from a slot-rendered component inside the React tree.
2. Keyed by `(pluginName, methodName)`, overwrite on re-register; latest wins.
3. Fold order: default first, then each override receives the previous result.
4. **No unregister.** An override outlives the component that registered it.
5. The consumer's default closure is evaluated per call (sees current props).

**Test surface today:** `setupTest.js` `jest.mock`s the module so
`usePluginsCallback` returns the default callback untouched — because
`initializeTestStore` never mounts a `plugins` reducer, the real hook would
throw on `Object.values(undefined)`. No test exercises the registry.

### D1 — A context provider holds the registry

`PluginOverridesProvider` keeps `Record<pluginName, Record<methodName, method>>`
in `useState`. Consumers read it with `useContext`. Mounted once at the route
root in `src/index.jsx` beside `ToastProvider` (any ancestor of the host
component works; root matches the sibling and means every slot in the app can
register). The alternative — a module-level `Map` with `useSyncExternalStore`,
which would let `registerOverrideMethod` stay a plain importable function — has
no precedent in this repo and would widen the contract to registration from
outside React; rejected for Stage 1.

### D2 — Plugins register imperatively from their own effect

Today's shape is "action creator + dispatch inside the plugin's `useEffect`".
The context equivalent is "setter from context + call inside the plugin's
`useEffect`": the plugin keeps ownership of *when* it registers (its deps), and
the payload `{ pluginName, methodName, method }` is unchanged.

```js
// before                                   // after
const dispatch = useDispatch();              const { registerOverrideMethod } = usePluginOverrides();
useEffect(() => {                            useEffect(() => {
  dispatch(registerOverrideMethod({ … }));     registerOverrideMethod({ … });
}, [deps]);                                  }, [deps]);
```

An effect-owning convenience hook (`useOverrideMethod(payload)`) can be added
later without breaking this; not in this PR.

### D3 — Additive `unregisterOverrideMethod`; no implicit cleanup

Today nothing removes an entry, and the host relies on that: `SequenceContent`
renders `<Unit key={unitId}>`, so `Unit` and the slot widget remount on every
unit change. For a registrar gated on an async fetch (the translations plugin),
implicit cleanup-on-unmount would mean unit B's first render has no override →
iframe loads untranslated → fetch resolves → register → URL changes → iframe
reloads. Today unit A's override (closing over course-level language values)
covers B's first render and the iframe loads once. So the provider does **not**
clean up on its own. It does expose `unregisterOverrideMethod({ pluginName, methodName })`
so a plugin that wants cleanup can return it from its effect — a React-idiomatic
opt-in that the Redux version had no room for.

### D4 — Names drop the Redux vocabulary, keep the plugin-facing verbs

- Directory `generic/plugin-store/` → **`generic/plugin-overrides/`** ("store"
  is the word we're removing; the break is happening anyway, so the import-path
  change rides along in the same major and is one line in the migration note).
- `PluginOverridesProvider`, `usePluginOverrides()` (context value:
  `{ overrideMethods, registerOverrideMethod, unregisterOverrideMethod }`).
- `registerOverrideMethod` keeps its **name and payload** — it describes what
  the plugin does, not the transport.
- Host-side `usePluginsCallback(methodName, defaultMethod)` **keeps its name
  and signature**; it was never Redux vocabulary and `Unit/index.jsx` stays
  byte-identical. (Renaming it is possible later; not a de-Redux concern.)

### D5 — Throw outside the provider

`usePluginOverrides` / `usePluginsCallback` outside `PluginOverridesProvider`
throw, matching `useToast`. The provider is mounted at the root in the app and
in `setupTest`'s `render` wrapper; the only tests rendering a real `Unit`
(`Unit/index.test.jsx`, the `Sequence`/`Course` suites) go through that
wrapper, and `CoursewareContainer.test.jsx` mocks `Unit`.

### D6 — Stack placement

A layer on top of #2081 in stack #2080, like every layer so far, rather than a
standalone PR off `master`. The only overlap with the open layers is the
`store.ts` reducer map (#2081 removes `courseHome`; this removes `plugins`) —
trivial either way.

## 1. `generic/plugin-overrides/PluginOverridesContext.tsx` (new, TS)

Replaces `plugin-store/{slice,hooks,index}.js`; `index.ts` re-exports. Shape,
following `ToastContext.tsx`:

```ts
export type OverrideMethod<T = unknown> = (previousResult: T) => T;
type OverrideMethods = Record<string /* pluginName */, Record<string /* methodName */, OverrideMethod<any>>>;

interface PluginOverridesContextValue {
  overrideMethods: OverrideMethods;
  registerOverrideMethod: (payload: { pluginName: string; methodName: string; method: OverrideMethod<any> }) => void;
  unregisterOverrideMethod: (payload: { pluginName: string; methodName: string }) => void;
}

export const PluginOverridesProvider = ({ children }) => {
  const [overrideMethods, setOverrideMethods] = useState<OverrideMethods>({});
  const registerOverrideMethod = useCallback(({ pluginName, methodName, method }) => {
    setOverrideMethods((current) => ({
      ...current,
      [pluginName]: { ...current[pluginName], [methodName]: method },
    }));
  }, []);
  const unregisterOverrideMethod = useCallback(({ pluginName, methodName }) => {
    setOverrideMethods((current) => {
      const { [methodName]: _removed, ...rest } = current[pluginName] ?? {};
      return { ...current, [pluginName]: rest };
    });
  }, []);
  const value = useMemo(() => ({ overrideMethods, registerOverrideMethod, unregisterOverrideMethod }), [overrideMethods]);
  return <PluginOverridesContext.Provider value={value}>{children}</PluginOverridesContext.Provider>;
};

export const usePluginOverrides = () => { /* useContext; throw outside provider */ };

export const usePluginsCallback = <T,>(methodName: string, defaultMethod: () => T): () => T => {
  const { overrideMethods } = usePluginOverrides();
  return () => Object.values(overrideMethods).reduce(
    (result, plugin) => (plugin[methodName] ? plugin[methodName](result) : result),
    defaultMethod(),
  );
};
```

- The fold is the same behavior as `hooks.js`: default first, insertion order,
  only plugins that define `methodName`.
- The setters are stable (`useCallback([])`), so a plugin effect listing them in
  deps does not loop. Re-registering (even the same function) produces a new
  `overrideMethods` object and re-renders consumers — parity with today's
  every-dispatch re-render; not optimized here.
- Typing: the registry is heterogeneous by nature (a method's value type is
  whatever the host's default returns for that name), so the stored map is
  `OverrideMethod<any>` at the boundary and `usePluginsCallback<T>` gives the
  host a typed `() => T`. That is a precise boundary cast at one place, not
  `as any` sprinkled at call sites.

## 2. Mount + store teardown

- `src/index.jsx`: `<PluginOverridesProvider>` wraps `<Routes>` alongside
  `<ToastProvider>`.
- `src/store.ts`: remove the `pluginsReducer` import and `plugins:` entry and
  delete the whole `middleware:` option (its sole purpose was the plugin
  exemption) → `configureStore({ reducer: { models, specialExams } })`.
  `RootState` shrinks accordingly.
- Delete `src/generic/plugin-store/` (`slice.js`, `hooks.js`, `index.js`).
- `Unit/index.jsx`: import path `@src/generic/plugin-store` →
  `@src/generic/plugin-overrides`; nothing else.

## 3. Docs — `generic/plugin-overrides/README.md` (new)

The plugin-author contract, since the consumers are external: what the
registry is (and that it is not FPF), `usePluginOverrides()` +
`registerOverrideMethod` with the `useEffect` example, the fold semantics
(default first, registration order, overwrite, no implicit cleanup, opt-in
`unregisterOverrideMethod`), the method names the host folds today
(`getIFrameUrl`, with the `Unit` default's inputs), and a **migration note**
for pre-#2017 plugins: the before/after in D2 plus the import path change.
Same convention as #2077's `progress-tab/README.md`.

## 4. BREAKING

`refactor!:` subject + `BREAKING CHANGE:` footer (semantic-release major):
`registerOverrideMethod` is no longer a Redux action creator and cannot be
`dispatch`ed; plugins obtain it from `usePluginOverrides()` (imported from
`@src/generic/plugin-overrides`); the `plugins` reducer is gone from the
store. Same class as #2077 (`examsData` → `useExamsData()`). Note for the
release/PR: the known external plugin also imports `@src/generic/model-store`
`useModel`, which #1977 breaks — coordinate one plugin release for both.

## Tests

- **`generic/plugin-overrides/PluginOverridesContext.test.tsx`** (new — the
  module has no tests today): `renderHook` under the provider.
  - default only → `defaultMethod()`'s value.
  - one override composes over the default; two plugins compose in
    registration order (each receives the previous result).
  - overwrite: re-registering the same `(pluginName, methodName)` replaces,
    doesn't chain.
  - a plugin without `methodName` is skipped; other method names don't leak.
  - default closure re-evaluated per call (change a captured value, call again).
  - persistence: a registrar that unmounts leaves its override in place (pins
    D3); `unregisterOverrideMethod` removes exactly that entry.
  - hooks outside the provider throw.
- **`src/setupTest.js`**: add `<PluginOverridesProvider>` to the `render`
  wrapper and **delete the `@src/generic/plugin-store` `jest.mock`** — with
  the provider present and an empty registry the real hook returns the
  default, which is what the mock faked. `Unit/index.test.jsx` (asserts iframe
  `src === getIFrameUrl(...)`) now runs the real fold.
- **`Unit/index.test.jsx`**: one added case — a tiny test widget registers a
  `getIFrameUrl` override; assert the iframe `src` reflects it (consumer-level
  proof the registry still reaches the URL).
- Expected green as-is (verify): `Sequence`, `SequenceContent`, `Course`,
  `CoursewareContainer` suites.

## Behavior changes

- Plugin API transport only (dispatch → context hook; import path). Payload,
  fold, overwrite and persistence semantics unchanged.
- New capability: opt-in `unregisterOverrideMethod`.
- Consumer re-render on registration: same as today.

## Decision doc (for the PR body)

1. Client state → context, not React Query (nothing is fetched); provider at
   the root beside `ToastProvider`.
2. Generic `(pluginName, methodName) → method` contract preserved — a public,
   unsearchable extension surface; functions in context state need no
   serializability exemption.
3. Plugins register imperatively from their own effect (`usePluginOverrides()`);
   payload unchanged; effect-owning hook deferred.
4. No implicit cleanup (mechanism: keyed `Unit` remount + async-gated
   registrar → double iframe load); additive `unregisterOverrideMethod`.
5. `plugin-store` → `plugin-overrides`; `usePluginsCallback` keeps its
   signature so `Unit/index.jsx` changes only an import path.
6. Throw outside the provider; `setupTest` provides it and the module mock is
   deleted so tests run the real code.
7. `store.ts` loses its only middleware option along with the reducer.
8. BREAKING footer; README with the migration note; coordinate with #1977's
   break of the same plugin.
9. Heterogeneous registry typed `OverrideMethod<any>` at the boundary,
   `usePluginsCallback<T>` typed at the host; no `as any` at call sites.

## Stack

Per D6: one new layer on top of #2081. Closes #2017.

## Verification

- `nvm use && npm run test -- src/generic/plugin-overrides src/courseware/course/sequence` (output to a file), then the full suite.
- `nvm use && npm run types` and `npm run lint:fix`.
- `git grep -n "plugin-store\|plugin-overrides\|registerOverrideMethod\|usePluginsCallback" src` → only the new module, its README/tests, `index.jsx` (provider), `Unit/index.jsx`, `setupTest.js` (provider); `git grep react-redux src/generic/plugin-overrides` empty; `plugins` absent from `store.ts`.
- Manual (tutor local): port the baseline `env.config.jsx` per D2 (swap the hook, drop `dispatch`, fix the import path), `npm run dev`, open a unit: the card renders, the switch adds/removes `show_title=1` on the iframe `src` and the in-iframe title appears/disappears; navigate to the next unit with the switch on and confirm the override applies on first load (D3). Record in `manual-testing-2017.md`.

</details>

brian-smith-tcril:
## Summary

Move the plugin override registry off Redux onto React context. `generic/plugin-store` let a plugin rendered in a slot register an override for a host-computed value (`registerOverrideMethod`), and the host folded the registered overrides over its default (`usePluginsCallback`); its one in-repo consumer is the unit content iframe URL. It was learning's own Redux and the only reason `store.ts` carried a middleware option. Nothing in it is server state, so it becomes a `PluginOverridesProvider` with the same contract — payload, fold order, overwrite, persistence — and only the transport changes. `store.ts` is now `models` + `specialExams`. Part of the Redux → React Query migration (#1946, Stage 1), stacked on #2081. Closes #2017.

## What changed

- **`generic/plugin-overrides/PluginOverridesContext.tsx` (new, replaces `generic/plugin-store/{slice,hooks,index}.js`).** `PluginOverridesProvider` holds `pluginName → methodName → method` in `useState`; `usePluginOverrides()` exposes `registerOverrideMethod` (same `{ pluginName, methodName, method }` payload as the old action creator, overwrite on re-register) and a new `unregisterOverrideMethod`; `usePluginsCallback(methodName, defaultMethod)` keeps its signature and fold (default first, then each registered override receives the previous result, in registration order). Both hooks throw outside the provider, like `useToast`. No `any`: the stored map is typed `(previousResult: never) => unknown`, which accepts any `OverrideMethod<T>` on registration without a cast and is uncallable until the host asserts its `T` in the fold — the single cast in the module, at the one place the host knows the type.
- **No implicit cleanup on unmount, by design.** `SequenceContent` renders `<Unit key={unitId}>`, so the slot widget remounts per unit; for a registrar gated behind an async fetch, cleanup would make the next unit's iframe load un-overridden and then reload. Today's behavior (the previous unit's override covers the first render) is preserved; plugins that want cleanup call `unregisterOverrideMethod` from their effect.
- **`src/index.jsx`** mounts `PluginOverridesProvider` around `Routes`, inside `ToastProvider`.
- **`src/store.ts`** drops the `pluginsReducer` and the entire `middleware` option (the serializable-check exemption existed for this slice alone).
- **`Unit/index.jsx`** changes only its import path.
- **`generic/plugin-overrides/README.md` (new):** the plugin-author contract (registering, fold semantics, the `getIFrameUrl` method the host folds today, adding a host method) and a before/after migration note from the Redux version.
- **Tests:** the module gets its first test file — default-only, composition order across two plugins, overwrite, skipping other method names, per-call default evaluation, `unregister` removing exactly one entry, persistence after the registrar unmounts, and the throw outside the provider. `setupTest.js`'s render wrapper provides `PluginOverridesProvider` and the module mock is deleted (it faked an empty registry because the test store never mounted the reducer), so `Unit/index.test.jsx`'s existing iframe-URL assertion now runs the real fold; it also gains a case where a sibling component registers a `getIFrameUrl` override and the iframe `src` reflects it.

## Breaking change

`registerOverrideMethod` is no longer a Redux action creator and cannot be `dispatch`ed; plugins import `usePluginOverrides` from `@src/generic/plugin-overrides` (was `@src/generic/plugin-store`) and call `registerOverrideMethod(payload)` from an effect. The `plugins` reducer is gone from the store. Same class as #2077. The one known external registrar (`@edx/unit-translation-selector-plugin`, see the baseline-example comment on #2017) also imports the model store's `useModel`, which #1977 will break — the two should ship as one plugin release.

## Testing

`npm run types` (0 errors), `npm run lint` (clean), full jest suite green at head (109 suites, 1112 passed / 3 pre-existing skips). Manual pass on tutor local with the baseline example plugin from the #2017 issue comment ported to the new API: all five by-hand checks passed; the store shape and override persistence rest on `store.ts` and the context tests respectively — see the details block below.

## Decisions

<details>

# Decisions — plugin override registry off Redux onto React context (#2017)

Entries 1–9 were settled in the plan (issue #2017 body, 2026-09-20) before
implementation; 10 onward landed with the code.

1. **Client state → React context, not React Query.** Nothing in the registry
   is fetched, cached or invalidated: plugins (descendants) contribute
   overrides and the host component (an ancestor) reads them. That is the
   context half of OEP-0067 ADR-0010. `PluginOverridesProvider` holds the
   registry in `useState` and is mounted once at the route root beside
   `ToastProvider`, the migration's established client-state pattern (state
   in the provider, use in the consumer).

2. **The generic contract is preserved; only the transport changes.** The
   `registerOverrideMethod({ pluginName, methodName, method })` payload, the
   fold (default first, then each registered override receives the previous
   result, in registration order), overwrite on re-register, and persistence
   after the registrar unmounts are unchanged. The registry is a public
   extension surface — operator `env.config.jsx` plugins are untracked and
   unsearchable — so replacing it with per-behavior typed hooks would break
   plugins we can't see. Functions in state were only awkward under Redux's
   serializability rule (the `store.ts` middleware exemption); context has no
   such rule, so keeping the generic shape costs nothing.

3. **Plugins register imperatively from their own effect.** Today's
   "action creator + dispatch inside the plugin's `useEffect`" becomes
   "setter from context + call inside the plugin's `useEffect`". The plugin
   keeps ownership of *when* it registers (its deps). An effect-owning
   convenience hook can be added later without breaking this; not in this PR.

4. **No implicit cleanup on unmount; additive `unregisterOverrideMethod`.**
   `SequenceContent` renders `<Unit key={unitId}>`, so `Unit` and the slot
   widget remount on every unit change. For a registrar gated behind an async
   fetch (the known external plugin renders nothing until its config fetch
   resolves), implicit cleanup would mean unit B's first render has no
   override → iframe loads un-overridden → fetch resolves → register → URL
   changes → iframe reloads. Today unit A's override covers B's first render
   and the iframe loads once. So the provider does not clean up on its own;
   it exposes `unregisterOverrideMethod({ pluginName, methodName })` for
   plugins that want cleanup from their effect.

5. **Names drop the Redux vocabulary; plugin-facing verbs stay.**
   `generic/plugin-store` → `generic/plugin-overrides` ("store" is the word
   being removed; the break is happening anyway so the import-path change
   rides in the same major). `PluginOverridesProvider` / `usePluginOverrides()`.
   `registerOverrideMethod` keeps its name — it describes what the plugin
   does, not the transport. Host-side `usePluginsCallback(methodName,
   defaultMethod)` keeps its name and signature; `Unit/index.jsx` changes only
   its import path.

6. **Throw outside the provider**, matching `useToast`. The provider is
   mounted at the root in the app and in `setupTest`'s `render` wrapper.

7. **`store.ts` loses its only middleware option along with the reducer.**
   The `serializableCheck` exemption existed for this slice alone; the store
   is now `configureStore({ reducer: { models, specialExams } })`.

8. **BREAKING (`refactor!:` + `BREAKING CHANGE:` footer), same class as
   #2077.** `registerOverrideMethod` can no longer be `dispatch`ed and the
   import path moves. A `generic/plugin-overrides/README.md` carries the
   plugin-author contract and the before/after migration note. The one known
   external registrar (`@edx/unit-translation-selector-plugin`) also imports
   the model store's `useModel`, which #1977 breaks — the two should ship as
   one plugin release.

9. **`never` in the stored map, one cast in the fold — no `any`.** The
   method-name set is open (decision 2) and each name's value type is the
   host's business, so the map can't name them. Two ways to say "unknown per
   key": `any` (instantly readable, two `no-explicit-any` disables, but
   unsound in both directions and silently — the fold would return `any` into
   the `T` accumulator with no visible assertion) or a parameter of `never`:
   `Record<string, Record<string, (previousResult: never) => unknown>>`. A
   function taking `string` is assignable to one taking `never`, so any
   `OverrideMethod<T>` registers without a cast (including from a typed
   plugin), and the stored method is uncallable until the host asserts its
   `T` — `(plugin[methodName] as OverrideMethod<T>)(result)` in
   `usePluginsCallback`. That forces the single trust point onto the page at
   the one place the host knows `T`, needs no lint disables, and doesn't
   propagate the way `any` does. Chosen over `any` for that visibility; the
   readability cost is paid once by the comment on the type. JS plugins
   (`env.config.jsx`, compiled npm packages) are unaffected: types are erased
   and `tsc` never includes root `.jsx` files.

10. **Tests run the real registry for the first time.** The module had no
    tests, and `setupTest.js` mocked `usePluginsCallback` because the test
    store never mounted the `plugins` reducer (the real hook would have thrown
    on `Object.values(undefined)`). The test `render` wrapper now provides
    `PluginOverridesProvider`, the module mock is deleted, and
    `Unit/index.test.jsx`'s existing iframe-URL assertion therefore exercises
    the real (empty) fold. New `PluginOverridesContext.test.tsx` pins:
    default-only, composition order across two plugins, overwrite, skip of
    other method names, per-call default evaluation, `unregister` removing
    exactly one entry, persistence after the registrar unmounts (decision 4),
    and the throw outside the provider. `Unit/index.test.jsx` gains a
    consumer-level case: a sibling component registers a `getIFrameUrl`
    override and the iframe `src` reflects it.

11. **Generic functions avoid shapes the autofixer breaks.** A generic arrow
    (`<T,>(…) =>`) in a `.tsx` file is autofixed by the repo's ESLint into
    `<T>(…)`, which the parser then reads as JSX and fails on; a named
    generic function expression inside `useCallback` gets rewritten to an
    arrow by `prefer-arrow-callback` and loses its `<T>` (and because Babel
    strips types, tests still pass — only `tsc` catches it, so run `lint:fix`
    *before* `types`). Hence `usePluginsCallback` is a `function`
    declaration, and `registerOverrideMethod` is typed through
    `useCallback<PluginOverridesContextValue['registerOverrideMethod']>(…)`,
    letting the arrow's parameter pick up `T` contextually.

</details>

## Manual testing

<details>
<summary>Checklist</summary>

# Manual testing — plugin override registry onto React context (#2017)

In-browser verification against a live backend (tutor local). This layer
claims **no user-visible change** for a learner: with no plugin registered the
unit iframe URL is exactly what `getIFrameUrl` builds, and with a plugin
registered the override applies the same way it did through Redux. The
observable changes are for plugin authors (the registration API) and in the
store (no `plugins` slice).

Course id: `course-v1:OpenedX+DemoX+DemoCourse`; any unit under
`/course/course-v1:OpenedX+DemoX+DemoCourse/...`.

- The baseline example plugin from the #2017 issue comment, ported to the new
  API, lives in the repo-root `env.config.jsx` (gitignored). It inserts a
  `DIRECT_PLUGIN` into `org.openedx.frontend.learning.unit_title.v1` that
  renders a Paragon Card with a switch; the switch registers a `getIFrameUrl`
  override that sets `show_title=1` (the `getIFrameUrl` default is `0`) or a
  pass-through.
- `nvm use && npm run dev`; the dev server picks up `env.config.jsx` at the
  repo root automatically (frontend-build resolves `env.config`).
- Inspect the iframe with the `#unit-iframe` element's `src` attribute in
  devtools; "the in-iframe title" means the unit title the LMS renders inside
  the xblock content when `show_title=1`.

## Verify by hand

- [x] **Example renders through the real provider** — open any unit: the
      "Plugin override example" card appears above the unit title, the
      default title and bookmark button still render (`keepDefault`), no
      console error from `usePluginOverrides`.
- [x] **Pass-through override leaves the default URL alone** — switch off:
      `#unit-iframe` `src` has `show_title=0`; no in-iframe title.
- [x] **Override applies** — switch on: `src` changes to `show_title=1` and
      the in-iframe title appears (the title now shows twice: MFE + LMS).
- [x] **Re-register overwrites** — switch off again: `src` back to
      `show_title=0`, in-iframe title gone. Toggle a few times; each flip
      takes effect (latest registration wins, no stacking).
- [x] **Courseware without the example** — remove/rename `env.config.jsx`,
      restart dev: a unit renders with the default title row and the iframe
      `src` has `show_title=0`; no errors (the provider is mounted with an
      empty registry).

- The store no longer carrying a `plugins` slice — not a runtime observation:
  the reducer map *is* `src/store.ts` (`models` + `specialExams`), `tsc`
  checks `RootState` against every remaining reader, and `git grep
  'state\.plugins'` over `src` is empty.
- Fold semantics — `PluginOverridesContext.test.tsx`: default-only,
  composition order across two plugins, overwrite, skipping other method
  names, per-call default evaluation, `unregisterOverrideMethod` removing
  exactly one entry, persistence after the registrar unmounts, throw outside
  the provider.
- The override reaching the iframe `src` — `Unit/index.test.jsx`
  (`applies a registered override to the iframe src`); the existing
  `generates correct iframeUrl` case now runs the real (empty) fold instead
  of the deleted `setupTest` mock.
- Two plugins composing on a live page — not exercised by hand (one example
  plugin); rests on the composition-order unit test.
- Override persistence across the unit remount (decision 4) —
  `PluginOverridesContext.test.tsx` (`keeps an override after the component
  that registered it unmounts`). A by-hand version was tried and dropped: the
  example widget keeps its switch in `useState`, so on the `<Unit key={unitId}>`
  remount it comes back off and re-registers the pass-through, overwriting the
  persisted override — the same thing the Redux slice would do. That tests the
  example's state handling, not the registry. A plugin that re-registers the
  same transform after a remount (the translations plugin keeps its language
  in `localStorage`) gets the first-render coverage decision 4 describes.

## Results

Env: tutor local, DemoX, local branch `bsmith/plugin-overrides-context` @
`0c924bda` (no PR yet). Run 2026-09-20.

- **Example renders through the real provider** — passed.
- **Pass-through override leaves the default URL alone** — passed.
- **Override applies** — passed.
- **Re-register overwrites** — passed.
- **Courseware without the example** — passed (example config commented
  out, dev server restarted).
- **Override persists across the unit remount** — tried and dropped from the
  by-hand list: after navigating to the next unit the iframe `src` was back to
  `show_title=0`, because the example's switch state reset on the remount and
  it re-registered the pass-through. That is the example overwriting its own
  entry, which the Redux slice would also have done; registry persistence is
  covered by the unit test named above.
- **No `plugins` slice in the store** — not run by hand (static property of
  `store.ts`; see above).

</details>


arbrandes (approved):
Approved.  The breaking change only requires minor changes to the only existing plugin.

We should discuss refactoring the registry (or doing away with it) for the move to frontend-base.  Claude helped me think this through:

---

The only real registrar is the published `@edx/unit-translation-selector-plugin`. What it needs:

1. Render in the unit title slot, with the slot props and the course language.
2. Stay hidden until an async config fetch confirms the feature and its languages.
3. Let the learner pick a language mid-session, persisted per course in localStorage.
4. Add `src_lang`/`dest_lang` to the iframe URL when the pick differs from the course language, without a page reload, on every change.
5. Keep applying across the `<Unit key={unitId}>` remount without a first-render gap.

None of that needs a function pipeline in Learning. We could use slots: the seam would be a unit content slot with `ContentIFrame` as its default widget. The plugin would `REPLACE` that widget with its own composition of the exported `ContentIFrame` and `getIFrameUrl`, adding `src_lang`/`dest_lang` from state it owns (requirement 4). Its selector stays a title-slot widget (1), gated on the config fetch (2), and the selection would live in a provider registered through `App.providers` (3). A provider mounted above the content widget has the value on its first render, so the reload-or-stale trade-off behind requirement 5 does not arise; and because the plugin's state stays in its own provider as data, neither does the re-render loop from #1330's `plugins/README.md` caveat.

What that asks of Learning: export `ContentIFrame` and pass the default `iframeUrl` on the slot, so the plugin composes rather than reimplements, and treat those as the slot's contract. Because the `REPLACE` is declared at config time and `condition` is synchronous, the plugin's widget has to self-gate, rendering the plain default when the feature is off or no language differs.

Generally speaking, frontend-base has no value-filter pipeline and I'm not sure we want one. Plus, ADR 0011's case against `Wrap` seems to apply to the override registry's function chain too. If the feature can be implemented as an app - even if we need to add a slot for it - it seems to be a way saner proposition.

brian-smith-tcril:
> The only real registrar is the published `@edx/unit-translation-selector-plugin`.

Claude kept trying to use that as justification for making a more narrow solution in this PR, and I had to push back a few times. The reality is that *we don't know* how the override functionality is being used by plugins at the moment, so keeping the pre-`frontend-base` version close to what currently exists was key here.

That being said, I 100% agree that this _isn't_ how we should do this in `frontend-base`. We can absolutely use the translations plugin and content iframe as a pattern setter, but we should also reach out to see if/how other plugins are using this functionality to see if there are any edge/corner cases.

Issue #2098, opened 2026-09-22
Stop the courseware gate queries refetching from components under the gate


brian-smith-tcril:
Part of #1946 — Redux → React Query migration (Stage 1). Layer on top of #2084.

**Goal:** fetch each of `useIsCourseLoaded`'s three queries once per navigation
into courseware, from `CoursewareContainer`, instead of once more for each
descendant that re-checks the gate.

**Context.** One outline → courseware navigation requests
`/api/course_home/course_metadata/`, `/api/courseware/course/` and
`/api/learning_sequences/v1/course_outline/` **three times each**, while
`/api/course_home/v1/navigation/` (`staleTime: Infinity`, outside the gate) is
requested once. A trace of the `course_metadata` requests shows the first from
`CoursewareContainer` and the other two from `UnitNavigation` and
`UnitNavigationEffortEstimate`, both through `useSequenceIds` →
`useIsCourseLoaded`. That hook is `CoursewareContainer`'s gate check, but it is a
fetching observer of all three queries, so every descendant that calls it (or
`useSequenceIds`) mounts after the data lands, finds it stale, and refetches.
There is no remount across the redirect. It is the same defect #2083 fixed for
the outline and dates tabs, and gets the same fix: the owner fetches, and the
gate check under it only reads.

**Tasks**
- Add `{ enabled }` to `useCourseHomeMeta`, `useCoursewareMetadata` and
  `useCoursewareOutline`.
- Make `useIsCourseLoaded` non-fetching: it passes `{ enabled: false }` to all
  three. `useSequenceIds` follows, since that hook is its only link to React
  Query. Neither takes an option, and the gated call sites (`CourseBreadcrumbs`,
  `useSequenceNavigationMetadata`, `UnitNavigationEffortEstimate`) are
  untouched. `CoursewareContainer` and `useCoursewareRedirects` keep fetching
  through the query hooks they already call directly.
- Component suites that render a gated component on its own mount the owner's
  fetches beside it (`src/tests/MountCourseQueryHooks.tsx`), rather than seeding
  the cache.
- Re-measure all four endpoints with the protocol below.

**Verify:** hard-reload the outline, wait for idle, clear the network log, click
a sequence title, wait for idle: each of the three gate endpoints is requested
once. Courseware, breadcrumbs, unit navigation and the redirect rules behave as
before.

**Out of scope:** `staleTime` (the owner's refetch on navigation is legitimate —
it is where `course_metadata` records the learner's daily streak); the
[...]

Issue #2085, opened 2026-09-21
Read courseHomeMeta from the query: tab-page, alerts, and course-home tabs


brian-smith-tcril:
Part of #1946 — Redux → React Query migration (Stage 1). Part of the #1977 model-store dissolution ([plan](https://github.com/openedx/frontend-app-learning/issues/1977)) — **Layer B2**. Depends on B1 (#2084).

**Goal:** convert the shared, course-home and alert readers of the `courseHomeMeta` model to the query. Half of the 53 `courseHomeMeta` readers; B3 (#2086) takes the rest and removes the bridge entry.

**Tasks** — 31 sites, each `useModel('courseHomeMeta', courseId)` → `useCourseHomeMeta(courseId, { enabled: false }).data ?? {}`:
- `tab-page/TabPage.tsx:76`, `tab-page/LoadedTabPage.tsx:39`.
- `alerts/` (9): `access-expiration-alert/hooks.js:30` (only its `'courseHomeMeta'` read — the `useModel(tab, …)` read is layer E / #1999), `active-enteprise-alert/hooks.js:8`, `course-start-alert/CourseStartAlert.jsx:26`, `course-start-alert/CourseStartMasqueradeBanner.jsx:16`, `course-start-alert/hooks.js:11,21,43`, `enrollment-alert/EnrollmentAlert.jsx:25`, `enrollment-alert/hooks.js:14`.
- `course-home/outline-tab/` (12): `OutlineTab.jsx:40`, `DateSummary.jsx:19`, five alert sites (`certificate-status-alert/hooks.js:37`, `course-end-alert/hooks.js:14`, `private-course-alert/PrivateCourseAlert.jsx:28` and `private-course-alert/hooks.js:11`, `scheduled-content-alert/hooks.js:22`), and five `widgets/` readers (`CourseDates`, `CourseTools`, `ProctoringInfoPanel`, `StartOrResumeCourseCard`, `WeeklyLearningGoalCard`).
- `course-home/progress-tab/` (5): `certificate-status/CertificateStatus.jsx:30`, `grades/course-grade/CourseGradeHeader.jsx:17`, `grades/detailed-grades/DetailedGrades.jsx:22`, `grades/detailed-grades/SubsectionTitleCell.jsx:25`, `related-links/RelatedLinks.jsx:16`.
- `course-home/dates-tab/` (2): `DatesTab.jsx:28`, `timeline/Day.jsx:29`.
[...]

PR #2099, opened 2026-09-22
refactor: take tab identity out of the course-home metadata query

### LINKED ISSUE #2084: Take tab identity out of the course-home metadata query
Part of #1946 — Redux → React Query migration (Stage 1). Part of the #1977 model-store dissolution ([plan](https://github.com/openedx/frontend-app-learning/issues/1977)) — **Layer B1**. Prerequisite for B2 (#2085) and B3 (#2086); no readers convert here.

**Goal:** take tab identity out of the course-home metadata query entirely, so the 53 `courseHomeMeta` readers can read it without naming a variant.

**Context.** `courseHomeQueryKeys.metadata(courseId, rootSlug)` caches the same endpoint twice — under `'outline'` for the course-home tabs and `'courseware'` for the courseware route. `rootSlug` never reaches the request. It changes exactly one field:

```js
// src/course-home/data/api.js
// The API uses "courseware" as a slug for both courseware and the outline tab.
slug: tab.tabId === 'courseware' ? rootSlug : tab.tabId,
```

That field is not server data — it is a pre-resolved answer to the nav's highlight check (`CourseTabLink.tsx:16`, `active: slug === activeTabSlug`). The LMS returns **one** tab entry for a destination learning splits into two pages, and that entry should be highlighted on both, so neither fixed value works and the fetch stamps whichever page asked. React Query keys on *what was requested*, while this field encodes *who was asking* — hence two cache entries for one response, and readers that would have to name a root.

Dropping `rootSlug` alone would leave the mapping as `slug: tab.tabId` — `camelCaseObject` has already produced `tabId`, and the endpoint sends only `tab_id`/`title`/`url`, so nothing is filtered either. The remaining effect would be to re-spell one field, which is the only reason tab identity is a concern of `src/course-home/data/` at all. The request is `GET /api/course_home/course_metadata/{courseId}?browser_timezone=…`: no tab is an input, and the response is identical for every page. Tab identity is *content* in the response body, never a parameter, so the mapping goes with the parameter.

**Tasks**
- `course-home/data/api.js`: drop the `rootSlug` parameter from `getCourseHomeCourseMetadata` and `normalizeCourseHomeCourseMetadata`, and delete the `tabs` mapping outright. `...data` carries tabs through as the LMS sent them.
- `course-home/data/queryKeys.ts`: `metadata(courseId)` loses the parameter; drop the argument at the ten call sites (`OutlineTab`, `DatesTab`, `ProgressTab`, `LiveTab`, `DiscussionTab`, `CourseAccessErrorPage`, `CoursewareContainer`, `redirects`, `useIsCourseLoaded`, `CourseExit`).
- Add `src/course-tabs/utils.ts` as the only module that names an LMS tab id: a named accessor per destination returning what callers use (`getCourseOutlineUrl`, `getDatesTabUrl`, `getProgressTabUrl`, `hasDiscussionTab`) over a module-private `getTab`, plus the `isActiveTab` predicate and a `getActiveTabTitle` wrapper carrying the two-pages rule:

```ts
// The LMS returns a single tab for both the course-home page and courseware,
// so it is the active tab on either of them.
export const isActiveTab = (slug: string, activeTabSlug?: string) => (
  slug === 'courseware'
    ? activeTabSlug === 'outline' || activeTabSlug === 'courseware'
    : slug === activeTabSlug
);
```

- Move the eight tab lookups onto those accessors (`CourseNonPassing`, `HiddenAfterDue`, `CertificateStatusAlert`, `CourseInProgress`, `RelatedLinks`, `DetailedGrades`, `widgets/discussions/widgetConfig.js`, `LoadedTabPage`), so no component outside `course-tabs/` names a tab field, holds a tab object, or writes a tab-id literal.
- Collapse the four duplicate `{ title, slug, url }` type declarations onto one `TabMetadata`. Required, not tidying — left alone they declare a field that no longer exists.
- Make `shared/data/__factories__/tab.factory.js` emit `tab_id`/`title`/`url` only. This is the safety net, not cleanup — see the findings below.
- Delete the two rootSlug labelling tests in `course-home/data/apiHooks.test.tsx`, invert the third, add unit tests for the accessors and for `discussionsPrefetch`, and update the pact test's expectation on the normalized field.

**Verify:** the Course tab is highlighted on **both** `/course/:courseId/home` and the courseware route, and on neither when another tab is active; the Helmet title still carries the tab name on both; the links built from a tab's url still resolve (the progress tab's "Course outline" and "Dates", the ungraded-assignments note, a hidden-after-due subsection's "progress page"); the discussions sidebar still prefetches topics; LMS-hosted tabs such as Teams still navigate away correctly.

**No request-count change.** Counting `course_metadata` hits across an outline → courseware crossing gives exactly 3 either way, five runs each. `useCourseHomeMeta` sets no `staleTime`, so a shared key still refetches for every observer that remounts on a stale entry — collapsing two keys into one turns a cache miss into a stale hit rather than removing a fetch. Reducing that count is #2098.

## Plan

> [!NOTE]

- **Ten call sites, not nine.** `useIsCourseLoaded` in `courseware/data/apiHooks.ts` passes `'courseware'` too, alongside `CoursewareContainer`, `CourseExit` and `redirects.ts`; the five course-home tabs and `CourseAccessErrorPage` pass `'outline'`.
- **`rootSlug` never reaches the LMS.** The request is `GET /api/course_home/course_metadata/{courseId}?browser_timezone=…` — `courseId` in the path, timezone in the query, nothing about a tab. The `tabs` array is content in the response body, the same for every page, so a cache key that varies by page is keying on our own interpretation rather than on the request.
- **The remaining mapping is a pure re-spelling.** After `rootSlug` goes, `tabs.map(tab => ({ slug: tab.tabId, title, url }))` renames one field and copies two. Per the pact contract the endpoint sends only `tab_id`, `title` and `url`, so it filters nothing either. Deleting it is what leaves the data layer with no opinion about tabs.
- **TypeScript cannot catch a missed reader.** `useModel('courseHomeMeta', courseId)` returns `any` — `generic/model-store/hooks.js` is JavaScript — and `useCourseHomeMeta` is typed as only `{ courseAccess }`. So the fixtures are the safety net. `shared/data/__factories__/tab.factory.js` currently emits `slug` **and** a derived `tab_id`, plus `priority`/`type` that this endpoint never sends; today's mapping is what strips them. Left alone, a consumer still reading `slug` would keep passing tests and break only in production.
- **`widgets/discussions/widgetConfig.js` is the highest-risk line in the change.** `discussionsPrefetch` has no test — every test calls `prefetchDiscussionTopics` directly, bypassing the gate, and the only test that renders the provider supplies `{ tabs: [] }`. Combined with the missing type coverage, getting it wrong is invisible: CI green, discussion-topic prefetching silently dead. It also legitimately receives `course?.tabs` as `undefined`, so the accessors tolerate that.
- **A named accessor per tab, not a generic `getTab(tabs, id)`.** Plain-versus-quirky lookups is a fact about the implementation, not about the caller, and `'dates'`/`'progress'`/`'discussion'` are the LMS's vocabulary appearing as bare literals in six components — the same kind of external knowledge these accessors were extracted for.
- **Three names disagree about the Course tab, so `utils.ts` holds that rather than the call sites.** The LMS's id is `courseware` and its title is **Course** (`CoursewareTab.title = gettext_noop('Course')`); this MFE splits it across the outline route `/course/:courseId/home` and the content routes `/course/:courseId/:sequenceId/:unitId`; and the tab's own `url` points at the outline, which the UI calls **Course outline** (the courseware sidebar, the progress tab's related links, its detailed-grades alert). Neither `courseware` nor `outline` appears in any URL — they are in-app labels carried by `activeTabSlug`. So the constant keeps the LMS's name, a comment states the disagreement once, and the export is `getCourseOutlineUrl(tabs)`, named for where it goes; both callers only ever wanted `.url`. `get` rather than `find` follows this repo's convention (`getResponseStatus`, `getErrorDetail`, `getSidebarId`), which uses `get` even for helpers returning `undefined`.
- **`tabs` is a server-owned nav menu, not a list of this app's pages** — which is why the payload carries URLs, and why only four accessors are needed. `openedx.course_tab` is an entry-point namespace and openedx-platform ships nineteen types; four are pages this MFE routes, and the rest are LMS-hosted (`wiki`, `instructor`, `teams`, `edxnotes`, the textbook tabs) or carry author-chosen URLs (`external_link` and `external_discussion` are `LinkTab` subclasses, `static_tab` has an author-chosen slug, the `lti_*` tabs are launch URLs). The MFE renders `title` → `url` for whatever arrives and inspects ids only for destinations it deep-links to from elsewhere — so the accessors return a URL or a boolean, never a tab, and `getActiveTabTitle` has to tolerate ids it has never heard of.
- **The rule goes in a predicate, not back into the data.** Rewriting the slug at display time in `LoadedTabPage` was considered and rejected: that component would then have to compute "which root am I on", so the `rootSlug` concept survives and every future tab-list reader has to go through one funnel.
- **Plugins are unaffected.** `CourseTabsNavigationSlot` passes no `pluginProps` and `CourseTabLinksSlot` passes only `activeTabSlug`, so the tabs array never reaches plugin code. `CourseTabLink`'s own `slug` prop is left alone — the slot README documents operators passing a custom tab name there, an identity whose domain is wider than LMS tab ids.
- **No re-render change.** `camelCaseObject` already deep-clones per fetch and `useModel` compares `tabs` by reference, so that reference was new on every fetch with or without the mapping.
- **Nobody designed this split — it is a 2020 bug fix that outlived its reason.** The outline became its own page in [#80](https://github.com/openedx/frontend-app-learning/pull/80) (2020-06-15), routed as `tab="outline"`, a *page* name. Ten days later [#93](https://github.com/openedx/frontend-app-learning/pull/93) created `course-home/data/api.js` emitting plain `slug: tab.tabId`, and from that point the Course tab was silently **not highlighted** on the outline page. Six months on, [#281](https://github.com/openedx/frontend-app-learning/pull/281) — a 31-file feature PR (+481/−203) building the anonymous/un-enrolled landing page, which among much else removed `requireAuthenticatedUser` — hit the dead highlight and fixed it in the data in two lines; its review thread says so outright: ["The "Course" tab wasn't active while in course home because our course metadata calls it "courseware""](https://github.com/openedx/frontend-app-learning/pull/281#discussion_r537750859). [#861](https://github.com/openedx/frontend-app-learning/pull/861) then turned that hardcode into the `rootSlug` parameter when it unified the tab source. The data fix was the cheap option, and incidental to that PR's subject, because `fetchTab` was a *per-page* thunk: fetch time and render time were the same moment, so "who is asking" was ambient, while fixing the comparison meant touching the nav component.
- **The duplicate request predates React Query, and `rootSlug` never caused it.** In 2020 the endpoint had one caller — courseware took its tabs from `/api/courseware/course/`, which is why only the outline's highlight was broken. `fetchCourse` began calling it too in `3fe5bb17` (AA-1018, 2022-01), and #861 then made those tabs the only source. Through all of that the thunks had no cache, so both callers refetched on every navigation regardless; `rootSlug` varied the shape of requests that were already happening. [#2023](https://github.com/openedx/frontend-app-learning/pull/2023) is the first point where a cache could have served both from one entry and `rootSlug` in the key prevented it — the first moment the 2020 decision had a price, and the first at which fixing it properly is cheaper than not.

<details>
<summary>Full plan</summary>

# Plan: #2084 — Take tab identity out of the course-home metadata query

Layer B1 of the model-store dissolution (#1977), stacked on layer A (#2093).
No reader converts here: this layer removes `rootSlug` so that B2 (#2085) and B3
(#2086) can convert 53 `courseHomeMeta` readers without any of them naming a
cache variant.

## Context

`courseHomeQueryKeys.metadata(courseId, rootSlug)` caches one endpoint twice —
under `'outline'` for the five course-home tabs and `CourseAccessErrorPage`, and
under `'courseware'` for `CoursewareContainer`, `CourseExit`, `redirects` and
`useIsCourseLoaded`. `rootSlug` never reaches the request. It changes exactly one
field, in `normalizeCourseHomeCourseMetadata` (`course-home/data/api.js:20`):

```js
// The API uses "courseware" as a slug for both courseware and the outline tab.
slug: tab.tabId === 'courseware' ? rootSlug : tab.tabId,
```

### Why that field exists

The LMS returns **one** tab entry, `tabId: 'courseware'`, for a destination
learning splits into two pages. Both pages render the same nav, and the nav
decides highlighting by string equality (`CourseTabLink.tsx:16`,
`active: slug === activeTabSlug`). Neither fixed value works — `'outline'` fails
to highlight on courseware, `'courseware'` fails on the course-home page — so
something has to know that this one tab is "the current page" in two different
situations. That knowledge can live in the data or in the comparison.

Nobody designed the split; a page name and a tab id collided and the nav broke:

| when | PR | what happened |
|---|---|---|
| 2020-06-15 | [#80](https://github.com/openedx/frontend-app-learning/pull/80) | The outline becomes its own page, routed as `tab="outline"` — a *page* name, not a tab id. |
| 2020-06-25 | [#93](https://github.com/openedx/frontend-app-learning/pull/93) | `course-home/data/api.js` is created emitting plain `slug: tab.tabId`. From here the Course tab is **not highlighted** on the outline page. |
| 2020-12-16 | [#281](https://github.com/openedx/frontend-app-learning/pull/281) | A 31-file feature PR building the anonymous/un-enrolled landing page notices the dead highlight in passing and fixes it in two lines. Its review thread states the bug outright — ["The "Course" tab wasn't active while in course home because our course metadata calls it "courseware""](https://github.com/openedx/frontend-app-learning/pull/281#discussion_r537750859). |
| 2022-03-10 | [#861](https://github.com/openedx/frontend-app-learning/pull/861) | Unifying the tab source means one normalizer serves both pages, so the hardcoded `'outline'` becomes the `rootSlug` parameter, passed `'outline'` by `fetchTab` and `'courseware'` by `fetchCourse`. |
| 2026-08-31 | [#2023](https://github.com/openedx/frontend-app-learning/pull/2023) | Converting the fetch to React Query means the queryFn's output varies by `rootSlug`, so it must go in the key — two cache entries for one response. |

#281's two-line fix was the cheap one, and incidental to that PR's actual subject.
`fetchTab` was a *per-page* thunk, so fetch time and render time were the same
moment and "who is asking" was ambient and free, while fixing the comparison meant
touching the nav component. Every consumer stayed a plain equality check.

It stops fitting because React Query keys on **what was requested** while this
field encodes **who was asking**.

### Why the whole mapping goes, not just the parameter

Removing `rootSlug` alone leaves the normalizer with this:

```js
tabs: data.tabs.map(tab => ({ slug: tab.tabId, title: tab.title, url: tab.url })),
```

`camelCaseObject` has already produced `tabId`, and per the pact contract the
endpoint sends only `tab_id`, `title` and `url` — so nothing is filtered either.
The map's entire remaining effect would be to re-spell one field, which is the
only reason tab identity is a concern of `src/course-home/data/` at all.

The request is `GET /api/course_home/course_metadata/{courseId}?browser_timezone=…`.
No tab is an input, and the response is identical for every page. Tab identity is
*content* in the response body, like `title` or `org` — never a parameter — so the
data layer should have no opinion about it. Leaving the map behind would also leave
the layer's central claim half-stated: the id is the id, and the page asking is not
part of it.

## The change

1. **`course-home/data/api.js`** — `getCourseHomeCourseMetadata(courseId)` and
   `normalizeCourseHomeCourseMetadata(metadata)` drop the parameter, and the
   `tabs` mapping is deleted outright. The normalizer keeps only `isMasquerading`;
   `...data` carries tabs through as the LMS sent them (`tabId`, `title`, `url`).
2. **`course-home/data/queryKeys.ts`** — `metadata: (courseId: string)`.
3. **`course-home/data/apiHooks.ts`** — `useCourseHomeMeta(courseId)`.
4. **Ten call sites** drop their argument: `DatesTab`, `DiscussionTab`,
   `LiveTab`, `OutlineTab`, `ProgressTab`, `CourseAccessErrorPage`,
   `CoursewareContainer`, `CourseExit`, `redirects.ts`, and `useIsCourseLoaded`
   in `courseware/data/apiHooks.ts`.
5. **New `src/course-tabs/utils.ts`** becomes the only module that names an LMS
   tab id. `getTab` stays module-private so no call site passes an id:

   ```ts
   export interface TabMetadata {
     tabId: string;
     title: string;
     url: string;
   }

   const getTab = (tabs: TabMetadata[] | undefined, tabId: string) => tabs?.find(tab => tab.tabId === tabId);

   export const getDatesTabUrl = (tabs?: TabMetadata[]) => getTab(tabs, 'dates')?.url;
   export const getProgressTabUrl = (tabs?: TabMetadata[]) => getTab(tabs, 'progress')?.url;
   export const hasDiscussionTab = (tabs?: TabMetadata[]) => !!getTab(tabs, 'discussion');

   export const getCourseOutlineUrl = (tabs?: TabMetadata[]) => getTab(tabs, 'courseware')?.url;

   export const isActiveTab = (slug: string, activeTabSlug?: string) => {
     if (activeTabSlug === slug) {
       return true;
     }

     // Course content renders at /course/:courseId/:sequenceId/:unitId and carries the slug `courseware`;
     // the course outline renders at /course/:courseId/home and carries `outline`. Users see one "Course" tab.
     if (activeTabSlug === 'outline' && slug === 'courseware') {
       return true;
     }

     return false;
   };

   export const getActiveTabTitle = (tabs?: TabMetadata[], activeTabSlug?: string) => (
     tabs?.find(tab => isActiveTab(tab.tabId, activeTabSlug))?.title
   );
   ```

   A helper per destination rather than a generic `getTab(tabs, id)`:
   plain-versus-quirky is a fact about the implementation, not about the caller,
   and the tab ids are the LMS's vocabulary appearing as bare literals in six
   components. Each returns what callers actually use — a URL, or a boolean —
   because every tab-object caller immediately wrote `tab && tab.url` and
   `widgetConfig.js` only tested truthiness. Even `getActiveTabTitle` returns a
   string, so nothing outside `utils.ts` handles a tab object at all.

   Only four destinations need an accessor, and that is not a coincidence. `tabs`
   is a server-owned nav menu, not a list of this app's pages: `openedx.course_tab`
   is an entry-point namespace and openedx-platform ships nineteen types, most of
   them LMS-hosted (`wiki`, `instructor`, `teams`, `edxnotes`, the textbook tabs)
   or carrying author-chosen URLs (`external_link`, `static_tab`, the `lti_*`
   tabs). The MFE renders `title` → `url` for whatever arrives and inspects ids
   only for the destinations it deep-links to from elsewhere.

   The Course tab gets a URL accessor instead, because three names disagree about
   it and no caller should have to hold that. The LMS's id is `courseware` and its
   title is "Course"; this MFE splits it across the outline route (`/home`) and the
   content routes; and the tab's own `url` points at the outline, which the UI calls
   **Course outline** (the courseware sidebar, the progress tab's related links, its
   detailed-grades alert). Neither `courseware` nor `outline` appears in any URL —
   they are in-app labels carried by `activeTabSlug`. So the constant keeps the LMS's
   name, the comment states the disagreement once, and the export is
   `getCourseOutlineUrl(tabs)` — named for where it goes, which is all either caller
   wanted. `get` rather than `find` follows the repo's own convention
   (`getResponseStatus`, `getErrorDetail`, `getSidebarId`), including for helpers
   that may return `undefined`.

   `isActiveTab` keeps a `slug` parameter because `CourseTabLink` passes it a value
   that may be a plugin's custom tab name rather than an LMS tab id. The optional
   `tabs` parameter matters too: `widgetConfig.js` passes `course?.tabs`, which is
   `undefined` before the model loads.

6. **Seven consumers** move to the helpers — one line each:
   `CourseNonPassing.jsx:28`, `HiddenAfterDue.jsx:15` and
   `CertificateStatusAlert.jsx:118` (`getProgressTabUrl`), `CourseInProgress.jsx:28`
   and `RelatedLinks.jsx:31` (`getDatesTabUrl`), `widgetConfig.js:10`
   (`hasDiscussionTab`), and `LoadedTabPage.tsx:47` (`getActiveTabTitle`).
   `DetailedGrades.jsx:42` and `RelatedLinks.jsx:29` take `getCourseOutlineUrl`,
   which collapses their two-line `tab && tab.url` dance into one.

   `CertificateStatusAlert.jsx:203-207` also documents the tab shape in PropTypes
   as `tab_id`, which is wrong both before and after — the camelCased field is
   `tabId`.

7. **One tab type.** `CourseTabsNavigation.tsx:11-15`,
   `CourseTabLinksList.tsx:5-9` and `CourseTabLinksSlot/index.tsx:5-9` each
   re-declare `{ title, slug, url }`; all three import `TabMetadata` instead. This is
   mandatory rather than tidying — left alone they declare a field that no longer
   exists. `TabMetadata`'s `title`/`url` are required, since `CourseTabLink`
   requires `url: string` and the optional form would not compile once the type is
   actually used; nothing depended on the optional form.

   `CourseTabLinksList` is the one seam: it destructures `tabId` and passes
   `slug={tabId}` to `CourseTabLink`, whose prop is a documented plugin contract
   with a wider domain than LMS tab ids.

8. **Faithful fixtures.** `shared/data/__factories__/tab.factory.js` emits `slug`
   *and* a derived `tab_id`, plus `priority`/`type` that this endpoint never sends;
   the mapping is what hides them today. It emits `tab_id`, `title` and `url` only,
   and the six `Factory.build('tab', …)` calls in
   `courseHomeMetadata.factory.js:38-123` pass `tab_id`.

   This is the change's safety net, not cleanup: `useModel('courseHomeMeta')`
   returns `any` (`generic/model-store/hooks.js` is JavaScript) and
   `useCourseHomeMeta` is typed as only `{ courseAccess }`, so **`tsc` cannot catch
   a missed reader**. Left alone, a consumer that still reads `slug` would keep
   passing tests and break only in production. Rosie silently accepts a stray
   `slug` override *and* falls back to its default `tab_id`, so all six convert in
   one edit with the built array's ids asserted once.

## Tests

- **`course-home/data/apiHooks.test.tsx`** has three tests pinning what this layer
  removes, added by #2023 when the convention was introduced: *labels the shared
  tab "courseware"*, *labels it "outline"*, and *keys the query by rootSlug so the
  two contexts do not share a cache entry*. The first two are **deleted** — they
  assert a data-layer opinion about tab ids, which is exactly what this layer
  removes, and only compile through a `data as { tabs: … }` cast. The third
  **inverts**: both contexts now share one cache entry, which is the improvement
  this layer makes, so it is worth pinning in that direction.
- **New unit tests for the helpers** (`course-tabs/utils.test.ts`): the courseware
  tab is active under both `'outline'` and `'courseware'`; an ordinary tab matches
  only its own id; each named helper finds its entry; all of them tolerate absent
  tabs.
- **New unit test for `discussionsPrefetch`** (`widgets/discussions/`). Nothing
  exercises it today — every test calls `prefetchDiscussionTopics` directly, and
  the only test that renders the provider supplies `{ tabs: [] }`. With no type
  coverage on that path either, a mistake there is invisible: CI green,
  discussion-topic prefetching silently dead.
- **`course-tabs/CourseTabsNavigation.test.jsx:65`** passes
  `activeTabSlug: tabs[0].slug`, which becomes `tabs[0].tabId`. It still derives
  the page from a tab, but the fixture's ids (`test1`, `test2`) are neither route
  names nor LMS tab ids, so nothing real is conflated; a literal `'test1'` would
  behave identically and can land whenever the test is next touched.
- **`pact-tests/lmsPact.test.jsx:126-131`** expects the normalized `slug: 'outline'`
  and becomes `tabId: 'courseware'`. The API-side `tab_id: 'courseware'` at line 85
  is the contract with the LMS and does not move.

## Behaviour changes

- **One cache entry per course instead of two**, verified by the inverted test in
  `apiHooks.test.tsx`: both contexts return the same object and one controlled
  render makes one request.

  **This does not reduce requests in the browser, and the layer should not claim
  it does.** Measured in tutor dev on the outline → courseware crossing (clear the
  network log once the outline settles, click a sequence title, wait for idle,
  count `course_metadata`): **exactly 3, five runs each, on this layer and on
  #2093** — no variance either side. An earlier draft claimed "two requests become
  one"; that was reasoning from the key alone, and it is wrong.

  The reason is `staleTime`. `useCourseHomeMeta` sets none, so the entry is stale
  the moment it lands and `refetchOnMount` fires for each observer that mounts on
  it — the courseware side mounts three (`CoursewareContainer`, `useIsCourseLoaded`,
  `redirects`) across a route that remounts as the redirect resolves. Collapsing two
  keys into one turns a cache *miss* into a stale *hit*; it does not remove a fetch.
  Whether that changes what blocks `useIsCourseLoaded` was not measured.

  The tab bar is not a client-side path at all: `CourseTabLink` renders a plain
  `<a href>`, so every tab click is a full page load that discards the cache. The
  only crossings that reuse it are a sequence title on the outline (`SequenceTitle`
  renders a router `Link` into courseware) and the home breadcrumb in courseware
  (`CourseBreadcrumbs` links to `/home`).

  Reducing the count is a `staleTime` question rather than a key question, and this
  layer makes it tractable by leaving one key to tune instead of two.
- **Tabs carry the ids the LMS sent**, wherever the payload is read, including the
  `courseHomeMeta` model the bridge still writes. The bridge itself is a
  passthrough (`model: { id: courseId, ...data }`) and needs no change.
- **No re-render change.** `camelCaseObject` already deep-clones per fetch and
  `useModel` compares `tabs` by reference, so that reference was new on every fetch
  with or without the mapping. Deleting it removes one array allocation.

Plugins are unaffected: `CourseTabsNavigationSlot` passes no `pluginProps`, and
`CourseTabLinksSlot` passes only `activeTabSlug` — the tabs array never reaches
plugin code. The slot README's example renders its own
`<CourseTabLink slug="custom-link" activeTabSlug={…} />`, which is untouched.

## Not in scope

- **`CourseTabLink`'s `slug` prop and the slot README** — that identity is a
  route comparison whose domain includes plugin-supplied names, not an LMS tab id.
- **`generic/tabs/Tabs.jsx`** drops `className`/`style` when cloning
  `CourseTabLink`, so tab overflow does not work. Pre-existing since the slot was
  added; its own issue.
- **The stale `tabs` block in the generated `src/pacts/frontend-app-learning-lms.json`**
  under the courseware interaction, unasserted since #861.

## Verification

- Manual, on both root pages: the Course tab is highlighted on
  `/course/:courseId/home` **and** on the courseware route; the Helmet title
  still carries the tab name on both; the progress tab's "Course Outline" and
  "Dates" links resolve; the certificate not-passing alert's grades button
  resolves; a past-due gated subsection's "progress page" link resolves; the
  course-exit dates and grades buttons resolve; the discussions sidebar still
  prefetches topics; and the Network tab shows **one** `course_metadata` request
  when moving between courseware and a tab, where there were two.

## Stack

One layer on top of #2093, closing #2084. B2 (#2085) and B3 (#2086) follow.

</details>

brian-smith-tcril:
## Summary

`/api/course_home/course_metadata/` sends **one** tab, `courseware`, for a destination this MFE renders as two separate pages: the course outline at `/course/:courseId/home`, and the courseware content routes. The nav decides which tab to highlight by comparing a tab's id against the page you are on, so that single entry has to count as "current" in two different situations.

The data layer used to settle that at fetch time by rewriting the tab's id. `getCourseHomeCourseMetadata` took a `rootSlug` — `'outline'` or `'courseware'`, whichever page was asking — and stamped the shared tab with it, so the nav's plain `===` would match. That made the response depend on its caller rather than on the request, so the query key had to carry `rootSlug` too, and one endpoint ended up cached twice per course. The arrangement dates to a 2020 bug fix rather than a design decision; decision 1 below has the chain, with links.

This removes `rootSlug` and the mapping with it. Tabs carry the id the endpoint sent, `src/course-home/data/` has no opinion about tabs at all, and the two-pages rule moves into `isActiveTab` in `src/course-tabs/utils.ts` — which becomes the only module that names a tab id. **No user-facing change.**

Part of the Redux → React Query migration (#1946, Stage 1), layer B1 of the model-store dissolution (#1977). Closes #2084. Stacked above #2093.

## What changed

- **`course-home/data/api.js`** — `getCourseHomeCourseMetadata(courseId)` and `normalizeCourseHomeCourseMetadata(metadata)` drop the parameter, and the `tabs` mapping is deleted outright rather than reduced to `slug: tab.tabId`. `camelCaseObject` already produces `tabId`, and per the pact contract the endpoint sends only `tab_id`/`title`/`url`, so what remained would have re-spelled one field and filtered nothing. The normalizer now computes only `isMasquerading`.
- **The key and hook** lose the parameter, along with the ten call sites of the hook and one more place that builds the key directly — `useIFrameBehavior`, which invalidates it and so does not appear in a search for the hook.
- **`course-tabs/utils.ts`** exports an accessor per destination returning what callers use — `getCourseOutlineUrl`, `getDatesTabUrl`, `getProgressTabUrl`, `hasDiscussionTab` — over a module-private `getTab`, alongside `isActiveTab` and `getActiveTabTitle`. Every one of the five tab-object callers immediately wrote `const tab = …; const url = tab && tab.url;`, and `widgetConfig.js` only tested truthiness, so no consumer wanted a tab. Eight lookups move onto them; no component outside `course-tabs/` now names a tab field or holds a tab.
- **`CourseTabLink` is untouched.** Its `slug` prop is a documented plugin contract whose domain is wider than LMS tab ids — the slot README has operators passing a name they chose, and `activeTabSlug` is a page name, not a tab id. `CourseTabLinksList` is the single seam.
- **One `TabMetadata` type.** `CourseTabsNavigation`, `CourseTabLinksList` and `CourseTabLinksSlot` each re-declared `{ title, slug, url }`; all three import it now. Mandatory rather than tidying — left alone they declare a field that no longer exists.
- **The tab factory is the safety net.** `useModel('courseHomeMeta')` returns `any` and `useCourseHomeMeta` is typed to only `{ courseAccess }`, so `tsc` cannot catch a missed reader. `tab.factory.js` emitted `slug` *and* a derived `tab_id`, plus `priority`/`type` that this endpoint has never sent — fields the mapping was silently stripping. It now matches the pact, so a consumer left unconverted fails in tests rather than only in production.
- **`discussionsPrefetch` gains the test it never had**, and loses an `edxProvider` local that named a provider check happening one layer down.

## Testing

`npm run types`, `npm run lint` and the full suite (113 suites, 1137 passed, 3 skipped) are green.

Manually verified on tutor dev. The nav highlight is the core of the change and was checked on both root pages, on every other tab, and across the two client-side crossings; the Helmet title, every link built from a tab's url, the discussions prefetch gate, an LMS-hosted tab (Teams), and a slot-inserted `CourseTabLink` all behave as before. Four checks were not run — three expensive course states that are repeat call sites of accessors already confirmed, and console noise — each listed with its reason.

**No request-count change, despite the collapsed key.** Counting `course_metadata` hits across an outline → courseware crossing gives exactly 3 either way, five runs each side with no variance. `useCourseHomeMeta` sets no `staleTime`, so a shared key still refetches for every observer that mounts on a stale entry — collapsing two keys into one turns a cache miss into a stale hit rather than removing a fetch. An earlier draft of this PR claimed the crossing saved a request; that was reasoning from the key alone and the measurement disproved it. Reducing the count is #2098, which this layer makes tractable by leaving one key to tune instead of two.

## Decisions

<details>

# Decisions — take tab identity out of the course-home metadata query (#2084)

Layer B1 of the model-store dissolution (#1977). Some entries were settled in the
plan on issue #2084; others landed with the code, and a few correct things the
plan had wrong — notably that no factory change was needed.

1. **The rule moves into the comparison, not back into the data.** `tabs[].slug`
   was never a property of the tab: the LMS returns **one** entry,
   `tabId: 'courseware'`, for a destination this MFE splits into two pages, and
   the normalizer stamped it with whichever page was asking so that the nav's
   `slug === activeTabSlug` check would highlight it. React Query keys on **what
   was requested** while the field encodes **who was asking**, which is what
   produced two cache entries for one response, so the knowledge moves to where
   the question is actually asked.

   **Nobody designed this split.** A page name and a tab id collided, the nav
   silently broke, and the cheapest fix available at the time was to bend the
   payload:

   | when | PR | what happened |
   |---|---|---|
   | 2020-06-15 | [#80](https://github.com/openedx/frontend-app-learning/pull/80) | The outline becomes its own page, routed as `tab="outline"` — a *page* name, not a tab id. |
   | 2020-06-25 | [#93](https://github.com/openedx/frontend-app-learning/pull/93) | `course-home/data/api.js` is created emitting plain `slug: tab.tabId`. From here the Course tab is **not highlighted** on the outline page: it carries `'courseware'`, the page passes `'outline'`, and the nav compares with `===`. |
   | 2020-12-16 | [#281](https://github.com/openedx/frontend-app-learning/pull/281) | Six months later, a 31-file feature PR (+481/−203) building the anonymous/un-enrolled landing page — public-course support, the private-course alert, the anonymous user menu, `requireAuthenticatedUser` removed — notices the dead highlight in passing and fixes it in two lines: `slug: tab.tabId === 'courseware' ? 'outline' : tab.tabId`. Its review thread states the bug outright — ["The "Course" tab wasn't active while in course home because our course metadata calls it "courseware""](https://github.com/openedx/frontend-app-learning/pull/281#discussion_r537750859). |
   | 2022-03-10 | [#861](https://github.com/openedx/frontend-app-learning/pull/861) | Unifying the tab source means one normalizer serves both pages, so the hardcoded `'outline'` becomes the `rootSlug` parameter. |
   | 2026-08-31 | [#2023](https://github.com/openedx/frontend-app-learning/pull/2023) | Converting the fetch to React Query means the queryFn's output varies by `rootSlug`, so it must go in the key — two cache entries for one response. |

   Each step was locally right. #281 chose the two-line fix in a function that
   already knew which page was asking, because `fetchTab` was a *per-page* thunk:
   fetch time and render time were the same moment, so "who is asking" was ambient
   and free, while fixing the comparison meant touching the nav component. It was
   also an incidental correction inside a large feature PR, which is a good part of
   why the approach was never weighed.

   **It cost nothing until caching arrived**, which is why five years passed
   without anyone revisiting it. In December 2020 there was exactly one caller of

   | when | callers of `course_metadata` | why the rewrite was free |
   |---|---|---|
   | 2020-12 ([#281](https://github.com/openedx/frontend-app-learning/pull/281)) | course-home pages only | Courseware took its tabs from `/api/courseware/course/`, which sent `slug: 'courseware'` natively — which is why only the outline's highlight was broken. The rewrite changed the sole consumer of the endpoint. |
   | 2022-01 (`3fe5bb17`, AA-1018 "api refactor") | course-home **and** `fetchCourse` | The courseware thunk adds `getCourseHomeCourseMetadata(courseId)` to its `Promise.allSettled`, so there are two callers. Harmless: courseware's nav still reads the courseware endpoint's tabs, so the hardcoded `'outline'` is never compared there. |
   | 2022-03 ([#861](https://github.com/openedx/frontend-app-learning/pull/861)) | both | The courseware endpoint stops sending tabs, so courseware's nav must use these — and the hardcode breaks *its* highlight, which is what turns `'outline'` into the `rootSlug` parameter. Still free: thunks had no cache, so both callers refetched on every navigation regardless. |
   | 2026-08 ([#2023](https://github.com/openedx/frontend-app-learning/pull/2023)) | both, now through React Query | A cache exists that could serve both callers from one entry. `rootSlug` in the key is exactly what prevents it. **First moment the 2020 decision has a price.** |

   So `rootSlug` never caused an extra request; it varied the shape of requests
   that were already happening. React Query keys on the request, so a per-page
   value inside a per-course response has to be declared in the key — #2023 doing
   the honest thing, and thereby turning a six-year-old shape problem into a
   visible duplicate fetch. It is also the first point at which fixing it properly
   is cheaper than not.

   The name `slug` came from the other endpoint. Both metadata endpoints served
   `get_course_tab_list`, and each named the id differently: course-home sent
   `tab_id`, while `/api/courseware/course/` built its own list and called the
   same value `slug`:

   ```python
   # openedx-platform, CoursewareMeta.tabs, deleted by openedx/openedx-platform#30023
   for priority, tab in enumerate(get_course_tab_list(self.effective_user, self.overview)):
       tabs.append({
           'title': _(title),
           'slug': tab.tab_id,
           'priority': priority,
           'type': tab.type,
           'url': tab.link_func(self.overview, reverse),
       })
   ```

   So `slug` was never a separate concept — it was the courseware API's alias for
   `tab_id`, and `priority` was the `enumerate()` index of a list that already
   arrived in order. Once #861 moved the MFE onto the course-home endpoint, the
   normalizer's `slug: tab.tabId` was reproducing a rename the courseware API had
   been doing server-side, and kept doing it for four years after that API stopped.
   Deleting the mapping is therefore not choosing a new name for the field; it is
   dropping an inherited alias from an endpoint that no longer exists and using
   the one name the surviving endpoint has always used.

2. **Rejected: stamping the slug at display time.** `LoadedTabPage` could rewrite
   the courseware entry before rendering the nav, keeping the comparison a plain
   `===`. That works, but the component then has to compute "which root am I on",
   so the `rootSlug` concept survives, and every future reader of the tab list
   has to go through that one funnel.

3. **The whole tabs mapping goes, not just the parameter.** Dropping `rootSlug`
   alone leaves `tabs.map(tab => ({ slug: tab.tabId, title, url }))`.
   `camelCaseObject` has already produced `tabId`, and per the pact contract the
   endpoint sends only `tab_id`, `title` and `url`, so that map filters nothing
   either — its entire remaining effect is to re-spell one field, which is the
   only reason tab identity is a concern of `course-home/data/` at all. The
   request is `GET /api/course_home/course_metadata/{courseId}?browser_timezone=…`:
   no tab is an input, and the response is identical for every page, so tab
   identity is *content* in the body and never a parameter. `normalizeCourseHome
   CourseMetadata` now computes only `isMasquerading`.

4. **`tabs` is a server-owned nav menu, not a list of this app's pages.** Worth
   stating because it explains why the payload carries URLs at all, and why so few
   accessors are needed. `openedx.course_tab` is an entry-point namespace — any
   installed Django app can register a type — and openedx-platform ships nineteen:

   ```
   ccx  courseware  dates  discussion  edxnotes  external_discussion  external_link
   html_textbooks  instructor  lti_discussion  lti_live  lti_tab  pdf_textbooks
   progress  static_tab  syllabus  teams  textbooks  wiki
   ```

   Four of those are pages this MFE routes. The rest are either LMS-hosted (`wiki`,
   `instructor`, `teams`, `edxnotes`, `syllabus`, the textbook tabs) or carry URLs
   the course author chose — `external_link` and `external_discussion` are
   `LinkTab` subclasses, `static_tab` has an author-chosen slug, and the `lti_*`
   tabs are launch URLs. The MFE could not construct those even in principle, and
   for the four it does own, the LMS is still the authority on whether a tab points
   here or at a legacy page, since that is a server-side rollout decision. Hence
   `url` in the payload, generated by `request.build_absolute_uri(tab.link_func(…))`.

   So the nav's job is `title` → `url` for whatever arrives, and the MFE inspects
   ids only for destinations it deep-links to from elsewhere. That set is exactly
   four, which is why `utils.ts` has four accessors and needs no general lookup.
   It is also why `getActiveTabTitle` must tolerate ids it has never heard of.

5. **A named accessor per destination, returning what callers use.** `getTab`
   stays module-private; `course-tabs/utils.ts` exports `getCourseOutlineUrl`,
   `getDatesTabUrl`, `getProgressTabUrl` and `hasDiscussionTab`. Plain-versus-quirky
   lookups is a fact about the implementation, not about the caller, so splitting
   the call sites on it exposes a distinction no consumer has reason to care
   about. And `'dates'`/`'progress'`/`'discussion'` are the LMS's vocabulary
   appearing as bare literals in six components — the same kind of external
   knowledge these accessors were extracted for. No component outside
   `course-tabs/` now names a tab field or a tab id.

   **They return values, not tabs.** Every one of the five tab-object callers
   immediately wrote `const tab = getXTab(tabs); const url = tab && tab.url;`, and
   `widgetConfig.js` only tested truthiness — so no consumer wanted a tab at all.
   Returning the URL (or a boolean) collapses each of those pairs to one line and
   means a component never holds a tab it has to know the shape of — including
   `getActiveTabTitle`, which returns the Helmet string rather than the tab.
   From a component's point of view "the dates tab" is not a meaningful object;
   the link destination is.

   `get`, not `find`. This repo has no `find*` helpers; the convention outside the
   API modules is `get*` — `getResponseStatus`, `getErrorDetail`,
   `getEnabledWidgets`, `getSidebarId`, `getReadableProctoringStatus`,
   `getAccessDeniedRedirectUrl`. The first two return `| undefined` and are still
   called `get`, so "might not be there" is not a distinction this codebase draws
   in a name. `find` also names the mechanism rather than the result: these would
   read the same if they were ever backed by a lookup map instead of
   `Array.prototype.find`.

   **The Course tab gets a URL accessor, not a tab accessor.** Three names
   disagree about this one nav entry, and containing that in `utils.ts` is worth
   more than naming it consistently at call sites. The LMS's id is `courseware`
   (`CoursewareTab.type`) and its title is **Course** (`CoursewareTab.title =
   gettext_noop('Course')`); this MFE splits it across the outline route
   `/course/:courseId/home` and the content routes `/course/:courseId/:sequenceId/:unitId`;
   and the tab's own `url` points at the outline, which is what the UI calls
   **Course outline** — `courseware/course/sidebar/sidebars/course-outline/`,
   `progress-tab/related-links`, `progress-tab/grades`. Neither `courseware` nor
   `outline` appears in any URL; they are in-app labels for which of the two pages
   is showing, which is what `activeTabSlug` carries.

   So a comment states the disagreement once, immediately above the one export it
   concerns, and that export is `getCourseOutlineUrl(tabs)` — named for where it
   goes. The tab ids stay inline string literals rather than named constants: each
   appears once or twice, the comment already carries the meaning a constant name
   would have gestured at, and `getTab(tabs, 'dates')` says more than
   `getTab(tabs, DATES_TAB_ID)`. Both callers only ever reached for `.url`, so this also collapses
   their `const tab = …; const url = tab && tab.url;` pairs into one line and
   retires `overviewTab`, a local from `edba1600` (2022) that matched nothing: not
   the LMS's id, not the UI's label, not the lookup beside it. An earlier pass named
   the accessor `getCourseOutlineTab`, which was wrong in the other direction — it
   gave a *page* name to the *tab*.

6. **All eight lookups move.** `CourseNonPassing`, `HiddenAfterDue` and
   `CertificateStatusAlert` take `getProgressTabUrl`; `CourseInProgress` and
   `RelatedLinks` take `getDatesTabUrl`; `widgetConfig.js` takes `hasDiscussionTab`;
   `LoadedTabPage` takes `getActiveTabTitle`; and `DetailedGrades` and `RelatedLinks`
   take `getCourseOutlineUrl`. The accessors accept `undefined` because
   `widgetConfig.js` legitimately passes `course?.tabs` before the model has
   loaded.

7. **`isActiveTab` keeps a `slug` parameter and `CourseTabLink` is untouched.**
   It takes a string rather than a tab because `CourseTabLink` passes it a value
   that need not be an LMS tab id: the slot README documents operators rendering
   their own `<CourseTabLink slug="custom-link" …/>`, and `activeTabSlug` is a
   route name — `OutlineTab` passes `'outline'`, which no tab is called. That
   identity's domain is strictly wider than the LMS's tab ids, so it is not a
   renamed `tabId` and does not move with this change. `CourseTabLinksList`'s
   `slug={tabId}` is the single seam, left uncommented: at that line the value is
   always an LMS tab id, since plugin links are rendered by the plugin inside the
   slot and never pass through that map, so a comment there would describe a case
   that cannot occur.

   **Two active-tab exports for one rule, on purpose.** `isActiveTab` has a single
   external caller (`CourseTabLink`), which on its own would argue for inlining it
   there. It stays in `utils.ts` because `getActiveTabTitle` is the rule's second
   consumer: `CourseTabLink` needs it for a CSS class and `LoadedTabPage` for the
   Helmet title, and those are different modules. Inlining into the component would
   leave `getActiveTabTitle` either importing from a component or carrying its own copy
   of the ternary — the duplication the helper was extracted to remove.

   Dropping `getActiveTabTitle` and letting `LoadedTabPage` write
   `tabs?.find(tab => isActiveTab(tab.tabId, activeTabSlug))` was considered. It
   trades two exports for one, but puts a tab field back in a component, which is
   the property the accessors exist to hold. Neither shape is obviously right; this
   one keeps the rule stated once and the field named nowhere outside
   `course-tabs/`.

   Also worth recording: making `isActiveTab` take only a `slug` and read the
   active page from context was rejected. `activeTabSlug` is not derivable where it
   is used — it appears in no URL, and is a hardcoded literal in each of the seven
   page components — and it is threaded for more than the nav (`ProductTours`,
   `InstructorToolbar`, `getAccessDeniedRedirectUrl`, `TabPage`'s tour button), so
   a context would be a second mechanism beside the existing prop rather than a
   replacement. It is also the slot's only `pluginProps`, and the README has
   operators forwarding it into their own `CourseTabLink`. The version of that idea
   worth doing is deriving `activeTabSlug` from the route app-wide — the mapping to
   `DECODE_ROUTES` is 1:1 — which deletes the literals and the threading, and is
   its own change.

8. **One `TabMetadata` type.** `CourseTabsNavigation`, `CourseTabLinksList` and
   `CourseTabLinksSlot` each re-declared `{ title, slug, url }`; all three now
   import `TabMetadata`. This was mandatory rather than tidying — left alone they
   declare a field that no longer exists. `title`/`url` became required, since
   `CourseTabLink` requires `url: string` and the optional form would not compile
   once the type was actually used; nothing depended on it, as `TabMetadata` had no
   importers.

9. **The factory is the safety net, not cleanup.** `useModel('courseHomeMeta')`
   returns `any` — `generic/model-store/hooks.js` is JavaScript — and
   `useCourseHomeMeta` is typed as only `{ courseAccess }`, so **`tsc` cannot
   catch a missed reader**. That makes the fixtures the only thing standing
   between a missed consumer and a silent production break, and the fixtures were
   describing a payload that does not exist.

   `/api/course_home/course_metadata/` has sent exactly `tab_id`, `title` and
   `url` since the endpoint was created
   ([`8a74bbd5fb`](https://github.com/openedx/openedx-platform/commit/8a74bbd5fb46d5e6920931e4378e79ca8a986ac6),
   AA-150, 2020-05-21). `CourseTabSerializer` is the only declaration of that
   shape and has never been modified since. It has never had `priority`, `type`,
   or a tab `slug`. The fixture's extra fields therefore did not drift out of
   date — they were never right for this endpoint:

   | | added | by | accurate then? |
   |---|---|---|---|
   | `tab.factory.js` created under `courseware/`, with `title`/`priority`/`slug`/`type`/`url` | 2020-07 | [#95](https://github.com/openedx/frontend-app-learning/pull/95) | **yes** — the courseware API really sent those |
   | moved to `shared/`, gains `.attr('tab_id', ['slug'], slug => slug)` so one factory serves both endpoints | 2021-03 | [#400](https://github.com/openedx/frontend-app-learning/pull/400) | it now emits the **union**, so correct for neither alone |
   | `tabs` added to `courseHomeMetadata.factory.js`, each build passing `priority`/`slug`/`type` | 2022-03-10 | [#861](https://github.com/openedx/frontend-app-learning/pull/861) | **no** — course-home never sent them |
   | courseware API's `tabs` deleted outright | 2022-03-11 | [openedx-platform#30023](https://github.com/openedx/openedx-platform/pull/30023) | the shape now exists nowhere |
   | remaining duplicate courseware fields deleted | 2022-03-18 | [openedx-platform#30079](https://github.com/openedx/openedx-platform/pull/30079) | |

   #861 and #30023 are the same ticket (AA-1207) a day apart: the MFE stopped
   reading tabs from the courseware API, then the platform removed them. Nothing
   was relocated to course-home — the fields were dropped. So #861 wrote a
   course-home fixture describing the courseware payload in the same commit that
   made that payload irrelevant. The mechanism looks mundane: it reused the
   existing courseware-shaped `tab` factory and passed its usual four-field
   literal, of which only `slug` was load-bearing, as the input `tab_id` derives
   from.

   Because `slug` was declared with `.attrs()` rather than `.option()`, it also
   leaked into the built object — so fixtures carried the very field the deleted
   mapping used to synthesise. A consumer left unconverted would have found
   `slug` in tests and `undefined` in the browser.

   The factory now emits `tab_id`, `title` and `url`, matching both the serializer
   and the pact, and the six `Factory.build('tab', …)` calls in
   `courseHomeMetadata.factory.js` moved with it in one edit — rosie silently
   accepts a stray `slug` override *and* falls back to its default `tab_id`, so a
   partial conversion would have produced six tabs all claiming `courseware`
   without throwing.

10. **Two of the three `rootSlug` tests are deleted, not rewritten.** #2023 added
   them when the convention was introduced: the tab is labelled `courseware`, the
   tab is labelled `outline`, and the query is keyed by rootSlug "so the two
   contexts do not share a cache entry". The first two asserted a data-layer
   opinion about tab ids — exactly what this layer removes — and only compiled
   through a `data as { tabs: … }` cast, so they have no owner once the mapping is
   gone. The third **inverts** into *serves both contexts from one cache entry*,
   asserting both hooks return the same object and that only **one** request was
   made, which pins the removal as the improvement it is.

11. **The pact test's expectation changes; the contract does not.** Its
    `slug: 'outline'` was an assertion about our *normalized* output and becomes
    `tabId: 'courseware'`; the `tab_id: 'courseware'` it asserts against the LMS
    is the actual contract and is untouched.

12. **`useIFrameBehavior`'s invalidation was an eleventh site, beyond the ten hook
    calls.** It builds the key directly —
    `courseHomeQueryKeys.metadata(eventCourseId, 'courseware')` — rather than
    calling the hook, so it does not show up in a search for the hook. Its test
    asserts the same key and moves with it.

13. **`discussionsPrefetch` got the only new test of a pre-existing path.**
    Nothing exercised it: every test calls `prefetchDiscussionTopics` directly,
    bypassing the tab gate, and the only test that renders the provider supplies
    `{ tabs: [] }`. With no type coverage on that path either, a mistake there
    would have been invisible — CI green, discussion-topic prefetching silently
    dead. It is the one site in this change where that was true.

    Its `edxProvider` local was dropped rather than renamed, because the name made
    a claim the line never checked. `DiscussionTab.is_enabled`
    (`lms/djangoapps/discussion/plugins.py`) returns `False` when
    `DiscussionLtiCourseTab` is enabled, so the `discussion` tab's presence means
    discussions are enabled and **not** LTI-provided — it says nothing about which
    provider. `openedx` and `legacy` both produce this tab; `prefetchDiscussionTopics`
    makes that distinction itself, one layer down (*"Only load topics for the openedx
    provider, the legacy provider uses the xblock"*). Authoring's vocabulary agrees:
    provider ids are `openedx`, `legacy`, or an LTI app, and `AppList.jsx` draws the
    line as `!['openedx', 'legacy'].includes(activeAppId)`. So `hasDiscussionTab(course?.tabs)`
    inlined into the condition states the real gate, and "edx provider" appears
    nowhere.

    A plugin could in principle render a nav entry of its own with a `discussion`
    slug, which would not make this prefetch fire: the gate reads `course.tabs` from
    the metadata payload, and a slot-inserted link never enters that array.

14. **The accessors take `tabs`, not `courseId`, and the hook façade waits for
    B2.** The destination is components that never handle a tab list at all —
    `const progressTabUrl = useProgressTabUrl();` — following the zero-argument
    pattern already established by `useProgressData()` and `useExamsData()` in
    `course-home/progress-tab/hooks.jsx`, which take `courseId` from `useParams()`.
    That is recorded on #2085; the hooks are thin wrappers over these accessors,
    so nothing here is throwaway.

    The suffix is load-bearing: `useDatesTabData`, `useOutlineTabData`,
    `useLiveTabData` and `useProgressTabData` already exist for the tab *content*
    endpoints, so a bare `useProgressTab()` would sit next to `useProgressTabData()`
    returning something entirely different. Naming each hook for what it returns
    keeps them apart: `Data` is what the tab shows, `Url` is a link to it. An
    earlier draft used a `Meta` suffix, from when the accessors returned tab
    objects; they return URLs, so the hooks follow.

    It waits because the source is still moving. Five of the six tab readers
    destructure `tabs` in the same `useModel('courseHomeMeta', courseId)` call as
    `org`, `title` or `verifiedMode`, so converting tabs alone would leave a
    component reading one payload through two mechanisms — the query for tabs, the
    store for the rest. B2 and B3 rewrite those destructures wholesale, which is
    where the hooks cost nothing extra. It is a coherence constraint rather than a
    technical one: the bridge keeps both paths valid, so a hook built now would
    return correct values.

    A `tabs` parameter is also the shape that survives either way.
    `widgets/discussions/widgetConfig.js` keeps calling `hasDiscussionTab` directly
    and permanently — `discussionsPrefetch({ courseId, course, queryClient })` is a
    widget-lifecycle function, not a component, and cannot call a hook.

15. **`CertificateStatusAlert`'s PropTypes were wrong in both directions.** They
    documented `tab_id` while the runtime objects carried `slug`; the camelCased
    field is `tabId`, so the fix is not a consequence of this change so much as a
    thing it made visible.

## What actually changes in the payload

The complete cross-product of what the API sends and what each caller asked for.
One cell moves:

| `tab_id` (from the API) | `rootSlug` (from the caller) | field before | field now |
|---|---|---|---|
| `courseware` | `outline` — the five course-home tabs, `CourseAccessErrorPage` | `slug: `**`outline`** | `tabId: `**`courseware`** |
| `courseware` | `courseware` — `CoursewareContainer`, `CourseExit`, `redirects`, `useIsCourseLoaded` | `slug: courseware` | `tabId: courseware` |
| `dates`, `progress`, `discussion`, `wiki`, `instructor`, `lti_live` | `outline` | `slug: ` same as `tab_id` | `tabId: ` same as `tab_id` |
| `dates`, `progress`, `discussion`, `wiki`, `instructor`, `lti_live` | `courseware` | `slug: ` same as `tab_id` | `tabId: ` same as `tab_id` |

Rows 2–4 were already identity: `rootSlug` was consulted only for the
`courseware` tab, and only on a course-home page did it produce a value
different from `tab_id`. So the old ternary did nothing at all on courseware
routes — it replaced `'courseware'` with `'courseware'`.

That single relabelling existed to make one comparison work. The courseware tab
is the nav entry for both root pages, and consumers asked "is this the page I am
on?" with `slug === activeTabSlug`, which cannot express "active on either of
two pages". Pre-stamping the payload per fetch made the plain equality true.
`isActiveTab` states the same fact directly, so the payload no longer has to
carry it — and once the payload is page-independent, one cache entry serves both
contexts:

| page | before: `slug` vs `activeTabSlug` | after: `isActiveTab('courseware', activeTabSlug)` |
|---|---|---|
| outline | `outline` = `outline` → active | `activeTabSlug === 'outline'` → active |
| courseware | `courseware` = `courseware` → active | `activeTabSlug === 'courseware'` → active |
| dates / progress / discussion / live | `outline` ≠ `dates` … → inactive | neither branch matches → inactive |

## Behaviour changes

- **One cache entry per course instead of two**, verified by the inverted test in
  `apiHooks.test.tsx`: both contexts return the same object and one controlled
  render makes one request.

  **This does not reduce requests in the browser, and the layer should not claim
  it does.** Measured in tutor dev on the outline → courseware crossing (clear the
  network log once the outline settles, click a sequence title, wait for idle,
  count `course_metadata`): **exactly 3, five runs each, on this layer and on
  #2093** — no variance either side. An earlier draft claimed "two requests become
  one"; that was reasoning from the key alone, and it is wrong.

  The reason is `staleTime`. `useCourseHomeMeta` sets none, so the entry is stale
  the moment it lands and `refetchOnMount` fires for each fetching observer that mounts
  after the owner's fetch has settled. Collapsing two keys into one turns a cache
  *miss* into a stale *hit*; it does not remove a fetch. Whether that changes what
  blocks `useIsCourseLoaded` was not measured. An earlier version of this entry
  attributed the three requests to `CoursewareContainer`, `useIsCourseLoaded` and
  `redirects` around a remount across the redirect; #2098 traced them and found no
  remount, those three deduped in one render, and the extra two coming from
  `UnitNavigation` and `UnitNavigationEffortEstimate` through `useSequenceIds`.

  The tab bar is not a client-side path at all: `CourseTabLink` renders a plain
  `<a href>`, so every tab click is a full page load that discards the cache. The
  only crossings that reuse it are a sequence title on the outline (`SequenceTitle`
  renders a router `Link` into courseware) and the home breadcrumb in courseware
  (`CourseBreadcrumbs` links to `/home`).

  Reducing the count is a `staleTime` question rather than a key question, and this
  layer makes it tractable by leaving one key to tune instead of two.
- **Tabs carry the ids the LMS sent** wherever the payload is read, including the
  `courseHomeMeta` model the bridge still writes until #1999. The bridge is a
  passthrough (`model: { id: courseId, ...data }`) and needed no change.
- **No re-render change.** `camelCaseObject` already deep-clones per fetch and
  `useModel` compares `tabs` by reference, so that reference was new on every
  fetch with or without the mapping. Deleting it removes one array allocation.
- **Fields the endpoint adds in future now reach consumers.** The mapping was the
  only allowlist on a tab; `...data` already passed every other field of the
  payload through unfiltered, so this makes tabs consistent with the rest.

Plugins are unaffected: `CourseTabsNavigationSlot` passes no `pluginProps` and
`CourseTabLinksSlot` passes only `activeTabSlug`, so the tabs array never reaches
plugin code.

## Left alone

- **`generic/tabs/Tabs.jsx`** drops `className`/`style` when cloning
  `CourseTabLink` for the overflow menu, so tab overflow does not work. Pre-dates
  this work; touching `CourseTabLink` here was tempting and declined.
- **The stale `tabs` block in the generated `src/pacts/frontend-app-learning-lms.json`**
  under the courseware interaction, carrying the old `slug`/`priority`/`type`
  shape. Unasserted by any test since [#861](https://github.com/openedx/frontend-app-learning/pull/861),
  and describing a response the platform deleted in
  [openedx-platform#30023](https://github.com/openedx/openedx-platform/pull/30023).
  openedx-platform's own `courseware_api/tests/pacts/` fixture carries the same
  fossil, so fixing it properly is a two-repo change.

- **`tab.factory.js`'s home in `src/shared/`.** It was moved there by
  [#400](https://github.com/openedx/frontend-app-learning/pull/400) to serve both
  the courseware and course-home fixtures — and the same commit deleted the
  courseware factory's `tabs` attribute, so it has had exactly one consumer from
  the moment it became "shared". Moving it under
  `course-home/data/__factories__/` would finish the thought, but it is churn this
  layer does not need.

</details>

## Manual testing

<details>
<summary>Checklist</summary>

# Manual testing — take tab identity out of the course-home metadata query (#2084)

In-browser verification against a live backend (tutor dev). The claim is **no
user-visible change at all** — including request counts, which were measured and are
unchanged (see below). Earlier drafts of this doc claimed the layer saved a request;
that was wrong.

Worth knowing while testing: the **tab bar is not** a client-side path.
`CourseTabLink` renders a plain `<a href>` with a server-supplied absolute URL, so
every tab click is a full page load that discards the cache. The two client-side
crossings are a **sequence title on the outline** (`SequenceTitle` renders a router
`Link` into courseware) and the **home breadcrumb in courseware**
(`CourseBreadcrumbs` links back to `/home`).

Two things are specific to this layer and carry the most risk:

- **the nav highlight** is the whole point of the change. `isActiveTab` replaces a
  payload that was pre-stamped per page, so the Course tab being lit on *both* root
  pages — and not lit anywhere else — is the check that matters most;
- **every deep link built from a tab's `url`** now comes from an accessor rather
  than an inline `find`. Those are spread across five components and several are
  only reachable in awkward course states, so they are listed individually below.

Nothing here is type-checked: `useModel('courseHomeMeta')` returns `any`, so a
missed reader compiles and passes CI. The fixtures were corrected for that reason,
but a live click is the only real proof.

## Setup

An ordinary course is enough for most of it — one with **dates**, **progress** and
**discussion** tabs enabled, plus at least one LMS-hosted tab (**wiki** or
**instructor** as staff) to confirm entries this MFE does not route still render.

Harder states, each needed by exactly one check:

- **Hidden-after-due link** — a subsection with "hide after due date" set and a due
  date in the past.
- **Certificate not-passing alert** — a course that has ended where the learner has
  a failing grade, in a verified-style mode.
- **Course exit pages** — `/course/:courseId/course-end`, reachable by completing
  the course (in progress vs non-passing give different buttons).

### The nav highlight (the core of this layer)

- [x] **Course tab is highlighted on the outline** — `/course/:courseId/home`.
- [x] **Course tab is highlighted on courseware** — any unit page. Before this
      layer these were two different cache entries stamped with different values;
      now one entry serves both and `isActiveTab` does the work.
- [x] **Course tab is *not* highlighted on other tabs** — dates, progress,
      discussion, live. Only the tab you are on is lit.
- [x] **Each other tab highlights on its own page** — dates on dates, progress on
      progress, and so on.
- [x] **The highlight is correct after client-side navigation** — from the outline,
      click a **sequence title** to enter courseware, then the **home breadcrumb**
      back. The nav unmounts and reappears each way (`TabPage` shows a loading state
      while the courseware queries resolve), so this is not a continuity check: it is
      that the Course tab is highlighted once each page settles. These are the only
      two transitions that reuse the cache, so they are the only ones where a stale
      cached value could show through.

### The page title

- [x] **Helmet title carries the tab name** on the outline and on courseware
      (both should read "Course | …"), and on dates/progress ("Dates | …").
      This is `getActiveTabTitle`, which now returns the string rather than the tab.

### Request count — measured, no difference

Protocol: hard reload the outline, wait for the Network tab to go idle, clear the
log, click a **sequence title**, wait for idle again, count `course_metadata`. The
last wait matters — cutting it short undercounts, which is what produced some
inconsistent early readings.

Result: **exactly 3, five runs each, on this layer and on `4750d1fc` (#2093)**. No
variance on either side. Three hits because fetching observers under
`CoursewareContainer` mount after its fetch has settled and refetch the stale entry.
#2098 traced the extra two to `UnitNavigation` and `UnitNavigationEffortEstimate`,
through `useSequenceIds`; there is no remount across the redirect.

- [x] **No measurable change in request count.** The layer does not reduce requests
      and should not claim to. `useCourseHomeMeta` sets no `staleTime`, so a shared
      key still refetches for every observer that mounts on a stale entry;
      collapsing two keys into one turns a cache miss into a stale hit rather than
      removing a fetch.
Not worth checking: whether the crossing feels faster. `TabPage` gates on
`useIsCourseLoaded`, which requires `useCoursewareMetadata`, `useCoursewareOutline`
**and** `useCourseHomeMeta` to succeed. This layer turns the third from a cache miss
into a stale hit, so it stops holding the spinner — but the first two are misses on
that crossing regardless, since the outline page never fetches them. One fewer
blocker out of three, where the other two still block, is not observable.

### Links built from a tab's url

- [x] **Progress tab → "Course outline"** in Related Links resolves to
      `/course/:courseId/home`. (`getCourseOutlineUrl`)
- [x] **Progress tab → "Dates"** in Related Links resolves to the dates tab.
      (`getDatesTabUrl`)
- [x] **Progress tab → "Course outline" link in the ungraded note** — the line
      reading *"For progress on ungraded aspects of the course, view your Course
      outline."* below the grades table; its inline link resolves. Renders only when
      `showUngradedAssignments()` is false. (`getCourseOutlineUrl` in
      `DetailedGrades`)
- [x] **Hidden-after-due alert → "progress page"** link resolves to the progress
      tab. (`getProgressTabUrl`) — **the one worth setting up**, because it is the
      only live check of `getProgressTabUrl` and the cheapest state to reach. Set it
      up entirely in the authoring MFE: **Course outline → subsection → Configure →
      Visibility**, middle radio option. On an **instructor-paced** course that reads
      "Hide content after due date" and needs a subsection **due date in the past**
      (Basic tab of the same modal); on a self-paced one it reads "Hide content after
      end date" and keys off the course end date instead. View as a **learner** —
      staff bypass the check entirely, and masquerading as a *specific* student
      deliberately shows the content with a banner rather than hiding it, so use the
      generic Learner role or a real learner account.

      If the alert does not appear, check `is_hidden_after_due` in the
      `/api/courseware/sequence/<usage key>` response — that is exactly what
      `Sequence.jsx:161` branches on, and `false` means the course config is not
      taking effect rather than the MFE being wrong.

Deliberately skipped — each is a repeat call site of an accessor already confirmed
above, in a state that costs a lot to reach. With the hidden-after-due check done,
every accessor has a live confirmation: `getCourseOutlineUrl` and `getDatesTabUrl`
from Related Links, `getProgressTabUrl` from the hidden-after-due alert.

- [ ] ~~Certificate not-passing alert → "view grades"~~ (`getProgressTabUrl`) —
      needs enrolled + a verified-family enrollment mode + course ended + not passing
      + no alerting cert status. *Not run.*
- [ ] ~~Course exit (in progress) → "view course schedule"~~ (`getDatesTabUrl`) —
      needs `hasScheduledContent && !userHasPassingGrade` and the course-end page.
      *Not run.*
- [ ] ~~Course exit (non-passing) → "view grades"~~ (`getProgressTabUrl`) — needs
      `isEligibleForCertificate && !userHasPassingGrade && canImmediatelyViewCertificate`.
      *Not run.*

### Discussions

- [x] **Discussion topics still prefetch** — with `DISCUSSIONS_MFE_BASE_URL` set
      and the course using built-in discussions, open a unit and confirm the
      discussions sidebar works as before. This gate lost its `edxProvider` local
      and is now `hasDiscussionTab(course?.tabs)`; it had no test at all before
      this layer, so a live check is worth more here than elsewhere.
- [x] **A course with discussions disabled** still loads units with no errors and
      no discussion topics request.

### Tabs this MFE does not route

- [x] **LMS-hosted tabs still render and navigate** — checked with **Teams**.
      Clicking it leaves the MFE entirely for the LMS teams page, which renders its
      own tab bar matching the one the MFE shows, and every tab there navigates to
      the right place.

      Stronger than the check was aiming for: both sides render from the same
      `get_course_tab_list`, so matching tab sets and working links confirm the URLs
      round-trip correctly rather than only that ours render. Teams also exercises
      `getActiveTabTitle` tolerating an id this MFE has never heard of.

      (Teams is the easiest subject — `TeamsTab` is an `EnrolledTab` with
      `view_name = "teams_dashboard"`, so it reverses to an LMS route and needs no
      staff account. Wiki or instructor work too.)

### Plugin links

- [x] **A slot-inserted `CourseTabLink` still renders and highlights as coded.**
      Checked with a temporary `env.config.jsx` inserting links into
      `org.openedx.frontend.learning.course_tab_links.v1`. They render in the nav,
      the real tabs are unaffected, and `activeTabSlug` reaches the widget as before.
      This layer changes nothing about the slot's props or `CourseTabLink`.

### General

- [ ] **No new console noise** on any of the above. *Not run — the MFE is noisy
      enough at baseline that a small addition would not be distinguishable without
      a before/after capture, which was not worth setting up for this layer.*

</details>


arbrandes (approved):
👍🏼

Issue #2103, opened 2026-09-23
Stop the progress tab data refetching from components under the tab


brian-smith-tcril:
Part of #1946 — Redux → React Query migration (Stage 1). Found while working on #2098; the same defect #2083 fixed for the outline and dates tabs and #2098 fixed for the courseware gate. Layer on top of #2104.

**Goal:** fetch `/api/course_home/progress/` once per progress-tab load, from `ProgressTab`, instead of once more when the components under it mount.

**Context.** One hard reload of `/course/:courseId/progress` requests `/api/course_home/progress/` **twice** (measured on tutor dev). `ProgressTab` (`progress-tab/ProgressTab.jsx:52`) fetches it with `useProgressTabData(courseId, targetUserId)` and passes it to `TabPage` as `courseStatus`. The rest of the tab reads it through `useProgressData()` (`progress-tab/hooks.jsx:5-8`), which is also a fetching observer of that query:

```js
export function useProgressData() {
  const { courseId, targetUserId } = useParams();
  return useProgressTabData(courseId, targetUserId).data;
}
```

Its 22 call sites (21 files in `course-home/progress-tab/`, including `useExamsData` in `hooks.jsx`, plus `plugin-slots/ProgressTabGradeBreakdownSlot/index.jsx`) all render under `TabPage`'s gate, so they mount after the data has landed, find it stale at the default `staleTime` of 0, and refetch. React Query dedupes the readers against each other, which is why the count is two and not twenty-two.

**Tasks**
- Add `{ enabled }` to `useProgressTabData` (the course-home `QueryOptions`, in the two-positional-plus-options shape `useProctoringInfoData` has); plain `enabled`, no `!!courseId` guard, since the hook has never had one.
- Make `useProgressData()` non-fetching, as #2098 did for `useIsCourseLoaded`: it passes `{ enabled: false }` and takes no option. `useExamsData()` reads through it and inherits that. `ProgressTab` keeps fetching.
- Add one sentence to `progress-tab/README.md` under `useExamsData()`: it reads the progress tab's own data, so it is for widgets in progress-tab slots; outside the tab it returns `null` unless something rendered on the page fetches that data (for example a widget calling `useProgressTabData(courseId)`).
[...]

PR #2105, opened 2026-09-23
refactor: move the progress tab's legacy-page redirect out of the query and into ProgressTab

### LINKED ISSUE #2104: Move the 404 redirect to the legacy progress page out of `getProgressTabData` and into `ProgressTab`
Part of #1946 — Redux → React Query migration (Stage 1). Found while planning #2103, and the layer before it.

**Goal:** make fetching `/api/course_home/progress/` side-effect free, so that only `ProgressTab` — the component that owns that fetch — can send the learner to the legacy progress page.

**Context.** `getProgressTabData` (`course-home/data/api.js:169-174`), the `queryFn` of `useProgressTabData`, handles a 404 by navigating the whole page:

```js
  } catch (error) {
    const httpErrorStatus = error?.response?.status;
    if (httpErrorStatus === 404) {
      global.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/progress`);
      return {};
    }
```

The redirect itself is a feature, from the original progress tab (#391, 2021). The endpoint 404s when the MFE progress tab is not active for the course — `course_home_mfe_progress_tab_is_active` in `openedx-platform` `lms/djangoapps/course_home_api/progress/views.py` is false for an Old Mongo course, when the `course_home.course_home_mfe_progress_tab` course waffle flag is off, or when `DisableProgressPageStackedConfig` disables it — and for a bad `student_id`. The legacy `courses/<id>/progress` view is the mirror image: it redirects learners to the MFE when the tab *is* active. So a course that opted out of the MFE progress tab bounces to the page that works.

The problem is where the redirect lives. Because it runs inside the `queryFn`, **anything that fetches this query redirects the page**, not just the progress tab. Today that includes a plugin widget calling `useExamsData()` (`progress-tab/hooks.jsx:11`) from a slot on another tab: it fetches through `useProgressData()`, and on a course with the MFE progress tab disabled, the learner is bounced off the outline (say) to the legacy progress page. After #2103 makes those hooks non-fetching, its migration note tells a plugin outside the progress tab to mount `useProgressTabData(courseId)` itself, which walks into the same redirect. The side effect belongs to the owner, not to the query, the same ownership line #2083, #2098 and #2103 draw for the fetch itself.

**Tasks**
- `getProgressTabData`: drop the 404 branch, so a 404 propagates like every unhandled status. The 401 and 403 branches (resolve to `{}`; access is handled via the metadata request) stay as they are. `shouldRetryQuery` (`src/queryClient.ts:34-43`) already skips retries on 4xx.
- `useProgressTabData`: add `logStatusAs: { 404: 'silent' }` to its `meta`, through the mechanism `useCoursewareOutline` uses for its expected 403, so the global `QueryCache.onError` (#2022) does not start logging a 404 that was never logged before (today the query succeeds with `{}`).
- `ProgressTab` (`progress-tab/ProgressTab.jsx:49-63`): when `tabDataQuery` fails with a 404 (`getResponseStatus` from `@src/data/http-error`), call `global.location.replace(...)` with the same URL from an effect, and return `PageLoading` — announcing `TabPage`'s existing `messages.loading` via `useIntl` — instead of `TabWithTimer`, so the page shows a spinner rather than `TabPage`'s error view while the browser navigates. Redirect on every 404, as today.
- Tests: the hook case *redirects to the legacy progress page and resolves to an empty object on a 404* (`course-home/data/apiHooks.test.tsx:272-288`, with its `window.location` swap) becomes *surfaces a 404 as an error without navigating*; `ProgressTab.test.jsx` gains the redirect case (404 reply, `replace` called with the legacy URL, loading indicator present, failure text absent), rendered without `fetchAndRender`'s wait for the indicator to disappear.

**Verify:** on a course with the MFE progress tab disabled (`course_home.course_home_mfe_progress_tab` off for the course, via Django admin), loading `/course/:courseId/progress` still lands on `/courses/:courseId/progress` on the LMS, with a spinner and no error page in between. On that same course, a plugin widget on the outline tab that mounts `useProgressTabData(courseId)` (the #2103 migration shape) stays on the outline and reads an error. On a normal course the progress tab is unchanged, including `/progress/:targetUserId` as staff.

**Behaviour change.** The redirect moves from before the query settles to the render after it errors: one render later, and only when `ProgressTab` is the fetcher. Today the query resolves to `{}` and the tab renders its content against an empty payload until the browser navigates; after this it renders a spinner instead. The 404 stays unlogged, as today. Nothing changes on any other response.

**Out of scope:** the refetch from readers under the tab (#2103, the layer above this one); the 401/403 handling; guarding the redirect on `targetUserId` (the endpoint's two 404 causes have identical bodies, and the current code redirects on both); and whether the legacy progress page is still a meaningful destination on current `openedx-platform` (it is, on `master`).

## Plan

> [!NOTE]

- **The destination is right; the trigger is wrong.** The endpoint 404s when the course opted out of the MFE progress tab (flag `toggle_default: False`; tutor's MFE plugin turns it on) or for a bad `student_id`, and the legacy view redirects the other way when the tab is active. The redirect stays a progress-tab feature; only what fires it changes.
- **Any fetcher redirects.** The side effect sits in the `queryFn`, so a plugin off the tab that fetches this query — today through `useExamsData()`, after #2103 by mounting `useProgressTabData` as its migration note says — bounces the learner to the legacy progress page from whatever tab they were on.
- **Throwing the 404 has one new consequence: logging.** Retries already skip 4xx. The global `onError` would log it at error, and today it is not logged at all (the query succeeds), so the hook declares `logStatusAs: { 404: 'silent' }` to keep that; `Register.queryMeta` already types `ModelStoreMeta & { logStatusAs? }` and the outline query already carries both kinds of entry.
- **Not `TabPage`'s error view during the navigation interval.** With the query in error, `deriveView` picks `isError`, and `getErrorDetail` has prose only for a 403, so the page would read "There was an error loading this course." until the LMS serves the legacy page — which computes grades and is often seconds. The repo's redirect precedents (`TabPage`'s `<Navigate>` for denied access, `courseware/RedirectPage.tsx`) render nothing; a spinner is the small honest deviation. Reusing `TabPage`'s loading string keeps the screen-reader announcement identical to the loading state that preceded it (the string is a prop to `PageLoading`, rendered `sr-only`).
- **The header drops out when the 404 lands.** Keeping it would mean holding `TabPage` in its loading view with an errored query; the jump is not worth that.
- **Redirect on every 404, as today.** The two 404 causes are indistinguishable from the response, and guarding on `targetUserId` would be a behaviour change unrelated to where the side effect lives.
- **The happy path is untouched.** On success `error` is null, the flag is false, the effect body is skipped and `ProgressTab` returns the same tree. 401/403 keep resolving to `{}`; other errors were already thrown.
- **`refactor:`, not breaking.** A plugin that fetched this query off the tab loses an unintended page navigation.

<details>
<summary>Full plan</summary>

# Plan: #2104 — Move the 404 redirect to the legacy progress page out of `getProgressTabData` and into `ProgressTab`

Next layer in the running stack, on top of #2098 (PR #2102). #2103 stacks on
this, so that its plugin migration advice (mount `useProgressTabData` yourself)
is side-effect free.

### How it works today

`useProgressTabData` (`course-home/data/apiHooks.ts:133-137`) has
`getProgressTabData` as its `queryFn`. That function's catch block
(`course-home/data/api.js:169-187`) turns three statuses into a successful
empty result, and one of them also navigates the page:

```js
  } catch (error) {
    const httpErrorStatus = error?.response?.status;
    if (httpErrorStatus === 404) {
      global.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/progress`);
      return {};
    }
    if (httpErrorStatus === 401) { … return {}; }   // unenrolled / unauthenticated: courseAccess handles it
    if (httpErrorStatus === 403) { … return {}; }   // access error: the metadata request carries the redirect
    throw error;
  }
```

So on a 404 the query *succeeds* with `{}`, `TabPage` opens its gate, and the
tab renders its content against an empty payload while the browser navigates
away. The redirect is from the original progress tab (#391, 2021).

### When the endpoint 404s

`lms/djangoapps/course_home_api/progress/views.py` (`openedx-platform`
`master`) raises 404 when:

- `course_home_mfe_progress_tab_is_active(course_key)` is false: an Old Mongo
  course, the `course_home.course_home_mfe_progress_tab` course waffle flag off
  for the course (`toggle_default: False`; tutor's MFE plugin turns it on), or
  `DisableProgressPageStackedConfig` disabled for it.
- `student_id` is not an integer, names no user, or the requester is not staff
  and asked for someone else.

The legacy `courses/<id>/progress` view is the mirror image: with the MFE tab
active it redirects non-staff to the MFE. So the destination is right, and the
redirect stays a feature of the progress tab.

### Why it has to move

The side effect lives in the `queryFn`, so **whatever fetches the query
redirects the page**, not just the progress tab. Today a plugin widget calling
`useExamsData()` from a slot on the outline fetches through `useProgressData()`
(both `progress-tab/hooks.jsx`), and on a course with the MFE progress tab
disabled the learner is bounced off the outline to the legacy progress page.
#2103 makes those hooks non-fetching and tells a plugin outside the tab to call
`useProgressTabData(courseId)` itself, which is the same trap. Ownership is the
line #2083, #2098 and #2103 draw for the fetch; this draws it for the fetch's
side effect: `ProgressTab` owns the fetch, so `ProgressTab` owns the redirect.

## The change

1. **`getProgressTabData`: drop the 404 branch.** A 404 falls through to
   `throw error` like every unhandled status. The 401 and 403 branches and
   their comments stay. Nothing else in `api.js` changes.

2. **`useProgressTabData`: keep the 404 silent.** `shouldRetryQuery`
   (`src/queryClient.ts:34-43`) already skips retries on 4xx, so the only new
   consequence of throwing is that the global `QueryCache.onError` (#2022) logs
   it, at `error` by default. Today the 404 is not logged at all — the query
   succeeds with `{}` — so a faithful port keeps it that way: the hook declares
   `'silent'`, the `LogLevel` the courseware outline query's `logStatusAs`
   mechanism already offers (`courseware/data/apiHooks.ts` uses `'info'` for
   its expected 403):

   ```ts
   meta: { modelType: 'progress', courseId, logStatusAs: { 404: 'silent' } },
   ```

   `Register.queryMeta` is already `ModelStoreMeta & { logStatusAs? }`
   (`queryClient.ts:17-21`), so the two coexist without a type change.

3. **`ProgressTab` redirects.** `progress-tab/ProgressTab.jsx:49-63` becomes:

   ```jsx
   const ProgressTab = () => {
     const { courseId, targetUserId } = useParams();
     const tabDataQuery = useProgressTabData(courseId, targetUserId);
     const redirectToLegacyProgress = getResponseStatus(tabDataQuery.error) === 404;

     useEffect(() => {
       if (redirectToLegacyProgress) {
         global.location.replace(`${getConfig().LMS_BASE_URL}/courses/${courseId}/progress`);
       }
     }, [redirectToLegacyProgress, courseId]);

     if (redirectToLegacyProgress) {
       return <PageLoading srMessage={…} />;
     }
     return (
       <TabWithTimer …>
   ```

   `getResponseStatus` is `src/data/http-error.ts:5-7`, already what `TabPage`
   and the query cache use to read a status. The URL string moves verbatim; as
   today it does not append `targetUserId` (the legacy view redirects a
   non-staff learner back to their own MFE progress page, which is the
   existing fallback chain).

   **Effect, not render-time.** `courseware/RedirectPage.tsx` calls
   `location.assign` during render. Here an effect keeps the render pure, and
   under `StrictMode`'s development double-invoke (`src/index.jsx`) the worst
   case is `replace` called twice with the same URL, which is a no-op.

4. **What the page shows while the browser navigates** (open question 1).
   `TabPage` (`tab-page/TabPage.tsx:141-149`) renders `HeaderSlot`, then one
   of loading / content / error, then `FooterSlot`. With the query now in
   error, rendering `TabWithTimer` as usual would show the header, the generic
   failure text ("There was an error loading this course") and the footer
   until the new document arrives. The draft above instead returns
   `PageLoading` from `ProgressTab`: a spinner with no header or footer, for
   the same interval. Neither is what today shows (content against `{}`,
   which for the nested destructures under the tab is a crash), so both are
   improvements; the question is which look to pick.

## Tests

- **`course-home/data/apiHooks.test.tsx`**, `describe('useProgressTabData')`:
  the case *redirects to the legacy progress page and resolves to an empty
  object on a 404* (`:272-288`) is replaced by *surfaces a 404 as an error
  without navigating*. It keeps the `window.location` swap (jsdom's `replace`
  is non-configurable, so the whole object is swapped in a `try/finally`) and
  asserts `isError`, `getResponseStatus(result.current.error) === 404`, and
  `replace` not called. The `logStatusAs` mechanism is covered generically in
  `src/queryClient.test.ts:37-46`; no hook-level log assertion, matching the
  outline 403.

- **`progress-tab/ProgressTab.test.jsx`**, new describe *when the progress
  request 404s*: mock `progressUrl` to reply 404 (`reply(404, {})` gives
  `error.response.status`, which is all `getResponseStatus` reads), swap
  `window.location`, render, and assert `replace` was called with
  `${LMS_BASE_URL}/courses/${courseId}/progress`, the loading indicator
  (`role="status"`) is present, and the failure text is not. `fetchAndRender`
  (`:73-91`) waits for the loading indicator to disappear, which here it
  never does, so this case renders through a variant that skips that wait.

- **`api.js` has no direct suite** for `getProgressTabData`; the hook suite is
  its coverage today and stays so.

## Behaviour changes

- The redirect fires one render later: after the query has errored, from the
  owner's effect, instead of inside the `queryFn` before the query settles.
- Only `ProgressTab` redirects. Any other fetcher of the progress query reads
  an error and stays on its page.
- During the navigation interval the page shows a loading (or error) view
  instead of the tab rendered against an empty payload.
- The 404 is still not logged (`logStatusAs: { 404: 'silent' }`), as today.

Not breaking: no plugin contract changes. A plugin that fetched this query off
the tab loses an unintended page navigation, which is the point.

## Commit type

`refactor:`. Nothing a learner on the progress tab sees is broken today, and
the change is where a side effect lives. Same call as `decisions-2098.md`
entry 2.

## Not in scope

- **The 401 / 403 branches** in `getProgressTabData`. They resolve to `{}`
  deliberately (access is decided from the metadata request) and have no side
  effect.
- **`useProgressData` / `useExamsData` fetching** — that is #2103, the layer
  above this one.
- **Appending `targetUserId` to the legacy URL.** Not done today; the fallback
  chain through the legacy view works without it.

## Open questions

1. **Settled: `PageLoading` during the navigation interval, announcing
   `TabPage`'s `messages.loading` ("Loading course page…").** `PageLoading`
   has no text of its own; `srMessage` is a required prop rendered in an
   `sr-only` span, so sighted learners see only the spinner. `ProgressTab`
   does what `TabPage.tsx:113-115` does: import `messages` from
   `tab-page/messages`, call `useIntl()`, pass
   `intl.formatMessage(messages.loading)`. Reusing the string keeps the
   screen-reader announcement identical to the loading state that preceded
   it; a progress-specific message can be added later if wanted.
   Rendering `TabWithTimer` as usual would show `TabPage`'s error view —
   `deriveView` reads `tabDataQuery.isError`, and `getErrorDetail` has prose
   only for a 403, so the 404 falls back to "There was an error loading this
   course." — for as long as the LMS takes to serve the legacy page, which
   computes grades and is often seconds. The repo's redirect precedents
   (`TabPage`'s `<Navigate>` for denied access, `courseware/RedirectPage.tsx`)
   render nothing; the spinner is the small honest deviation. The header,
   visible during the preceding loading state, drops out when the 404 lands;
   keeping it would mean holding `TabPage` in its loading view with an errored
   query, and the jump is not worth that.
2. **Settled: redirect on every 404, as today.** The endpoint 404s for two
   unrelated reasons (tab disabled for the course; a bad or unauthorized
   `student_id`) with an identical body, and the current code redirects on
   both without looking. Guarding on `targetUserId` would be a behaviour
   change unrelated to where the side effect lives — scope creep for this
   issue; if that experience should change it gets its own issue.

## Verification

- Manual (tutor dev). Turn the MFE progress tab off for one course: Django
  admin → Waffle → course flag `course_home.course_home_mfe_progress_tab`, add
  a course override with the flag off (or `DisableProgressPageStackedConfig`
  for the course). Then:
  - Load `/course/:courseId/progress` as a learner: lands on
    `/courses/:courseId/progress` on the LMS; the interval before it shows
    the chosen view, not an error and not a crash.
  - On that course, a widget in an outline-tab slot (via `env.config.jsx`)
    that calls `useProgressTabData(courseId)` and renders `String(isError)`:
    the outline stays put and the widget reads `true`; the logging service
    records nothing for the 404.
  - Turn the flag back on: the progress tab loads as before, including
    `/progress/:targetUserId` as staff.

## Stack

One layer on top of #2098 (PR #2102), closing #2104. #2103 follows, then B2
(#2085).

</details>

brian-smith-tcril:
## Summary

`getProgressTabData`, the `queryFn` behind `useProgressTabData`, answered a 404 from `/api/course_home/progress/` by calling `location.replace` to the LMS's legacy progress page and resolving to `{}`. The destination is right: the endpoint 404s when the MFE progress tab is not active for the course (the `course_home.course_home_mfe_progress_tab` course flag off, an Old Mongo course, or `DisableProgressPageStackedConfig`), and the legacy view is the page that works there. The trigger is wrong: because the redirect sat inside the query, **anything that fetched the query navigated the whole page**, not just the progress tab. A plugin widget calling `useExamsData()` from a slot on another tab fetches through it, and on such a course bounced the learner off the outline to the legacy progress page.

The side effect now belongs to the fetch's owner. `getProgressTabData` lets the 404 propagate like every other unhandled status, and `ProgressTab`, the component that fetches this query, redirects from an effect when its query fails with a 404, showing a spinner while the browser navigates. Any other fetcher of the progress query reads an error and stays on its page. Nothing changes on any other response.

Part of the Redux → React Query migration (#1946, Stage 1). Closes #2104. Stacked above #2102 (#2098); #2103 stacks on this one, so that its plugin migration note — mount `useProgressTabData(courseId)` yourself — is side-effect free.

## What changed

- **`getProgressTabData` drops its 404 branch.** The 401 and 403 branches, which resolve to `{}` because access is decided from the metadata request, are unchanged.
- **`useProgressTabData` declares `logStatusAs: { 404: 'silent' }`.** Letting the 404 throw has one new consequence — the global `QueryCache` `onError` would log it as an error — and before this change the 404 was not logged at all (the query succeeded with `{}`), so the hook keeps it silent, through the same `logStatusAs` mechanism the courseware outline query uses for its expected 403. Retries already skip 4xx.
- **`ProgressTab` owns the redirect.** It reads the status off `tabDataQuery.error` with `getResponseStatus`, calls `location.replace` from an effect with the same URL as before, and returns `PageLoading` — announcing `TabPage`'s existing loading message — instead of `TabWithTimer`. Rendering `TabWithTimer` would have shown `TabPage`'s error view ("There was an error loading this course.") until the legacy page, which computes grades, finished loading. The redirect still fires on every 404, as before: the endpoint's two 404 causes (tab disabled; bad or unauthorized `student_id`) are indistinguishable from the response, and this change is about where the redirect lives, not when.
- **Tests.** The hook's 404 case now asserts the query errors without navigating (it keeps its `window.location` swap to prove `replace` was not called). `ProgressTab.test.jsx` gains the redirect case: 404 reply, `replace` called with the legacy URL, loading indicator present, failure text absent. Its `fetchAndRender` took a `{ waitForLoaded }` option because the loading indicator it normally waits out is the expected view here. Both new cases fail against the pre-change source.

## Testing

`npm run types`, `npm run lint` and the full suite (113 suites, 1141 passed, 3 skipped) are green.

Manually verified on tutor dev as staff, producing real 404s with a `targetUserId` that matches no user (`/progress/999999`) — the endpoint answers that exactly as it answers an opted-out course, and the frontend takes the same path from there. The bad-id progress URL shows the bare spinner and lands on the legacy page with no error view in between; browser back returns cleanly; an outline-tab probe widget that mounts `useProgressTabData` with the bad id stays on the outline and reads the 404; one request per load, no retry, nothing logged; the plain progress URL, viewing a real learner's progress, and masquerade all render as before. The console was not compared against a before capture. A course with the MFE progress tab actually disabled was not set up — same status, same body, same frontend path.

## Decisions

<details>

# Decisions — move the 404 redirect to the legacy progress page out of `getProgressTabData` and into `ProgressTab` (#2104)

Layer on top of #2098 (PR #2102); #2103 stacks on it. The diagnosis, the
ownership move and the two open questions were settled in the plan on issue
#2104; entries here record choices made while implementing.

1. **What the page shows while the browser navigates: `PageLoading`,
   announcing `TabPage`'s own loading message.** With the 404 now an error,
   rendering `TabWithTimer` as usual would put `TabPage` in its error view:
   `deriveView` reads `tabDataQuery.isError`, and `getErrorDetail` has
   backend prose only for a 403, so a 404 falls back to "There was an error
   loading this course." That text would stay up until the LMS serves the
   legacy progress page, which computes grades and is often seconds, for a
   learner who is about to land on a page that works. The repo's redirect
   precedents (`TabPage`'s `<Navigate>` for denied access,
   `courseware/RedirectPage.tsx`) render nothing; the spinner is the small
   honest deviation. `PageLoading` has no text of its own — `srMessage` is a
   required prop rendered `sr-only` — so `ProgressTab` formats
   `tab-page/messages`' `loading` ("Loading course page…") itself, the way
   `TabPage.tsx:113-115` does. That keeps the screen-reader announcement
   identical to the loading state that preceded it; a progress-specific
   string can be added later if wanted. The header, visible during that
   preceding state, drops out when the 404 lands; keeping it would mean
   holding `TabPage` in its loading view with an errored query, and the jump
   is not worth that.

2. **Redirect on every 404, as today.** The endpoint 404s for two unrelated
   reasons — the MFE progress tab disabled for the course, or a bad or
   unauthorized `student_id` — with an identical body, and the current code
   redirects on both without looking. Guarding on `targetUserId` would be a
   behaviour change unrelated to where the side effect lives; if that
   experience should change it gets its own issue.

3. **Effect, not render-time.** `courseware/RedirectPage.tsx` calls
   `location.assign` during render. Here a `useEffect` keyed on the derived
   boolean and `courseId` keeps the render pure; under `StrictMode`'s
   development double-invoke the worst case is `replace` called twice with
   the same URL, a no-op.

4. **`logStatusAs: { 404: 'silent' }` on the hook.** The only new consequence of
   letting the 404 throw is logging: `shouldRetryQuery` already skips 4xx, and
   the global `QueryCache.onError` (#2022) logs at `error` unless the query
   says otherwise. Before this change the 404 was not logged at any level —
   `getProgressTabData` caught it and the query succeeded with `{}` — so the
   faithful port keeps it silent. `'info'` was the first draft, following the
   courseware outline query's treatment of its expected 403 and on the
   grounds that the redirect is now a decision the app makes; review
   (2026-09-23) preferred preserving the old behaviour exactly, and `'silent'`
   is a `LogLevel` the mechanism already supports. The mechanism is covered in
   `src/queryClient.test.ts` (including a `'silent'` case); no hook-level log
   assertion, matching the outline.

5. **Tests.** The hook case *redirects to the legacy progress page and
   resolves to an empty object on a 404* became *surfaces a 404 as an error
   without navigating*; it keeps the `window.location` swap so it can assert
   `replace` was not called. `ProgressTab.test.jsx` gained the redirect case
   in a describe of its own. `fetchAndRender` took a `{ waitForLoaded }`
   option (the shape `OutlineTab.test.jsx`'s helper has) because its default
   wait — for the loading indicator to disappear — never ends when the
   indicator *is* the expected view. Both new cases fail against the
   pre-change source (`replace` called from the `queryFn`; query succeeds
   with `{}`) and pass with it.

6. **The sequence on a 404, render by render.** The early return and the
   effect divide the work, and it is worth being exact about which does what:

   1. *Progress query pending.* `tabDataQuery.error` is null, so
      `getResponseStatus` returns undefined and `redirectToLegacyProgress` is
      false. `ProgressTab` renders `TabWithTimer`; `TabPage` derives its
      loading view — header, spinner, footer. The effect registers; its body
      is skipped.
   2. *Metadata resolves.* Still pending on progress; same view.
   3. *The 404 lands.* React Query moves the query to its error state and
      `ProgressTab` re-renders. `getResponseStatus` now returns 404, the flag
      is true, and **in that same render** `ProgressTab` takes the early
      return and renders `PageLoading` directly. `TabWithTimer` unmounts, so
      the header and footer go and a bare spinner remains. `TabPage` never
      sees the errored query — it stops being rendered in the render where
      the error first appears — which is what keeps its error view off the
      screen without any change to `TabPage`.
   4. *After that commit, the effect runs.* Its dependency went false → true,
      so the body executes and calls `location.replace`. `useEffect` runs
      after paint, so the bare spinner is painted before navigation starts.
   5. *The browser navigates.* The current document — the bare spinner —
      stays visible until the LMS responds with the legacy progress page.

   So the early return decides what is on screen from step 3 until the new
   page arrives, and the effect decides when navigation starts. On the happy
   path the flag is false at every render: the early return is never taken
   and the effect body never runs, so `ProgressTab` returns the same tree it
   did before this change.

7. **Commit type `refactor:`, not breaking.** Nothing a learner on the
   progress tab sees is broken today, and the change is where a side effect
   lives. A plugin that fetched this query off the tab loses an unintended
   page navigation, which is the point, and no documented contract changes.

</details>

## Manual testing

<details>
<summary>Checklist</summary>

# Manual testing — move the 404 redirect to the legacy progress page out of `getProgressTabData` and into `ProgressTab` (#2104)

What changed: `getProgressTabData` no longer navigates the page on a 404; the
query errors instead (and stays unlogged, as before). `ProgressTab`, the
component that owns that fetch, redirects to `/courses/:courseId/progress` on
the LMS when its query fails with a 404, rendering a bare spinner from the
render in which the 404 lands until the browser gets there. Any other fetcher
of the progress query reads an error and stays put.

**The bug this layer could introduce** is a course that used to fall back to the
legacy progress page and now shows an error view, or a spinner forever. The
first group below is for that. The second group is the reason for the layer: a
fetcher that is not the progress tab must no longer move the page.

## Setup

**Producing a real 404 without changing any config.** The endpoint 404s for a
bad `student_id` as well as for an opted-out course, and the frontend cannot
tell the two apart — same status, same empty body, same code path from the
response onward. So run the "opted-out" checks below as **staff** against an
ordinary course, using a `targetUserId` that matches no user:

    /course/:courseId/progress/999999

The alternative is the real thing: Django admin → Waffle → *Waffle flag course
overrides* → `course_home.course_home_mfe_progress_tab` for one course,
override **off** (or `Disable progress page stacked configs`). Not needed for
the frontend path; use it only if you want the learner-facing framing.

**Do not block the endpoint in the Network tab for this.** A blocked request
fails at the network level with no `response`, so `getResponseStatus` reads
undefined, the flag stays false, the query retries three times (only 4xx skips
retries) and `TabPage` shows its ordinary error view. That is correct and
unchanged for a non-404 failure, but it never reaches the redirect.

**One check needs a widget that fetches the progress query from the outline.**
Add it to `env.config.jsx` (untracked — check a stale local copy is not
carrying other overrides). It asks for the bad id so it gets the 404 too:

```jsx
import { PLUGIN_OPERATIONS, DIRECT_PLUGIN } from '@openedx/frontend-plugin-framework';
import { useParams } from 'react-router-dom';
import { useProgressTabData } from './src/course-home/data/apiHooks';

const ProgressProbe = () => {
  const { status, error } = useProgressTabData(courseId, '999999');
  return <p data-testid="progress-probe">progress query: {status} {error?.response?.status ?? ''}</p>;
};

const config = {
  pluginSlots: {
    'org.openedx.frontend.learning.course_outline_tab_notifications.v1': {
      plugins: [{
        op: PLUGIN_OPERATIONS.Insert,
        widget: { id: 'progress-probe', type: DIRECT_PLUGIN, RenderWidget: ProgressProbe },
      }],
    },
  },
};
export default config;
```

That slot is rendered by `OutlineTab.jsx` in the side column (`CourseOutlineTabNotificationsSlot`); `course_home_section_outline.v1` is the only other slot on the outline tab.
This is the #2103 migration shape (a widget mounting `useProgressTabData`
itself), so it doubles as a preview of what that layer will tell plugins to do.

Run everything as **staff** on one ordinary course; the 404 checks use the
bad-id URL above, the happy-path checks use the plain progress URL.

### The redirect still works for the progress tab (404 via bad id)

- [x] **Load `/course/:courseId/progress/999999`.** Header and spinner while
      the queries are in flight, then a bare spinner (no header or footer)
      once the 404 lands, then the LMS legacy progress page at
      `/courses/:courseId/progress`. "There was an error loading this course."
      never appears. (Staff are not bounced back to the MFE by the legacy
      view, so it renders the legacy page for the staff user's own progress —
      same as before this change.)
- [x] **Browser back from the legacy page** lands where you came from, not in
      a redirect loop. `location.replace` adds no history entry, so the MFE
      `/progress/999999` URL should not be in the back stack.

### Nothing but the progress tab redirects (404 via bad id)

- [x] **Load the outline with the probe widget.** The outline renders and
      stays. The probe reads `progress query: error 404`. Nothing navigates.
- [x] **Load the dates tab** (no probe there): unaffected — it never fetched
      this query.

### Request and logging (404 via bad id)

- [x] **One `/api/course_home/progress/…/999999/` request per load**, answered
      404, on both the progress tab and the outline-with-probe. No retry — 4xx
      is not retried.
- [x] **Nothing logged for the 404.** The logging service (or the console, if
      the mock logger prints there) records no error and no info line for it;
      before this change the 404 was never logged either.

### The happy path is unchanged (same course, valid URLs)

- [x] **Load `/course/:courseId/progress`.** Grades, completion donut,
      certificate status, related links render as before; header present
      throughout; no bare-spinner frame.
- [x] **Load `/course/:courseId/progress/:targetUserId`** for a real enrolled
      learner: "Course progress for <username>", as before.
- [x] **Masquerade as a learner** (masquerade bar) and open the progress tab:
      renders as before.

### General

- [ ] **No new console errors** on any of the above. In particular no React
      warning from `ProgressTab` about hooks or effects, and no "Not
      implemented: navigation" — that one is jsdom-only. (not sure, doesn't seem like it)

Not covered:

- **A course with the MFE progress tab actually disabled** (waffle flag off,
  `DisableProgressPageStackedConfig`, or an Old Mongo course): the endpoint
  returns the same 404 with the same empty body, and the frontend takes the
  same path from there, so the bad-id runs above cover it. The difference is
  only the learner-facing framing — a learner, not staff, and the legacy view
  then rendering their real progress rather than redirecting.
- **A non-staff learner loading `/progress/<someone else>`**: the endpoint
  404s, the redirect fires as before, and the legacy view bounces a non-staff
  learner back to their own MFE progress page. Unchanged by this layer and
  needs a second learner account; the staff bad-id check exercises the same
  frontend path.

## Results

Run 2026-09-23 on tutor dev, as staff on one ordinary course, 404s produced
via the bad-id URL. Nine of ten checks pass as written: the bad-id progress
URL shows the bare spinner and lands on the legacy page with no error view;
browser back returns cleanly; the outline with the probe stays put and the
probe reads `error 404`; the dates tab is unaffected; one request per load,
no retry, nothing logged; the plain progress URL, a real learner's id and
masquerade all render as before. The console check is left unchecked: nothing
stood out, but the console was not compared against a before capture.

</details>


arbrandes (approved):
👍🏼

PR #2106, opened 2026-09-23
refactor!: stop the progress tab data refetching from components under the tab

brian-smith-tcril:
## Summary

One load of the progress tab requested `/api/course_home/progress/` **twice**. `ProgressTab` fetches it and gates on it; everything under the gate reads it through `useProgressData()`, which was also a *fetching* observer of the same query — 22 call sites, 21 files under `course-home/progress-tab/` plus the grade-breakdown slot. They mount in one commit after the data has landed, find it stale at the default `staleTime` of 0, and refetch. React Query dedupes them against each other (two requests, not twenty-two) but not against the owner's already-settled fetch.

This is the defect #2083 fixed on the outline and dates tabs and #2098 fixed on the courseware gate, and it gets the same fix: the owner fetches, the readers under it subscribe without fetching. `useProgressData()` never fetches; `ProgressTab` keeps fetching. One request per load.

Part of the Redux → React Query migration (#1946, Stage 1). Closes #2103. Stacked above #2105 (#2104), which moved the progress query's 404 redirect out of the `queryFn` so that the migration step below — a plugin mounting the fetch itself — has no side effect.

## What changed

- **`useProgressTabData` takes `{ enabled }`**, the course-home `QueryOptions`, in the two-positional-plus-options shape `useProctoringInfoData` has. Plain `enabled`, no `!!courseId` guard — the hook never had one and the route always supplies a `courseId`.
- **`useProgressData()` passes `{ enabled: false }` and takes no option of its own** — the shape #2098 settled for `useIsCourseLoaded`. Every caller renders under `ProgressTab`, which fetches the same key from the same route params, so no caller wants the reader to fetch; a default-on option would let a new reader silently reintroduce the duplicate. A disabled observer returns the query's current state and re-renders when the owner's fetch lands, so every reader sees the same value as before.
- **No `staleTime`.** It would hide the reader fetches without deciding who owns them, and would change the owner's freshness too: the progress payload changes server-side with no app action, and `useRequestCert`, which renders under this tab, relies on refetch-on-mount to resync.
- **README.** `progress-tab/README.md` now says `useExamsData()` reads the progress tab's own data, is for widgets in progress-tab slots, and returns `null` elsewhere unless something rendered on the page fetches that data.
[...]

PR #962, opened 2022-08-15
Updated frontend-build to v12

BilalQamar95:
**Ticket:**
[42: Upgrade eslint to v8.x](https://github.com/openedx/frontend-wg/issues/42)

**What changed?**
- Updated `frontend-build `to v12 (Eslint was updated in `frontend-build` version resulting in it's version being bumped to v12. This PR updates `frontend-build` to reciprocate eslint version update)
- Resolved `eslint` issues post update

PR #2108, opened 2026-09-23
test: make the LoadedTabPage streak test actually assert

### LINKED ISSUE #2107: Make the `LoadedTabPage` streak test actually assert: an unconditional mock from #354 and a fixture inert since AA-1018
Part of #1946 — Redux → React Query migration (Stage 1). Not a conversion: a test-hygiene fix, filed here because B2 (#2085) rewrites how `LoadedTabPage.test.jsx` gets its data, and the suite's streak case should assert something before that layer changes the source.

**What's wrong.** *shows streak celebration modal* in `src/tab-page/LoadedTabPage.test.jsx` cannot fail, for two independent reasons:

- **The mock renders unconditionally.** The suite replaces `StreakCelebrationModal` with a component that returns `<div data-testid="StreakModal" />` whatever its props, and `LoadedTabPage` always renders `StreakModal` (`LoadedTabPage.tsx:71-77`), open or closed. The assertion — the mock is in the document — is true on every render of `LoadedTabPage`.
- **The fixture never reaches the component.** The case builds the *courseware* metadata factory with `celebrations: { streakLengthToCelebrate: 3 }`. `LoadedTabPage` reads `celebrations` from `courseHomeMeta`, which `initializeTestStore` fills from the separate `courseHomeMetadata` factory, built with `celebrations: null` (`buildSimpleCourseAndSequenceMetadata`, `sequenceMetadata.factory.js`). `streakLengthToCelebrate` is `null` in the component for the whole test.

Deleting the `celebrations` attribute from the fixture leaves the test green.

**Provenance.** The mock has been unconditional since the case was written (2525805a, #354, 2021-02-22). The fixture stopped reaching the component in 3fe5bb17 (AA-1018, 2022-01-14), which moved `LoadedTabPage`'s read from the courseware model to `courseHomeMeta` and gave the test store its own `courseHomeMetadata` factory; the case was not updated. b3d33667 (#962, 2023) only reformatted the mock for a lint rule.

**Precedent.** #2078 (PR #2079) and #2100 (PR #2101) each fixed tests that could not fail as their own `test:` layer, below the layer that would otherwise have had to touch them, so that layer stayed about its own change.

**Tasks**
- Build the fixture on the factory the component reads: `Factory.build('courseHomeMetadata', { celebrations: { streak_length_to_celebrate: 3 } })`, passed as `initializeTestStore`'s `courseHomeMetadata` option (the shape the suite's staff case already uses).
- Make the mock record what it is given — a `jest.fn()` component the `jest.mock` factory delegates to — and assert on the props `LoadedTabPage` derives from `celebrations`: `isStreakCelebrationOpen: true` and `streakLengthToCelebrate: 3`.
- Add the contrast in *renders correctly* (default fixture, `celebrations: null`): the modal is rendered closed, `isStreakCelebrationOpen: false`. Without it the new assertion could be satisfied by a mock that is always called open.
- Test-only; `LoadedTabPage.tsx` is not touched.

**Verify:** the suite is green; removing `celebrations` from the streak fixture makes the streak case fail, and swapping `!!isStreakCelebrationOpen` for `true` on the `StreakModal` prop makes the contrast fail.

> [!NOTE]

brian-smith-tcril:
## Summary

Make the *shows streak celebration modal* case in `tab-page/LoadedTabPage.test.jsx` actually assert. It could not fail: the `StreakModal` mock rendered a bare div whatever its props, and `LoadedTabPage` renders `StreakModal` on every render, open or closed; and its fixture set `celebrations` on the courseware metadata factory, while the component has read `celebrations` from `courseHomeMeta` since AA-1018 (2022), which the test store fills from a separate factory with `celebrations: null`. Removing `celebrations` from the fixture left the test green. Test-only; `LoadedTabPage.tsx` is untouched. Stacked on #2106 as the stack top; below B2 (#2085), which rewrites this suite's data source and should land on a case that checks something. Part of #1946 as test hygiene on code the migration touches. Closes #2107.

## What changed

- **The fixture goes on the `courseHomeMetadata` factory**, through the `initializeTestStore` option the suite's staff case already uses, so the streak reaches the model the component reads, camel-cased by the real normalizer.
- **The modal is no longer mocked.** Closed, `StreakCelebrationModal` renders nothing; open, it renders a dialog titled with the streak length. The case asserts a dialog reading "3 day streak" is on the page. Rendering it costs what its own suite pays: the analytics module mocked and an `act` wrapper around the render.
- **`renders correctly` asserts the contrast**: no dialog on the default fixture. A modal that is always open cannot satisfy the streak case's assertion.

## Testing

`npm run types` and `npm run lint` clean; `LoadedTabPage.test.jsx` 3/3 with no console output. Negative checks on this layer: removing `celebrations` from the fixture fails the streak case (no element with role `dialog`); forcing the modal open in the component fails the contrast (a dialog titled "null day streak"). The full suite was not run for a one-file test change.

## Provenance

The mock has been unconditional since the case was written in #354 (2021-02-22). The fixture stopped reaching the component in AA-1018 (3fe5bb17, 2022-01-14), which moved `LoadedTabPage`'s read to `courseHomeMeta` and gave the test store its own `courseHomeMetadata` factory. Found while planning B2 (#2085), which rewrites how this suite gets its data; split out as its own layer, like #2078 / #2079 and #2100 / #2101, so B2's diff to the suite is only the data-source swap.

## Decisions

<details>

# Decisions — make the LoadedTabPage streak test actually assert (#2107)

Test-only layer between #2103 (PR #2106) and B2 (#2085). Entries 1 and 2 were
settled while reviewing the #2085 plan; the rest landed with the code.

1. **A layer of its own, below B2.** B2 rewrites how `LoadedTabPage.test.jsx`
   gets its data (the query instead of the store), so it would have had to
   touch this case anyway, and the first draft of the B2 plan fixed it there.
   Rejected in review for the #2078 / #2100 shape: a small issue and a tiny
   `test:` layer first, so B2's diff to the suite is only the data-source
   swap and the repair of a 2021 test is reviewed as what it is. It sits on
   #2103 because that is the stack top; nothing in it depends on #2103.

2. **Two defects, either of which alone keeps the case green.**
   - **The mock rendered unconditionally.** `StreakCelebrationModal` was
     mocked as a component returning `<div data-testid="StreakModal" />`
     whatever its props, and `LoadedTabPage` always renders `StreakModal`
     (`LoadedTabPage.tsx:71-77`), open or closed. "The mock is in the
     document" is true on every render. So since the case was written
     (2525805a, #354, 2021-02-22).
   - **The fixture never reached the component.** The case built the
     courseware metadata factory with `celebrations`, but `LoadedTabPage` has
     read `celebrations` from `courseHomeMeta` since 3fe5bb17 (AA-1018,
     2022-01-14), and `initializeTestStore` fills that model from a separate
     `courseHomeMetadata` factory built with `celebrations: null`
     (`buildSimpleCourseAndSequenceMetadata`, `sequenceMetadata.factory.js`).
     The case was not updated then. b3d33667 (#962, 2023) only reformatted
     the mock for a lint rule.

   Removing `celebrations` from the fixture left the case green, on this
   layer's base with no other change.

3. **The fixture goes on the `courseHomeMetadata` factory, through the
   `initializeTestStore` option.** `initializeTestStore({ courseHomeMetadata })`
   is the shape the suite's staff case already uses
   (`original_user_is_staff: true`), and the store fills `courseHomeMeta`
   from the mocked `course_metadata` endpoint through the real normalizer, so
   `streak_length_to_celebrate` arrives camel-cased exactly as production
   data does. The course id comes from `getTestStoreIds(testStore)` rather
   than a hand-built `courseMetadata`, since the case no longer builds one.

4. **The modal is not mocked; the case asserts on the dialog.** A first
   version kept a mock and made it a `jest.fn` component
   (`UnitSuspense.test.jsx:20-21`'s shape), asserting on the props
   `LoadedTabPage` passed it — `isStreakCelebrationOpen: true`,
   `streakLengthToCelebrate: 3`. Rejected in review: that tests that
   `LoadedTabPage` hands a stub an open flag, not that anything is shown,
   which is what the case is named for. The real `StreakCelebrationModal`
   returns `null` when closed (`StreakCelebrationModal.jsx:117-119`) and
   renders a Paragon `ModalDialog` titled
   `${streakLengthToCelebrate} day streak` when open (`:143-161`), so the
   case now asserts `getByRole('dialog')` has text `3 day streak`: both the
   open flag and the length flow through the real component to what a
   learner sees. Rendering it costs what its own suite pays
   (`StreakCelebrationModal.test.jsx:18-19,56`): `jest.mock('@edx/frontend-platform/analytics')`,
   because opening fires `recordStreakCelebration` → `sendTrackEvent`, and an
   `act` wrapper around the render for the modal's effects. Its own
   `useModel('courseHomeMeta')` read (`org`, `username`) is served by the
   same store the fixture fills; its discount effect is skipped because
   `streakDiscountCouponEnabled` is `null` on the default fixture.

5. **`renders correctly` asserts the contrast.** On the default fixture
   (`celebrations: null`) there is no dialog on the page
   (`queryByRole('dialog')` is null). Without it the streak case's assertion
   could be satisfied by a modal that is always open — the mirror of
   defect 1. It also pins the `!!` coercion on the prop
   (`LoadedTabPage.tsx:75`): forcing the prop to `true` in the component
   renders a dialog titled "null day streak" and fails this case.

6. **Test-only, one file.** `LoadedTabPage.tsx` is untouched; the diff is
   `src/tab-page/LoadedTabPage.test.jsx` alone (10 insertions, 8 deletions).
   Suite 3/3 with no console output; `npm run lint` and `npm run types`
   clean. Negative checks on this layer: removing `celebrations` from the
   fixture fails *shows streak celebration modal* (no element with role
   `dialog`); replacing `!!isStreakCelebrationOpen` with a bare
   `isStreakCelebrationOpen` on the `StreakModal` prop fails *renders
   correctly*. The full suite was not run for a one-file test change.

</details>


arbrandes (approved):
👍🏼

